Issue #79, opened 2020-11-03
HTML tags in Product description


websupporter:
### Describe the bug
When the product description contains HTML tags (e.g. to separate different content paragraphs with headings), these HTML tags are visible in the PayPal dialogue.

### To reproduce
1. Create a product with HTML tags like headings at the beginning of the description
1. Click the PayPal button on the product page
1. Go to the product list in the PayPal dialogue.

**Screenshots**

![image](https://user-images.githubusercontent.com/6458412/97986569-94ba8c00-1de2-11eb-926c-b4067593d6fc.png)

### Expected behaviour
The HTML tags should not be present in the PayPal dialogue.

**Environment (please complete the following information):**
 - WordPress Version: latest
 - WooCommerce Version: latest
 - WooCommerce PayPal Payments Plugin Version: 1.0.0
 - Browser [e.g. chrome, safari] and Version: all

Issue #76, opened 2020-10-22
purchase_units -> amount -> value should be string


websupporter:
https://developer.paypal.com/docs/api/orders/v2/?mark=amount#definition-amount_with_breakdown

When send to the API, the value property of the amount object should be a string, but it is a float:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Entity/class-amount.php#L77

The money object does this string conversion towards the API here
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Entity/class-money.php#L68

and we should follow this pattern and adhere to the APIs definition.

Issue #72, opened 2020-10-16
Remove ability to change shipping address on PayPal from checkout page


Chaithi:
Mexico team pointed out an issue with the checkout page when checking out with PayPal.

After customer has entered their address on Woo and then submit to PayPal, they can still change their address within the PayPal modal, which can potentially cause issues with Chargebacks.

Best practice is to set the following in the [Application Context](https://developer.paypal.com/docs/api/orders/v2/#definition-order_application_context) object:

`"shipping_preference": "SET_PROVIDED_ADDRESS"`

Note this should only be done via the Checkout page when the buyer has already provided a shipping address. When using the shortcut flows (cart, mini-cart, product) the behavior shouldn't change.

Current behavior:
![image (2)](https://user-images.githubusercontent.com/6610999/96279106-e0280a00-0f9b-11eb-8d14-a86f9f613a9e.png)

Expected behavior:

![image (3)](https://user-images.githubusercontent.com/6610999/96279124-e61deb00-0f9b-11eb-842c-5b6b91275922.png)

Issue #67, opened 2020-10-15
Attempting to checkout without country selected results in unexpected error message


chickenn00dle:
Attempting to checkout without selecting a country results in the following error: https://d.pr/i/tsrqzK

All other required fields result in more specific errors, for example:

```
<strong>Billing Email address</strong> is a required field.
```

Note: Even these errors are a bit broken as we are able to see the HTML strong tags.

**Steps to reproduce**

1. Make sure PayPal is connected
2. Add an item to cart and go to checkout
3. Fill out all billing details except country
4. Attempt to pay with PayPal
5. You will see the above error

PR #81, opened 2020-11-04
Release/1.0.1

Dinamiko:
* Fix - PayPal Smart buttons don't load when using a production/live account and `WP_Debug` is turned on/true. #66
* Fix - [Card Processing] SCA/Visa Verification form loads underneath the Checkout blockUI element. #63
* Fix - Attempting to checkout without country selected results in unexpected error message. #67
* Fix - Remove ability to change shipping address on PayPal from checkout page. #72
* Fix - Amount value should be a string when send to the api. #76
* Fix - "The value of a field does not conform to the expected format" error when using certain e-mail addresses. #56
* Fix - HTML tags in Product description. #79

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,13 @@
 *** Changelog ***
 
+= 1.0.1 - 2020-11-04 =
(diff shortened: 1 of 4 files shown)


jorgeatorres (approved):
LGTM.

PR #89, opened 2020-11-19
Check purchase units payments captures status is not declined before …

### LINKED ISSUE #88: Order with Payment received when Hosted Fields transaction is declined
When a Hosted Fields transaction is declined, the API call responds with a 201 status code, but will include purchase_units -> payments -> captures -> status ([Documentation link](https://developer.paypal.com/docs/api/orders/v2/#definition-capture))

Since this acts differently than PayPal payments, it looks like this wasn’t implemented to check for, so these transactions are being marked as completed. This can cause merchants to ship out items, but not having been paid.

### To reproduce
1. In Settings Payments enable PayPal Card Processing
2. Add product to cart and go to Checkout page
3. In Billing details First name use: CCREJECT-INVALID
4. Select credit card as payment and fill the fields
5. Place order 

### Expected behavior
Once order is received the order status should be On hold.

### What happened instead
Once order is received the order status is Processing with a Payment received Order note.  

**Environment:**
 - WordPress Version 5.5.3
 - WooCommerce Version 4.6.1
 - WooCommerce PayPal Payments Plugin Version 1.0.2

Dinamiko:
Fixes #88

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
@@ -10,6 +10,7 @@ declare( strict_types=1 );
 namespace WooCommerce\PayPalCommerce\WcGateway\Gateway;
 
 use WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
 
 /**
  * Trait ProcessPaymentTrait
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
```diff
@@ -71,7 +72,15 @@ public function process_payment( $order_id ) {
 			}
 
 			$this->session_handler->destroy_session_data();
+		} catch ( RuntimeException $error ) {
+			$this->session_handler->destroy_session_data();
+			wc_add_notice(
+				__( 'Payment provider declined the payment, please use a different payment method.', 'woocommerce-paypal-payments' ),
+				'error'
+			);
+			return null;
```
websupporter: I would put the message as the exception message. This way, if we need to throw exceptions for other reasons, later on, we would not have to touch this piece of code anymore.

`throw new RuntimeException(__( 'Payment provider declined the payment, please use a different payment method.', 'woocommerce-paypal-payments' ));`

`wc_add_notice($error->getMessage())`

websupporter (approved):
Left one remark, otherwise good to go imo.

modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
Dinamiko: Done, good point, thanks.

PR #100, opened 2021-01-13
Update `dhii\module-interface`

Dinamiko:
Fixes the problem when there is more than one instance of `dhii\module-interface` composer package by updating it to the latest version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,8 +4,9 @@
     "description": "PayPal Commerce Platform for WooCommerce",
     "license": "GPL-2.0",
     "require": {
-        "dhii/module-interface": "0.1",
+        "dhii/module-interface": "0.2.x-dev",
         "psr/container": "1.0.0",
+        "oomphinc/composer-installers-extender": "^1.1",
         "container-interop/service-provider": "^0.4.0",
         "dhii/containers": "v0.1.0-alpha1",
         "dhii/wp-containers": "v0.1.0-alpha1",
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -4,7 +4,7 @@
     "description": "PayPal Commerce Platform for WooCommerce",
     "license": "GPL-2.0",
     "require": {
-        "dhii/module-interface": "0.1",
+        "dhii/module-interface": "0.2.x-dev",
```
AlexP11223:
```suggestion
        "dhii/module-interface": "^0.2",
```

Maybe better to use this, that's what we do in other projects.

Dinamiko: @AlexP11223 done, thanks for the info 👍

websupporter:
`dhii/module-interface": "0.1"` was introduced in #2

0.2 requires PHP 7.1 (see  https://github.com/Dhii/module-interface/blob/v0.2.0-alpha1/composer.json#L15), while #1 requires the plugin to be compatible with PHP7.0

XedinUnknown:
@Dinamiko, this package has already been upgraded to use `dhii/module-interface` `0.3.x` [here][1]. XedinUnknown/woocommerce-paypal-payments#1 will be merged into woocommerce/woocommerce-paypal-payments#120.

[1]: https://github.com/XedinUnknown/woocommerce-paypal-payments/pull/1/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R8

jorgeatorres (approved):
Had to rebase to fix conflicts and also update `PluginModule::run()` to follow `ModuleInterface::run()`. Other than that, this looks good 👍.

Issue #96, opened 2021-01-07
Payment buttons only visible on order-pay site when Mini Cart is enabled; payment fails


Dinamiko:
### Describe the bug
As a user, I can only see the payment buttons in the order-pay site when the checkbox for "Enable on Mini Cart" has been enabled. The order-pay page can be accessed from within my customer account when an unpaid order exists: /my-account/orders/
When the buttons are visible, I get an error when trying to pay for the order. After clicking “Pay with PayPal”, the PayPal window pops up but closes itself instantly without the chance to log in. No logs are created for this attempted payment.

### Expected behavior
Scenario **Pay for order with PayPal Button with Terms and Conditions checkbox**
**Given** there is an order pending to pay in My Account
**And** Enable on Mini Cart is disabled in Payment settings
**And** a page is selected on Settings /  Advanced / Terms and conditions
**And** I select a pending order by clicking Pay action in My Account / Orders list
**And** I check “I have read and agree …” checkbox
**When** I click PayPal Button and pay
**Then** I should see Order received correctly

Scenario **Pay for order with Credit Card with Terms and Conditions checkbox**
**Given** there is an order pending to pay in My Account
**And** Enable on Mini Cart is disabled in Payment settings
**And** a page is selected on Settings /  Advanced / Terms and conditions
**And** I select a pending order by clicking Pay action in My Account / Orders list
**And** I check “I have read and agree …” checkbox
**When** I fill credit cart data and click Pay for order
**Then** I should see Order received correctly

Scenario **Pay for order with PayPal Button without Terms and Conditions checkbox**
**Given** there is an order pending to pay in My Account
**And** Enable on Mini Cart is disabled in Payment settings
**And** there is no page selected on Settings /  Advanced / Terms and conditions
**And** I select a pending order by clicking Pay action in My Account / Orders list
**When** I click PayPal Button and pay
**Then** I should see Order received correctly

Scenario **Pay for order with Credit Card without Terms and Conditions checkbox**
**Given** there is an order pending to pay in My Account
**And** Enable on Mini Cart is disabled in Payment settings
**And** there is no page selected on Settings /  Advanced / Terms and conditions
**And** I select a pending order by clicking Pay action in My Account / Orders list
**When** I fill credit cart data and click Pay for order
**Then** I should see Order received correctly

**Environment**
PayPal Payments 1.0.3
WordPress 5.6
WooCommerce 4.8
Storefront Theme 3.0.0
PHP 7.4

Issue #93, opened 2020-12-15
high volume of failed calls to /v1/notifications/webhooks


Dinamiko:
There is a large number of failed calls to the POST /v1/notifications/webhooks API, it seems that register method gets called also when the webhook is already registered: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-webhooks/src/class-webhookregistrar.php#L69-L92

When registration fails this code tries to reschedule it again: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-webhooks/src/class-webhookregistrar.php#L86-L89

It could be that there are some fails, which are not related to e.g. timeout, but are bound to the paypal account (maybe there are accounts, which simply can't register a webhook). 

One solution could be investigate the specific error and seperate different reasons for failing to decide if reschedule or not. 

Another solution could be remove the rescheduling logic and let it fail, the response error(s) are already logged here:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Endpoint/class-webhookendpoint.php#L106
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Endpoint/class-webhookendpoint.php#L124

Issue #91, opened 2020-12-02
GB country has ACDC blocked


Chaithi:
### Describe the bug
Due to the code [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/4187fbb93cc22fa90105499da0b785529a0d2efb/modules/ppcp-button/src/Assets/class-smartbutton.php#L734-L745) the **disable-funding=card** is added to the SDK for all GB merchants, meaning they cannot use ACDC features.

The code snippet I outlined checks if the store country is GB, if so, it adds **disable-funding=card** to the JavaScript SDK query parameters.

What this does is disables the black **Credit or Debit Card** button, but also prevents the hosted fields that allow for processing direct credit cards.

This means any GB merchant that onboards to the Credit Card Processing section should have buyers have a Credit Card option to enter their CC number, exp date, CVV, but because of this code block, those fields will never render.

Looking back at our discussions during implementation, this was supposed to be **disable-funding=credit** as PayPal Credit was not supported in the UK

### To reproduce

1. Have store country set to GB.
2. Onboard to ACDC.

### Expected behavior
ACDC form should show.

PR #101, opened 2021-01-18
Release/1.0.4

Dinamiko:
* Fix - Check if WooCommerce is active before initialize. #99
* Fix - Payment buttons only visible on order-pay site when Mini Cart is enabled; payment fails. #96
* Fix - High volume of failed calls to /v1/notifications/webhooks #93
* Fix - GB country has ACDC blocked. #91


jorgeatorres (approved):
LGTM.

PR #120, opened 2021-02-02
Building on multiple PHP versions, and remove auth token

XedinUnknown:
- Will solve problems described in #100.
- Now requiring PHP 7.1.
- Build on PHP 7.1 through 8.0
- Now using PSR-4.
- Add lockfile for consistency between dep installations.
- Now using a distributed module system with Composer Merge Plugin.
- Now using modularity standard version 0.3.
- Configured PHPCS.
- Cleaned unused deps and other things.
- Add Psalm (allowed to fail for now).
- Remove auth token.
- Bring workflow in line with the rest of Inpsyde plugins.


XedinUnknown:
Due to a series of factors, I stronly recommend using PHP 7.1 for this project. Not only will this fix many existing incompatibility issues, but will also make the project significantly more maintainable, and compatible with a wider range of libraries and tools that make maintainability much cheaper.

XedinUnknown:
Right now, the [build] is failing. This is because there isn't a Composer lockfile, so it tries to update on its own - which is not something that should happen in a plugin. And I cannot add the lockfile, because I cannot install deps: the project of PHP 7.0 (this was not previously declared) depends on PHP 7.1 libs.

[build]: https://github.com/XedinUnknown/woocommerce-paypal-payments/actions/runs/531502364

XedinUnknown:
Nevertheless, this PR brings things like Docker to the mix, which will make installing deps on the correct version of PHP much easier. I will add Gulp instructions too, when the lib compatibility problem is resolved. For overall picture of improvements, see [`wp-oop/plugin-boilerplate`][].

[`wp-oop/plugin-boilerplate`]: https://github.com/wp-oop/plugin-boilerplate

XedinUnknown:
One important question - maybe for @websupporter - is: why not use PSR-4? Is the WP class file naming convention a hard requirement of this project?

jorgeatorres:
Hi @XedinUnknown!

Thanks for the PR. A couple of comments:

- While not overly critical, this is a WooCommerce (and thus WordPress) plugin so we definitely want it to adhere to the WordPress standards as much as possible. Unless there's a valid reason not to do that.
- Regarding the PHP 7.0 vs PHP 7.1 issue, this is something we're discussing internally. WooCommerce core recommends 7.3 but the requirement is 7.0. PayPal Payments should work wherever WC core does, but if you could elaborate on some of the reasons for possibly requiring 7.1, that could help steer the decision.
- Some of the changes you're bringing into the plugin look great (Docker, etc.) and definitely help with testing, etc. Only thing worth mentioning is that the GitHub actions aren't supported by this repo right now, so the Travis integration shouldn't be removed.

XedinUnknown:
Thanks for the quick response, and sorr for the delay! Here are my notes to your comments:

- We understand that you want to adhere to the WordPress standards and of course we will do so too. The reason why I suggested PSR-4 and why I would prefer it is because PSR-4 is technically better for autoloading and would be more convenient for development. If the current classmap approach is important, I am happy to leave it as is.
- I know that there is already an on-going discussion about the minimal PHP requirement for this project. Adding to any points that may have already been mentioned, in my opinion it could be a good idea to upgrade to PHP 7.1 because many of the newer tools, standards, and libraries that could greatly reduce the amount of time spent on development, debugging, and code reviews require at least PHP 7.1. One of them is Psalm, and it would allow us to automatically verify the correctness of the code in many cases. The same can be said about interop standards, and other libraries that would remove the need for duplicate effort.
- As far as I know, Github Actions are automatically supported in every Github repo - free for any public project, including this one. I am aware that for this repo Actions may have been disabled. As we are working with Github Actions in all of our other plugins, we would prefer that. I understand if you would like to keep using Travis. Thus my suggestion is to keep using Travis for what it is already used for (PHPCS checks), and do the rest in Github Actions. Would that work for you?

jorgeatorres:
Hi @XedinUnknown!

> [...] because PSR-4 is technically better for autoloading and would be more convenient for development. If the current classmap approach is important, I am happy to leave it as is.

I don't feel strongly about this and we can always exclude some of the rules from the WP standard if the benefit is worth it. Could you give some examples of how some classnames/directories would have to change?

> Adding to any points that may have already been mentioned, in my opinion it could be a good idea to upgrade to PHP 7.1 [...]

Got it. We've only found 7 extensions on the marketplace that explicitly require PHP > 7.0. While this is something we're taking into account, it's not necessarily the only way to look at this matter. We're trying to obtain some stats from current users to decide.
Quick question: if we bump the requirement to 7.1, would this allow building the extension under PHP 8.x? Right now, dependencies prevent this (Composer can't resolve everything successfully).

> As far as I know, Github Actions are automatically supported in every Github repo - free for any public project, including this one. I am aware that for this repo Actions may have been disabled.

We were seeing some errors with actions in this repo due to the repo being in a legacy plan, but it seems those were from the time the repo was private. I did a quick test and it is true that actions are available now that it is public.
As long as we have a workflow for running unit tests and PHPCS, feel free to make the changes, including removing the Travis integration.

Thanks for working on this!

XedinUnknown:
@jorgeatorres, please accept my apologies for the delay. I have taken the liberty to implement the suggested changes in a [separate branch][1] for your approval.

Here, you can see that the directory tree has changed in the following way:
class file names have been changed to match the class names 1:1, as per the PSR-4 spec, e.g. `class-disabledsmartbutton.php` -> `DisabledSmartButton.php`.

The package now requires PHP 7.1, and supports PHP 8, while building on all minor versions throughout, e.g. 7.1, 7.2, 7.3, 7.4, 8.0 - on Github Actions.

Other upgrades have been made, and you can find their description in the PR. They include upgrading standards for wider compatibility.

Please let me know what you think. Looking forward to your reply.

[1]: https://github.com/XedinUnknown/woocommerce-paypal-payments/pull/1

jorgeatorres:
Hi @XedinUnknown,

Thanks for replying back. I'll have a look at your PR as soon as possible 🤞. Thanks for working on it!
I'm also going to bring up the PHP 7.0 vs 7.1 issue with my team again and will let you know what the decision is.

jorgeatorres:
Hi @XedinUnknown,

After discussing with the team, we've decided it's ok to bump the PHP requirement to 7.1. I also don't see a problem with skipping some WP-specific naming rules as long as that makes development and testing easier.
In particular, I hope bumping the PHP requirement would allow us to support building the extension in PHP 8.0 and address some of the conflicts/issues with Composer (or lack of support for Composer 2). Would that be the case?

Given we're near the [1.3.0](#144) release, we believe this would be a good time to enforce the PHP 7.1 requirement (updating the plugin header and adding a check just in case someone installs under PHP < 7.1). As for the other changes (renaming of classes, tests, etc.) as found in your branch, we would be happy to merge those _after_ 1.3.0 is released.

XedinUnknown:
This builds for me in XedinUnknown/woocommerce-paypal-payments#1.

XedinUnknown:
@Dinamiko, after synchronizing with upstream, [build][] is now failing. Would you be able to take a look, please?


[build]: https://github.com/XedinUnknown/woocommerce-paypal-payments/runs/2978604717?check_suite_focus=true

Dinamiko:
Already included after `1.6.1` release.

PR #134, opened 2021-03-10
Introduce client-side vaulting and allow Subscription renewals through payment tokens

Dinamiko:
### Description
This PR adds client-side vaulting functionality to both PayPal and Credit Card payments. It also allows WooCommerce Subscriptions product renewals through PayPal payment tokens.


modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
```diff
@@ -85,4 +85,19 @@ public function process_payment( $order_id ) {
...
+
+	/**
+	 * Checks if vault setting is either enabled for PayPal or Credit Card.
+	 *
+	 * @return bool Whether vault settings are enabled or not.
+	 * @throws \WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException When a setting hasn't been found.
+	 */
+	protected function vault_settings_enabled(): bool {
+		if ( ! $this->config->has( 'vault_enabled' ) && ! $this->config->has( 'dcc_vault_enabled' )
+			|| ! $this->config->get( 'dcc_vault_enabled' ) && ! $this->config->get( 'dcc_vault_enabled' ) ) {
```
strangerkir: I noticed the condition on this line is duplicated. Also, I guess something wrong with the condition. If I understand the idea correctly, you meant something like 

```php 
return $vault_enabled || $dcc_vault_enabled;          
```

woocommerce-paypal-payments.php
```diff
@@ -28,6 +28,8 @@
 define( 'PAYPAL_SANDBOX_API_URL', 'https://api.sandbox.paypal.com' );
 define( 'PAYPAL_INTEGRATION_DATE', '2020-10-15' );
 
+define( 'PPCP_FLAG_SUBSCRIPTION', true );
+
 ! defined( 'CONNECT_WOO_CLIENT_ID' ) && define( 'CONNECT_WOO_CLIENT_ID', 'AcCAsWta_JTL__OfpjspNyH7c1GGHH332fLwonA5CwX4Y10mhybRZmHLA0GdRbwKwjQIhpDQy0pluX_P' );
 ! defined( 'CONNECT_WOO_SANDBOX_CLIENT_ID' ) && define( 'CONNECT_WOO_SANDBOX_CLIENT_ID', 'AYmOHbt1VHg-OZ_oihPdzKEVbU3qg0qXonBcAztuzniQRaKE0w1Hr762cSFwd4n8wxOl-TCWohEa0XM_' );
 ! defined( 'CONNECT_WOO_MERCHANT_ID' ) && define( 'CONNECT_WOO_MERCHANT_ID', 'K8SKZ36LQBWXJ' );
```
strangerkir, line 35: I think everything here is ok, but just in case: are these values the same always and not secret?

strangerkir, line 35: I think everything here is ok, but just in case: are these values the same always and not secret?

strangerkir (commented):
For me, it looks like a great job!

But probably there is one accidental error with the vault enabled check that should be fixed.

modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
Dinamiko: Good catch here, thanks! yes the idea is to return true as soon as one of them (PayPal or credit card) has the setting enabled. There were some typos getting the settings, also I´ve reversed the conditional, I think now is a bit more clear.

woocommerce-paypal-payments.php
Dinamiko: Yup, those are the IDs, always the same, are public data and ok to be there.

strangerkir (approved):
Looks good to me.

strangerkir (approved):
Looks good to me.

PR #135, opened 2021-03-10
Remove the addition of Venmo to the disable-funding variable when loading the SDK

Dinamiko:
### Description
Removes add Venmo to disable funding by default, adds it as `disable_funding` option.


strangerkir (approved):
Looks good to me.

PR #142, opened 2021-03-31
Pay Later disclaimers consolidation

Dinamiko:
In the admin panel, next to the option for merchants to enable Pay Later messaging, please show the following disclaimers based on region: United States, Germany, United Kingdom, France, Australia

**Displays Pay Later messaging for available offers. Restrictions apply. Click here to learn more.**

"Click here to learn more" should be hyperlinked to the following links:
US: https://developer.paypal.com/docs/commerce-platforms/admin-panel/woocommerce/us/
UK: https://developer.paypal.com/docs/commerce-platforms/admin-panel/woocommerce/uk/
DE: https://developer.paypal.com/docs/commerce-platforms/admin-panel/woocommerce/de/
AU: https://developer.paypal.com/docs/commerce-platforms/admin-panel/woocommerce/au/
FR: https://developer.paypal.com/docs/commerce-platforms/admin-panel/woocommerce/fr/


Dinamiko:
Hi @strangerkir I´ve created this branch from `feature/PCP-88-payment-token-vaulting` because vaulting has some conditionals for Pay Later messaging, the only code in this PR that you have to review is this commit: https://github.com/woocommerce/woocommerce-paypal-payments/pull/142/commits/b5b10a0dfc4168890256282da98d37d854d140c2

modules/ppcp-wc-gateway/services.php
```diff
@@ -815,7 +859,7 @@ static function ( $container ): AuthorizeOrderActionNotice {
 				),
 				'requirements' => array( 'messages' ),
 				'gateway'      => 'paypal',
-				'description'  => str_replace( '<a>', '<a href="https://www.paypal.com/us/business/buy-now-pay-later">', __( 'Customize the appearance of <a>Pay Later messages</a> on checkout to promote special financing offers, which help increase sales.', 'woocommerce-paypal-payments' ) ),
+				'description'  => str_replace( '<a>', '<a href="' . $messages_disclaimers->link_for_country() . '" target="_blank">', __( 'Displays Pay Later messaging for available offers. Restrictions apply. <a>Click here to learn more.</a>', 'woocommerce-paypal-payments' ) ),
```
strangerkir, line 862: I think it shouldn't be like so. For the countries which shouldn't have restrictions, it displays the link with an empty 'href' attribute.

strangerkir (changes requested):
I guess something is wrong with the disclaimer message and its link. If I understand it correctly, the disclaimer should be displayed in certain countries only.

The rest looks good to me.

modules/ppcp-wc-gateway/services.php
Dinamiko: @strangerkir Pay Later settings are only displayed on the countries that allows it: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Helper/class-messagesapply.php#L23-L28

So for example if you set Store Address Country / State to "Spain" it does not show any Pay Later setting.

strangerkir: Oh, you are right. 

But anyway, it gives me the <a> tag with the empty `href` attribute when the country is the UK.

UPD: I found the root of the problem now. The `wc_get_base_location()` returns 'GB' for 'country' when store location is United Kingdom. But the array with messages uses 'UK'. Not sure what exactly should be fixed in this case.

Dinamiko: Good catch! :) I´ve just fixed now: https://github.com/woocommerce/woocommerce-paypal-payments/pull/142/commits/3d7fb7fc536871f70719f18502450093697dd41a?file-filters%5B%5D=.php&hide-deleted-files=true#diff-c20665203423aa892f69af009a83dc777d23631ec30ebb7776746ddb341a6f5fR28

when you have time please try it again and let me know, thanks.

strangerkir: Thanks! Works well for me now.

strangerkir (approved):
Looks good to me.

Dinamiko:
Merged into `feature/PCP-88-payment-token-vaulting`.

PR #144, opened 2021-04-08
Release 1.3.0

Dinamiko:
* Add - Client-side vaulting and allow WooCommerce Subscriptions product renewals through payment tokens. #134
* Add - Send transaction ids to woocommerce. #125
* Fix - Validate checkout form before sending request to PayPal #137
* Fix - Duplicate Invoice Id error. #143
* Fix - Unblock UI if Credit Card payment failed. #122
* Fix - Detected container element removed from DOM. #123
* Fix - Remove disabling credit for UK. #127
* Fix - Show WC message on account creating error. #136


Dinamiko:
Hi @jorgeatorres @mattallan 1.3.0 release is ready.

mattallan:
Hi @Dinamiko,
Is it possible to test Vault/Subscriptions support with a sandbox account?

Here's what I'm seeing when I try to test subscription purchases/renewals using Credit Card Processing:

1. When I enable the vault setting, purchase a subscription I see the following error in my logs:

```
WARNING [FORBIDDEN] No permission for the requested operation https://developer.paypal.com/docs/api/vault/v2/#error-PERMISSION_DENIED
```

2. I went into my Sandbox App and enabled vault under the Accept Payments options (https://d.pr/i/EcEuW1), reconnected my account, but now I'm seeing the following console log when trying to submit the checkout:

```
POST https://cors.api.sandbox.paypal.com/v2/checkout/orders/00L281812B584171P/validate-payment-method
[HTTP/1.1 404 Not Found 1043ms]
```

Any ideas?

Dinamiko:
Hi @mattallan, 

Thanks for testing, here is what we´ve found so far: 

Regarding the first one it looks like it fails because of incorrect authorization token when it calls the vault endpoint, we are not sure yet if it´s just because vaulting takes some time to be ready and the authorization token was created without vaulting capabilities at that time. I checked and it worked for me after sometime when auth token was automatically generated afterwards. We are currently checking this, once I´ve more info I´ll let you know.

Here are credit cards numbers just in case you´re not already using them for testing vaulting: https://developer.paypal.com/docs/limited-release/vault-payment-methods/vault-sdk/#test-cards
 
About the second issue it´s because after reconnecting it uses the old client token stored in the browser session, removing it or leave it expire fixes the issue:

We are also working on this issue to fix it ASAP.

Dinamiko:
Hi @mattallan,

We did a couple of commits on this release branch that could fix the aforementioned issues, the first one adds vaulting to api configuration and the second ensures that the client token is generated on each page rendering, that should fix problems like using an old one when reconnecting.

When you have time please give it a try and let us know, thanks.

mattallan:
Hi @Dinamiko thanks for the updates and the details on the vault test cards! 🙇 

I'm still having issues with testing Subscription support on test stores. I'm not sure if it's a bug or if I'm doing something wrongso I'm curious if my steps are different from yours.

Here's what I'm doing:
1. Purchase a subscription (i used 4111 card)
2. Test a renewal order by go into the Edit subscription page and use the Process Renewal action : https://d.pr/i/bfSeBw
3. Click update
4. Notice the subscription goes on-hold (no successful renewal payment) and the new renewal is pending payment

When I view the logs I see:

```
2021-04-15T05:18:44+00:00 WARNING Could not create order.
2021-04-15T05:18:44+00:00 ERROR An error occured while trying to renew the subscription for order 100541: Could not create order.
```

I tried to purchase another subscription using the saved card option dropwon: https://d.pr/i/XlXoax

The checkout immediately fails with:


There's also a typo in these error messages "Unfortunatly" should be "Unfortunately".

After seeing this message, I tried a different card and then I get "Could not create order.": https://d.pr/i/NmcaN1

I'm testing on both local dev site and on a public test site but cannot get anything working 😢 Any guidance appreciated!

Dinamiko:
Hi @mattallan,

Thanks for noticing the `Unfortunatly` typo, I´ve just did a commit with the fix.

Regarding `Could not create order.` issue while triggering a Subscription renewal I had the same from time to time while I was developing, in my case it was because the request timeout, can you please temporary increase the request timeout to confirm that it´s because of that:
```
add_filter('http_request_args', function($r) {
    $r['timeout'] = 15;
    return $r;
}, 100, 1);

add_action('http_api_curl', function($handle) {
    curl_setopt( $handle, CURLOPT_CONNECTTIMEOUT, 15 );
    curl_setopt( $handle, CURLOPT_TIMEOUT, 15 );
}, 100, 1);
```

About not supported card error, it looks like not all card numbers works from that page, in my case the ones that worked for me were MasterCard and Visa numbers.

jorgeatorres:
Hi @Dinamiko!

Some issues I've found:

1. Similarly to @mattallan, I haven't been able to test subscription renewals. I have enabled vaulting for my sandbox app and also on the plugin settings (both for PayPal Checkout and Credit Card Processing) and reconnected through onboarding before testing. While purchasing a subscription, I didn't notice anything different, and when trying to renew (via "Process Renewal" on the admin) the subscription status goes to "On Hold" and the renewal order remains "Pending Payment". The logs show this:

   > 2021-04-15T19:05:09+00:00 INFO Start moneytransfer for order 207
   > 2021-04-15T19:05:10+00:00 WARNING [INTERNAL_SERVER_ERROR] An internal server error has occurred. https://developer.paypal.com/docs/api/vault/v2/#error-INTERNAL_SERVER_ERROR
   > 2021-04-15T19:05:10+00:00 ERROR No payment token found for customer 1
   > 2021-04-15T19:05:10+00:00 INFO Moneytransfer for order 207 is completed.

   It seems to me, @mattallan managed to at least get this partially working because my error logs don't even have the "An error occured while trying to renew the subscription for order..." part. It might be that my setup is incorrect somehow. Do you know why this might be happening? Am I missing something in my setup?
3. Related to the above, if the renewal didn't work (in my case an error is even logged), would it be possible to tell the admin? With the subscription silently going to on-hold and a 'pending payment' renewal order being generated it looks as if you were dealing with a "Manual Payment" subscription and not a **failed** renewal.
2. On the settings screen, help text under titles is now bold and in a smaller font size than in 1.2.x:
   settings
   This apparently is because 1.3.0 is using `<th>` instead of `<td>`, but it looks out of place for a WP settings screen. Also, the font size under the main header ("PayPal Checkout") is _smaller_ than under the other subtitles (this also happened in 1.2.x, btw).
3. When looking at an order's detail and hovering over the "?" icon next to the gateway name, something akin to a "shortcode" is displayed instead of the gateway ID (happens for both Checkout and Credit Card Processing):
   Screen Shot 2021-04-15 at 13 53 08
4. When not connected, the Credit messaging vs Vaulting notice should not appear:
   Screen Shot 2021-04-15 at 13 57 31

Dinamiko:
Hi @jorgeatorres,

1. We also had `INTERNAL_SERVER_ERROR` yesterday in some other endpoints, but then it worked after some time. It could be that something was not working on Sandbox at the time you tested yesterday. So when you have time, can you please give it another try to see if the issue persist or not? thanks.

Not related to the above but important for testing is that we increased the request timeout to 15-20 seconds in our dev environments while testing on Sandbox, as it sometimes failed because of request timeout reached.

2. We are currently working on improving the logs in this part, regarding telling the admin about it, do you mean just displaying an admin message or do you mean also doing something else like changing the order status or similar? thanks.

3. Done: https://github.com/woocommerce/woocommerce-paypal-payments/pull/144/commits/e3cb93f632e5419a69e6e134d6e6ff7a9adbd31a

4. We are currently working on this, once fixed I´ll let you know.

5. Done: https://github.com/woocommerce/woocommerce-paypal-payments/pull/144/commits/ced4336c331ab06cf75206edf2602fa671161449

Dinamiko:
@jorgeatorres, regarding point 4 I disabled the plugin and then enabled **PayPal Standard** payment, I created a subscription using that payment, I can see the same behavior in that field, it displays the gateway ID wrapped in square brackets., in this case it displays `paypal`:

jorgeatorres:
Hi @Dinamiko,

Thanks for your response and for those quick fixes.

(1) I'm going to test again and will let you know what I find.

(2) I was, in particular, referring to the situation I experience while testing: an internal error prevented the payment token from being obtained, but as an admin all I saw was a "Subscription updated" notice, which gives the wrong impression: it looks as if everything went fine when in fact the opposite is true and you now have a subscription "on hold" with a pending payment renewal order. The latter is the correct behavior, but communicating to the admin that things didn't go as planned would be a valuable addition.

(4) Fair enough. I don't know why it looks like a "shortcode" but if it's not specific to PayPal Payments, no need to worry about it.

Not related to the above, have you had the chance to look [at this](https://github.com/woocommerce/woocommerce-paypal-payments/pull/120#issuecomment-819704138)?

mattallan:
@Dinamiko I'm still having trouble testing subscription renewals on both with Smart Buttons and Card Processing.

I'm still seeing the following errors in logs:

```
2021-04-20T05:06:20+00:00 INFO Start moneytransfer for order 100604
2021-04-20T05:06:26+00:00 WARNING Could not create order.
2021-04-20T05:06:26+00:00 ERROR An error occured while trying to renew the subscription for order 100604: Could not create order.
```

I tried creating a new sandbox account and app but still not having any luck with this so I've reached out to the PayPal team to see if they can help.

---

What's the correct way of testing the saved card functionality on the PayPal Card Processing form?

Here's what I see on the checkout: https://d.pr/i/avrfGd

When I select a card from this drop down and click "Place order" I see the following error notice:


I'm assuming the card form shouldn't be empty? Or should be hidden when I've selected a saved card.

Dinamiko:
Hi @mattallan,

I went able to reproduce the not supported credit card error, it was because it was not using the correct flow. I´ve just did a commit with the fix, it should work now. You tested it correclty, to use a saved credit card just select a saved one and then click "Place order" button.

Regarding subscription renewals error message, I´ve added now the error response to it, so it should now display more information about the concrete error.

Important before continuing testing, please do a git pull and then compile the JS assets to be sure you have the latest changes, thanks.

mattallan:
Thanks @Dinamiko for the latest fixes! I've been able to start successfully testing this today:

**Credit card processing**
- [x] Regular product purchase
- [x] Subscription purchase with vaulting enabled
- [x] Subscription purchase with vaulting disabled (confirmed the payment method isn't loaded)
- [x] Process subscription renewal
- [x] Mixed checkout purchase
- [x] Save a new card
- [x] Use saved card
- [x] process renewal on subscription paid with saved card
- [ ] Changing payment methods on the subscription (:x: when i try testing changing payment methods, the button disappears on the page. Here's a recording: https://d.pr/v/owuiaC)
- [ ] Managing saved cards (i.e. edit/delete) :x:
   - I couldn't find any way to edit/delete saved cards. Is this not supported yet and is there any plans to do so? For some context. In other gateways like stripe and square, there's a "Payment Methods" menu item on the My Account page and lists all the customers saved cards and allows them to remove ([screenshot](https://d.pr/i/bOKbJo))

While testing I noticed this error was logged:

```
21-Apr-2021 05:30:27 UTC] PHP Notice:  Undefined index: ppcp-credit-card-gateway in /Users/matt/local/woo/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Checkout/class-disablegateways.php on line 83
```

Unfortunately, I only noticed this in the middle of testing and I wasn't able to figure out what steps I took or what I did to reproduce this error but I thought I'd report it anyway 😞 

**Smart buttons**
- I haven't been able to test this yet and still running into issues (not sure if it's the code or on PayPal's side)

   - With vaulting and save paypal account enable, when I click on Debit and Credit Card button the checkout, the PayPal pop-up always asks me to create a new account.
   - When I try to use an existing sandbox customer email it tells me to log in, then it redirects me back to creating a new account ([recording](https://d.pr/v/2L9Ejt)).
   - When I try use a unique email the PayPal pop-up closes but nothing happens on the checkout

- While trying to test PayPal Checkout with vaulting disabled, I noticed the script args were still loading with `&vault=true`. I realised this is because I still had vaulting enabled under the disabled Card Processing settings (https://d.pr/i/VvKgYb). I think we should update `vault_settings_enabled()` to check if DCC is enabled first then check `dcc_vault_enabled`

Dinamiko:
Hi @mattallan, thanks for testing!

Here is what I have so far:

**Credit card processing**

Changing payment methods on the subscription, I disabled the switch of Place order button on that page, so now it should not hide the button.

Managing saved cards was not planned for this first vaulting iteration but will be great to have, I take note of this, thanks.

**Smart buttons**

Debit and Credit Card button on the checkout, I tried (logged in and logged out) but not able to reproduce the issue at the moment, still trying it, once I´ve more info I´ll let you know.

PayPal Checkout with vaulting disabled, I´ve added the check for dcc enabled first as you proposed, it should disable vaulting on PayPal Checkout when dcc is disabled.

jorgeatorres:
H @Dinamiko,

Once again, thanks for all your hard work on this. I've reported some other issues in #147. I didn't add those here to reduce clutter and confusion.

Dinamiko:
Hi @jorgeatorres @mattallan,  

I´ve added latest fixes to this branch and replied in #147 issue, if is there anything else pending from our side please let me know, thanks.

jorgeatorres:
Hi @Dinamiko,

Things are looking good. Some minor things I noticed while testing:

- It'd be good to make the "you must enable vaulting on your account" on the settings screen a link to the documentation on docs.woocommerce.com. The exact link would be this: https://docs.woocommerce.com/document/woocommerce-paypal-payments/#enable-vaulting-on-your-live-account.
- As you yourself have noticed, sometimes it takes a while to process renewals (or even the initial purchase). I've had success with your suggestion of increasing the timeout. I'm not sure if this "slowness" only affects sandbox but maybe we should add this workaround directly in PayPal Payments and only for PayPal Payments requests, as this doesn't seem to happen with other gateways. How hard would that be?
- When a renewal fails (because there's no token, for instance) it is not clear why it fails unless you check the logs (which might be confusing to most merchants) and that's only if you have logging enabled. Stripe, for instance, adds a note to the subscription telling the merchant the payment failed:
 
   Screen Shot 2021-04-27 at 10 07 31

   Also, the order status is not "Pending Payment" but "Failed":

   Screen Shot 2021-04-27 at 10 11 36
   
   Arguably, this is clearer than what PPCP currently does. I'm not sure if we need to add a note telling the merchant exactly why things failed (something along the lines of "no payment token found for customer"), but at least indicating on the subscription notes that the payment failed and maybe using "Failed" as status instead of "Pending Payment" could help. I'm pinging @mattallan to see if he has a better idea.

mattallan:
Hi @Dinamiko, I've ran through another round of testing and I couldn't see any issues with the checkout flow for purchasing regular products and subscriptions as well as subscription renewal process.

I did find some issues with trying to change payment method on a subscription:

**Change payment method**

 - [ ] Changing the payment method on a subscription from Stripe to a PPCP saved card results in a "CANNOT BE ZERO OR NEGATIVE" error: https://d.pr/i/DnGNeO

To reproduce:
1. purchase a subscription with another payment gateway (I used stripe)
2. view the subscription via My Account > Subscriptions
3. click the "Change payment" button
4. Try change to a PPCP saved card
5. click "change payment method" button
5. Notice error (https://d.pr/i/DnGNeO)

 - [ ] Changing the payment method Stripe to PPCP (new card) results in a page refresh.

To reproduce:
1. purchase a subscription with another payment gateway (I used stripe)
2. view the subscription via My Account > Subscriptions
3. click the "Change payment" button
4. Fill in the credit card form using a new card
5. Notice payment refreshes and subscription has not been switched to PayPal Payments

**Subscription renewals always use the last saved card**

Using Card A to purchase one subscription and then Card B to purchase second subscription results in both subscriptions charging the Card B for all renewals. Subscriptions renewals should use the same card they were purchased with so I"m curious if this is planned to be implemented later on?

---

> indicating on the subscription notes that the payment failed and maybe using "Failed" as status instead of "Pending Payment" could help

I agree with this suggestion.

Looking at other gateways, Square marks the order as failed and adds a note like this:


I think something like this would be very helpful, especially if stores don't have logging enabled.

Another question I have is, is "Vaulting" a well-known term to store managers? When reading "To use vaulting features, you must enable vaulting on your account." doesn't explain what this setting is doing.

Maybe this could be updated to:
"Enable saved cards and subscription features on your store. To use vaulting features, you must enable vaulting on your account."

Dinamiko:
Hi @jorgeatorres,

> It'd be good to make the "you must enable vaulting on your account" on the settings screen a link to the documentation on docs.woocommerce.com. The exact link would be this: https://docs.woocommerce.com/document/woocommerce-paypal-payments/#enable-vaulting-on-your-live-account.

Done: https://github.com/woocommerce/woocommerce-paypal-payments/pull/144/commits/9db88cacb93825e1f8f4bcebe75c2ebf4bc427a9

> As you yourself have noticed, sometimes it takes a while to process renewals (or even the initial purchase). I've had success with your suggestion of increasing the timeout. I'm not sure if this "slowness" only affects sandbox but maybe we should add this workaround directly in PayPal Payments and only for PayPal Payments requests, as this doesn't seem to happen with other gateways. How hard would that be?

The recommendation to increase the timeout on Sandbox came from PayPal, we didn't get related issues in support at the moment so we assume that the timeout limitation is only for Sandbox.

> When a renewal fails (because there's no token, for instance) it is not clear why it fails unless you check the logs (which might be confusing to most merchants) and that's only if you have logging enabled. Stripe, for instance, adds a note to the subscription telling the merchant the payment failed:

Done: https://github.com/woocommerce/woocommerce-paypal-payments/pull/144/commits/17f1a912412697570652076d0062eef644d7b9af

Dinamiko:
Hi @mattallan, thanks for testing!

> Change payment method
> - Changing the payment method on a subscription from Stripe to a PPCP saved card results in a "CANNOT BE ZERO OR NEGATIVE"
> -  Changing the payment method Stripe to PPCP (new card) results in a page refresh.

We are able to reproduce the issue but we´ll need time to fix them, we´ve only implemented subscription renewals using payment tokens at the moment. i´m going to create a dedicated issue for this.

What we propose after this release is to check the Subscription plugin integration and create issues to be sure both plugins works smoothly  together.

> Subscription renewals always use the last saved card
> Using Card A to purchase one subscription and then Card B to purchase second subscription results in both subscriptions charging the Card B for all renewals. Subscriptions renewals should use the same card they were purchased with so I"m curious if this is planned to be implemented later on?

We have planed to allow customer to select a payment as part of managing saved cards (edit/delete saved payment tokens)  discussed previously that will we included in the next release. 

> Another question I have is, is "Vaulting" a well-known term to store managers? When reading "To use vaulting features, you must enable vaulting on your account." doesn't explain what this setting is doing.
> Maybe this could be updated to:
> "Enable saved cards and subscription features on your store. To use vaulting features, you must enable vaulting on your account."

We´ve not added it at the moment because there is already the tooltip explaining what the setting is doing, we can add it is just that I´m not sure if we have to because is already there in the tooltip, just let us know, thanks:
vaulting-tooltip

PR #154, opened 2021-05-03
Filter PayPal-supported language codes

Dinamiko:
### Description
When selecting a language in WordPress like German [formal] or Norsk [Bokmål] banners and buttons do not appear on the selected pages and also there will be an error after trying to buy a product on the checkout page.

### Steps to test:
1. Select German [formal] in WordPress general settings.
2. Enable all banners and buttons in the plugin settings and save it.
3. Check the product pages.
4. Buy a product.


modules/ppcp-button/src/Assets/class-smartbutton.php
```diff
@@ -726,7 +726,7 @@ private function url(): string {
 		$params = array(
 			'client-id'        => $this->client_id,
 			'currency'         => get_woocommerce_currency(),
-			'locale'           => get_user_locale(),
+			'locale'           => $this->valid_locale_code(),
```
strangerkir: I'd suggest not pass the `locale` at all, as PayPal detects it automatically, and [their advice](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#query-parameters) (the `locale` parameter description in the table below) is to not use this parameter unless it's really needed.

But if we really need this, then it maybe makes sense to map possible WP locales to the ones accepted by PayPal. For example, when the Bulgarian language is set for the site, then the WP locale is `bg_BG`, but [PayPal accepts](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#locale) `en_BG` only. A similar problem appears for the Belarus language, and should be for some others I haven't checked. I tested it, and in my case when the locale is not accepted by PayPal, the JS SDK script is not loaded. Instead, I've had HTTP error code 400.

PCP-138-locale-issue

modules/ppcp-api-client/src/Repository/class-applicationcontextrepository.php
```diff
@@ -59,4 +59,20 @@ public function current_context(
...
+	protected function valid_bcp47_code() {
+		$locale = str_replace( '_', '-', get_user_locale() );
+		$parts  = explode( '-', $locale );
+
+		if ( count( $parts ) < 3 ) {
+			return $locale;
+		}
+
+		return substr( $locale, 0, strrpos( $locale, '-' ) );
+	}
```
strangerkir: I have some doubts about this. I believe this will work in most cases but I expect it fail when `get_user_locale()` returns locale containing 3 letters with no delimiter (like `bel` when Belarusian site language is chosen,  also `ceb` for Cebuano, `ary`, `dsb`,  etc.).

PayPal documentation [states](https://developer.paypal.com/docs/api/orders/v2/#definition-order_application_context) they use the `^[a-z]{2}(?:-[A-Z][a-z]{3})?(?:-(?:[A-Z]{2}))?$` regex to validate the language. I tested a few of possible WP locales and it worked for the `xx-XX` format (like `de-DE`), also for `xx` format (like `ar`), but not for `xxx` format (like the `bel` and others I provided above). This function will return three-letters code as is (which is valid BCP 47, if I understand it correctly). But it doesn't meet the regex from PayPal docs provided above. 

I have no good suggestion about this issue. Probably, it would make sense to use that regex to validate the result and fallback to default if validation wasn't passed.

strangerkir (changes requested):
Maybe some changes are needed.

Dinamiko:
Hi @strangerkir, 

I´ve removed `locale` on smart button, also added validation and fallback for bcp47, when you have time please take a look, thanks.

strangerkir (approved):
Looks good to me :)

PR #160, opened 2021-05-10
Add woocommerce inbox variant

moon0326:
Add WooCommerce inbox variant option.

#Changes proposed in this Pull Request:

P2: paJDYF-1uJ-p2
Fixes https://github.com/Automattic/woocommerce.com/issues/10117

Description: This PR adds a new option `woocommerce_inbox_variant_assignment ` for the Remote Inbox Notification on plugin activation or upgrade.

* This PR does not clean up `woocommerce_inbox_variant_assignment ` option on uninstallation so that we don't create a different value in case user decides to install the plugin again.  

* I'm calling `woocommerce_inbox_variant_assignment` method in the `plugins_loaded` hook since I wasn't able to find a way to trigger a method when the plugin gets upgraded. It's not ideal, but performance should be ok since the option is autoloaded. 

# Testing instructions

**Testing Activation**
1. Ensure `woocommerce_inbox_variant_assignment` option does not exist. If it does, delete it before testing.
2. Deactivate the plugin then activate it again
3. `woocommerce_inbox_variant_assignment` option should be created.
4. Repeat step 2 and ensure that the value of `woocommerce_inbox_variant_assignment` does not change.


woocommerce-paypal-payments.php
```diff
@@ -140,3 +141,19 @@ function( $links ) {
...
+
+
+/**
+ * Add woocommerce_inbox_variant for the Remote Inbox Notification.
+ *
+ * P2 post can be found at https://wp.me/paJDYF-1uJ.
+ */
+if ( ! function_exists( 'add_woocommerce_inbox_variant' ) ) {
+	function add_woocommerce_inbox_variant() {
+		$config_name = 'woocommerce_inbox_variant';
```
becdetat: This should be `$option` to match `get_option` and `update_option`'s signatures

becdetat (approved):
Looks great. I just had one comment but I'll pre-approve.

woocommerce-paypal-payments.php
moon0326: Updated in https://github.com/woocommerce/woocommerce-paypal-payments/pull/160/commits/aac19bb8a1173e2e5fa736ddbad5e9077b296dc7

psealock:
https://github.com/woocommerce/woocommerce-paypal-payments/pull/160/commits/ae12b285a35547aed04ffa0f96448728a6a76939 changes the option name to `woocommerce_inbox_variant_binary` and its variant assignment to 1 or 2 because the experiment parameters have changed. Since the original option name is already out in the wild in WC Subscriptions 3.1.1, a new one is required.

psealock:
Upon further thought, https://github.com/woocommerce/woocommerce-paypal-payments/pull/160/commits/67af7c932bde6fb1b46d6b8e019c403dac1e3a22 makes the variants 1-12 for greater possibilities of variant numbers: 2, 3, 4, 6, or 12

LCmry:
Hi @psealock the option is `woocommerce_inbox_variant_assignment` now. Is it supposed to be `binary` or `assignment`?

psealock:
>Is it supposed to be binary or assignment?

`assigment` is correct. There was some back and forth but the dust has settled on `woocommerce_inbox_variant_assignment`

LCmry (approved):

psealock:
Seeing as 5.4 of WC is due to be released tomorrow and has the same code, I will just go ahead and close this PR. Thanks for the reviews.

Issue #156, opened 2021-05-03
Subscriptions aren't renewed using the same source used for purchasing 


jorgeatorres:
From [#144](https://github.com/woocommerce/woocommerce-paypal-payments/pull/144#issuecomment-828155856):

> #### Subscription renewals always use the last saved card
> 
>Using Card A to purchase one subscription and then Card B to purchase second subscription results in both subscriptions charging the Card B for all renewals. Subscriptions renewals should use the same card they were purchased with so I"m curious if this is > planned to be implemented later on?

jorgeatorres:
Fixed in #161.

PR #161, opened 2021-05-17
Improve Subscription plugin support

### LINKED ISSUE #155: Changing the payment method on a subscription results in error
From [#144](https://github.com/woocommerce/woocommerce-paypal-payments/pull/144#issuecomment-828155856):

>  #### Changing the payment method on a subscription from Stripe to a PPCP saved card results in a "CANNOT BE ZERO OR NEGATIVE" error
>  https://d.pr/i/DnGNeO
> 
> To reproduce:
> 1. purchase a subscription with another payment gateway (I used stripe)
> 2. view the subscription via My Account > Subscriptions
> 3. click the "Change payment" button
> 4. Try change to a PPCP saved card
> 5. click "change payment method" button
> 5. Notice error (https://d.pr/i/DnGNeO)
> 
>  #### Changing the payment method Stripe to PPCP (new card) results in a page refresh.
> 
> To reproduce:
> 1. purchase a subscription with another payment gateway (I used stripe)
> 2. view the subscription via My Account > Subscriptions
> 3. click the "Change payment" button
> 4. Fill in the credit card form using a new card
> 5. Notice payment refreshes and subscription has not been switched to PayPal Payments

Dinamiko:
Fixes #155, #156.

**Note**: Currently it only supports change payment to an existing saved payment, the reason is because it´s currently not possible to vault via JavaScript SDK without purchase, as soon as this feature is added we´ll add it here as well.

Scenario **renewal subscription uses the correct payment**
**Given** Vault is enabled
**And** I purchase a Subscription product saving Credit card payment 
**And** I purchase another Subscription product using another credit card
**When** I renew the first Subscription product
**Then** it should use the same credit card used in the purchase

Scenario **change subscription payment from another payment to an existing saved credit card**
**Given** Vault is enabled
**And** there is a Subscription product using anothert payment (like PayPal Standard or Stripe)
**And** I am on Change payment method screen
**And** I select an existing saved credit card
**When** click "Change payment method" button
**Then** I should see payment updated correctly
**And** I can renew the Subscription correctly


modules/ppcp-subscription/src/class-renewalhandler.php
```diff
@@ -203,6 +203,19 @@ private function get_token_for_customer( \WC_Customer $customer, \WC_Order $wc_o
 			);
 		}
 
+		$subscription = wcs_get_subscription( $wc_order->get_meta( '_subscription_renewal' ) );
```
strangerkir: Is there any chance this may be used incorrectly somehow when this function is not defined? I'd suggest checking if this function exists if unexpected usage is possible.

strangerkir (approved):
Looks good to me.

PR #163, opened 2021-05-18
Use order_number instead of internal id when creating invoice Id

### LINKED ISSUE #162: Does not show the correct invoice number
### Describe the bug
The user facing order ID can differ from the internally used Post ID. Its the difference between `\WC_Order::get_order_number()` and `\WC_Order::get_id()`. For the invoice number, the plugin sends to PayPal the prefix + `get_id()`, which can result in the following unexpected behavior:
image (8)

### To reproduce
1. Add this to a mu-plugin or somewhere `add_filter('woocommerce_order_number', function($id) { return '9900000' . $id; });`
2. Create a order through PayPal
3. See the invoice number in PayPal

### Expected behavior
I would expect the default behavior of the plugin to reflect the order number in the invoice no. As this might be up to debate (invoice numbers are not necessarily identical with order numbers), additionally I would propose a filter for developers to alter this data before it gets send to PayPal.

websupporter:
**Issue**: #162

---

### Description
Uses the order number instead of the internal post id when creating the invoice number

### Steps to test:
See #162

### Filter
This PR also includes a filter. Not sure if this is wanted and if so, about naming.

Closes #162 .

### DIFF
--- a/modules/ppcp-api-client/src/Factory/class-purchaseunitfactory.php
+++ b/modules/ppcp-api-client/src/Factory/class-purchaseunitfactory.php
@@ -135,7 +135,7 @@ class PurchaseUnitFactory {
 			$soft_descriptor
 		);
 		return apply_filters(
-			'woocommerce-paypal-payments.purchase-unit.from-wc-order',
+			'woocommerce_paypal_payments_purchase_unit_from_wc_order',
 			$purchase_unit,
 			$order
 		);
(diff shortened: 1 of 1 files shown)


Dinamiko:
Hi @websupporter we´ve tested it and it works like a charm, I´ve renamed the filter name to use underscores because of phpcs warning.

Dinamiko (approved):

PR #150, opened 2021-04-29
Disable vault setting if vaulting feature is not available

Dinamiko:
Fixes #151

PR #166, opened 2021-05-24
UI/UX for vaulting settings

### LINKED ISSUE #157: UI/UX for vaulting settings
- As suggested in [#144](https://github.com/woocommerce/woocommerce-paypal-payments/pull/144#issuecomment-828155856), maybe we should do a better job at explaining what vaulting does, as it might be too technical for most merchants. The tooltip content is ok, but it doesn't work on mobile and it's easy to miss. Maybe the description should be moved next to the checkbox itself:

   > Another question I have is, is "Vaulting" a well-known term to store managers? When reading "To use vaulting features, you must enable vaulting on your account." doesn't explain what this setting is doing.
   > Maybe this could be updated to: "Enable saved cards and subscription features on your store. To use vaulting features, you must enable vaulting on your account."

- Given vaulting and Pay Later messaging are incompatible, maybe instead of a notice (or in addition to it) we should indicate next to the relevant settings why the setting is disabled. If you miss the notice, and scroll down, you'll see settings that are not available for no 'apparent' reason.
- Similarly to the above, to enable vaulting you first have to uncheck _all_ messaging settings (otherwise the vaulting checkbox is disabled). Maybe a simpler UX would be that clicking the vaulting checkbox (while having at least one messaging setting on) asks the customer whether they want to disable all messaging, and automatically handle unchecking of those checkboxes if the answer is affirmative.

Dinamiko:
Fixes #157.

**Note**: Automatically unchecking Pay Later checkboxes when Vault setting is checked is not implemented because Vault setting is also disabled when vaulting features are not enabled in PayPal (see #150), independently of how many Pay Later checkboxes are checked.

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -1807,7 +1807,6 @@ return array(
 			unset( $fields['disable_funding']['options']['card'] );
 		}
 
-		$dcc_applies = $container->get( 'api.helpers.dccapplies' );
 		/**
 		 * Depending on your store location, some credit cards can't be used.
 		 * Here, we filter them out.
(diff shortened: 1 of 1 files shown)


strangerkir (approved):
Looks good to me.

PR #146, opened 2021-04-20
Fix/pcp 130 fix cc fields

strangerkir:
**Issue**:  [PCP-130](https://inpsyde.atlassian.net/browse/PCP-130).

---

### Description

Fix Credit Card form fields placeholder and label.

`MM\YYYY` placeholder is replaced with the `MM\YY` to be the same as the label, and also as the Woocommerce and many other systems use by default.

The `Card code` label is replaced with the `CVV` to avoid customer confusion.

### Steps to test:

1. Go to the checkout page, select Credit Card as payment method.
2. Check the credit card fields, they should have labels and placeholders like these:

PCP-130-done

### Documentation

- [ ] This PR needs documentation (has the "Documentation" label).

* Fix - Credit card fields placeholder and label.

### Related issues
Closes #158

PR #170, opened 2021-06-01
Release 1.3.2

Dinamiko:
- Fix - Improve Subscription plugin support. #161
- Fix - Disable vault setting if vaulting feature is not available. #150
- Fix - Cast item get_quantity into int. #168
- Fix - Fix Credit Card form fields placeholder and label. #146
- Fix - Filter PayPal-supported language codes. #154
- Fix - Wrong order status for orders with contain only products which are both virtual and downloadable. #145
- Fix - Use order_number instead of internal id when creating invoice Id. #163
- Fix - Fix pay later messaging options. #141
- Fix - UI/UX for vaulting settings. #166

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,16 @@
 *** Changelog ***
 
+= 1.3.1 - TBD =
+* Fix - Improve Subscription plugin support. #161
+* Fix - Disable vault setting if vaulting feature is not available. #150
+* Fix - Cast item get_quantity into int. #168
+* Fix - Fix Credit Card form fields placeholder and label. #146
+* Fix - Filter PayPal-supported language codes. #154
+* Fix - Wrong order status for orders with contain only products which are both virtual and downloadable. #145
+* Fix - Use order_number instead of internal id when creating invoice Id. #163
+* Fix - Fix pay later messaging options. #141
+* Fix - UI/UX for vaulting settings. #166
+
 = 1.3.1 - 2021-04-30 =
 * Fix - Fix Credit Card fields for non logged-in users. #152
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "woocommerce-paypal-payments",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "description": "WooCommerce PayPal Payments",
   "repository": "https://github.com/woocommerce/woocommerce-paypal-payments",
   "license": "GPL-2.0",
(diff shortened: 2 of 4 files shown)


Dinamiko:
Hi @jorgeatorres @mattallan,

Release 1.3.2 PR is ready, it includes vaulting and Subscription improvements and fixes among other bug fixes. If you have any doubt please let me know, thanks.

Note that it does not include the release date yet on `changelog.txt`.

mattallan:
Hi @Dinamiko,

I tested this release today and didn't notice any major issues with this. Here's what I tested:

- [x] `* Fix - Improve Subscription plugin support`
  - [x] Scenario change subscription payment from another payment to an existing saved credit card
    - To test: Add two saved cards and tried to change the payment method from my account page
  - [x] Scenario renewal subscription uses the correct payment`
    - To test: After changing the payment method to a different card, trigger a renewal and log into PP Sandbox to check the last 4 card number (there's probably a simpler way of testing this)
 - [x] `* Fix - Credit Card form fields placeholder and label`
 - [ ] `* FIx - UI/UX for vaulting settings`
   - How do I test this? I tried creating a new Sandbox account with vaulting not enabled but when I connected it to my store, vaulting gets enabled by default. There were some SB accounts I've connected and the vaulting option in settings was disabled but I couldn't reproduce this recently.

---

I had a bit of trouble testing this release, but it turns out the test cards I was using were causing the problems...

Here's the only [two vault cards ](https://developer.paypal.com/docs/limited-release/vault-payment-methods/vault-sdk/#test-cards)that worked for me (and maybe for you too):
`4500600000000061`
`4012888888881881`

Dinamiko:
Hi @mattallan, thanks for testing!

> FIx - UI/UX for vaulting settings
> How do I test this? I tried creating a new Sandbox account with vaulting not enabled but when I connected it to my store, vaulting gets enabled by default. There were some SB accounts I've connected and the vaulting option in settings was disabled but I couldn't reproduce this recently.

We´ve added `VAULT` in api integration (https://github.com/woocommerce/woocommerce-paypal-payments/blob/cc1cd30bb52e101604ab9f6ad83aa27b2ee3f0d9/modules/ppcp-api-client/src/Repository/class-partnerreferralsdata.php#L109) so now vaulting will be automatically enabled on PayPal at onboarding time, this is a bit tricky to test because it will disable the setting only if it fails when enabling it on PayPal, so if the vaulting setting is enabled in your case, it means that is correctly enabled on PayPal otherwise it would be disabled.

> Here's the only two vault cards that worked for me (and maybe for you too):

Same, for me only MasterCard and Visa numbers from that Vaulting documentation page seems to work.

jorgeatorres:
Hi @Dinamiko!

This looks good. Thank you!
I believe we can finetune certain UI details related to the vaulting/Pay Later settings but I don't want to hold off the release because of that. I'll open a new issue so that we can discuss.

Once again, thank you for all the work that went into this release.

jorgeatorres (approved):
LGTM.

PR #171, opened 2021-06-04
Fix phpunit tests

Dinamiko:
This PR fixes unit tests on `trunk` branch.


XedinUnknown (approved):
LGTM, but I'm a little cautious about the implementation changes: plugin logic has been modified, not just tests. So if you think this is good, then I'm fine with it.

PR #179, opened 2021-06-16
Remove `merchant-id` query parameter in JSSDK

Dinamiko:
The `merchant-id` query parameter in the JSSDK is designed for doing third party calls. This parameter is extraneous for first-party calls when the Client ID is owned by the merchant. This has caused some issues previously, so should be removed.


strangerkir (approved):
Looks good to me.

I'd suggest removing `payee_repository` from class properties and constructor dependencies since it's not used anymore.

PR #182, opened 2021-06-18
Add `enable-funding=paylater` as a query parameter when enabled

Dinamiko:
This PR adds `enable-funding=paylater` as a query parameter when messaging settings are enabled. It also displays a Pay Later button disclaimer in messaging settings.


strangerkir (approved):
Looks good to me.

PR #184, opened 2021-06-21
Add 30 seconds timeout by default to all requests

Dinamiko:
In order to prevent timeout errors due to default WordPress 5 seconds, this PR adds 30 seconds timeout by default to all its requests to PayPal.


strangerkir (approved):
Looks good to me.

PR #196, opened 2021-07-14
Automatically import PayPal Checkout settings during initial setup

jorgeatorres:
**Do not review yet. This PR is part of the work we're doing to smooth migration from PayPal Standard/Checkout to PayPal Payments.**

### Description

This PR adds support for populating PayPal Payments settings with compatible settings taken from PayPal Checkout if PayPal Checkout is in use.

Right now, this only happens the first time PPCP settings are initialized. As such, the change doesn't affect current users of PayPal Payments.

For discussing: we might want to revisit the above condition and let the importer run for users that haven't completed onboarding, even if the PayPal Payments database option already exists.

### Steps to test:

1. Install PayPal Checkout and configure it however you like.
1. Delete `woocommerce-ppcp-settings` from the `wp_options` table to simulate a "clean install" of PayPal Payments.
1. Go to WooCommerce > Payments > PayPal.
1. Enter your credentials.
1. Most settings with a corresponding PayPal Checkout one, should have the value from PayPal Checkout. Specifically: you could confirm with basic settings such as "Brand Name", "Prefix" or "Logging" but also with the different context specific settings (buttons, Pay Later, etc.).

### Documentation

- [X] This PR needs documentation (has the "Documentation" label).

Automatically import PayPal Checkout settings when initializing PayPal Payments settings.

### DIFF
--- a/modules/ppcp-compat/extensions.php
+++ b/modules/ppcp-compat/extensions.php
@@ -1,0 +1,12 @@
+<?php
+/**
+ * The compatibility module extensions.
+ *
+ * @package WooCommerce\PayPalCommerce\Compat
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Compat;
+
+return array();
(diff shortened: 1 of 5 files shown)


jorgeatorres:
I didn't realize renaming a branch would automatically close the PR 😅 . Re-submitted as #198.

PR #198, opened 2021-07-15
[Legacy] Automatically import PayPal Checkout settings during initial setup

jorgeatorres:
### Description

This PR adds support for populating PayPal Payments settings with compatible settings taken from PayPal Checkout if PayPal Checkout is in use.

Right now, this only happens the first time PPCP settings are initialized. As such, the change doesn't affect current users of PayPal Payments.

For discussing: we might want to revisit the above condition and let the importer run for users that haven't completed onboarding, even if the PayPal Payments database option already exists.

### Steps to test:

1. Install PayPal Checkout and configure it however you like.
1. Delete `woocommerce-ppcp-settings` from the `wp_options` table to simulate a "clean install" of PayPal Payments.
1. Go to WooCommerce > Payments > PayPal.
1. Enter your credentials.
1. Most settings with a corresponding PayPal Checkout one, should have the value from PayPal Checkout. Specifically: you could confirm with basic settings such as "Brand Name", "Prefix" or "Logging" but also with the different context specific settings (buttons, Pay Later, etc.).

### Documentation

- [X] This PR needs documentation (has the "Documentation" label).

Automatically import PayPal Checkout settings when initializing PayPal Payments settings.

### DIFF
--- a/modules/ppcp-compat/services.php
+++ b/modules/ppcp-compat/services.php
@@ -30,4 +30,10 @@ return array(
 		return new PPEC\SubscriptionsHandler( $ppcp_renewal_handler, $gateway );
 	},
 
+	'compat.ppec.settings_importer'     => static function( $container ) : PPEC\SettingsImporter {
+		$settings = $container->get( 'wcgateway.settings' );
+
+		return new PPEC\SettingsImporter( $settings );
+	},
+
 );
--- a/modules/ppcp-compat/src/PPEC/class-settingsimporter.php
+++ b/modules/ppcp-compat/src/PPEC/class-settingsimporter.php
@@ -1,0 +1,235 @@
+<?php
+/**
+ * PayPal Checkout settings importer.
+ *
+ * @package WooCommerce\PayPalCommerce\Compat\PPEC
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Compat\PPEC;
+
+use WooCommerce\PayPalCommerce\WcGateway\Settings\Settings;
+
+/**
+ * Handles import of settings from PayPal Checkout into PayPal Payments.
+ */
+class SettingsImporter {
+
+	/**
+	 * PPCP settings.
+	 *
+	 * @var Settings
+	 */
+	private $ppcp_settings;
+
+	/**
+	 * PayPal Checkout database option.
+	 *
+	 * @var array
+	 */
+	private $ppec_settings;
+
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Settings $settings PPCP settings instance.
+	 */
+	public function __construct( Settings $settings ) {
+		$this->ppcp_settings = $settings;
+		$this->ppec_settings = (array) get_option( PPECHelper::PPEC_SETTINGS_OPTION_NAME, array() );
+	}
+
+	/**
+	 * Sets up WP hooks to import PayPal Checkout settings into PPCP when needed.
+	 *
+	 * @return void
+	 */
+	public function maybe_hook() {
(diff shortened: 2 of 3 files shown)


jorgeatorres:
This should be ready for review.

LCmry:
Hi @jorgeatorres did an initial run through. For the Pay Later/Credit Messaging on checkout a couple settings didn't transfer. Are these the right correspondences?
Credit Messaging Logo position: Right --> Pay Later Logo Position: Left
Credit Messaging Text Color: Monochrome --> Pay Later Text Color: Black

Otherwise it looked like it transferred where there were equivalencies and not where there weren't. For example, on the single product page pay later messaging layout, in PPEC I had "graphic" but in PPCP there is not option for that, the other option is "flex".

jorgeatorres:
Hi @lcmry!

Thanks for testing! I tried to reproduce your issue with the Credit Messaging settings for the Checkout section and couldn't reproduce it. Could you confirm a couple of things?

- Did you have "Configure settings specific to regular checkout" enabled in PPEC or were those the "global" defaults?
- Did you have "Enable PayPal Credit messages" on in PPEC?

I'm sorry if those are obvious questions, but the way the importer works, it doesn't import settings that depend on a disabled setting (even if they had a value in the database). For instance, if you make customizations to PayPal Credit messages for the checkout page but Credit messages is disabled at the time of the import (in PPEC), you'll get the defaults. Just want to confirm if that's what happened here.

LCmry:
Just noting here that the above issues mentioned turned out not to be issues. Testing as expected.

mattallan:
I'm a little lost with what I'm seeing.

The PPCP invoice prefix is being set as "decbad-" on import: https://d.pr/lLhpus
I have no idea where this string is coming from 🤔 It doesn't exist anywhere in my DB or code.

Ohhh okay I just found this issue is happening when PPEC is deactivated. If I activate PPEC and then test the importing of settings, everything looks good. But if I deactivate PPEC and test settings import it has this weird `decbad-` as the invoice prefix

I'm sure there's probably other issues but this was the most obvious.

Is it intended that stores have PPEC remain active for this import of settings to work?

jorgeatorres:
Hi @mattallan!

Thanks giving this a a try 😸 .

> The PPCP invoice prefix is being set as "decbad-" on import: https://d.pr/lLhpus
> I have no idea where this string is coming from 🤔 It doesn't exist anywhere in my DB or code.

When PPEC is not enabled, the settings importer doesn't do anything (which is also related to your other question). As such, the prefix option gets its default value, which is generated from an [md5 hash of your site's URL](https://github.com/woocommerce/woocommerce-paypal-payments/blob/78fa4a5437356c3ceb92b6d9f2dbcefecb2a6e75/modules/ppcp-wc-gateway/services.php#L682-L687). This is just how PayPal Payments generates the default value and has nothing to do with the settings importer or PPEC.

> Is it intended that stores have PPEC remain active for this import of settings to work?

My rationale behind this was that the sole presence of the PPEC option in the database didn't necessarily indicate that the gateway (PPEC) was "in use" before you installed PPCP (i.e. you were *migrating* to PPCP). It could've been left over from a previous installation of the plugin or something.
That said, I have nothing against changing this to always attempt to import the settings or use a different logic.

----

Let me know what your thoughts on the above are or if you find other issues while testing. Thanks again!

mattallan:
Thanks @jorgeatorres ! That makes sense.

I definitely understand your reasoning with only importing PPEC if the gateway is enabled/activated as we don't want to import old settings that were used years ago when the store had PPEC installed. I don't think many stores will do what I did anyway with deactivating PPEC before enabling/configuring PPCP so I think your line of thinking is fine :)

Merging

PR #202, opened 2021-07-21
Do not send decimals when currency does not support them

Dinamiko:
### Description
Currency Japananese Yen/JPY is not functional in PPP because it does not support decimals, see https://developer.paypal.com/docs/reports/reference/paypal-supported-currencies/

This PR ensures sending rounded values without decimals for currencies that does not allow decimals.

### Steps to test:
- Set webstore to JPY (Japanese Yen)
- Purchase item and try to pay at any step
- Getting the errors:
Something went wrong. Please try again or choose another payment source. (on basked PP payment)[UNPROCESSABLE_ENTITY] The requested action could not be performed, semantically incorrect, or failed business validation. https://developer.paypal.com/docs/api/orders/v2/#error-DECIMALS_NOT_SUPPORTED (on checkout PP payment)

### DIFF
--- a/modules/ppcp-api-client/src/Entity/class-amount.php
+++ b/modules/ppcp-api-client/src/Entity/class-amount.php
@@ -29,6 +29,13 @@ class Amount {
 	private $breakdown;
 
 	/**
+	 * Currencies that does not support decimals.
+	 *
+	 * @var array
+	 */
+	private $currencies_without_decimals = array( 'HUF', 'JPY', 'TWD' );
+
+	/**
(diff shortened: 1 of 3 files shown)


strangerkir (approved):
Looks good to me.

Is it needed to show a notice to users/store admins about the price will be rounded? I checked, and it looks like WooCommerce itself doesn't prevent using decimals for prices in Yen. I guess it may be useful to discuss it with product owners.
Woocommerce-Yen-With-Decimals

PR #214, opened 2021-08-04
Cast gateway description to string

Dinamiko:
When `woocommerce_gateway_description` filter receives `$description` as `null` it throws an error because private method `display_saved_paypal_payments` expects a string, this PR fixes this problem by casting `$description` as string when is passed along to the private method.

### DIFF
--- a/modules/ppcp-subscription/src/class-subscriptionmodule.php
+++ b/modules/ppcp-subscription/src/class-subscriptionmodule.php
@@ -80,7 +80,7 @@ class SubscriptionModule implements ModuleInterface {
 				$settings                 = $container->get( 'wcgateway.settings' );
 				$subscription_helper      = $container->get( 'subscription.helper' );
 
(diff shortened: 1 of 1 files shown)


modules/ppcp-subscription/src/class-subscriptionmodule.php
```diff
@@ -80,7 +80,7 @@ function ( $description, $id ) use ( $container ) {
 				$settings                 = $container->get( 'wcgateway.settings' );
 				$subscription_helper      = $container->get( 'subscription.helper' );
 
-				return $this->display_saved_paypal_payments( $settings, $id, $payment_token_repository, $description, $subscription_helper );
+				return $this->display_saved_paypal_payments( $settings, $id, $payment_token_repository, (string) $description, $subscription_helper );
```
strangerkir: I suggest casting to string also the second parameter, $id. Normally, it should be a string already, but as this value comes from the filter and can be modified by any other plugin, it's safer to cast it too.

strangerkir (approved):

PR #215, opened 2021-08-06
[Legacy] Upgrade inbox notes

jorgeatorres:
### Description

This PR adds an inbox note with upgrade messaging.

#### No Subscriptions messaging
1. Check out this branch.
1. Install and enable PayPal Express Checkout (if necessary).
1. Make sure there aren't any subscriptions on the site running via PayPal Express Checkout.
1. An inbox note inviting you to deactivate PayPal Express Checkout should appear (on WC > Orders > Inbox, for example).
1. Follow the link to deactivate PayPal Express Checkout (or do it via Plugins).
1. The inbox note should no longer appear.

#### Subscriptions messaging
1. Disable PayPal Payments (to prevent the "no subscriptions" note from re-appearing).
1. Clear note information from the database. SQL query: `DELETE FROM wp_wc_admin_notes WHERE name='ppcp-disable-ppxo-note'`.
1. Re-enable PayPal Express Checkout.
1. Either purchase a subscription using PayPal Express Checkout or create a manual subscription and set its payment method to PayPal Express Checkout.
1. Re-enable PayPal Payments.
1. You should see an inbox note inviting you to deactivate PayPal Express Checkout, but with a specific mention of subscriptions support.
1. Follow the link to deactivate PayPal Express Checkout (or do it via Plugins).
1. The inbox note should no longer appear.

Add messaging for PayPal Express Checkout users.

### DIFF
--- a/modules/ppcp-compat/src/PPEC/class-deactivatenote.php
+++ b/modules/ppcp-compat/src/PPEC/class-deactivatenote.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * Deactivate PayPal Express Checkout inbox note.
+ *
+ * @package WooCommerce\PayPalCommerce\Compat\PPEC
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Compat\PPEC;
+
+use Automattic\WooCommerce\Admin\Notes\Note;
+use Automattic\WooCommerce\Admin\Notes\Notes;
+use Automattic\WooCommerce\Admin\Notes\NoteTraits;
+
+/**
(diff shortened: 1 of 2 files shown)


LCmry:
Hi @jorgeatorres, thanks for the clear testing instructions. Everything tested as expected.
A couple notes:
- When a note is dismissed and the plugin is deactivated and reactivated, it does not come back. I think this is expected, but wanted to note.
- When Subscriptions has been activated, then deactivated, the note that details subscriptions shows, not the other. I also think this is probably not a problem, since there were subscriptions on the site.

PR #220, opened 2021-08-10
Add action when app container is built

### LINKED ISSUE #219: Feature: Add ability to utilize the builded Module Container
### Describe the Feature
It would be very useful, to implement a filter or action that exposes the builded modules container, like:

```php
apply_filters('woocommerce_paypal_payments_builded_container', $container);
```

If you have security/stability concerns, an only readable Container would make also huge sense, so you can just grab configured components out of the container, so there is no space for changing components that could result in Plugin errors.

strangerkir:
**Issue**: #219 

---

### Description
Add action to allow read access to the app container.

### Documentation

### Changelog entry
* Add - Action when the app container is built.

Closes #219.

### DIFF
--- a/woocommerce-paypal-payments.php
+++ b/woocommerce-paypal-payments.php
@@ -89,6 +89,7 @@ define( 'PPCP_FLAG_SUBSCRIPTION', true );
 			$proxy     = new ProxyContainer();
 			$provider  = new CompositeCachingServiceProvider( $providers );
 			$container = new CachingContainer( new DelegatingContainer( $provider ) );
+			apply_filters('woocommerce_paypal_payments_built_container', $container);
 			$proxy->setInnerContainer( $container );
 			foreach ( $modules as $module ) {
 				/* @var $module ModuleInterface module */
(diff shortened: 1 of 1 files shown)


woocommerce-paypal-payments.php
```diff
@@ -89,6 +89,7 @@ function() {
 			$proxy     = new ProxyContainer();
 			$provider  = new CompositeCachingServiceProvider( $providers );
 			$container = new CachingContainer( new DelegatingContainer( $provider ) );
+			$container = apply_filters( 'woocommerce_paypal_payments_built_container', $container );
 			$proxy->setInnerContainer( $container );
```
brianvarskonst: I would suggest use a Action (`do_action`) with the Container after Line 93 and pass the proxy Container, because this Container is already readonly. I would only allow to change the Container if you provide a Module, then you can subscribe to the modules hook, where you can change the Container via a Module.

brianvarskonst (changes requested):

woocommerce-paypal-payments.php
strangerkir: Ok, updated.

brianvarskonst (approved):
LGTM

Dinamiko (approved):

PR #222, opened 2021-08-11
WooCommerce orders left in “Pending Payment” after a decline

Dinamiko:
### Description
WooCommerce orders left in “Pending Payment” after a decline.

### Steps to test:
Use this filter to return a `INSTRUMENT_DECLINED` error in the response. 
```
        $args['headers']['PayPal-Mock-Response'] = '{"mock_application_codes": "INSTRUMENT_DECLINED"}';
    }
    return $args;
}, 10, 2);
```

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
@@ -141,6 +141,11 @@ trait ProcessPaymentTrait {
 			}
 		} catch ( PayPalApiException $error ) {
 			if ( $error->has_detail( 'INSTRUMENT_DECLINED' ) ) {
+				$wc_order->update_status(
+					'failed',
+					__( 'Instrument declined.', 'woocommerce-paypal-payments' )
+				);
+
 				$this->session_handler->increment_insufficient_funding_tries();
 				$host = $this->config->has( 'sandbox_on' ) && $this->config->get( 'sandbox_on' ) ?
 					'https://www.sandbox.paypal.com/' : 'https://www.paypal.com/';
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
```diff
@@ -141,6 +141,11 @@ public function process_payment( $order_id ) {
 			}
 		} catch ( PayPalApiException $error ) {
 			if ( $error->has_detail( 'INSTRUMENT_DECLINED' ) ) {
+				$wc_order->update_status(
+					'failed',
+					__( 'Instrument declined.', 'woocommerce-paypal-payments' )
+				);
+
```
strangerkir, line 148: This should work, but I have some doubts about it. 
The code below gives 3 tries, from what I can see, before suggesting the customer using another payment method. Does it make sense if the order is already marked as failed?
Another thing I'm not sure about is why it checks for the `INSTRUMENT_DECLINED` only? I know, it's out of the scope of this PR, but just want to make sure there will be no problems with it. It looks like the order will be marked as failed only in case of this error :thinking: Is it the expected result?

strangerkir (commented):
Looks ok, but I have a few questions to be absolutely sure.

modules/ppcp-wc-gateway/src/Gateway/class-processpaymenttrait.php
Dinamiko: Hi @strangerkir thanks for the feedback, 

Regarding retrying, if payment is accepted on second or third try order status changes from `failed` to `on-hold` first and then to `processing`, I´ve checked and it works like so now.

About the `INSTRUMENT_DECLINED` only, I have just added `failed` order status to the other two failing cases, thanks for the hint :)

strangerkir (approved):

PR #226, opened 2021-08-18
Run CI for PHP 7.1-7.4

AlexP11223:
Looks like we only needed to allow `phpunit` 7/8 for that, and remove one 7.2+ constant in a test.

For 8.0 there are more complex issues with deps: https://github.com/woocommerce/woocommerce-paypal-payments/runs/3361390659


Dinamiko (approved):
Looks good, I´ll merge it after 1.5.1 release, thanks.

PR #231, opened 2021-08-23
Do not send payee email

AlexP11223:
### Description

Do not include the merchant email in the `payee`  object when sending API requests. It seems to be not needed since we already send ID, and email causes validation issues in some cases.

I did that by modifying `Payee::to_array` to not include email when we have ID (is it possible to not have ID?), not sure if that's the best approach, I guess ideally we would want to decouple this serialization logic somehow (e.g. in case we want to have email for some cases).

### Steps to test:
1. Perform a checkout.
2. It seems like we cannot see the body of the request in the sandbox history... So besides adding some debug logging, the best thing we can do is probably just check the response of ajax `ppc-create-order` request in DevTools, inside `payee` it should have only id without email. Or ideally we could try finding out what was causing the original issue and check that it cannot be reproduced with this fix.

### Changelog entry

- Do not send the merchant email in API requests if we send the merchant ID.

### DIFF
--- a/modules/ppcp-api-client/src/Entity/class-payee.php
+++ b/modules/ppcp-api-client/src/Entity/class-payee.php
@@ -68,11 +68,11 @@ class Payee {
 	 * @return array
 	 */
 	public function to_array(): array {
-		$data = array(
-			'email_address' => $this->email(),
-		);
+		$data = array();
 		if ( $this->merchant_id ) {
 			$data['merchant_id'] = $this->merchant_id();
+		} else {
+			$data['email_address'] = $this->email();
 		}
 		return $data;
 	}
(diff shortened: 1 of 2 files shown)


XedinUnknown (approved):
This looks good to me. However, following the problem description in the PR, if you want to be able to control how serialization happens - then I believe the right way is to create a serializer object, and switch it in configuration whenever necessary. The serializer is also probably something to inject into the payee factory. If you still need to use `to_array()`, or otherwise keep the `Payee` responsible for serializing itself (which is not necessarily a good idea IMHO, but I don't know enough details to present an argument), then you can simply inject the new serializer into the `Payee`.

Dinamiko (approved):

Issue #87, opened 2020-11-12
Provide option for Credit Card to appear above PayPal in checkout


Chaithi:
### Describe the bug
Since the module is not broken into two methods, the user doesn't have control over which payment method (between PayPal and Credit Card) appears first in checkout. 

### Additional details
Provide toggle in the Credit Card Processing section to allow merchant to set Credit Cards first in checkout.

ethanclevenger91:
For some reason (likely user experience related), the plugin authors elected to toss out the default payment gateway configuration settings screens that come with a payment gateway, and instead tagged the credit card gateway within the main PayPal gateway.

If you toss out the screen check [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/1.5.1/modules/ppcp-wc-gateway/src/class-wcgatewaymodule.php#L218), the credit card payment gateway appears as expected in the list of gateways and can be re-ordered as expected, but the action button in the right-hand column goes to an unused screen.

A naive PR would do just that, but would need, at a minimum, a redirect set up from the default action URL that WooCommerce generates to the custom setting screen this plugin creates.

ethanclevenger91:
Same details essentially rehashed [here](https://wordpress.org/support/topic/checkout-page-payment-options-order/) on the WP.org support forums.

jorgeatorres:
Fixed in #236.

PR #232, opened 2021-08-24
Removes screen check for payment gateway.

ethanclevenger91:
**Issue**: #87 

---

### Description

Removes the screen check for adding the credit card payment gateway, allowing it to be reordered for display to end users just like any other payment gateway.

**However,** the default action button generated by WooCommerce in this interface goes to a settings page that this plugin doesn't utilize. Instead, the credit card gateway settings are currently configured _within_ the PayPal gateway settings. To say nothing of the merits of this strategy.

Looking for input on how to handle this. Simplest solution seems to introduce a redirect from the done-for-you settings screen to the sub-setting screen within the main PayPal gateway settings. It's already configured with an error-level notice to send folks to the main PayPal gateway settings screen if they haven't yet connected to PayPal.

Alternate solution would be to move the credit card gateway settings to the done-for-you settings screen from WooCommerce for the gateway.

### Steps to test:

1. Navigate to WooCommerce > Settings > Payments
2. See the PayPal Credit Card gateway is now in the list of available gateways

### Documentation

Exposes the credit card payment gateway on the default WooCommerce payment settings screen.

### DIFF
--- a/modules/ppcp-wc-gateway/src/class-wcgatewaymodule.php
+++ b/modules/ppcp-wc-gateway/src/class-wcgatewaymodule.php
@@ -206,16 +206,12 @@ class WcGatewayModule implements ModuleInterface {
 				$methods[]   = $container->get( 'wcgateway.paypal-gateway' );
 				$dcc_applies = $container->get( 'api.helpers.dccapplies' );
 
-				$screen = ! function_exists( 'get_current_screen' ) ? (object) array( 'id' => 'front' ) : get_current_screen();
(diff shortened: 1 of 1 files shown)


AlexP11223:
Done in #236

jorgeatorres:
Hi @ethanclevenger91,

Thanks for your PR! A fix for this was released with [version 1.6.0](https://github.com/woocommerce/woocommerce-paypal-payments/releases/tag/1.6.0). See #236 for the details.

PR #233, opened 2021-08-25
Fix Project Configuration

XedinUnknown:
This extracts the changes related to PSR-4 refactoring from #120.

Adds some missing configuration to the projects:

1. PHPStorm config.
2. Composer lockfile.
3. Git attributes for correct archiving and line endings.
4. Fix line endings that caused Docker problems.


AlexP11223:
`composer.lock` management for such projects sounds a bit confusing to me.
Do we need to commit it after creating it with the min PHP version we support (7.1)? What if someone commits with wrong version?
And not much benefits because in CI for each PHP version we would need `update` anyway.

XedinUnknown:
Not sure how this project is any different to most other projects we manage in that regard. The lockfile needs to reflect what is installed on the PHP version that the project is built and shipped on, like for everything else. Someone won't commit the wrong version, because the Docker setup includes the same version for everyone by default. If they do commit deps on the wrong version, then they will have made a mistake - like any other mistake.

Yep, due to the `update` it's not very beneficial for CI. But deps are also installed locally, and this is especially important when running `docker-compose build` for the first time. The main benefit, however, is that it eliminates a certain class of problem - the "Which deps did you install?" kind - and the lockfile allows us to always be on the same page with this. From this PoV, it also relates to CI, because if a CI build passes, we can confidently say "Our stuff works under _these_ conditions", and the lockfile should be part of those conditions.

I cannot think of a reason not to include the lockfile in every project that uses Composer.

Dinamiko:
@XedinUnknown can we just not include any PHPStorm configuration at all? from other projects where this is added I see commits with plenty of `.idea` files on it, that makes PRs confusing. Also if we follow the same logic why not adding configuration for other IDEs like VSCode for example?

Dinamiko:
@jorgeatorres before merging this PR I need your feedback regarding including `composer.lock` file in the repo, do you know if there is any reason to not include it from your side? thanks.

AlexP11223:
> Also if we follow the same logic why not adding configuration for other IDEs like VSCode for example?

Because we use mainly PHPStorm.

Though I personally don't care whether it is included or not. I usually run commands via terminal (`yarn docker:.....` in this project) and PHPStorm auto-detects many things, like the min PHP version from `composer.json`.

One useful thing I remember in some projects is excluding some auto-generated files from the project (e.g. the duplicated modules dir in Zettle), but here it does not seem to be needed for now.

XedinUnknown:
@Dinamiko, we can, but why shouln't we? It's very helpful. Out of the box: Composer, PHP, Psalm integration, DB client, Git client, PHPUnit integration, etc etc. PHPStorm is de facto company stanadard, and most of our projects (all new ones, AFAI) include PHPStorm configs, since Docker containers give us the possibility of knowing all of the settings beforehand.

A PR will contain many changes to files in `.idea` if a lot of things in PHPStorm config change - which happens in PRs like this one, and not much in others. This is IMHO quite literally part of project configuration, and has (or _should_ have) no effect on anything distributed - because ae5fd9e excludes `.idea` from Github's auto dist archive, and the build script should remove it too - same as e.g. `docker`.

Confirm please.

Dinamiko:
@XedinUnknown I understand, but I would prefer to not include them for now, thanks.

XedinUnknown:
@Dinamiko, I've removed PHPStorm configs from this PR. Now, PHPStorm regularly loses track of project configuration, and every time this happens I have to re-configure it.

I've also taken the liberty to cherry-pick some changes from the later branches, because they were also related to project configuration - please see.

jorgeatorres:
> @jorgeatorres before merging this PR I need your feedback regarding including composer.lock file in the repo, do you know if there is any reason to not include it from your side? thanks.

Hi @Dinamiko! No, feel free to remove it from gitignore.

Dinamiko (commented):
@XedinUnknown why are these 5 png files included in this PR?
pngs

XedinUnknown:
@Dinamiko, yep. It is very strange indeed. This is, believe it or not, line-endings that got scrambled due to `git checkout` and associated configuration on Windows. If we assume that PNGs for some reason have line endings in them, then this corrects them, and git will not complain again. See 559f843.

_Update_: apparently, this breaks the images. Reverted in 758ad55.

Dinamiko (approved):

phpcs.xml.dist
```diff
@@ -19,10 +19,21 @@
 		<exclude-pattern>tests/</exclude-pattern>
 	</rule>
 
+	<rule ref="WordPress">
+		<exclude name="WordPress.Files.FileName.NotHyphenatedLowercase" />
+		<exclude name="WordPress.Files.FileName.InvalidClassFileName" />
```
AlexP11223, line 24: Can we also exclude weird stuff like `WordPress.PHP.DisallowShortTernary`? :D 

Looks like WP added this rule because of some weird reasons about not being useful in their use-cases and trying not to scare beginners. https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

Dinamiko: @AlexP11223 agreed, disallow short ternary does not make much sense now that we are on PHP 7.1.

XedinUnknown: ✔️ b8f35c8

Dinamiko (approved):

PR #235, opened 2021-08-26
Refactor for PSR-4

XedinUnknown:
⚠️ Requires #233.

This extracts the changes related to PSR-4 refactoring from #120.

- Add package configuration to every module, and includes them all in main project - a more distributed modular system.
- Fix some class names.
- Project now fully complies with PSR-4.
- Update some constraints, and upgrade dependencies.
- Platform requirements now defined - requires PHP 7.1.


.github/workflows/php.yml
```diff
@@ -1,27 +1,25 @@
...
     - uses: actions/checkout@v2
 
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
-        php-version: ${{ matrix.php-versions }}
+          php-version: ${{ matrix.php-versions }}
+
+    - name: Packagist.com Auth
+      run: composer config --global --auth http-basic.repo.packagist.com token ${{ secrets.PACKAGIST_TOKEN }}
```
AlexP11223: Do we need this here? I think all packages are public.

.github/workflows/php.yml
```diff
@@ -31,7 +29,8 @@ jobs:
...
 
-    - name: Run test suite
-      run: ./vendor/bin/phpunit
-    - name: Run Woocommerce coding standards
-      run: ./vendor/bin/phpcs
+    - name: Run PHPUnit
+      run: vendor/bin/phpunit
+
+    - name: Run PHPCS
+      run: ./vendor/bin/phpcs--runtime-set ignore_warnings_on_exit 1 src modules woocommerce-paypal-payments.php --extensions=php
```
AlexP11223: Why ignore warnings? Currently there are no warning. :)

docker/Dockerfile_wp
```diff
@@ -74,9 +74,18 @@ RUN docker-php-ext-install intl json && \
     docker-php-ext-enable intl json
 
 # Install Node
-RUN curl -sL https://deb.nodesource.com/setup_15.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
 RUN apt-get install -y nodejs
 
+# Install Phive
```
AlexP11223: Why install phive if not used?

docker/Dockerfile_wp
```diff
@@ -64,7 +64,7 @@ RUN apt-get install -y \
     wget
 
 RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
-    php composer-setup.php --install-dir=/usr/bin --filename=composer --version=1.10.19 && \
+    php composer-setup.php --install-dir=/usr/bin --filename=composer --version=1.10.22 && \
```
AlexP11223: I think it can be just 2 (default).

.github/workflows/php.yml
XedinUnknown: No problem to remove, if extra. It was just a useful thing from the source project.

.github/workflows/php.yml
XedinUnknown: Because whether they are or aren't there, they should not cause the task to fail.

docker/Dockerfile_wp
XedinUnknown: It's another thing from the source project that I thought would be useful. No problem to remove, if not needed.

docker/Dockerfile_wp
XedinUnknown: It cannot, unfortunately, due to Composer plugin dependency constraints.

`wikimedia/composer-merge-plugin:1.4` requires `composer-plugin-api:^1.0`. `wikimedia/composer-merge-plugin:^2.0.1` supports `composer-plugin-api:^2.0`, but requires PHP 7.2, which is higher than this project's minimal requirement.

.github/workflows/php.yml
AlexP11223: Why? It makes sense to ignore warnings if phpcs is added to an old project with lots of warnings, but here we have no warnings right now, so why not keep it this way? 

If there are bad warning rules, then these rules should be disabled, not ignored, because this will result in polluted list of warnings and it will be difficult to see new warnings.

Also because of the weird WP indent rules we often use `phpcbf` here, so if we ignore warnings it may result in lots of changes of unrelated files, making PRs difficult to read.

XedinUnknown: Because a Warning severity does not indicate failure. It indicates something suspicious that should be brought to our attention, but not something that is outright wrong. In other words, our PHPCS standards are a specification, and if there are only warnings, it means that we still comply with it. But we should still see warnings, as they indicate that something _may_ be wrong.

`phpcbf` and other auto-fixing stuff is just there to help. I always try to commit those changes separately. If you are talking about indenting rules, then I guess you are talking about whitespace. Github PRs have the [option to hide whitespace changes][1], so this is not reallhy a problem.

[1]: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

Dinamiko:
Hi @jorgeatorres before merging this PR I need your feedback on renaming files to follow PSR-4 convention, we firmly believe this is a good move but we´ll need your OK before merging it, when you have time, please let us know which are your thought or if you see any potential problem in doing so, thanks.

docker/Dockerfile_wp
XedinUnknown: Removed in 1866efc.

modules/ppcp-compat/src/PPEC/PpecHelper.php
```diff
@@ -11,7 +11,7 @@
  * Helper class with various constants associated to the PayPal Express Checkout plugin, as well as methods for figuring
  * out the status of the gateway.
  */
-class PPECHelper {
+class PpecHelper {
```
jorgeatorres: Would something break if we keep this class as `PPECHelper`. Keep in mind that `PPEC` is an acronym (PayPal Express Checkout) and it looks a bit weird capitalized that way.
Alternatively, we could just rename it to `PayPalCheckoutHelper` or something like that.

modules/ppcp-onboarding/src/OnboardingRestController.php
```diff
@@ -16,7 +16,7 @@
 /**
  * Exposes and handles REST routes related to onboarding.
  */
-class Onboarding_REST_Controller {
+class OnboardingRestController {
```
jorgeatorres: Similar to the above, REST is an acronym.

jorgeatorres:
>  I need your feedback on renaming files to follow PSR-4 convention, we firmly believe this is a good move but we´ll need your OK before merging it

@Dinamiko: I believe we'll have to ignore some naming rules in the PHPCS config file, but I'm ok with this change. This convention is already being used in other WooCommerce packages so it wouldn't be the first time we deviate from the naming convention in the WooCommerce standards.

That said, and given this PR contains a ~180 files changeset, is there anything else I should be taking a closer look at?
I added a couple of comments about instances of certain acronyms (such as REST and PPEC) being capitalized differently, which I believe is odd, but not sure if it's needed for the autoloader to work or something. Please let me know 😃.

Dinamiko:
@jorgeatorres thanks for the review,

My idea was not to assing the PR to you because it contains a lot of files :) it was just to know your thoughs on PSR-4 renaming.

Regarding acronyms I´ve just discused with the team and there is nothing wrong on using them, nothing will be broken, it was just done this way to follow the standard strictly, using capitalization for word separation.

So let´s use acronyms for these two (REST and PPEC) and I´ll check if there are more there.

Dinamiko:
@XedinUnknown PR looks good, just a little adjustment about acronyms on class and file names and we are good to merge:
When you have time, can you please rename these:
-  `PpecHelper` to `PPECHelper` 
- `OnboardingRestController` to `OnboardingRESTController`
- `DccProductStatus` to `DCCProductStatus`
- `WcGatewayModule` to `WCGatewayModule` or `WooCommerceGatewayModule`

XedinUnknown:
I confirm that this kind of class names are explicitly allowed by point 2.2.5 of [PSR-4][]. The requested classes have been renamed as requested, in 40b5783, 9f2496a, 6078de2, and 72216ba.

[PSR-4]: https://www.php-fig.org/psr/psr-4/

modules/ppcp-onboarding/src/OnboardingRestController.php
XedinUnknown: ✔️ 9f2496a

modules/ppcp-compat/src/PPEC/PpecHelper.php
XedinUnknown: ✔️ 40b5783

AlexP11223 (approved):

Dinamiko:
Reopening PR

Dinamiko (approved):

PR #236, opened 2021-08-27
Show CC gateway in admin payments list

AlexP11223:
### Description

Adds CC gateway in the payments list in WC settings, to allow changing order, enabling/disabling it separately.

https://user-images.githubusercontent.com/5680466/131116506-040028b1-fd87-462a-9900-e8f087dfe232.mp4

However, currently it is not possible to disable the PayPal gateway and continue using the CC gateway. I am not sure what was the reason for that and whether we can make them separate now.

### Steps to test:
1. Open payments in WC settings.
2. Try changing order, enabling/disabling, editing settings.
3. Check that everything changes as expected and reflected during the user checkout.

### Changelog entry

- Added the PayPal Credit Card gateway in the payments list in WC settings to allow changing order.


XedinUnknown (approved):
Excellent improvements!

jorgeatorres:
Hi @AlexP11223!

Can you provide some context on why these changes are being made?

> I am not sure what was the reason for that and whether we can make them separate now.

Unless this has changed recently, DCC requires PayPal to be enabled and they also share the same credentials, so having both on the same settings screen makes configuration easier.
Also, this was discussed a while ago and consensus was that it was easier if we didn't show the gateways as different to reduce confusion. While this could be revisited, I haven't seen any new discussions on that.

AlexP11223:
I think the main reason was to allow re-ordering of gateways, that is e.g. put DCC before PP.

#87 

There was a discussion recently mentioned in PCP-178.

jorgeatorres:
> I think the main reason was to allow re-ordering of gateways, that is e.g. put DCC before PP.

That makes sense and I'm not against this change personally, but I do remember this was the way it worked initially and after some discussion it was decided that both should be on the same screen, thus my question.
Given this also has some implications to documentation, I think it should be good to discuss again before merging (in Slack, for example).

> There was a discussion recently mentioned in PCP-178.

Right now, we don't have access to your internal system.

AlexP11223:
> Right now, we don't have access to your internal system.

There was just a comment from Isabella saying that during a meeting on August 5 it was decided to do it this way for DCC.

Dinamiko:
Hi @jorgeatorres,

Just to clarify this a bit, the idea was not to remove DCC from PayPal settings screen but just display it also in Payments methods screen so it can be reordered for example.

Regarding DCC requires PayPal to be enabled, yes we could make it clear for the merchant, maybe an admin message or similar? does it makes sense to you? thanks.

AlexP11223:
Currently there is only a hint on the enable/disable checkbox saying that it must be enabled to use either of them.

I guess we can add an admin notice if only DCC is enabled.

AlexP11223:
Added the notice.

image

The current way of adding notices is a bit strange, such as why have different method names in the *Notice classes instead of the same interface (+ loop), and probably would be better to use extensions for adding them instead of the filter. But for now added it in the same way.

modules/ppcp-wc-gateway/src/Notice/class-dccwithoutpaypaladminnotice.php
```diff
@@ -0,0 +1,77 @@
...
+	 */
+	public function message(): ?Message {
+		if ( ! $this->should_display() ) {
+			return null;
+		}
+
+		$message = sprintf(
+			/* translators: %1$s the gateway name. */
+			__(
+				'PayPal Card Processing cannot be used without the PayPal gateway. <a href="%1$s">Enable  the PayPal Gateway</a>.',
```
Dinamiko: There is an extra space between **Enable** and **the**.

Dinamiko:
@AlexP11223 PR looks great, good refactoring job here! 
Just a little thing, can we limit the display of admin notice to only these three screens: Payments, PayPal and PayPal Card Processing? thanks.

modules/ppcp-wc-gateway/src/Notice/class-dccwithoutpaypaladminnotice.php
AlexP11223: :heavy_check_mark:

isabellainpsyde:
@jorgeatorres the solution was discussed and approved by @Chaithi. The initial request from RT3 was this one: https://github.com/woocommerce/woocommerce-paypal-payments/issues/87. 
As Emili mentioned, the idea was not to remove DCC from PayPal settings screen but just display it also in Payments methods screen so it can be reordered for example.

Dinamiko (approved):

PR #238, opened 2021-08-31
Remove travis

AlexP11223:
We don't need it, GitHub Actions do the same and more.
And it used PHP 7.0.


Dinamiko:
Hi @jorgeatorres, 

Now that we have GH actions properly working: https://github.com/woocommerce/woocommerce-paypal-payments/actions

I just want to confirm with you if we can remove travis or if should we leave it here for some reason, can you please confirm it? thanks.

jorgeatorres:
Hi @Dinamiko!

Yes, as long as the GH action checks for coding standard violations and runs the unit tests (which I see it does), that's fine. Thanks!

Dinamiko (approved):

PR #251, opened 2021-09-10
Improve UX for stored credit card payments

### LINKED ISSUE #242: Saved credit card does not auto fill
### Describe the bug
When I try to auto fill the credit card field by selecting the saved credit card, I am not able to do so. 

### To reproduce
1. Make a payment with credit card processing and mark that credit card to be saved before the payment.
2. Make a payment again but this time with a saved credit card by selecting that saved card from the drop down list
3. Observe, after the saved card selection, credit card field are not populated

## Suggested solution
Auto fill credit card fields from stored payment is not possible because we do not have access to the complete card number but only the last four digits.

What we can do however is improve the UX to make it clear that to place the order is only needed to select a saved payment from the select, see attached video:
https://user-images.githubusercontent.com/456223/132839195-309a8f00-cd2c-4f03-8f16-55f8759dc7f1.mp4

**Screenshots**
image-2021-08-18-12-12-29-657

 - WordPress Version: 5.8
 - WooCommerce Version: 5.6
 - WooCommerce PayPal Payments Plugin Version: 1.5.1

Dinamiko:
Fixes #242.


modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
```diff
@@ -24,9 +24,11 @@ class CheckoutBootstap {
...
           })
 
-        jQuery('#saved-credit-card').on('change', () => {
-            this.displayPlaceOrderButtonForSavedCreditCards()
-        })
+        setTimeout(() => {
+            jQuery('#saved-credit-card').on('change', () => {
+                this.displayPlaceOrderButtonForSavedCreditCards()
+            })
+        }, 3000)
```
AlexP11223: What's the purpose of this delay?

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
```diff
@@ -100,13 +102,45 @@ class CheckoutBootstap {
...
+        jQuery('#ppcp-credit-card-gateway-card-number').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-gateway-card-number').attr("disabled", true);
+        jQuery('label[for="ppcp-credit-card-gateway-card-expiry"]').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-gateway-card-expiry').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-gateway-card-expiry').attr("disabled", true)
+        jQuery('label[for="ppcp-credit-card-gateway-card-cvc"]').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-gateway-card-cvc').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-gateway-card-cvc').attr("disabled", true)
+        jQuery('label[for="vault"]').css({ opacity: 0.4 })
+        jQuery('#ppcp-credit-card-vault').css({ opacity: 0.4 })
```
AlexP11223: Maybe for opacity it would be better and cleaner to add a class via JS and specify these styles in a CSS file.  Also I think it is enough to specify opacity just for their parent container?

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
Dinamiko: It seems that sometimes the listener is not added because the element is not loaded yet, that´s the reason of the delay here. I´m not a big fan of delays but I didn´t found a better way at the moment.

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
Dinamiko: I tried adding a css class with opacity via JS but it´s overridden by the inline styles:

hosted-css

Seems adding the property via JS (using jQuery in this case) overwrites the inline styles.

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
AlexP11223: I am not sure how this element is added, if you cannot fire some kind of event/callback after it is rendered, you can use [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) (for example simply attach it to some parent and when it fires check that this elements exists).

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
AlexP11223: You probably can add `!important` in the class style rules.

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
Dinamiko: Done.

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
Dinamiko: Done.

AlexP11223: hm, the class name is strange, missing `-disabled` like on the image above.

Also you can probably just add these classes once on init and then only set `disabled` attribute and use `myclass:disabled` selector in CSS. https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

Dinamiko: `-disabled` added to class name. Regarding css `disabled` attribute, I´m also using the class on labels which are not disabled. I would prefer to use only one approach for both (labels and inputs) if possible.

AlexP11223 (approved):

PR #253, opened 2021-09-13
Improve logging

### LINKED ISSUE #252: Improve logging
Currently logger does not print much useful information, ideally it should display more information about PayPal API requests and responses.

It is important that we do so by not compromising sensible information like access token.

Expected logging entry could look like this:
```
2021-09-14T10:36:47+00:00 INFO POST https://api.sandbox.paypal.com/some-endpoint
Request Body: {"foo":"bar"}
Response Debug ID: abc123
Response: Array
(
    [code] => 200
    [message] => OK
)
```

Dinamiko:
Fixes #252.


modules/ppcp-api-client/src/Endpoint/class-requesttrait.php
```diff
@@ -36,6 +36,31 @@ private function request( string $url, array $args ) {
 			$args['headers']['PayPal-Partner-Attribution-Id'] = 'Woo_PPCP';
 		}
 
-		return wp_remote_get( $url, $args );
+		$response = wp_remote_get( $url, $args );
+
+		$this->logger->log( 'info', '--------------------------------------------------------------------' );
```
AlexP11223: Why log it as separate log entries? You can just add newlines. But both of these may not be a good idea because of the [DB logging mode](https://developer.woocommerce.com/2017/01/26/improved-logging-in-woocommerce-2-7/) (newlines are removed there iirc, and separate log entries will look weird + there are only like 10 entries per page). Maybe try putting it into a more concise format without `----------` etc., just `GET URL ...., status code: ..., ...`. You can still use newlines so that it would look better at least in non-db mode, and in db view they should be displayed as spaces.

modules/ppcp-api-client/src/Endpoint/class-requesttrait.php
```diff
@@ -36,6 +36,31 @@ private function request( string $url, array $args ) {
...
+		$this->logger->log( 'info', 'URL: ' . wc_print_r( $url, true ) );
```
AlexP11223: Why not just `->info`?

Dinamiko: Done.

modules/ppcp-api-client/src/Endpoint/class-requesttrait.php
Dinamiko: Done.

modules/ppcp-api-client/src/Endpoint/class-requesttrait.php
```diff
@@ -36,6 +38,14 @@ private function request( string $url, array $args ) {
 			$args['headers']['PayPal-Partner-Attribution-Id'] = 'Woo_PPCP';
 		}
 
-		return wp_remote_get( $url, $args );
+		$response = wp_remote_get( $url, $args );
+
+		if ( $this->logger instanceof WooCommerceLogger ) {
+			$this->logger->logRequestResponse( $url, $args, $response );
```
AlexP11223: Would be great to put it somewhere else instead of the logger. Maybe just some trait returning a string.

modules/woocommerce-logging/src/Logger/class-woocommercelogger.php
```diff
@@ -14,6 +14,7 @@
 
 use Psr\Log\LoggerInterface;
 use Psr\Log\LoggerTrait;
+use WP_Error;
```
AlexP11223: can be removed here.

modules/ppcp-api-client/src/Endpoint/class-requesttrait.php
```diff
@@ -36,6 +39,54 @@ private function request( string $url, array $args ) {
...
 
-		return wp_remote_get( $url, $args );
+		$response = wp_remote_get( $url, $args );
+
+		if ( is_wp_error( $response ) ) {
+			$this->logger->error( $response->get_error_code() . ' ' . $response->get_error_message() );
+			return $response;
+		}
+
+		$this->logger->info( $this->request_response_string( $url, $args, $response ) );
```
AlexP11223: Maybe better to use some lower level like `->debug`, so that user can filter logs.

Dinamiko: Done.

AlexP11223 (approved):

Issue #246, opened 2021-09-06
Webhook Status


Dinamiko:
### Feature description
Currently, subscriptions to webhooks is done on the backend during onboarding and is invisible to the user. If there is anything wrong, it is difficult to troubleshoot.

Proposal: Webhooks status page and process to subscribe.

Create a status page that:
- Checks against the webhooks list to verify all desired webhooks are subscribed to.
- If webhooks are not subscribed to, subscribe merchant to the needed events.
- Add re-register button.

**Important:** We should add an FAQ in the documentation of steps to try if you’re not seeing any webhooks (server blocking incoming connection, etc)

Dinamiko:
Merged into trunk.

PR #259, opened 2021-09-16
Webhooks status page

AlexP11223:
Implements #246 except mock webhooks.

image

Some things are based on the old code from the onboarding, gateway modules and not perfect due to lack of time.


Dinamiko (approved):
Great job! I´ve just checked it and works like a charm.

Issue #260, opened 2021-09-16
Creates a mock webhook to test if it is received by the listener


Dinamiko:
As part of #246, we decided to exclude it from 1.6.0 release and add it to the next one.

Creates a [mock webhook](https://developer.paypal.com/docs/api/webhooks/v1/#simulate-event) to test if it is received by the listener. This can help check for any firewall or network issues.

Dinamiko:
Merged into trunk.

PR #273, opened 2021-09-23
Webhook simulation

AlexP11223:
Adds #260 

At first I was thinking to simply save the time of the last webhook and display it here, but then I figured out that we can detect the simulated webhook by the ID we get after the API request. So it is saved into options and checked in the webhook listener. This way we get a bit better UX and no confusing logs because of failed webhook verifications (simulated webhook cannot be verified).

The status is checked every 2 sec via ajax, after 30 sec it displays an error but still tries for the next 30 sec.

https://user-images.githubusercontent.com/5680466/134470344-cb0725fd-16e9-4e3a-a117-21f05408296c.mp4

If failed:
image


Dinamiko (approved):
Awesome work!

Issue #268, opened 2021-09-17
[GlobalStep] Payment gets failed renewal of subscription manually/automatically (288)


gglobalstep:
## Bug Description:
Payment gets failed renewal of subscription manually/automatically.

## Environment:
Woocommerce Version : WooCommerce 5.6.0
WooCommerce PayPal Payments 1.6.0 - test
WordPress version: v5.8.1

PC:
Windows 10, Mac 10.14.6
Chrome(Version 93.0.4577.82) 
Firefox(Version 92.0 (64-bit))
Safari: v14.1.2

## Steps To Reproduce:
1. Create any test site using JN site.
1. Install and activate all the required plugins.
1. Complete the setup wizard.
1. Connect the store to Pay Pal payment.
1. As a customer, purchase a subscription product and take note of the subscription number.
1. Go to WooCommerce > Subscriptions and click on the subscription from step 5.
1. In the “Subscription actions” drop down (to the right) choose “Process renewal” and agree to the dialog.
1. Observe that status changed to "On-hold" due to payment failed.

## Note:
Same scenario observed for renewal of subscription automatically.

## Actual Result:
Payment gets failed renewal of subscription manually/automatically.

## Expected Result:
Payment should get successfully paid renewal of subscription manually/automatically.

## Screenshot:

![#268_1](https://user-images.githubusercontent.com/41110392/133833678-ea52bb47-8d46-473f-915b-96faf95a898d.png)
![#268_2](https://user-images.githubusercontent.com/41110392/133833684-2c57654a-1310-4245-b903-5d389347e1f3.png)

*Isolating the problem (mark completed items with an [x]):*
- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

<details>

`
### WordPress Environment ###

WC Version: 5.6.0
REST API Version: ✔ 5.6.0
WC Blocks Version: ✔ 5.5.1
Action Scheduler Version: ✔ 3.2.1
WC Admin Version: ✔ 2.5.1
Log Directory Writable: ✔
WP Version: 5.8.1
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

[...]

PR #274, opened 2021-09-23
Allow customers to see and delete their saved payments in My Account.

Dinamiko:
Relates to #268.

https://user-images.githubusercontent.com/456223/134516579-08ebee34-e698-489d-acd7-c2c23771a51a.mp4


modules/ppcp-vaulting/src/Endpoint/class-deletepaymenttokenendpoint.php
```diff
@@ -0,0 +1,84 @@
...
+							wp_send_json_success();
+							return true;
+						}
+
+						wp_send_json_error( 'Could not delete payment token.' );
+						return false;
+					}
+				}
+			}
+		} catch ( Exception $error ) {
```
AlexP11223, line 88: should log.

modules/ppcp-vaulting/src/class-vaultingmodule.php
```diff
@@ -0,0 +1,116 @@
...
+
+	/**
+	 * Check if is payments page.
+	 *
+	 * @return bool Whethen page is payments or not.
+	 */
+	private function is_payments_page(): bool {
+		global $wp;
+		$request = explode( '/', $wp->request );
+		if ( end( $request ) === 'ppcp-paypal-payment-tokens' ) {
```
AlexP11223, line 110: Maybe should use `parse_url`?

I guess this will fail if e.g. there is `?` or `#` at the end.

modules/ppcp-vaulting/src/Endpoint/class-deletepaymenttokenendpoint.php
Dinamiko: Done.

modules/ppcp-vaulting/src/class-vaultingmodule.php
Dinamiko: Done.

AlexP11223 (approved):

PR #291, opened 2021-09-29
Update JS packages 

Dinamiko:
Update all JS packages in `modules` to latest versions, by doing so it makes mandatory building the plugin using the latest Node version (currently 16+).

### DIFF
--- a/modules/ppcp-button/package.json
+++ b/modules/ppcp-button/package.json
@@ -7,13 +7,13 @@
     "@babel/core": "^7.9.0",
     "@babel/preset-env": "^7.9.5",
     "babel-loader": "^8.1.0",
-    "cross-env": "^5.0.1",
-    "file-loader": "^4.2.0",
-    "node-sass": "^4.13.0",
-    "sass-loader": "^8.0.0",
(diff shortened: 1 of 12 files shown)


modules/ppcp-button/package.json
```diff
@@ -7,13 +7,13 @@
...
-    "file-loader": "^4.2.0",
-    "node-sass": "^4.13.0",
-    "sass-loader": "^8.0.0",
-    "webpack": "^4.42.1",
-    "webpack-cli": "^3.1.2",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0"
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "cross-env": "^7.0.3",
+    "file-loader": "^6.2.0",
+    "node-sass": "^6.0.1",
```
AlexP11223: I think it should be replaced with `sass`, and looking at other projects this change should not break anything.

Dinamiko: Done.

AlexP11223 (approved):

Issue #293, opened 2021-09-30
Transactions cannot be voided


AlexP11223:
### Problem
When the transaction is an authorization, there is "Order Action" to capture. There is also option to "Refund". Selecting Refund will cause an API error as you are unable to refund an authorization via PayPal.

If a transaction is an authorization, then it cannot be refunded. It can only be captured or voided.

### Proposed Solution

When transaction is an authorization, replace "Refund" option with "Void" option. Documentation to void an authorization can be found here: https://developer.paypal.com/docs/api/payments/v2/#authorizations_void

Dinamiko:
Fixed in #299.

PR #299, opened 2021-10-04
Void transaction

AlexP11223:
Void non-captured authorizations instead of refunding #293

It is done in the gateway refund handler, I am not sure if there is an easy way to add `Void` button and  hide `Refund`. I guess we would need to do that via JS + ajax, but also I think there are other ways to trigger refunds in WC, so using the refund handler should be more reliable. However, currently it is possible to "refund" (void) while not specifying the whole money amount (I think we cannot affect it from the handler) and order status will not change. Should we set the `Refunded` status in this case? Also maybe we should add an order note about voiding.

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/class-paymentsendpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/class-paymentsendpoint.php
@@ -195,10 +195,11 @@ class PaymentsEndpoint {
 	 *
 	 * @param Refund $refund The refund to be processed.
 	 *
-	 * @return bool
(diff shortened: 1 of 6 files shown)


Dinamiko (approved):

AlexP11223:
> However, currently it is possible to "refund" (void) while not specifying the whole money amount (I think we cannot affect it from the handler) and order status will not change. Should we set the `Refunded` status in this case?

Added this. So now after voiding (doing refund from WC when intent was authorize) WC order should have Refunded status in any case.

PR #319, opened 2021-10-11
Do not show capture order action after capture, refund/cancel/fail

### LINKED ISSUE #318: "Capture" order action shown after capture
"Capture authorized PayPal payment" order action is still shown after capture, and refund.

## Steps to reproduce

1. Set **Intent** to **Authorize**.
2. Make an order.
3. Capture via order actions, or refund.

## Expected behaviour

"Capture authorized PayPal payment" order action should be shown only when it is possible to capture.

AlexP11223:
Fixes #318

Now the action is not shown if already captured or if the order status is refunded/cancelled/failed.

The capture column/label may still be shown for cancelled/failed orders (if it was successfully created, but then the status was changed manually), I think it may be useful to know if e.g. a cancelled order was captured.


Dinamiko (changes requested):
Looks good, please fix file names as we have just merged PSR-4 to trunk, thanks.

Dinamiko (approved):

PR #322, opened 2021-10-13
Add psalm

Dinamiko:
Extracts changes from https://github.com/XedinUnknown/woocommerce-paypal-payments/pull/3


AlexP11223:
Does not run in actions :thinking: 

> Could not locate a config XML file in path /home/runner/work/woocommerce-paypal-payments/woocommerce-paypal-payments/.

Dinamiko:
> Does not run in actions 🤔

@AlexP11223 yes I noticed that, I think this PR was just for adding psalm but not for fixing its errors, if you rename/copy ` psalm.xml.dist` to `psalm.xml` and run psalm there are currently 114 errors found plus 466 other issues.

I suggest to merge this PR and then create a new one for fixing all the errors, it will take time so in the meantime we could use psalm locally on our own PR files so we can progressivelly fix things.

What do you think? does it makes sense to you?

AlexP11223:
No, for not fixing errors there is the baseline file. It should be updated and then only new errors will be reported. Looks like you had some kind of mistake in the `.dist` part of the file name.

AlexP11223:
I have a strange performance issue however :thinking: 

It takes 10 minutes to run (`docker-compose run --rm test vendor/bin/psalm --show-info=false --threads=8 --diff`).

In other projects with similar setup it was much faster.

`--debug` shows that it gets stuck on `Parsing /app/vendor/php-stubs/wordpress-stubs/wordpress-stubs.php`

XedinUnknown:
That's odd. Sometimes, Psalm crashes consistently on my machine while no prob on CI. In this case, running it in debug mode "fixes" it. Maybe try that?

AlexP11223:
with `--debug` it was stuck forever on wordpress-stubs.php :)

hm, weird, seems to be fine after restarting docker (`docker-compose down`, `docker-compose up -d wp_dev`)

AlexP11223 (approved):

PR #324, opened 2021-10-13
Run ci only on pull requests and trunk

Dinamiko:
Run CI only on pull requests and trunk.


Dinamiko:
Not needed for now, as GH usage is free for public repositories: 
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration
https://github.community/t/for-public-repositories-is-there-a-monthly-limit-on-minutes/129017/2

PR #327, opened 2021-10-14
Prevent subscription from being purchased when saving payment fails

### LINKED ISSUE #308: Prevent subscription from being purchased when saving payment fails (327)
### Description
Subscription renewals fails because customer does not have saved payments, it could happen when purchase went processed correctly but for some reason the payment was not saved for the customer.

### Steps to reproduce
- Add a subscription product and go to Checkout.
- Select PayPal Card Processing gateway.
- Use a credit card that is NOT ready for vauling (you can use any generated on PayPal developer site / credit card generator) 
- Place the order.

Currently there is no indication that payment is not saved correctly on PayPal but order is processed correctly.

### Expected behaviour
Repeat the same steps above but using the package with the fix.
The expected behavior should be a message indicating that was not possible to save the payment, the parent order with Failed status and subscription Cancelled.

Now repeat the steps again but this time use a vaulting ready credit card (you can use any of the ones in this page: https://developer.paypal.com/docs/limited-release/vault-payment-methods/vault-sdk/#test-cards).
The expected behavior should be the Order processed correctly, parent order with Processing status and subscription Active.

### Possible solution
We could implement a way to detect when this happens and prevent the subscription from being purchased. Otherwise, none of the subscription actions will work as expected.

When there is a subscription product in the cart **force it as an Authorization instead of Sale**. That way we can void the Authorization instead of Refunding a payment.

Place the order and after that we check if the payment is saved correctly (by calling v2/vault/payment-tokens API:
- If it's not saved then void the Authorization and mark the order as Failed, also display a message/note explaining the reason.
- If it´s saved then capture the authorized order.

 - WordPress Version 5.8
 - WooCommerce Version 5.7
 - WooCommerce PayPal Payments Plugin Version 1.6.0

Dinamiko:
Fixes #308.


modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
```diff
@@ -258,4 +346,14 @@ protected function handle_failure( \WC_Order $wc_order, Exception $error ): void
...
 	abstract protected function environment(): Environment;
+
+	/**
+	 * Checks whether the authorization can be voided.
+	 *
+	 * @param Authorization $authorization The authorization to check.
+	 * @return bool
+	 */
+	private function is_voidable_authorization( Authorization $authorization ): bool {
+		return $authorization->status()->is( AuthorizationStatus::CREATED );
```
AlexP11223: Pending is voidable too I think?

Probably should use/extract `is_voidable_authorization` from `RefundProcessor`. Maybe just add `is_voidable` to `ApiClient\Entity\Authorization`.

Dinamiko: Done.

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -741,15 +741,18 @@ private function payerData() {
 	 * @throws \WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException If a setting was not found.
 	 */
 	private function url(): string {
+		$intent = ( $this->settings->has( 'intent' ) ) ? $this->settings->get( 'intent' ) : 'capture';
```
AlexP11223, line 744: hm, something weird with letter case inconsistency.

For example, here it was upper: 
https://github.com/woocommerce/woocommerce-paypal-payments/blob/79e2dd2ef5f9b46ce6f6b8a74c28be1bc38f36e4/modules/ppcp-wc-gateway/extensions.php#L73

modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
AlexP11223: If the logic is the same, why duplicate it?

https://github.com/woocommerce/woocommerce-paypal-payments/pull/331

Dinamiko: OK I´ve merged #331, thanks!

modules/ppcp-button/src/Assets/SmartButton.php
Dinamiko: Good catch! done :)

AlexP11223: It's in other places too.

https://github.com/woocommerce/woocommerce-paypal-payments/blob/79e2dd2ef5f9b46ce6f6b8a74c28be1bc38f36e4/modules/ppcp-wc-gateway/src/Admin/OrderTablePaymentStatusColumn.php#L22

https://github.com/woocommerce/woocommerce-paypal-payments/blob/ba061ce15bf4e9a102b359c70e7a2997b1fdf63e/modules/ppcp-wc-gateway/services.php#L583-L591

https://github.com/woocommerce/woocommerce-paypal-payments/blob/ba061ce15bf4e9a102b359c70e7a2997b1fdf63e/modules/ppcp-compat/src/PPEC/SettingsImporter.php#L126

https://github.com/woocommerce/woocommerce-paypal-payments/blob/ba061ce15bf4e9a102b359c70e7a2997b1fdf63e/modules/ppcp-button/src/Assets/SmartButton.php#L754

https://github.com/woocommerce/woocommerce-paypal-payments/blob/ba061ce15bf4e9a102b359c70e7a2997b1fdf63e/modules/ppcp-onboarding/assets/js/settings.js#L287

Not sure if all of them can be fixed without breaking compatibility (such as if they are saved to db), but it may be causing bugs if we e.g. compare values case sensitively.

Dinamiko: @AlexP11223 I reverted `capture` in lowecase on `SmartButton.php` as it is used in the JS script.

I think the reason of letter case inconsistency is because lowecase is used in JS while uppercase is used for API calls:
https://developer.paypal.com/docs/platforms/checkout/add-capabilities/auth-capture/#1-change-approval-intent
https://developer.paypal.com/docs/platforms/checkout/add-capabilities/auth-capture/#2-create-order

So the rule to follow seems to be `intent` setting stored in lowercase and then use `strtoupper` when needed, for example here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/79e2dd2ef5f9b46ce6f6b8a74c28be1bc38f36e4/modules/ppcp-wc-gateway/extensions.php#L73

AlexP11223: hm, in this example it's unclear why they suddenly decided to convert to upper if non-API values like our settings are supposed to be lower.

But yeah we probably should always do case-insensitive comparison of intents unless we are really sure about case.

Or treat them as completely separate things (because e.g. we can add some other options to our intent settings) but it is difficult to enforce because they are just strings.

AlexP11223 (approved):

PR #332, opened 2021-10-15
Change shipping preference if purchase units does not contain shipping

### LINKED ISSUE #330: Can't checkout to certain countries with optional postcode (319)
### Description
Same issue as #224: can't checkout to certain countries with optional postcode, that was not fixed on 1.6.0. Countries that have no postal code encounter an issue during checkout, since it still required a postal code to be included.

Error:
```
[UNPROCESSABLE_ENTITY] The requested action could not be performed, semantically incorrect, or failed business validation. https://developer.paypal.com/docs/api/orders/v2/#error-MISSING_SHIPPING_ADDRESS
```
### Steps to reproduce
- Create a shipping zone that includes countries like UAE or Ireland
- On checkout set address to United Arab Emirates or Ireland
- Use PayPal to pay

Error appears: `MISSING_SHIPPING_ADDRESS`

### Expected behaviour
As with the PayPal Checkout plugin, the checkout should work for countries with optional postcodes.

### Environment
 - WordPress Version: 5.8
 - WooCommerce Version: 5.8
 - WooCommerce PayPal Payments Version 1.6.1

Dinamiko:
Fixes #330.


modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -572,4 +576,20 @@ public function patch_order_with( Order $order_to_update, Order $order_to_compar
...
 		return $new_order;
 	}
+
+	/**
+	 * Checks if items contains shipping.
+	 *
+	 * @param array $items The items.
+	 * @return bool Whether items contains shipping or not.
+	 */
+	private function items_without_shipping( array $items ): bool {
```
AlexP11223: This name sounds like it returns the items. So it should either do that or be renamed into something like `has_items_without_shipping`.

modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -572,4 +576,20 @@ public function patch_order_with( Order $order_to_update, Order $order_to_compar
 		$new_order = $this->order( $order_to_update->id() );
 		return $new_order;
 	}
+
+	/**
+	 * Checks if items contains shipping.
```
AlexP11223: contain

or better "Checks if there is at least one item without shipping."

modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -193,6 +193,10 @@ static function ( $item ) use ( &$contains_physical_goods ): bool {
 				: ApplicationContext::SHIPPING_PREFERENCE_GET_FROM_FILE
 			: ApplicationContext::SHIPPING_PREFERENCE_NO_SHIPPING;
 
+		if ( $this->items_without_shipping( $items ) ) {
+			$shipping_preferences = ApplicationContext::SHIPPING_PREFERENCE_NO_SHIPPING;
```
AlexP11223, line 197: What if there are some items with shipping and some without? Is it ok to have `SHIPPING_PREFERENCE_NO_SHIPPING` for the whole order then?

Dinamiko: Yup, it´s ok to have `SHIPPING_PREFERENCE_NO_SHIPPING` for the whole order because as soon the country does not match PayPal business rules `shipping` is not going to be included in `purchase_units`: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php#L164

In this case we need to change `shipping_preference` to avoid this response error: 
**The shipping address is required when `shipping_preference=SET_PROVIDED_ADDRESS`.**

AlexP11223 (approved):

PR #341, opened 2021-10-27
Do not validate checkout form when creating PayPal order

### LINKED ISSUE #335: Order of WooCommerce checkout actions causing incompatibility with AvaTax address validation (375)
### Description
PayPal Payments is causing the `woocommerce_checkout_order_processed` action to run before `woocommerce_after_checkout_validation` when processing payment, which is reverse of how those actions are usually ran in WC.

AvaTax stores its `wc_avatax_address_validated` flag in session data, and this is later checked during `woocommerce_after_checkout_validation` . However, session data is reset during `woocommerce_checkout_order_processed` , so by calling that action first, PayPal Payments is causing the validation flag to be cleared before it can be verified. 

**This prevents payment from being processed** and causes the “Please validate your address to continue” message to appear at checkout, even if the address had already been validated.

The issue occurs consistently on a default theme with only WooCommerce, PayPal Payments, and AvaTax plugins enabled.

A default checkout option like COD runs the actions in the correct order.

I also tested with a SkyVerge gateway plugin that has a similar off-site redirect to that of PayPal Payments, and the actions ran in the expected order there, allowing payment to process with a validated address.

### To reproduce

1. Install and activate PayPal Payments alongside SkyVerge's [WooCommerce AvaTax plugin](https://woocommerce.com/products/woocommerce-avatax/)
1. Navigate to the AvaTax plugin settings: **WooCommerce > Settings > Tax > AvaTax** 
1. Enable AvaTax address validation and set **Require for Tax Calculation** to true
1. Add a product to your cart and proceed to checkout
1. Enter a valid billing address, validate w/ AvaTax's "Validate Address" button
1. Attempt to submit payment via PayPal Payments. After submitting payment information, the checkout page will return with a "Please validate your address to continue" error message

### Expected behavior

**For customers:** When my address is validated at checkout, address validation should not prevent me from completing payment.

**For AvaTax & other plugins:** PayPal Payments should call WooCommerce checkout actions in the expected order, specifically `woocommerce_checkout_order_processed` should only run after `woocommerce_after_checkout_validation`

**Environment:**
Tested on:
 - WordPress version 5.8.1
 - WooCommerce version 5.7.0 and 5.8.0
 - WooCommerce PayPal Payments plugin version 1.5.1 and 1.6.1
 - Chrome version 94.0.4606.81
 - WooCommerce AvaTax version 1 .13.2

### Additional details

A8C Ref: 4322337-zen
SV Ref: 179556

Dinamiko:
Fixes #335.

### Possible cause
We are currently calling WC `process_checkout` twice, the first time it's called before creating the PayPal order to validate WC checkout form.

This is causing third party plugins not working because they are using `woocommerce_after_checkout_validation` for validating its own logic and after that cleaning session data on `woocommerce_checkout_order_processed`, when our plugin calls `process_checkout` the second time it calls again `woocommerce_after_checkout_validation` making it fail because there is no session stored available, it was already cleared on `woocommerce_checkout_order_processed`.

### Suggested solution
We can simply skip WC checkout form validation when creating the PayPal order, that's in fact how it works when we create the PayPal order from other pages like Product or Cart.

By not calling `process_checkout` when creating the PayPal order we are going to probably also fix some of those WC Failed orders because the WC order is going to be created only after PayPal order is created correctly and WC checkout form is validated.

### DIFF
--- a/modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
+++ b/modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
@@ -187,10 +187,26 @@ class CreateOrderEndpoint implements EndpointInterface {
 
 			if ( 'checkout' === $data['context'] ) {
 				if ( isset( $data['createaccount'] ) && '1' === $data['createaccount'] ) {
-					$this->process_checkout_form_when_creating_account( $data['form'], $wc_order );
+					try {
+						$order = $this->create_paypal_order( $wc_order );
+					} catch ( Exception $exception ) {
+						$this->logger->error( 'Order creation failed: ' . $exception->getMessage() );
+						throw $exception;
+					}
+					wp_send_json_success( $order->to_array() );
 				}
 
-				$this->process_checkout_form( $data['form'] );
+				try {
+					$order = $this->create_paypal_order();
+				} catch ( Exception $exception ) {
+					$this->logger->error( 'Order creation failed: ' . $exception->getMessage() );
+					throw $exception;
+				}
+
+				if ( ! $this->early_order_handler->should_create_early_order() ) {
+					wp_send_json_success( $order->to_array() );
+				}
+				$this->early_order_handler->register_for_order( $order );
 			}
 			if ( 'pay-now' === $data['context'] && get_option( 'woocommerce_terms_page_id', '' ) !== '' ) {
 				$this->validate_paynow_form( $data['form'] );
@@ -320,7 +336,6 @@ class CreateOrderEndpoint implements EndpointInterface {
 		return $payer;
 	}
 
-
 	/**
 	 * Sets the BN Code for the following request.
 	 *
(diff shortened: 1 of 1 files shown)


AlexP11223:
But what was the reason to call it twice? Is everything still working?

Looks like we do not have WC validation now before creating the order here, is it ok?

Dinamiko:
@AlexP11223 

> But what was the reason to call it twice? 

The reason to call `process_checkout` the first time I think was only for validating the WC checkout form before creating the PayPal order.

Here is how it currently works;
We call `process_checkout` the first time here before creating the PayPal order: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php#L380
Then after PayPal order is approved we call it again here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js#L25

`process_checkout` calls `validate_checkout` internally: https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/class-wc-checkout.php#L1161

which defines `woocommerce_after_checkout_validation` action: https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/class-wc-checkout.php#L887

We are not validating WC checkout form when creating the PayPal order on other pages like Single product or Cart, that´s the reason I proposed this solution, it does separate PayPal order creation from WC order creation. It also opens the door for writing our own PayPal specific validation for order creation.

> Is everything still working?

For what it tested so far it works, but we´ll need to do more acceptance to ensure nothing is broken for all cases.

> Looks like we do not have WC validation now before creating the order here, is it ok?

For creating the PayPal order we are not going to have WC checkout form validation now but it still does validation when creating the WC order:
https://user-images.githubusercontent.com/456223/139224476-f5581fdc-6eec-406e-9a79-d647deab689a.mp4

AlexP11223 (approved):

PR #343, opened 2021-10-28
Disable PayPal gateways if subscription product in cart and customer not logged in

### LINKED ISSUE #342: Guest users must checkout twice for subscriptions, no smart buttons loaded (350)
### Description
When there is a subscription product in the cart, guest users need to perform the checkout twice.

### To reproduce
1. Check “Allow customers to create an account during checkout” in WooCommerce/Settings/Accounts & Privacy
2. Log out
3. Add subscription product to cart
4. Visit checkout

- Smart buttons are not loaded; instead regular “Sign up now” button appears
- Clicking the “Sign up now” button creates a Failed order, reloads the page, and the user will be logged in automatically
- After the page reloads, the smart buttons are loading and the user could actually purchase the subscription

### Possible cause
We are currently vaulting both payments (PayPal and Advanced Credit Card) with the JavaScript SDK: https://developer.paypal.com/docs/limited-release/vault-payment-methods/vault-sdk/#how-it-works

This means that we **need to generate a client token identifying the buyer by a unique customer ID**. The client token is generated and added to the JSSDK script on page load meaning that **when the plugin generates the client token there is no user logged in** yet and therefore the client token is generated without including customer_id.

### Suggested solution
1) When there is a subscription product in the cart and the customer is not logged in yet then hide both PayPal gateways.
2) Create a new issue for allowing non logged in customers to purchase subscription products.

### Environment
 - WordPress Version: 5.8
 - WooCommerce Version: 5.8

Dinamiko:
Fixes #342.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
@@ -184,7 +184,7 @@ class CreditCardRenderer {
             const save_card = this.defaultConfig.save_card ? true : false;
             let vault = document.getElementById('ppcp-credit-card-vault') ?
                 document.getElementById('ppcp-credit-card-vault').checked : save_card;
-            if(this.defaultConfig.enforce_vault) {
+            if (this.defaultConfig.enforce_vault) {
                 vault = true;
             }
             const contingency = this.defaultConfig.hosted_fields.contingency;
--- a/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
@@ -182,8 +182,11 @@ class CreditCardRenderer {
 
         if (this.formValid && this.cardValid) {
             const save_card = this.defaultConfig.save_card ? true : false;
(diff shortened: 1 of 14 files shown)


Dinamiko:
@AlexP11223 any idea why psalm is reporting errors on files that does not belong to this PR? https://github.com/woocommerce/woocommerce-paypal-payments/runs/4079028799?check_suite_focus=true

AlexP11223:
Psalm updated to new version.

Fixed in #346

modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -182,8 +182,11 @@ class CreditCardRenderer {
 
         if (this.formValid && this.cardValid) {
             const save_card = this.defaultConfig.save_card ? true : false;
-            const vault = document.getElementById('ppcp-credit-card-vault') ?
+            let vault = document.getElementById('ppcp-credit-card-vault') ?
                 document.getElementById('ppcp-credit-card-vault').checked : save_card;
+            if(this.defaultConfig.enforce_vault) {
```
AlexP11223: space

AlexP11223 (approved):

PR #351, opened 2021-11-05
Add option for flushing rewrite rules

### LINKED ISSUE #350: PayPal payments page empty after switching packages (390)
### Describe the bug
When user updates to a new WooCommerce PayPal payments package, PayPal Payments page in the My Account section in the frontend will be empty. 

### To reproduce
1. Deactivate and activate the plugin.
2. Visit "My Account / PayPal payments" page.

### Expected behavior
"My Account / PayPal payments" page should be accessible.

### Possible cause
Rewrite rules is executed before creating the endpoint.

### Suggested solution
Flush rewrite rules using a flag after creating the endpoint. 

### Environment
 - WordPress Version: 5.8
 - WooCommerce Version: 5.8

Dinamiko:
Fixes #350.

### DIFF
--- a/woocommerce-paypal-payments.php
+++ b/woocommerce-paypal-payments.php
@@ -81,6 +81,16 @@ define( 'PPCP_FLAG_SUBSCRIPTION', true );
 		'plugins_loaded',
 		function () {
 			init();
+
+			if ( ! function_exists( 'get_plugin_data' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/plugin.php';
+			}
+			$plugin_data    = get_plugin_data( __DIR__ . '/woocommerce-paypal-payments.php' );
+			$plugin_version = $plugin_data['Version'] ?? null;
+			if ( get_option( 'woocommerce-ppcp-upgrading-version' ) !== $plugin_version ) {
+				do_action( 'woocommerce_paypal_payments_gateway_migrate' );
+				update_option( 'woocommerce-ppcp-upgrading-version', $plugin_version );
+			}
 		}
 	);
 	register_activation_hook(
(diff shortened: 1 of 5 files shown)


AlexP11223:
I think it may break on re-activation/re-installation?

Deactivated, then some other plugin calls `flush` (or the user on the permalinks page), and after the next activation we do not have this rule.

I would add a "migration"/upgrading event like in Zettle. At the end of `plugins_loaded` compare the current version with the version saved in an option, if not matches - save the current one and fire event (`woocommerce_paypal_payments_gateway_migrate`).

And we need to flush in this event + in `register_activation_hook`.

In `register_activation_hook` I guess we also need to call `add_rewrite_endpoint` before `flush`.

So then just add 3 handlers in the vaulting module's `run`:

- `woocommerce_paypal_payments_gateway_migrate` - `add` + `flush`
- `woocommerce_paypal_payments_gateway_activate` - `add` + `flush`
- `init` - `add` (I guess it's in case something else calls `flush` later)

woocommerce-paypal-payments.php
```diff
@@ -81,14 +81,23 @@ function() {
...
 			init();
+
+			if ( ! function_exists( 'get_plugin_data' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/plugin.php';
+			}
+			$plugin_data    = get_plugin_data( __DIR__ . '/woocommerce-paypal-payments.php' );
+			$plugin_version = $plugin_data['Version'] ?? null;
+			if ( get_option( 'woocommerce-ppcp-upgrading-version' ) !== $plugin_version ) {
+				do_action( 'woocommerce_paypal_payments_gateway_migrate' );
+				update_option( 'woocommerce-ppcp-upgrading-version', $plugin_version );
```
AlexP11223: maybe it should be just `woocommerce-ppcp-version`?

Dinamiko: Done :)

AlexP11223 (approved):

PR #352, opened 2021-11-08
Check if terms checkbox is available

### LINKED ISSUE #294: Payment not possible in pay for order form because of terms checkbox missing (371)
### Problem
The PayPal payment cannot be startet in the pay for order form because this plugin checks for the _terms_ checkbox which might not be shown in combination with third party plugins.

### To reproduce
1. Use Plugin like "WooCommerce Germanized" and "WooCommerce Germanized Pro" which has a checkbox management where _terms_ checkbox is disabled or not shown on the pay for order form.
2. Try to pay an order via pay for order form.
3. Error "Please read and accept the terms and conditions to proceed with your order." occurs without having seend any _terms_ checkbox.

### Expected behavior
Validation of terms checkbox should only be triggered if checkbox was shown - like in the WooCommerce core.

**Environment**
 - WordPress 5.7.3
 - WooCommerce 5.5.1
 - WooCommerce PayPal Payments Plugin Version 1.6.0
 - WooCommerce Germanized and WooCommerce Germanized Pro

Dinamiko:
Fixes #294.

### DIFF
--- a/modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
+++ b/modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
@@ -381,7 +381,7 @@ class CreateOrderEndpoint implements EndpointInterface {
 	 */
 	private function validate_paynow_form( string $form_values ) {
 		$parsed_values = wp_parse_args( $form_values );
-		if ( ! isset( $parsed_values['terms'] ) ) {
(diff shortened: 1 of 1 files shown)


AlexP11223 (approved):
seems to match WC behavior

https://github.com/woocommerce/woocommerce/blob/a6df312732fd444c4b8a69459699b85583276eb5/plugins/woocommerce/includes/class-wc-form-handler.php#L411-L412

PR #353, opened 2021-11-08
introduce woocommerce_paypal_payments_product_supports_prb filter

### LINKED ISSUE #234: [Request] Disable PayPal Smart buttons for specific product types
### Describe the feature/request

Our team, SomewhereWarm, is looking to add an integration between our plugins and the new PayPal Payments. Some of the extensions we develop are Product Bundles, Gift Cards, Composite Products and All Products for WooCommerce Subscriptions -- all are available on WooCommerce.com.

These 4 extensions allow customers to make configurations in the single product page that can result to 
1) a price change or,
2) additional data that need to be posted, such as Gift Card meta (sender, recipient, message).

We have found that currently PayPal Smart buttons do not parse these prices/data correctly when they are clicked. Therefore, to prevent any issues we thought to add an integration that disables these buttons for specific product pages -- we used to have a similar integration for PayPal Checkout. 

Taking a look at the code, we noticed that you are already hiding these buttons for specific product types, such as Grouped and External: https://github.com/woocommerce/woocommerce-paypal-payments/blob/0afe6cd2e490c03fa6879b0f34d8db214616d22f/modules/ppcp-button/src/Assets/class-smartbutton.php#L426

Ideally, we would like to filter the return value of these two functions: https://github.com/woocommerce/woocommerce-paypal-payments/blob/0afe6cd2e490c03fa6879b0f34d8db214616d22f/modules/ppcp-button/src/Assets/class-smartbutton.php#L421 and https://github.com/woocommerce/woocommerce-paypal-payments/blob/0afe6cd2e490c03fa6879b0f34d8db214616d22f/modules/ppcp-button/src/Assets/class-smartbutton.php#L439 
Do you think it is possible to add a filter there that has the `$product` object as an argument? 

So far, we have tried a couple of approaches to disable these buttons. However, none of them was robust or performant enough! 

1. 

The first thing we tried was to remove this action: https://github.com/woocommerce/woocommerce-paypal-payments/blob/0afe6cd2e490c03fa6879b0f34d8db214616d22f/modules/ppcp-button/src/Assets/class-smartbutton.php#L336 -- however, there wasn't any function we could use to retrieve the `SmartButton` object.

2.

Another thing we tried was to use the `woocommerce_paypal_payments_single_product_renderer_hook` and return an empty string. However, the `$product` object wasn't available at this point and running `wc_get_product()` on every page load wasn't a performant solution. 

In case there is any way to achieve what we are after without introducing new filters, please let us know! Thank you in advance!

helgatheviking:
### Description

Add a filter that will allow other plugins to conditionally disable the payment request buttons for single products that cannot support them.

### Steps to test:

```
add_filter( 'woocommerce_paypal_payments_product_supports_prb', '__return_false' );
```

Should turn off payment request buttons on _all_ single products even if they are enabled in the admin.

Because the `$product` object is passed, can also disable conditionally, for example, in the case of Mix and Match products

```
/**
 * Disable support for payment request buttons
 * 
 * @param bool $supports_prb
 * @param WC_Product $product
 * @return bool
 */
function kia_test_buttons( $supports_prb, $product ){
    if ( $product->is_type( 'mix-and-match' ) ) {
        $supports_prb = false;
    }
    return $supports_prb;
    
}
add_action( 'woocommerce_paypal_payments_product_supports_prb', 'kia_test_buttons', 10, 2 );
```

### Changelog entry

Add `woocommerce_paypal_payments_product_supports_prb` filter

Closes #234.

cc @jimjasson

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -420,11 +420,10 @@ class SmartButton implements SmartButtonInterface {
 	 */
 	public function button_renderer() {
 		$product = wc_get_product();
+
 		if (
 			! is_checkout() && is_a( $product, \WC_Product::class )
-			&& (
-				$product->is_type( array( 'external', 'grouped' ) )
-				|| ! $product->is_in_stock()
+			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_prb', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
 			)
 		) {
 			return;
@@ -437,6 +436,15 @@ class SmartButton implements SmartButtonInterface {
 	 * Renders the HTML for the credit messaging.
 	 */
 	public function message_renderer() {
+		$product = wc_get_product();
+
+		if (
+			! is_checkout() && is_a( $product, \WC_Product::class )
+			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_prb', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
+			)
+		) {
+			return;
+		}
 
 		echo '<div id="ppcp-messages" data-partner-attribution-id="Woo_PPCP"></div>';
 	}
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -423,7 +423,7 @@ class SmartButton implements SmartButtonInterface {
 
 		if (
 			! is_checkout() && is_a( $product, \WC_Product::class )
-			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_prb', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
+			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_payment_request_button', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
 			)
 		) {
 			return;
@@ -440,7 +440,7 @@ class SmartButton implements SmartButtonInterface {
 
 		if (
 			! is_checkout() && is_a( $product, \WC_Product::class )
-			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_prb', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
+			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_payment_request_button', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
 			)
 		) {
 			return;


helgatheviking:
Hi team, can we get some eyes on this?

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -420,11 +420,10 @@ public function enqueue(): bool {
 	 */
 	public function button_renderer() {
 		$product = wc_get_product();
+
 		if (
 			! is_checkout() && is_a( $product, \WC_Product::class )
-			&& (
-				$product->is_type( array( 'external', 'grouped' ) )
-				|| ! $product->is_in_stock()
+			&& ( ! apply_filters( 'woocommerce_paypal_payments_product_supports_prb', ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock(), $product )
```
leonardola: It would be better if we did `apply_filters( 'woocommerce_paypal_payments_product_supports_prb', $product)` instead. This would allow the plugin adding the filter to have access to more info in the product. Also we should add the apply filter as the last part of the if statement. So we don't call the hook unless the other checks passed.

leonardola (changes requested):
Thanks for the contribution. I left a note about some changes that would make the hook more flexible for other plugins.

modules/ppcp-button/src/Assets/SmartButton.php
leonardola: Also I tend to dislike abbreviations as sometimes they usually need more context to understand. So I would suggest to change the hook name to `woocommerce_paypal_payments_product_supports_payment_request_button` even though it's quite long.

helgatheviking: Not sure I understand the first comment. Do you mean something like:
```
$supports = ! is_checkout() && is_a( $product, \WC_Product::class ) && ! $product->is_type( array( 'external', 'grouped' ) ) && $product->is_in_stock();
$supports = apply_filters( 'woocommerce_paypal_payments_product_supports_payment_request_button', $product ); 

if ( ! $supports ) {
	return;
}
```

leonardola: Oops. I misread the `)` here. So we are actually good to go. Just changing the filter name would be enough.

leonardola:
I've tested this and I can still see the PayPal button in the cart when a product that does not support it is present. It should probably not show up there right?

helgatheviking:
>It should probably not show up there right?

As far as I know, it's fine there. For example, in my Name Your Price plugin... the payment request buttons don't work on the single product page because there's an extra input where the user enters their price. So users can click the PRB before entering a price and bypass the minimum price checks! Ideally, I'd love to enable/disable the buttons when the price is valid, but that's not possible yet. 

Once the Name Your Price product is in the cart though... it _has_ a price. So in the cart the PRB don't cause any trouble.

leonardola (approved):
Thanks @helgatheviking for contributing in this one. Keep rockin'

helgatheviking:
Glad to see this merged! @leonardola one more thing... what would be the easiest way to test this plugin is active? Like a `class_exists( "something" )` or equivalent?

leonardola:
I usually check the class like you posted. But you can also use [this function](https://developer.wordpress.org/reference/functions/is_plugin_active/). 

I'll probably do a release tomorrow so keep an eye open to update any code running on that hook on your sites

leonardola:
Just got pinged that we'll need to postpone the release to after May 20

helgatheviking:
No problem, I'm not in a rush. Which class would you test for? `WooCommerce\PayPalCommerce` ?

leonardola:
You could check [this](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/src/Plugin.php) class as it seems to be loaded on every request

PR #359, opened 2021-11-12
Do not pass `customer_id` if vaulting is disabled

### LINKED ISSUE #358: Do not display saved payments on PayPal buttons if vault option is disabled (419)
### Describe the bug
When Vault option is disabled and there are saved PayPal payments, the payment is still showing in PayPal button while it should not.

### To reproduce
Save a PayPal payment and disable Vault option afterwards, then visit a page containing PayPal buttons.

### Expected behavior
PayPal buttons should not display saved payment if Vault option is disabled.

### Environment
 - WordPress Version: 5.8
 - WooCommerce Version: 5.9

Dinamiko:
Fixes #358.

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/IdentityToken.php
+++ b/modules/ppcp-api-client/src/Endpoint/IdentityToken.php
@@ -14,6 +14,7 @@ use WooCommerce\PayPalCommerce\ApiClient\Entity\Token;
 use WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException;
 use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
 use Psr\Log\LoggerInterface;
+use WooCommerce\PayPalCommerce\WcGateway\Settings\Settings;
 
 /**
  * Class IdentityToken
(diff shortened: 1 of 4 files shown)


modules/ppcp-api-client/src/Endpoint/IdentityToken.php
```diff
@@ -84,7 +94,9 @@ public function generate_for_customer( int $customer_id ): Token {
 				'Content-Type'  => 'application/json',
 			),
 		);
-		if ( $customer_id && defined( 'PPCP_FLAG_SUBSCRIPTION' ) && PPCP_FLAG_SUBSCRIPTION ) {
+		if ( $customer_id
+			&& ( $this->settings->has( 'vault_enabled' ) && $this->settings->get( 'vault_enabled' ) )
+			&& defined( 'PPCP_FLAG_SUBSCRIPTION' ) && PPCP_FLAG_SUBSCRIPTION ) {
 			$args['body'] = wp_json_encode( array( 'customer_id' => $this->prefix . $customer_id ) );
```
AlexP11223, line 102: `$args...` looks like part of the condition.

I think phpcs rules allow something like

image


which is more readable.

AlexP11223 (approved):

PR #365, opened 2021-11-16
Use order ID instead of "number" for PayPal custom field

### LINKED ISSUE #354: Payments (sometimes) fail when using custom order numbers (401)
### Describe the bug
Background Info: We use the (pretty common and widely used) WC Jetpack/Booster plugin to generate custom order numbers for our orders instead of using the default WP Post ID's. That generates continuous order numbers for us, these are in no way related to the internal post ID's.

In #163 the behaviour was changed to use the `get_order_number()` function to hand over the invoice_id and custom_id to Paypal instead of using the `get_id()`. Which is great for the UX of the user to see the actual order number generated by our system. Everything works fine until after the payment. 

Once the Paypal webhooks of "CHECKOUT.ORDER.COMPLETED" or "CHECKOUT.ORDER.APPROVED" hit the store, it's getting the custom order numbers in those webhooks again as "custom_id" (which makes sense, since thats what we told Paypal in the first place to use), BUT the handler in the plugin expects this "custom_id" to be a WP Post ID and looks for the order by trying "post__in = custom_id", which will obviously not work with a custom order number. The handler should check again for the order number and NOT the ID. 

So as a result we see a lot of those messages in the log:
> 2021-11-02T03:24:23+00:00 WARNING Order for PayPal order XXXXXXXXXXXXXX not found.
> 2021-11-02T03:24:23+00:00 INFO Webhook has been handled by CHECKOUT.ORDER.APPROVED

Here is an example incoming webhook that fails (personal and actual order information redacted):
```
2021-11-02T03:24:24+00:00 DEBUG POST https://api.paypal.com/v1/notifications/verify-webhook-signature
Request Body: {"transmission_id":"455ac6c0-...","transmission_time":"2021-11-02T03:23:38Z","cert_url":"https:\/\/api.paypal.com\/v1\/notifications\/certs\/CERT-...","auth_algo":"SHA256withRSA","transmission_sig":"PrKFYC3p8x8Dlp...","webhook_id":"27...","webhook_event":{"id":"WH-05Y23463...","event_version":"1.0","create_time":"2021-11-02T03:23:35.171Z","resource_type":"checkout-order","resource_version":"2.0","event_type":"CHECKOUT.ORDER.APPROVED","summary":"An order has been approved by buyer","resource":{"create_time":"2021-11-02T03:22:28Z","purchase_units":[{"reference_id":"default","amount":{"currency_code":"USD","value":"582.10","breakdown":{"item_total":{"currency_code":"USD","value":"562.10"},"shipping":{"currency_code":"USD","value":"20.00"},"tax_total":{"currency_code":"USD","value":"0.00"}}},"payee":{"email_address":"...","merchant_id":"...","display_data":{"brand_name":"..."}},"description":"","custom_id":"CM-G97487","invoice_id":"CM-G97487","items":[{"name":"...","unit_amount":{"currency_code":"USD","value":"7.90"},"tax":{"currency_code":"USD","value":"0.00"},"quantity":"5","description":"...","sku":"...","category":"PHYSICAL_GOODS"},{"name":"...","unit_amount":{"currency_code":"USD","value":"9.90"},"tax":{"currency_code":"USD","value":"0.00"},"quantity":"1","description":"...","sku":"...","category":"PHYSICAL_GOODS"},{"name":"...","unit_amount":{"currency_code":"USD","value":"512.70"},"tax":{"currency_code":"USD","value":"0.00"},"quantity":"1","description":"","sku":"..."","category":"PHYSICAL_GOODS"}],"shipping":{"name":**redacted**}}],"links":[{"href":"https:\/\/api.paypal.com\/v2\/checkout\/orders\/...","rel":"self","method":"GET"},{"href":"https:\/\/api.paypal.com\/v2\/checkout\/orders\/...","rel":"update","method":"PATCH"},{"href":"https:\/\/api.paypal.com\/v2\/checkout\/orders\/...\/capture","rel":"capture","method":"POST"}],"id":"...","intent":"CAPTURE","status":"APPROVED"},"links":[{"href":"https:\/\/api.paypal.com\/v1\/notifications\/webhooks-events\/WH-05Y...","rel":"self","method":"GET"},{"href":"https:\/\/api.paypal.com\/v1\/notifications\/webhooks-events\/WH-05Y23...\/resend","rel":"resend","method":"POST"}]}}
```

I'm not entirely sure when these webhooks are actually used, because most orders run through just fine, but once these webhooks are involved, they fail. But the worst part is: The payment actually goes through, but the order fails in our system, resulting in very pissed off customers (and they have every right to).

Related lines in the webhook handlers I mentioned:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/1a5e5d88fc6f1f1460b75292c8f582d054f5a42d/modules/ppcp-webhooks/src/Handler/CheckoutOrderCompleted.php#L112
https://github.com/woocommerce/woocommerce-paypal-payments/blob/1a5e5d88fc6f1f1460b75292c8f582d054f5a42d/modules/ppcp-webhooks/src/Handler/CheckoutOrderApproved.php#L169

### To reproduce
1. Use Jetpack/WC Booster (or similar plugin) to enable custom order numbers instead of using the WP Post ID's as order number
2. Please an order and pay
3. Paypals Webhooks coming back with "custom_id" set to your custom order number
4. Plugin trying to find order with "post__in" using that value, which cannot find the order
5. Payment was actually done on Paypal, but fails on store

### Expected behavior
Either always use post ID's in the whole process or make sure the it looks up the correct order with the order number again instead of the ID.

 - WordPress Version: 5.8.1
 - WooCommerce Version: 5.8.0
 - Browser [e.g. chrome, safari] and Version: Irrelevant
 - Any other plugins installed: WC Jetpack/Booster

AlexP11223:
Fixes #354 

There are 2 fields: invoice ID and custom. It seems like we can just put the real ID into the second one.

image

### DIFF
--- a/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
+++ b/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
@@ -119,9 +119,8 @@ class PurchaseUnitFactory {
 		$reference_id    = 'default';
 		$description     = '';
 		$payee           = $this->payee_repository->payee();
-		$wc_order_id     = $order->get_order_number();
-		$custom_id       = $this->prefix . $wc_order_id;
-		$invoice_id      = $this->prefix . $wc_order_id;
+		$custom_id       = $this->prefix . (string) $order->get_id();
+		$invoice_id      = $this->prefix . $order->get_order_number();
 		$soft_descriptor = '';
 		$purchase_unit   = new PurchaseUnit(
 			$amount,
--- a/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
+++ b/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
@@ -119,7 +119,7 @@ class PurchaseUnitFactory {
 		$reference_id    = 'default';
 		$description     = '';
 		$payee           = $this->payee_repository->payee();
-		$custom_id       = $this->prefix . (string) $order->get_id();
+		$custom_id       = (string) $order->get_id();
 		$invoice_id      = $this->prefix . $order->get_order_number();
 		$soft_descriptor = '';
 		$purchase_unit   = new PurchaseUnit(
--- a/modules/ppcp-webhooks/src/Handler/PrefixTrait.php
+++ b/modules/ppcp-webhooks/src/Handler/PrefixTrait.php
@@ -31,7 +31,10 @@ trait PrefixTrait {
 	 */
 	private function sanitize_custom_id( string $custom_id ): int {
 
-		$id = str_replace( $this->prefix, '', $custom_id );
+		$id = $custom_id;
+		if ( strlen( $this->prefix ) > 0 && 0 === strpos( $id, $this->prefix ) ) {
+			$id = substr( $id, strlen( $this->prefix ) );
+		}
 		return (int) $id;
 	}
 }
--- a/tests/PHPUnit/ApiClient/Factory/PurchaseUnitFactoryTest.php
+++ b/tests/PHPUnit/ApiClient/Factory/PurchaseUnitFactoryTest.php
@@ -18,13 +18,14 @@ use function Brain\Monkey\Functions\expect;
 
 class PurchaseUnitFactoryTest extends TestCase
 {
+	private $wcOrderId = 1;
+	private $wcOrderNumber = '100000';
 
     public function testWcOrderDefault()
     {
-        $wcOrderId = 1;
         $wcOrder = Mockery::mock(\WC_Order::class);
-        $wcOrder
-            ->expects('get_order_number')->andReturn($wcOrderId);
+        $wcOrder->expects('get_order_number')->andReturn($this->wcOrderNumber);
+        $wcOrder->expects('get_id')->andReturn($this->wcOrderId);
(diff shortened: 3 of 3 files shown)


Dinamiko (approved):

modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
```diff
@@ -119,9 +119,8 @@ public function from_wc_order( \WC_Order $order ): PurchaseUnit {
 		$reference_id    = 'default';
 		$description     = '';
 		$payee           = $this->payee_repository->payee();
-		$wc_order_id     = $order->get_order_number();
-		$custom_id       = $this->prefix . $wc_order_id;
-		$invoice_id      = $this->prefix . $wc_order_id;
+		$custom_id       = $this->prefix . (string) $order->get_id();
```
danielhuesken: I think prefix make no sense for the custom id. For invoice id it is needed.

Also the setting say it is only for invoice.

AlexP11223: :heavy_check_mark:

Dinamiko (approved):

PR #371, opened 2021-11-18
Add retry counter meta to order to avoid duplicate invoice error on c…

### LINKED ISSUE #367: [RC] When paying for a subscription and vaulting fails, cart is cleared (423)
### Describe the bug

When trying to pay for a subscription with DCC using a credit card number that doesn't permit vaulting (for instance, any card from the CC generator), the order is not allowed to continue (which was fixed in #308) but for some reason the whole cart is cleared, which can result in a bunch of strange errors should the user attempt to re-submit the form.

### To reproduce

1. Setup DCC with vaulting enabled.
1. Add a subscription product to the cart.
1. Go to https://developer.paypal.com/developer/creditCardGenerator and generate some CC details.
1. Go to Checkout and enter the previously generated card details.
1. Click "Place Order".
1. Confirm that placing order fails because the method wasn't stored in the vault correctly.
1. Click "Place Order" again to see a weird error message.
1. Refresh the page and see the cart has been emptied, preventing users from re-attempting payment.

**Screenshots**
Screencast:

screencast1

### Expected behavior

Users should be allowed to use other CC details to complete payment. Card shouldn't be cleared.

Dinamiko:
Fixes #367.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
@@ -253,13 +253,19 @@ trait ProcessPaymentTrait {
 						}
 					}
 
+					// Adds retry counter meta to avoid duplicate invoice id error on consequent tries.
+					$wc_order->update_meta_data( 'ppcp-retry', (int) $wc_order->get_meta( 'ppcp-retry' ) + 1 );
+					$wc_order->save_meta_data();
+
 					$this->session_handler->destroy_session_data();
 					wc_add_notice( $error_message, 'error' );
 
 					return $failure_data;
 				}
 
+				WC()->cart->empty_cart();
 				$this->session_handler->destroy_session_data();
+
 				return array(
 					'result'   => 'success',
 					'redirect' => $this->get_return_url( $wc_order ),
--- a/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
+++ b/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
@@ -121,9 +121,11 @@ class PurchaseUnitFactory {
 		$payee           = $this->payee_repository->payee();
 		$wc_order_id     = $order->get_order_number();
 		$custom_id       = $this->prefix . $wc_order_id;
-		$invoice_id      = $this->prefix . $wc_order_id;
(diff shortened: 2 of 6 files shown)


modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
```diff
@@ -253,13 +253,19 @@ function ( Authorization $authorization ): bool {
...
+					$wc_order->update_meta_data( 'ppcp-retry', (int) $wc_order->get_meta( 'ppcp-retry' ) + 1 );
+					$wc_order->save_meta_data();
+
 					$this->session_handler->destroy_session_data();
 					wc_add_notice( $error_message, 'error' );
 
 					return $failure_data;
 				}
 
+				WC()->cart->empty_cart();
```
AlexP11223, line 266: Why do we do this? :thinking:

AlexP11223 (commented):
I don't know if this solution is ok, not sure how these invoices work, what happens with the previous "errored" invoices, and whether it is ok to have IDs like this.

modules/ppcp-wc-gateway/src/Gateway/ProcessPaymentTrait.php
Dinamiko: I´ve moved it from `OrderProcessor::process` to here because otherwise there were no chance to retry the payment when the payment is not saved on PayPal.

Dinamiko: @AlexP11223

In order to prevent subscriptions renewals to fail because no payment stored for customer we are introducing a check after the payment to see if the payment is saved on PayPal or not, if it´s not saved then we set the WC order as failed.

The problem is that if the customer tries again with the same order it gets a **DUPLICATE_INVOICE_ID**, here is how I currently implemented it:
1. POST order with Auth intent
2. PATCH
3. Authorize
4. Check if payment is saved, if so capture auth order.

The above works, however if point 4 fails, we void the auth order and start again:
1. POST order with Auth intent
2. PATCH
3. Authorize, here we get **DUPLICATE_INVOICE_ID**

The error is because as the PayPal authorization was successful, that order ID is “used”, so we have to increment or otherwise alter the invoice ID.

I´ve implemented the later by adding a counter at the end, so if `WC-001` is used, we alter it to `WC-001-1`.

AlexP11223 (approved):

PR #375, opened 2021-11-22
Enhance the UX of the Pay Later and Vaulting features

### LINKED ISSUE #174: Vaulting & Pay Later UI/UX
Progress has been made since #157, but I believe there are still some aspects that need fine-tuning and it'd be great if we could start discussing approaches or ideas.

1. Enabling vaulting is still cumbersome as you have to uncheck each and every _Pay Later_ checkbox one by one. This is not very user-friendly, especially considering that _Pay Later_ comes enabled by default. It might seem like a minor detail, but there's no global _Pay Later_ switch and the settings are scattered around the settings page, making things unnecessarily difficult.

   I'm no longer convinced the approach I suggested in #157 (displaying a confirmation dialog and doing the unchecking automatically with JS) is the best one, but I still think we should make things easier. Maybe the vaulting setting should only be disabled when the token doesn't have the necessary scope, but in every other case it should be possible to enable vaulting (without asking for confirmation) as long as it's clear that enabling it will disable Pay Later. The setting could look something like this:

   Screen Shot 2021-06-10 at 08 07 08

   Obviously the "warning" would only appear when vaulting hasn't been enabled.

2. It's not immediately clear _why_ the vaulting setting might be disabled. This notice at the top does explain why but it is easy to miss. It also seems a bit superfluous given _Pay Later_ is enabled by default, so it's a notice that almost every merchant is getting even if they don't need or want to use vaulting:
  
   Screen Shot 2021-06-10 at 07 55 13

   We should probably move the "explanation" next to the vaulting checkbox itself, so you only see it when you actually want to enable vaulting:
   
   Screen Shot 2021-06-10 at 07 54 52

   To make things worse, the disabled state plus the link to enable vaulting give the idea that the setting is disabled because you haven't done something with your PayPal account and not because some _Pay Later_ checkbox is checked off below.

   That said, if we go with what I suggested in (1), it wouldn't be necessary to disable the vaulting checkbox at all, except in the specific case of the token not containing the required scopes. Only in that particular case, we would display the link to "enable vaulting on your account".

3. _Pay Later_ settings are better at indicating why you can't use them, but the messaging remains even when you've disabled vaulting:

   screencast

   The messaging should reflect the current configuration (even if unsaved) and not be tied to the values in the database. Also, the warning should have an anchor link to the vaulting setting so that you can easily toggle it.

To summarize my thoughts: (I) as long as the scope is in the token, it should always be possible to enable vaulting (with a small warning indicating that doing so will disable _Pay Later_ messaging) (II) the UI in general (labels, warnings, etc.) should always reflect the current configuration as seen by the user (even if it hasn't been saved to the database).

Please let me know what you think.

a-danae:
**Issue**: #174

---

### Description

This PR implements the improvements suggested in #174:

- Disables the Vaulting checkbox only when vaulting isn't in the PayPal account's scope
- Displays the copy about enabling Vaulting in the PayPal account only when it's not in the account's scope
- Unchecks and disables the Pay Later checkboxes when Vaulting is enabled in the plugin, and vice versa
- Removes the admin notice mentioning that Pay Later is disabled due to Vaulting being enabled
- Removes the admin notice mentioning that Vaulting is disabled due to Pay Later being enabled
- Updates the UI dynamically following the selected options without having to reload the page

### Steps to test:

PayPal account's vaulting scope changes:
1. Disable vaulting on your PayPal account (I didn't find a way to do it on my sandbox account, so I made the method that checks whether vaulting is enabled in the scope return 'false' in the plugin)
2. Go to the WooCommerce -> Settings -> Payments -> PayPal -> Manage (or Set up if it isn't enabled)
3. Make sure the "Vaulting" checkbox is disabled, and that its label includes "To use vaulting features, you must enable vaulting on your account" 
4. Go to WooCommerce -> Settings -> Payments -> PayPal Card Processing -> Manage (or Set up if it isn't enabled)
5. Make sure step 3 is also true on this screen

1. Enable vaulting on your PayPal account
2. Go to the WooCommerce -> Settings -> Payments -> PayPal -> Manage (or Set up if it isn't enabled)
3. Make sure the "Vaulting" checkbox is enabled, and that its label is "Enable saved cards and subscription features on your store." alone, without "To use vaulting features, you must enable vaulting on your account" 
4. Go to WooCommerce -> Settings -> Payments -> PayPal Card Processing -> Manage (or Set up if it isn't enabled)
5. Make sure step 3 is also true on this screen

PayPal's settings page UI changes (do all of this without saving the settings):
1. Make sure Vaulting is enabled on your PayPal account
2. Go to the WooCommerce -> Settings -> Payments -> PayPal -> Manage (or Set up if it isn't enabled)
3. Make sure the Vaulting setting is disabled
4. Confirm it now has a description saying "This will disable all Pay Later messaging on your site"
5. Enable the "Enable message on Checkout", "Enable message on Single Product", and "Enable message on Cart" checkboxes
6. Confirm that the label for these checkboxes is "Enable on {location}"
7. Disable the Vaulting feature
8. Confirm that the checkboxes for "Enable message on Checkout", "Enable message on Single Product", and "Enable message on Cart" are now unchecked and disabled
9. Confirm that the labels for these settings changed from "Enable on {location}" to "You have PayPal vaulting enabled, that's why Pay Later Messaging options are unavailable now. You cannot use both features at the same time."
10. Confirm that their child settings are now hidden

This behavior should persist on each step when saving and reloading the page.

1. Make sure vaulting is enabled on your account
2. Go to the WooCommerce -> Settings -> Payments -> PayPal -> Manage (or Set up if it isn't enabled)
3. Disable vaulting on the plugin settings
4. Enable the Pay Later settings
5. Save the settings
6. Disable vaulting on your PayPal account
7. Go to the plugin's settings (or reload the page you were on)
8. Make sure the behavior is consistent with when the vaulting setting is disabled in the plugin

### Documentation

[This section](https://woocommerce.com/document/woocommerce-paypal-payments/#section-6) from WooCommerce docs would behave differently than what's described there after this PR is merged.
Under `To make Vault work with WooCommerce Subscriptions, follow the steps below.`, steps 2 and 3 won't be needed. You can now enable Vaulting without having to disable Pay Later and saving first.

Closes #174


modules/ppcp-wc-gateway/services.php
```diff
@@ -2026,4 +2009,44 @@ static function ( ContainerInterface $container ): AuthorizeOrderActionNotice {
...
+		}
+
+		$vaulting_label .= '<p class="description">';
+		$vaulting_label .= __( 'This will disable all Pay Later messaging on your site.', 'woocommerce-paypal-payments' );
+		$vaulting_label .= '</p>';
+
+		return $vaulting_label;
+	},
+
+	'button.helper.pay-later-label'                => static function ( ContainerInterface $container ): string {
```
a-danae: @jorgeatorres I wasn't sure about the convention for naming the keys and introducing items in this array of functions. I tried to follow what I saw but I might have missed the point. Glad to update it.

modules/ppcp-wc-gateway/services.php
```diff
@@ -2026,4 +2009,44 @@ static function ( ContainerInterface $container ): AuthorizeOrderActionNotice {
...
+				'<a
+					href="https://docs.woocommerce.com/document/woocommerce-paypal-payments/#enable-vaulting-on-your-live-account"
+					target="_blank"
+				>',
+				'</a>'
+			);
+		}
+
+		$vaulting_label .= '<p class="description">';
+		$vaulting_label .= __( 'This will disable all Pay Later messaging on your site.', 'woocommerce-paypal-payments' );
```
a-danae, line 2168: @jorgeatorres I understood that this text should be displayed when Vaulting is included in the PayPal account's scope, but I decided to always display it instead to avoid surprises after going through the process of including vaulting in the account.

Like, if having Pay Later is a deal-breaker as a merchant, I think it'd be more convenient to know about its incompatibility with vaulting early on. As always, glad to change this behavior.

modules/ppcp-wc-gateway/services.php
AlexP11223: There is no clear convention, but since it seems to be related to the settings fields (`wcgateway.settings.fields`), then maybe it should be something like `wcgateway.settings.fields.pay-later-label`.

btw if you are confused by all this container/modularity stuff, this may help: https://gist.github.com/gmazzap/0f9d658c183ff0aaf057a1a9229c3280

a-danae: Makes sense, function name updated.

Thanks for the suggestion and for the reasoning behind the container/modularity approach, @AlexP11223!

a-danae:
@AlexP11223 @Dinamiko  I noticed a couple of issues when testing the latest pre-release package: 

1. When Vaulting was enabled and you disabled it, the description in the Messaging options remained saying that they were disabled because Vaulting was enabled, at least until you reload the page.
2. The checkbox for Vaulting is disabled in the PayPal settings page when Messaging is enabled, but when you go to the Card Processing settings, the checkbox for Vaulting can be toggled.

[Here's a video](https://d.pr/v/IRUiI7) with both behaviors.

I was about to create an issue for that, but I recalled this PR of mine and noticed these behaviors weren't present here. Would any of you please review it? I've updated it with trunk. I know other things relevant to this PR might have changed, glad to update whatever is needed.

AlexP11223 (approved):

a-danae:
@AlexP11223 should I merge this, or will you handle it?

Dinamiko (approved):

Dinamiko:
Hi @a-danae, we´ll merge it, thanks.

PR #381, opened 2021-11-24
Fix error handler clear messages

### LINKED ISSUE #366: [RC] Error messages are not cleared even when checkout is re-attempted (DCC) (424)
### Describe the bug
When using DCC and paying with a credit card that has been disabled or a card number known to fail, the error messages at the top of the checkout form are not cleared between attempts, which shouldn't happen, particularly when the error messages are the same.

### To reproduce

1. Enable and configure DCC.
1. Initiate the purchase workflow.
1. On the checkout screen, use credit card number "4111 1111 1111 1112", an expiration date in the future and any 3-digit number as CVV.
1. Click "Place Order".
1. Confirm you get an error message with text "Unfortunately, your credit card details are not valid."
1. Click "Place Order" again.
1. Confirm that the error messages from previous attempts persist.

The same happens when you try to pay with a credit card that is not being allowed in WC > Settings > Payments > PayPal Card Processing > "Disable specific credit cards".

**Screenshots**

image

### Expected behavior

Error messages should be cleared between attempts.

Dinamiko:
Fixes #366.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ErrorHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ErrorHandler.js
@@ -73,11 +73,13 @@ class ErrorHandler {
 
     clear()
     {
-        if (! this.wrapper.classList.contains('woocommerce-error')) {
+        if (this.messagesList === null) {
             return;
         }
-        this.wrapper.classList.remove('woocommerce-error');
-        this.wrapper.innerText = '';
+
+        while( this.messagesList.firstChild ){
+            this.messagesList.removeChild( this.messagesList.firstChild );
+        }
     }
 }
 
--- a/modules/ppcp-button/resources/js/modules/ErrorHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ErrorHandler.js
@@ -77,9 +77,7 @@ class ErrorHandler {
             return;
         }
 
-        while( this.messagesList.firstChild ){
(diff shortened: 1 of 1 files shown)


modules/ppcp-button/resources/js/modules/ErrorHandler.js
```diff
@@ -73,11 +73,13 @@ class ErrorHandler {
...
-        if (! this.wrapper.classList.contains('woocommerce-error')) {
+        if (this.messagesList === null) {
             return;
         }
-        this.wrapper.classList.remove('woocommerce-error');
-        this.wrapper.innerText = '';
+
+        while( this.messagesList.firstChild ){
+            this.messagesList.removeChild( this.messagesList.firstChild );
+        }
```
AlexP11223: maybe cleaner to just `.innerHTML = ''`?

https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#replacing_the_contents_of_an_element

AlexP11223 (changes requested):

modules/ppcp-button/resources/js/modules/ErrorHandler.js
Dinamiko: Done :)

AlexP11223 (approved):

Issue #377, opened 2021-11-23
New additions for system report status (254)


Dinamiko:
This information should be added to the PCP status in the WooCommerce system report. Ordered by relevance (most important at the top):

- Reference Transactions: Yes/No - whether or not Reference Transactions are enabled for the connected account
- Logging enabled: Yes/No - whether or not logging is enabled
- Webhook status: OK/Unknown - whether or not the last webhook was successful
- Checkout: Yes/No - whether or not the PayPal Checkout gateway was configured previously (wp_options entry remains and could be checked)
- WooCommerce currency supported: Yes/No - whether or not PayPal supports the default store currency

Dinamiko:
Added in #382.

PR #382, opened 2021-11-25
Improve status report

AlexP11223:
Implements #377

Also added separate field for `export-label` so that the label in the exported report (when sending to support) is always in English.

And replaced Yes/No to :heavy_check_mark: icon or dash –, because that's how WC does it for most of the other items in the Status, and to avoid translations (looks like there is no anything like `export-value`).

image

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
@@ -1,0 +1,138 @@
+<?php
+/**
+ * The billing agreements endpoint.
+ *
+ * @package WooCommerce\PayPalCommerce\ApiClient\Endpoint
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\ApiClient\Endpoint;
+
+use stdClass;
+use WooCommerce\PayPalCommerce\ApiClient\Authentication\Bearer;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Webhook;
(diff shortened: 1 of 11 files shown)


modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
```diff
@@ -0,0 +1,138 @@
...
+ * @package WooCommerce\PayPalCommerce\ApiClient\Endpoint
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\ApiClient\Endpoint;
+
+use stdClass;
+use WooCommerce\PayPalCommerce\ApiClient\Authentication\Bearer;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Webhook;
```
Dinamiko: Regarding IDE seems some classes are not used:
```
use WooCommerce\PayPalCommerce\ApiClient\Entity\Webhook;
use WooCommerce\PayPalCommerce\ApiClient\Entity\WebhookEvent;
use WooCommerce\PayPalCommerce\ApiClient\Factory\WebhookEventFactory;
use WooCommerce\PayPalCommerce\ApiClient\Factory\WebhookFactory;
```

Dinamiko (approved):

modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
AlexP11223: :heavy_check_mark:

PR #385, opened 2021-11-30
Ensure hosted fields `cardholderName` is submitted to JavaScript SDK

### LINKED ISSUE #378: Customer details not available in order and name gets replaced by xxx@dcc2.paypal.com (359)
Customers who choose PayPal Card Processing, sometimes the customer Name doesn't show up in the PayPal Dashboard and Transactions (but these details show up fine in the WooCommerce Orders and Emails). For Name, it will often be **example@dcc2.paypal.com**. Also have 1 example where the name shows up as “CHASE VISA CARDHOLDER”.

### To reproduce
Try placing an order in PayPal Card Processing gateway without adding First and Last name in checkout form.
It does create the PayPal order and displays WooCommerce checkout errors afterwards.

### Expected behavior
Cardholder´s name should be transmitted to PayPal, even when the user checks out without a user account.
When there is no Cardholder´s name available it should not create the PayPal order and display the corresponding error message.

### Environment
 - WordPress Version: 3.8
 - WooCommerce Version: 3.9
 - WooCommerce PayPal Payments: 1.6.2

Dinamiko:
Fixes #378.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
@@ -199,8 +199,12 @@ class CreditCardRenderer {
                 hostedFieldsData.cardholderName = this.defaultConfig.payer.given_name + ' ' + this.defaultConfig.payer.surname;
             }
             if (!hostedFieldsData.cardholderName) {
-                const firstName = document.getElementById('billing_first_name') ? document.getElementById('billing_first_name').value : '';
-                const last = document.getElementById('billing_last_name') ? document.getElementById('billing_last_name').value : '';
+                const firstName = document.getElementById('billing_first_name') ? document.getElementById('billing_first_name').value
+                    : document.getElementById('shipping_first_name') ? document.getElementById('shipping_first_name').value
+                    : '';
+                const lastName = document.getElementById('billing_last_name') ? document.getElementById('billing_last_name').value
+                    : document.getElementById('shipping_last_name') ? document.getElementById('shipping_last_name').value
+                    : '';
 
                 if (!firstName || !lastName) {
                     this.spinner.unblock();
--- a/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
@@ -194,6 +194,23 @@ class CreditCardRenderer {
             if (contingency !== 'NO_3D_SECURE') {
                 hostedFieldsData.contingencies = [contingency];
             }
+
+            if (this.defaultConfig.payer) {
+                hostedFieldsData.cardholderName = this.defaultConfig.payer.given_name + ' ' + this.defaultConfig.payer.surname;
+            }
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -194,6 +194,27 @@ class CreditCardRenderer {
             if (contingency !== 'NO_3D_SECURE') {
                 hostedFieldsData.contingencies = [contingency];
             }
+
+            if (this.defaultConfig.payer) {
+                hostedFieldsData.cardholderName = this.defaultConfig.payer.given_name + ' ' + this.defaultConfig.payer.surname;
+            }
+            if (!hostedFieldsData.cardholderName) {
+                const firstName = document.getElementById('billing_first_name') ? document.getElementById('billing_first_name').value
+                    : document.getElementById('shipping_first_name') ? document.getElementById('shipping_first_name').value
```
AlexP11223: This looks weird, why do we send the name from shipping address as cardholder name?

Dinamiko: @AlexP11223 

> This looks weird, why do we send the name from shipping address as cardholder name?

From where else could we get it when the customer is not logged in?

It appears that when no cardholder name is sent with the transaction, it causes displaying the name in PayPal as the DCC email (ex. example@dcc2.paypal.com or “CHASE VISA CARDHOLDER”).

https://developer.paypal.com/docs/business/checkout/advanced-card-payments/integrate

This PR ensures `cardholderName` is sent to hosted fields JSSDK otherwise it displays an error message and does not submit the hosted fields.

Dinamiko: Another solution could be adding a new "Card Holder Name" field to hosted fields:

AlexP11223: Shipping address may not be related to payment. 

Why not just show error if no name in the billing address?

Dinamiko: > Shipping address may not be related to payment.

yup makes totally sense, I´ve just removed shipping as fallback, thanks.

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -684,17 +684,18 @@ private function localize_script(): array {
...
+					'fields_not_valid'         => __(
 						'Unfortunately, your credit card details are not valid.',
 						'woocommerce-paypal-payments'
 					),
-					'card_not_supported' => __(
+					'card_not_supported'       => __(
 						'Unfortunately, we do not support your credit card.',
 						'woocommerce-paypal-payments'
 					),
+					'cardholder_name_required' => __( 'Cardholder\'s first and last name required, please fill checkout form required fields.', 'woocommerce-paypal-payments' ),
```
AlexP11223: Maybe should add `the` before `checkout`.

And `are` before `required`.

AlexP11223 (approved):

PR #423, opened 2022-01-04
Add customer id for guest users purchasing subscriptions

### LINKED ISSUE #422: Allow guest users to purchase subscription products from checkout page (434)
### Describe the Bug

Currently it's not possible to save payments for guest users because we need to generate a client token identifying the buyer by a unique `customer_id`: https://developer.paypal.com/limited-release/vault-payment-methods/vault-sdk/#link-stepgenerateaclienttokenforyourbuyer

### To Reproduce

1. As a guest add a subscription product to the cart and try to place the order.

### Expected Behavior

Guest user should be able to purchase subscription products using PayPal or Credit Card payment gateways.

### Acceptance (Sandbox)

**Credit Card gateway**
- As a guest user add a subscription product to the cart and go to checkout. 
- Place the order using a vault ready credit card number (see https://developer.paypal.com/limited-release/vault-payment-methods/vault-sdk#link-testcards).
- As a customer check that the payment is saved correctly on My account / PayPal payments.
- As a customer check that WC order status is processing and subscription is active.
- As a customer try to renewal the subscription (by clicking Renew now button) using the saved credit card payment.
- As an admin go to subscription and process renewal (by selecting “process renewal” in Subscription actions).

**PayPal gateway**
Same steps as in credit card but using PayPal button. Please use a brand new PayPal personal account in this case, as it is not currently possible to vault on the same PayPal account using a different customer id.

### Environment

 - WordPress Version: 5.8
 - WooCommerce Version: 6.0
 - Plugin Version: 1.6.4

Dinamiko:
Fixes #422.

Because we need to send a customer ID to PayPal in order to generate the client token and it does not exist yet for guest customers, we could create a unique ID and add it the WC session.

Subscription plugin will create the customer during checkout, so we can add that unique ID stored in the session to the created customer as meta. Then we can use that ID stored in the customer meta to get its saved payments.


modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -177,10 +177,6 @@ public function __construct(
 	 */
 	public function render_wrapper(): bool {
 
-		if ( ! $this->can_save_vault_token() && $this->has_subscriptions() ) {
```
AlexP11223: Why it was removed?

AlexP11223 (commented):
hm, would be great to make the subscriptions plugin create the customer before handling the payment to avoid all this complexity. Or maybe at least somehow extract/isolate the ID retrieval to avoid duplicating the user meta checks every time we need the ID.

modules/ppcp-button/src/Assets/SmartButton.php
Dinamiko: @AlexP11223 

> make the subscriptions plugin create the customer before handling the payment

How can we do that?

AlexP11223: dunno, depends on how it does that. Maybe WC know something about that, or maybe they had issues like that before and there is a simpler solution?

Dinamiko: It does it by adding `createaccount` to the request :
```
	/**
	 * During the checkout process, force registration when the cart contains a subscription.
	 *
	 * @since 1.1
	 * @param $woocommerce_params This parameter is not used.
	 */
	public static function force_registration_during_checkout( $woocommerce_params ) {
		if ( WC_Subscriptions_Cart::cart_contains_subscription() && ! is_user_logged_in() ) {
			$_POST['createaccount'] = 1;
		}
	}
```

The problem I see is that we need the generated `data-client-token` added to JS script on page load:



let´s imagine the guest is visiting a subscription product page, even if it´s technically possible to create an user programatically at this point, what happens if the customer leaves the site, the merchant is going to have a bunch of customers created in the system for no reason...

Dinamiko: In fact at this point is not possible to create the customer as we don´t know the email, for example using:
`function wc_create_new_customer( $email, $username = '', $password = '', $args = array() ) {`

AlexP11223: hm, if we need it on page load, then yeah, I guess there is no simple way.

But then we should at least

> somehow extract/isolate the ID retrieval to avoid duplicating the user meta checks every time we need the ID.

and check that we did not miss any case. 

I think here we need this too?

https://github.com/woocommerce/woocommerce-paypal-payments/blob/21aa8980d642c0c3745d72b572e835aa4e97757e/modules/ppcp-subscription/src/RenewalHandler.php#L152

Dinamiko: @AlexP11223 I´ve extracted customer ID retrieval and reverted the conditional in `modules/ppcp-button/src/Assets/SmartButton::render_wrapper`.

Regarding cases where we need customer id from user, is only needed when requesting saved payments here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/e9cafc762883beb62aa010a377590334b21c05a9/modules/ppcp-api-client/src/Endpoint/PaymentTokenEndpoint.php#L96

and when generating the client token here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/e9cafc762883beb62aa010a377590334b21c05a9/modules/ppcp-api-client/src/Endpoint/IdentityToken.php#L108

AlexP11223 (approved):

developcrk:
Hi @AlexP11223 and @Dinamiko,

I'm not sure if I can add a comment here but I read that you are worried about the user creation. Well you shouldn't. On stripe gateway every time some one attempts to pay a user, a subscription and an order is created no matter if the payment failed.

So we end up with a bunch of users and on-hold orders anyway.

I hope this help.

Thanks for the great job you are doing!

PR #449, opened 2022-01-24
Generate unique id based on the length of the prefix

### LINKED ISSUE #448: [pre-release] client_id string too long when invoice prefix has >9 characters (511)
### Describe the Bug

When the invoice prefix is longer than 9 characters, then the smart buttons will not be loaded for logged-out guest users when Vaulting is enabled as the client_id would be too long.

### To Reproduce

1. Install 1.6.5-test2 pre-release (https://github.com/woocommerce/woocommerce-paypal-payments/releases/tag/1.6.5-test2)
2. Enable Vaulting
3. Set invoice prefix with 10 or more characters
4. Log out
4. Visit checkout or any page with a smart button as guest

### Expected Behavior

Smart buttons should appear for the user.

### Actual Behavior

Smart buttons will not load
```
details: [{field: "/customer_id", value: "invprefix-61eb040ddadba", location: "body",…}]
0: {field: "/customer_id", value: "invprefix-61eb040ddadba", location: "body",…}
description: "The value of a field is either too short or too long"
field: "/customer_id"
issue: "INVALID_STRING_LENGTH"
location: "body"
value: "invprefix-61eb040ddadba"
````

### Possible cause
The invoice prefix can contain a maximum of 15 characters.

The newly generated unique customer id for the guest users contains 13 characters. Example: 61eb040ddadba

The client_id can have a maximum length of 22 characters. (more info: https://inpsyde.slack.com/archives/CUNAX990E/p1638547814152300?thread_ts=1638544042.152200&cid=CUNAX990E)

When there are more than 9 characters in the invoice prefix, the generated client_id would exceed the limit of 22 characters for logged-out users.

### Suggested solution

Shorten the newly generated unique customer id from 13 characters to 7 at most. This seems better than reducing the invoice prefix character count.
When a merchant uses all 15 characters available for the invoice then the client_id could still be generated with 22 characters.

### Environment
 - WordPress Version: 5.8
 - WooCommerce Version: 6.1
 - Plugin Version: 1.6.5-test2

### PR DESCRIPTION
**Pull request by Dinamiko**
Fixes #448.

### DIFF
--- a/modules/ppcp-api-client/src/Repository/CustomerRepository.php
+++ b/modules/ppcp-api-client/src/Repository/CustomerRepository.php
@@ -13,6 +13,7 @@ namespace WooCommerce\PayPalCommerce\ApiClient\Repository;
  * Class CustomerRepository
  */
 class CustomerRepository {
+	const CLIENT_ID_MAX_LENGTH = 22;
 
 	/**
 	 * The prefix.
@@ -37,13 +38,14 @@ class CustomerRepository {
 	 * @return string
 	 */
 	public function customer_id_for_user( int $user_id ): string {
+		$unique_id = $this->generate_unique_id();
 		if ( 0 === $user_id ) {
 			$guest_customer_id = WC()->session->get( 'ppcp_guest_customer_id' );
 			if ( is_string( $guest_customer_id ) && $guest_customer_id ) {
 				return $guest_customer_id;
 			}
 
-			$unique_id = $this->prefix . uniqid();
+			$unique_id = $this->generate_unique_id();
 			WC()->session->set( 'ppcp_guest_customer_id', $unique_id );
 
 			return $unique_id;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/ppcp-api-client/src/Repository/CustomerRepository.php [hunk 1]
```diff
@@ -56,4 +58,17 @@ public function customer_id_for_user( int $user_id ): string {
...
+	 *
+	 * @return string
+	 */
+	protected function generate_unique_id(): string {
+		$offset = self::CLIENT_ID_MAX_LENGTH - strlen( $this->prefix );
+
+		return strlen( uniqid() ) > $offset
+			? $this->prefix . substr( uniqid(), $offset )
+			: $this->prefix . uniqid();
+	}
```
AlexP11223: This seems difficult to understand and not sure if correct. Also two `uniqid` calls, possibly returning different lengths.

Maybe better

    return substr( $$this->prefix . strrev( uniqid() ), 0, self::CLIENT_ID_MAX_LENGTH );

AlexP11223 (changes requested):

modules/ppcp-api-client/src/Repository/CustomerRepository.php
Dinamiko: @AlexP11223 that´s much better, thanks!

AlexP11223 (approved):

PR #463, opened 2022-02-01
Update 3D Secure (521)

Chaithi:
Change the 3D SECURE contingency to SCA_ALWAYS

**Issue**: The 3DS Contingency values have changed with introduction of 3DS v2.

**Ticket**:

**Slack Thread**:

---

### Description

Update 3D_SECURE contingency to SCA_ALWAYS

### Documentation

No documentation changes needed.

### Changelog Entry

Updates 3D Secure contingency,

Closes # .

PR #464, opened 2022-02-03
Ensure 3ds contingency uses `SCA_ALWAYS` instead of `3D_SECURE` (521)

Dinamiko:
Add improvements to #463.

This PR ensures that `SCA_ALWAYS` is used instead of `3D_SECURE` stored in the database: https://github.com/woocommerce/woocommerce-paypal-payments/pull/463/files#diff-01a2758cc519d8adb065af0350c9459edb64a45576b4a2e527e8d2059c7889cbL1922-R1922

### Testing package
[PCP-521-woocommerce-paypal-payments.zip](https://github.com/woocommerce/woocommerce-paypal-payments/files/8001383/PCP-521-woocommerce-paypal-payments.zip)

### Acceptance criteria
1. For new plugin users on PSD2 countries ([see list here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/services.php#L586-L613)) default 3DS contingency should be `SCA_WHEN_REQUIRED`.
2. For existing plugin users with `3D_SECURE` option saved, `SCA_ALWAYS` option value should be automatically updated*

*Note that only the value is updated, the option label is still the same (Always trigger 3D Secure).
Also note that in order to trigger the update the plugin needs to be upgraded, the provided package contains a higher plugin version which needs to be uploaded and installed from Plugins admin screen (replace existing with the new one).

### DIFF
--- a/modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
+++ b/modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
@@ -209,25 +209,6 @@ class SettingsListener {
 	}
 
 	/**
-	 * Ensure 3DS contingency use `SCA_ALWAYS` instead of `3D_SECURE`.
-	 *
-	 * @return void
-	 * @throws NotFoundException When a setting was not found.
-	 */
-	public function listen_for_3d_secure_contingency(): void {
-		if ( ! $this->is_valid_site_request() || $this->settings->get( '3d_secure_contingency' ) !== '3D_SECURE' ) {
-			return;
-		}
-
-		$this->settings->set( '3d_secure_contingency', 'SCA_ALWAYS' );
-		$this->settings->persist();
-
-		$redirect_url = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway' );
-		wp_safe_redirect( $redirect_url, 302 );
-		exit;
-	}
-
-	/**
 	 * Listens to the request.
 	 *
 	 * @throws \WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException When a setting was not found.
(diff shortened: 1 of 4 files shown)


modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
```diff
@@ -208,6 +208,25 @@ function () use ( $exception ) {
...
+	 */
+	public function listen_for_3d_secure_contingency(): void {
+		if ( ! $this->is_valid_site_request() || $this->settings->get( '3d_secure_contingency' ) !== '3D_SECURE' ) {
+			return;
+		}
+
+		$this->settings->set( '3d_secure_contingency', 'SCA_ALWAYS' );
+		$this->settings->persist();
+
+		$redirect_url = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway' );
```
AlexP11223: Maybe should redirect to the cards tab? `admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway&ppcp-tab=ppcp-credit-card-gateway`

Or maybe redirect not needed here at all.

Also as I understand it is executed only if the merchant saves settings. Maybe better to do that simply on some kind of plugin init/upgrade, such as `woocommerce_paypal_payments_gateway_migrate`.

modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -132,6 +132,17 @@ static function () use ( $c ) {
 				$endpoint->handle_request();
 			}
 		);
+
+		add_action(
+			'woocommerce_paypal_payments_gateway_migrate',
+			static function () use ( $c ) {
+				$settings = $c->get( 'wcgateway.settings' );
```
AlexP11223, line 140: I would add a type hint here to help psalm and IDE.

    assert( $settings instanceof Settings );

modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
Dinamiko: yup, much better on plugin upgrade, thanks!

AlexP11223 (changes requested):

modules/ppcp-wc-gateway/src/WCGatewayModule.php
Dinamiko: Done! :)

AlexP11223 (approved):

PR #474, opened 2022-02-07
Add transaction id to order note when refund is received through webhook

### LINKED ISSUE #473: Add transaction ID to WC order and order note when refund is received (537)
As part of #456 we added transaction ID into the order note, we need to add it again when a refund is received because the transaction ID is not the same as the previous one.

### Acceptance criteria
Create a refund on PayPal and then check that order and order note displays the correct transaction ID. Note that it can take some minutes until the webhook is received.

Dinamiko:
Fixes #473.


AlexP11223:
Maybe we should also set the transaction ID of the order? (using `TransactionIdHandlingTrait`)

Dinamiko:
@AlexP11223 Done, thanks.

AlexP11223 (approved):

Issue #467, opened 2022-02-03
PayPal Payments doesn't set transaction fee metadata (307)


danielbreves:
### Describe the Bug

PayPal Checkout used to set the transaction fee as an order metadata [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/85a478de146459418b23854cca4a60f073344dbf/includes/functions.php#L111), PayPal Payments doesn't.

### Expected Behavior

There should be some mechanism for plugins to read PayPal transaction fees.

### Environment

 - WordPress Version 5.9
 - WooCommerce Version 6.1.1

Dinamiko:
Implemented in #490.

PR #490, opened 2022-02-14
Add fees to meta and show on the order

### LINKED ISSUE #489: Show  PayPal fee information in order (307)
The old PayPal Checkout Gateway plugin referenced the PayPal Fee and actual payout amount after all applied fees:

image

PayPal Payments currently does not display these amounts in the WooCommerce order.

AlexP11223:
Fixes #489 and #467

The fees are added to the order meta, and rendered on the admin order page if the meta is available, similarly to the old plugin.

The fees are available in the API after completing capture, so we need to catch the capture completion. It happens in multiple places, but, unless I missed something, it seems like they all call `handle_capture_status` for updating the WC order. So I added firing of `woocommerce_paypal_payments_order_captured` action here, and in the handler I save the fees from the capture to the meta.


Dinamiko (approved):

danielbreves:
Hi @AlexP11223 @Dinamiko thanks for working on this. Any ETA on when it might get merged and released?

Dinamiko:
Hi @danielbreves, thanks to you for pointing it out and for the initial work :)

We´ve just merged it so if nothing goes wrong the idea is to release it during this week.

Issue #499, opened 2022-02-17
Hide smart buttons for free products and zero-sum carts (500)


Narek13:
### Describe the Bug

Paypal doesn't not support zero-sum payments(when they are $0) so the buttons should be hidden in the cart/checkout when user adds a **$0** item to their order

### To Reproduce

1. create product with $0 price
2. smart button on single product page & cart will load
3. checkout will not work as PayPal does not support zero-sum payments

### Screenshots

![1d241ae9-b114-4a58-b278-058ab687d450](https://user-images.githubusercontent.com/11319597/154488180-f4350c4f-7d82-486b-b1e8-6f934a129c2d.png)

### Expected Behavior

Buttons should not load when the order total is $0.

### Actual Behavior

Buttons are loaded

Dinamiko:
Fixed in #500.

PR #500, opened 2022-02-17
Check if price total is not zero before showing buttons

Narek13:
**Issue**: #499

---

### Description

Paypal doesn't not support zero-sum payments(when they are $0) so the buttons should be hidden in the cart/checkout when user adds a **$0** item to their order.
The PR will add a check to see if the product price amount is not zero before showing the buttons.

### Steps to Test

1. create product with $0 price
2. smart button on single product page & cart will load
3. checkout will not work as PayPal does not support zero-sum payments

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -331,7 +331,6 @@ class SmartButton implements SmartButtonInterface {
 		if (
 			( is_product() || wc_post_content_has_shortcode( 'product_page' ) )
 			&& ! $not_enabled_on_product_page
-			&& ! $this->is_product_price_total_zero()
 		) {
 			add_action(
 				$this->single_product_renderer_hook(),
@@ -1085,21 +1084,5 @@ class SmartButton implements SmartButtonInterface {
 	protected function is_cart_price_total_zero(): bool {
         // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison
 		return WC()->cart->get_cart_contents_total() == 0;
-	}
-
-	/**
-	 * Check if current product price total is 0.
-	 *
-	 * @return bool true if is 0, otherwise false.
-	 */
-	protected function is_product_price_total_zero(): bool {
-		if ( ! is_product() ) {
-			return false;
-		}
-
-		$product = wc_get_product();
-
-        // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison
-		return $product && $product->get_price() == 0;
 	}
 }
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -331,6 +331,7 @@ private function render_button_wrapper_registrar(): bool {
 		if (
 			is_cart()
 			&& ! $not_enabled_on_cart
+			&& ! $this->is_cart_price_total_zero()
```
AlexP11223: Maybe better just check once at the start of the function and `return`?

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -568,7 +572,7 @@ public function dcc_renderer() {
 			return;
 		}
 
-		// phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
+        // phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
```
AlexP11223, line 576: :thinking:

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1070,4 +1074,30 @@ private function mini_cart_button_renderer_hook(): string {
...
 	}
+
+	/**
+	 * Check if cart product price total is 0.
+	 *
+	 * @return bool true if is 0, otherwise false.
+	 */
+	protected function is_cart_price_total_zero(): bool {
+		$cart_total = WC()->cart->get_cart_contents_total();
+		return ! ( $cart_total > 0 );
```
AlexP11223: Why not just `== 0` or `<= 0`? :)

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: That will not be better I think because there is also another [action for single_product_renderer_hook](https://github.com/woocommerce/woocommerce-paypal-payments/blob/164b74fdcba08b31c6543b224752bf9fde9eb7fc/modules/ppcp-button/src/Assets/SmartButton.php#L353) and for this one, another, [is_product_price_total_zero](https://github.com/woocommerce/woocommerce-paypal-payments/blob/164b74fdcba08b31c6543b224752bf9fde9eb7fc/modules/ppcp-button/src/Assets/SmartButton.php#L351) function is used to check

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: is done by `yarn docker:fix-lint`

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: efca8ba27dde17fea3dc027f732bb2ac5306e635 done.

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: anyway I have improved this a little here: 6de1aa2ae8f24966b4becea855ccb172d0c0ad52
I have just put action for `single_product_renderer_hook` at the top of the function and then used early return for the rest of the function

AlexP11223 (approved):

Dinamiko (approved):

PR #504, opened 2022-02-21
Improve detection of click event handler

### LINKED ISSUE #503: DCC orders randomly failing (505)
### Describe the Bug

This issue is about cases where the PayPal Card Processing would fail.

Following update 1.7.0, we got a new error message that occurs when the behavior occurs:
`No PayPal order found in the current WooCommerce session`

The above error message is displayed because for some reason “Place order” button does not call create order endpoint but WC process checkout.

It seems like switching other payment gateways and then switching back to PayPal Card Processing results in some kind of reload, and then our event listeners are not added (https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js#L125) because `formSubmissionSubscribed` is true.

### To Reproduce
1. Enable PayPal Card Processing
2. Enable other payment gateways
3. Add product to cart, visit checkout
4. Switch over some payment gateways
5. Switch back to PayPal Card Processing gateway 
6. Input credit card details and hit Place order

### Expected Behavior
Order processes properly.

### Actual Behavior
Order instantly fails with error above.

### Environment
 - WordPress Version: 5.9
 - WooCommerce Version: 6.2
 - Plugin Version: 1.7.0-test2

Dinamiko:
Fixes #503.


modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -122,16 +122,18 @@ class CreditCardRenderer {
...
-                document.querySelector(wrapper + ' button').addEventListener(
-                    'click',
-                    event => {
-                        event.preventDefault();
-                        this._submit(contextConfig);
-                    }
-                );
-                this.formSubmissionSubscribed = true;
-            }
+           const checkoutForm = document.forms.checkout;
```
AlexP11223: hm, does it work with mollie now? I thought we would need to add the attribute to something that gets reloaded during this switch (that is the card form).

modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -122,16 +122,18 @@ class CreditCardRenderer {
...
+           if(checkoutForm.getAttribute('data-ppcp-subscribed') !== true) {
```
AlexP11223: missing space

modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -122,16 +122,18 @@ class CreditCardRenderer {
...
 
-            if (!this.formSubmissionSubscribed) {
-                document.querySelector(wrapper + ' button').addEventListener(
-                    'click',
-                    event => {
-                        event.preventDefault();
-                        this._submit(contextConfig);
-                    }
-                );
-                this.formSubmissionSubscribed = true;
```
AlexP11223, line 133: `this.formSubmissionSubscribed = false;` in constructor can be removed now.

modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
Dinamiko: yup, it works adding the attribute to checkout form. 

I tried adding it to the hosted fields form wrapper (https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js#L75) but it does not work.

It works on the hosted fields button wrapper (https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js#L15) so this is what I used finally.

AlexP11223 (approved):

PR #515, opened 2022-02-25
Ensure `disable_funding` setting exist before getting it

Dinamiko:
Fixes an onboarding issue when onboarding in a fresh plugin install.

### DIFF
--- a/modules/ppcp-onboarding/src/Endpoint/LoginSellerEndpoint.php
+++ b/modules/ppcp-onboarding/src/Endpoint/LoginSellerEndpoint.php
@@ -147,7 +147,7 @@ class LoginSellerEndpoint implements EndpointInterface {
 
 			$accept_cards    = (bool) ( $data['acceptCards'] ?? true );
 			$funding_sources = array();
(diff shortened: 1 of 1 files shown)


modules/ppcp-onboarding/src/Endpoint/LoginSellerEndpoint.php
```diff
@@ -147,7 +147,7 @@ public function handle_request(): bool {
 
 			$accept_cards    = (bool) ( $data['acceptCards'] ?? true );
 			$funding_sources = array();
-			if ( $this->settings->get( 'disable_funding' ) ) {
+			if ( $this->settings->has( 'disable_funding' ) && $this->settings->get( 'disable_funding' ) ) {
```
AlexP11223: Why `get` here? You simply need to replace `get` with `has`, it was supposed to be `has`, just a mistype.

AlexP11223 (changes requested):

modules/ppcp-onboarding/src/Endpoint/LoginSellerEndpoint.php
Dinamiko: Done!

AlexP11223 (approved):

PR #526, opened 2022-03-04
Listening for payment created webhooks and action scheduled 

### LINKED ISSUE #525: Improve payment token checking for subscriptions (562)
Currently when purchasing a subscription product, just after the payment is processed we check if the payment is saved on PayPal and if it´s not then we update the order status as failed and the subscription as cancelled. Because vaulting a paymet through the JSSDK is asynchronous it could be that when we check it it´s not there yet and therefore we handle a false negative.

We could wait for the incoming webhook and also register an event for checking if the payment is saved on PayPal after some minutes, then if it is not there fail the order as we do now.

Here is how the workflow could look like:

PCP Issues drawio (1)

### Acceptance Criteria

- Add subscription product to the cart and go to Checkout.
- Click PayPal button and process payment.
- Immediatelly after order is placed succesfully, check that subscription status is on-hold.
- After one minute or so check subscription status is active and order status is processing.

Dinamiko:
Fixes #525.


modules/ppcp-vaulting/src/PaymentTokenChecker.php
```diff
@@ -0,0 +1,225 @@
...
+
+	/**
+	 * Captures authorized payments for the given WC order.
+	 *
+	 * @param WC_Order $wc_order The WC order.
+	 * @throws NotFoundException When there is a problem capturing the payment.
+	 */
+	private function capture_authorized_payment( WC_Order $wc_order ): void {
+		if ( $this->settings->has( 'intent' ) && strtoupper( (string) $this->settings->get( 'intent' ) ) === 'CAPTURE' ) {
+			$this->authorized_payments_processor->capture_authorized_payment( $wc_order );
```
AlexP11223: Looks like `capture_authorized_payment` returns `bool`. Maybe should either return it too or check and throw.

AlexP11223 (approved):

PR #529, opened 2022-03-09
Fix ajax URLs on multisite

### LINKED ISSUE #528: Multisite path doubled in ajax URLs (578)
In our ajax requests, the subfolder installation path of WordPress site is doubled.

double URL path pcp

### To Reproduce

- setup multisite subsite with PayPal Payments
- visit checkout, open browser console (Network tab) and click PayPal button, observe the ajax request URLs

AlexP11223:
Fixes #528

I think `home_url()` is not needed here, `/......` should work fine since we use ajax URLs only for requests performed from our site. WC does not use it too (e.g. `wc-ajax=update_order_review`).


Dinamiko (approved):

AlexP11223:
oh, missed that we pass URL of `ReturnUrlEndpoint` to PayPal.

Added `network_home_url` for it, for non-multisite it's [the same as](https://developer.wordpress.org/reference/functions/network_home_url/#source) `home_url`.

Dinamiko (approved):

PR #540, opened 2022-03-17
Fix the description counting when emoji is used

### LINKED ISSUE #491: emoji in product description causing INVALID_STRING_LENGTH error (347)
### The problem
Depending on how it's counted, emojis may count as more than one character. 
When the product description contains text and an emoji like this 🤔 in the first 127 characters,  then the API responds with `INVALID_STRING_LENGTH` as the emoji causes the string to exceed the maximum length.

### To Reproduce
1. create a product with an emoji in the description (has to be within the first 127 characters)
2. visit the single product page and click the PayPal button
3. alternatively, add the product to the cart and click the PayPal button from the cart or checkout page
3. payment fails: `[INVALID_REQUEST] Request is not well-formed, syntactically incorrect, or violates schema. https://developer.paypal.com/docs/api/orders/v2/#error-INVALID_STRING_LENGTH `

### Expected behavior
No error when checking out with a product with an emoji in the description.

### Actual Behavior
An error occurs when clicking the PayPal button and no PayPal order will be created.

### Workaround
Remove emoji from the first 127 characters of the product description. Emoji can be placed in later parts as only the first 127 characters are sent to PayPal.

### System information
PayPal Payments 1.7.0
WooCommerce 6.2.0
WordPress 5.9

Narek13:
**Issue**: #491

---

### Description

When the product description contains text and an emoji like this 😉 in the first 127 characters that are sent to PayPal,  then the API responds with `INVALID_STRING_LENGTH` because the emoji would make the string longer than the maximum.

The PR will fix the problem by changing the `mb_substr` function to `substr` function

### Steps to Test

1. Create a product with an emoji in the description (has to be in the part that’s sent to PayPal 😉)
2. Buy product from any page: single product page, cart, checkout
3. Payment fails

---

Closes #491 .

### DIFF
--- a/modules/ppcp-api-client/src/Factory/ItemFactory.php
+++ b/modules/ppcp-api-client/src/Factory/ItemFactory.php
@@ -61,7 +61,7 @@ class ItemFactory {
 					mb_substr( $product->get_name(), 0, 127 ),
 					new Money( $price_without_tax_rounded, $this->currency ),
 					$quantity,
-					mb_substr( wp_strip_all_tags( $product->get_description() ), 0, 127 ),
+					substr( wp_strip_all_tags( $product->get_description() ), 0, 127 ),
 					$tax,
 					$product->get_sku(),
 					( $product->is_virtual() ) ? Item::DIGITAL_GOODS : Item::PHYSICAL_GOODS
--- a/modules/ppcp-api-client/src/Factory/ItemFactory.php
+++ b/modules/ppcp-api-client/src/Factory/ItemFactory.php
@@ -61,7 +61,7 @@ class ItemFactory {
 					mb_substr( $product->get_name(), 0, 127 ),
 					new Money( $price_without_tax_rounded, $this->currency ),
 					$quantity,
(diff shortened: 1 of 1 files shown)


modules/ppcp-api-client/src/Factory/ItemFactory.php
```diff
@@ -61,7 +61,7 @@ function ( array $item ): Item {
 					mb_substr( $product->get_name(), 0, 127 ),
 					new Money( $price_without_tax_rounded, $this->currency ),
 					$quantity,
-					mb_substr( wp_strip_all_tags( $product->get_description() ), 0, 127 ),
+					substr( wp_strip_all_tags( $product->get_description() ), 0, 127 ) ?: '',
```
AlexP11223, line 64: What about non-latin languages? 

It sounds like e.g. for Russian and other cyrillic languages it will now send only about half of what was previously sent, probably worse for Chinese/Japanese if their symbols take more than 2 bytes.

Also it seems possible that the last letter may be broken? If it takes 2+ bytes and `substr` stops at the first byte.

Dinamiko (approved):

PR #572, opened 2022-04-05
Use `hasOwnProperty` instead of `length` for checking if property exist

### LINKED ISSUE #569: [RC] Errors in the Checkout page when paying with a disabled card (623)
### Describe the Bug

There are errors in the browser console on the Checkout page when paying with a disabled card.

This doesn't affect the payment flow as the customer still can't pay with a disabled card, as expected. But we shouldn't have errors in the console.

### To Reproduce

1. Make sure Card Processing is enabled
2. Add "Visa" to the "Disable specific credit cards" option
3. Go to the store and add a product to your cart
4. Select "Credit cards" as the payment method
5. Pay with a Visa card
6. Notice there're errors in the console. [Here's a video](https://d.pr/v/xX4p87) with the behavior

image

### Expected Behavior

There shouldn't be errors in the console.

### Actual Behavior

There're errors in the console.

### Environment

WordPress Version: 5.9.2
WooCommerce Version: 6.3.1
Plugin Version: 1.7.1-test2
Browser and Version: Chrome 99.0.4844.84
Any other plugins installed: None

Dinamiko:
Fixes #569.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
@@ -218,7 +218,8 @@ class CreditCardRenderer {
             }).catch(err => {
                 this.spinner.unblock();
                 this.errorHandler.clear();
-                if (err.details.length > 0) {
+
+                if (err.hasOwnProperty('details')) {
                     this.errorHandler.message(err.details.map(d => `${d.issue} ${d.description}`).join('<br/>'), true);
                 }
             });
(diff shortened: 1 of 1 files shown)


modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -218,7 +218,8 @@ class CreditCardRenderer {
             }).catch(err => {
                 this.spinner.unblock();
                 this.errorHandler.clear();
-                if (err.details.length > 0) {
+
+                if (err.hasOwnProperty('details')) {
```
AlexP11223: hm, I think `hasOwnProperty` is not intended for such purposes.

Maybe just `if (err.details)`?

Also you can check `length` like this: `err.details?.length > 0`

AlexP11223 (changes requested):

modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
Dinamiko: yup, much better with `err.details`, thanks.

AlexP11223 (approved):

PR #574, opened 2022-04-06
Check the cart total amount inside mini cart buttons renderer hook.

### LINKED ISSUE #573: PayPal buttons not displayed in minicart. (626)
### Describe the Bug

PayPal buttons are not added in minicart. adding product to cart which is a first product in cart then the buttons will not be show, until you add another product into cart

### To Reproduce

1. Go to Shop.
2. Add product to cart.
3. Check minicart.
4. Reload page and check minicart.

### Expected Behavior

The buttons are shown in mini cart if the total price amount is not 0

### Actual Behavior

The buttons are not shown

Narek13:
**Issue**: #573

---

### Description

PayPal buttons are not added in minicart. adding product to cart which is a first product in cart then the buttons will not be show, until you add another product into cart.

The PR will fix the problem by making a check about cart total amount inside the button renderer hook.

### Steps to Test

1. Go to Shop.
2. Add product to cart.
3. Check minicart.
4. Reload page and check minicart.

---

Closes #573 .

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -361,11 +361,11 @@ class SmartButton implements SmartButtonInterface {
 		) {
 			add_action(
 				$this->mini_cart_button_renderer_hook(),
-				static function () {
-                    // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison
-					if ( WC()->cart->get_cart_contents_total() == 0 ) {
+				function () {
+					if ( $this->is_cart_price_total_zero() ) {
 						return;
 					}
+
 					echo '<p
                                 id="ppc-button-minicart"
                                 class="woocommerce-mini-cart__buttons buttons"
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -354,6 +354,27 @@ class SmartButton implements SmartButtonInterface {
 
 		add_action( $this->pay_order_renderer_hook(), array( $this, 'button_renderer' ), 10 );
 
+		$not_enabled_on_minicart = $this->settings->has( 'button_mini_cart_enabled' ) &&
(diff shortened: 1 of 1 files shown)


modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -354,6 +354,27 @@ private function render_button_wrapper_registrar(): bool {
...
+		$not_enabled_on_minicart = $this->settings->has( 'button_mini_cart_enabled' ) &&
+			! $this->settings->get( 'button_mini_cart_enabled' );
+		if (
+		! $not_enabled_on_minicart
+		) {
+			add_action(
+				$this->mini_cart_button_renderer_hook(),
+				static function () {
+                    // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison
+					if ( WC()->cart->get_cart_contents_total() == 0 ) {
```
AlexP11223: Why not call `is_cart_price_total_zero`?

AlexP11223 (changes requested):

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: @AlexP11223 
I have tried to have a variable `$is_cart_price_total_zero = $this-> is_cart_price_total_zero()` and then `static function () use ($is_cart_price_total_zero )` but it returned wrong value, I guess because it was called outside of the hook. 

In other hands you can't call `$this` inside the `static` handler, right? or is there a better way?
Thanks

AlexP11223: You can if you either remove `static` here or make `is_cart_price_total_zero` `static`.

AlexP11223 (approved):

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: done 515a06542ed8aa652ac6d0b38b5e68de901f9c73
Thanks.

Dinamiko (approved):

Issue #580, opened 2022-04-12
Allow free trial subscriptions (PCP-475)


AlexP11223:
Currently it´s not possible to purchase a free subscriptions because it´s not possible to place an order with zero amount in PayPal. To provide a solution for this problem we could use the “one dollar auth and void afterwards” approach for cards and vaulting without payment for PayPal.

https://developer.paypal.com/limited-release/vault-payment-methods/vault-api/

Dinamiko:
Implemented in #581.

PR #581, opened 2022-04-12
Free trial

AlexP11223:
Implements #580 

Only in checkout for now. Should be easy to support in Pay for order pages, just need to figure out how to get the order ID.

For PayPal we reuse the standard smart button by disabling all funding sources/`messages`. The standard behavior is stopped via `return actions.reject()` in `onClick`, and we redirect to the account approval as described in the docs. One issue here is that the standard popup sometimes still appears for 0.5 sec.

The functions for free trial checks of cart/orders/products are implemented as trait. Not sure if we should prefer this or composition like with `SubscriptionHelper`, I decided to use trait because it seems more convenient in this case, we need it in many places including other traits. The only disadvantage I see is that it may be a bit more difficult to mock trait in tests, but still should be ok.

https://user-images.githubusercontent.com/5680466/162985971-1177c386-1b98-46a6-a3b1-02690057ae60.mp4


modules/ppcp-api-client/src/Entity/PaymentTokenActionLinks.php
```diff
@@ -0,0 +1,78 @@
...
+ * The links from CUSTOMER_ACTION_REQUIRED v2/vault/payment-tokens response.
+ *
+ * @package WooCommerce\PayPalCommerce\ApiClient\Entity
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\ApiClient\Entity;
+
+use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
```
Dinamiko: Seems not used in this class.

modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
```diff
@@ -13,6 +13,9 @@
 use Psr\Log\LoggerInterface;
 use WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Address;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Amount;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\AmountBreakdown;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Money;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Payer;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\PayerName;
```
Dinamiko: Seems these 3 are not used in the class:
```
use WooCommerce\PayPalCommerce\ApiClient\Entity\Address;
use WooCommerce\PayPalCommerce\ApiClient\Entity\AmountBreakdown;
use WooCommerce\PayPalCommerce\ApiClient\Entity\PayerName;
```

modules/ppcp-subscription/src/Helper/SubscriptionHelper.php
```diff
@@ -11,6 +11,8 @@
 
 namespace WooCommerce\PayPalCommerce\Subscription\Helper;
 
+use WC_Product;
+use WC_Subscription;
```
Dinamiko: Seems this is not used in the class.

modules/ppcp-vaulting/src/PaymentTokenChecker.php
```diff
@@ -15,8 +15,10 @@
...
 use WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint;
 use WooCommerce\PayPalCommerce\ApiClient\Endpoint\PaymentsEndpoint;
-use WooCommerce\PayPalCommerce\ApiClient\Entity\Authorization;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
+use WooCommerce\PayPalCommerce\ApiClient\Repository\OrderRepository;
+use WooCommerce\PayPalCommerce\Subscription\FreeTrialHandlerTrait;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\CreditCardGateway;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
 use WooCommerce\PayPalCommerce\WcGateway\Processor\AuthorizedPaymentsProcessor;
 use WooCommerce\PayPalCommerce\WcGateway\Settings\Settings;
```
Dinamiko, line 21: Not related to this PR but worth removing these 3 as seems not used:
```
use WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint;
use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
```

Dinamiko (approved):
It looks great, good job!

PR #621, opened 2022-04-28
Fix duplicated partial refunds

### LINKED ISSUE #522: Partial refunds via PayPal are created twice/double in WooCommerce order (583)
### Describe the Bug

PayPal partial refunds via WooCommerce backend result in two refund entries in the order details view of WooCommerce

### To Reproduce

1. Create a partial refund in the backend of WooCommerce and use "Refund via PayPal" button.
2. The refund entry is successfully created in the backend 
3. PayPal correctly refunds the amount to the customers bank account / debig card / credit card etc.
4. PayPal sends back a webhook about the successful refund
5. PayPal Payments plugin creates a new refund entry to the order instead of updating the existing one.
6. You now have two refunds in the order details of WooCommerce

### Screenshots

grafik

### Expected Behavior

Don't create a second refund entry if it already exists

### Actual Behavior

Creates the refund a second time.

### Environment

 - WordPress 5.9.1
 - WooCommerce 6.2.1
 - Plugin Version 1.7.0 (but happens to previous versions also, problem is therefore not new)
 - WooCommerce Germanized 3.8.3
 - WooCommerce Germanized Pro 3.4.1
 - Firefox

AlexP11223:
Fixes #522 

It may not be a perfect solution because theoretically we may get a race condition if we get webhook immediately after sending the refund to PayPal before the refund ID is saved. But it seems like currently the webhooks are sent a bit later, not during the `v2/payments/captures/.../refund` API request, so hopefully it's ok.

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/PaymentsEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/PaymentsEndpoint.php
@@ -198,11 +198,11 @@ class PaymentsEndpoint {
 	 *
 	 * @param Refund $refund The refund to be processed.
 	 *
-	 * @return void
+	 * @return string Refund ID.
 	 * @throws RuntimeException If the request fails.
 	 * @throws PayPalApiException If the request fails.
 	 */
-	public function refund( Refund $refund ) : void {
+	public function refund( Refund $refund ) : string {
 		$bearer = $this->bearer->bearer();
 		$url    = trailingslashit( $this->host ) . 'v2/payments/captures/' . $refund->for_capture()->id() . '/refund';
 		$args   = array(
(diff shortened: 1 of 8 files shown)


modules/ppcp-api-client/src/Endpoint/PaymentsEndpoint.php
```diff
@@ -223,12 +223,15 @@ public function refund( Refund $refund ) : void {
 		}
 
 		$status_code = (int) wp_remote_retrieve_response_code( $response );
-		if ( 201 !== $status_code ) {
+		$json        = json_decode( $response['body'] );
```
Dinamiko, line 225: `$json` variable already created on line 219.

Dinamiko (approved):

PR #631, opened 2022-05-04
PUI gateway visible when checkout currency or order total are not supported

Dinamiko:
### Description

The PUI gateway is visible for every currency as long as the Checkout county is Germany.

Payments must be made in EUR and be between 5€ and 2500€.
We should prevent errors whenever possible so it makes more sense it hide the gateway if the currency or checkout total is not supported.

### Steps to Test

1. set up PUI
2. change WooCommerce currency to any non-EUR currency
3. gateway remains visible but every Checkout attempt will fail
4. change currency back to EUR and change order total to anything outside of 5€-2500€
5. gateway remains visible but every Checkout attempt will fail

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -286,7 +286,7 @@ class PayUponInvoice {
 
 		$cart = WC()->cart ?? null;
 		if ( $cart ) {
-			$cart_total = (int) WC()->cart->get_totals()['total'];
+			$cart_total = (float) $cart->get_total( 'numeric' );
 			if ( $cart_total < 5 || $cart_total > 2500 ) {
 				return false;
 			}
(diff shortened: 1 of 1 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -270,6 +269,32 @@ function( array $methods ): array {
...
+			return false;
+		}
+
+		if ( 'EUR' !== get_woocommerce_currency() ) {
+			return false;
+		}
+
+		$cart = WC()->cart ?? null;
+		if ( $cart ) {
+			$cart_total = (int) WC()->cart->get_totals()['total'];
```
AlexP11223: I used `(float) ->get_total( 'numeric' )` instead of `get_totals` somewhere else, not sure if there is any difference.

AlexP11223: Also you should use the `$cart` variable here.

AlexP11223 (approved):

PR #637, opened 2022-05-09
Hide PUI gateway when merchant account not being enabled for PUI (683)

Dinamiko:
### Description
When merchant is onboarded but not elegible for PUI (using a merchant account that is not elegible https://developer.paypal.com/limited-release/orders-apms/pay-upon-invoice/#link-eligibility), the gateway should be hidden.

### Proposed solution
We could get the current seller status ([woocommerce-paypal-payments/PartnersEndpoint.php at trunk · woocommerce/woocommerce-paypal-payments](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php#L100) ) and then check if it contains PUI product. If it does not contain it then disable the PUI gateway.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -14,6 +14,8 @@ use WC_Order;
 use WooCommerce\PayPalCommerce\ApiClient\Endpoint\PayUponInvoiceOrderEndpoint;
 use WooCommerce\PayPalCommerce\Button\Exception\RuntimeException;
 use WooCommerce\PayPalCommerce\Onboarding\Environment;
+use WooCommerce\PayPalCommerce\Onboarding\State;
+use WooCommerce\PayPalCommerce\WcGateway\Helper\PayUponInvoiceProductStatus;
 use WooCommerce\PayPalCommerce\WcGateway\Processor\TransactionIdHandlingTrait;
 use WooCommerce\PayPalCommerce\WcGateway\Settings\Settings;
 use WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException;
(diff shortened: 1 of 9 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -75,6 +77,34 @@ class PayUponInvoice {
...
 
+	/**
+	 * The onboarding state.
+	 *
+	 * @var State
+	 */
+	protected $state;
+
+	/**
+	 * The is ppcp settings page.
```
AlexP11223: "The" looks strange here. :)

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceProductStatus.php
```diff
@@ -0,0 +1,107 @@
...
+namespace WooCommerce\PayPalCommerce\WcGateway\Helper;
+
+use WooCommerce\PayPalCommerce\ApiClient\Endpoint\PartnersEndpoint;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\SellerStatusProduct;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
+use WooCommerce\PayPalCommerce\WcGateway\Settings\Settings;
+use WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException;
+
+/**
+ * Class DccProductStatus
```
AlexP11223: Wrong name here and below.

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceProductStatus.php
```diff
@@ -0,0 +1,107 @@
...
+	) {
+		$this->settings          = $settings;
+		$this->partners_endpoint = $partners_endpoint;
+	}
+
+	/**
+	 * Whether the active/subscribed products support PUI.
+	 *
+	 * @return bool
+	 * @throws NotFoundException Should a setting not be found.
```
AlexP11223: Probably should be "When a setting not found".
But actually I don't see it thrown anywhere? If it is because of `$this->settings->get`, then it should not be possible after the `has` check before (only if it's some kind of bug). So this `@throw` should be removed.

AlexP11223 (changes requested):

AlexP11223 (approved):

PR #644, opened 2022-05-11
[PUI] Customer service instructions field not mandatory in gateway settings

Dinamiko:
### Description

The Customer service instructions field must be populated for the PUI gateway to work, so we might aswell make it a required field on the gateway settings page to ensure it’s populated.

### Steps to Test

1. configure and enable PUI
2. leave field Customer service instructions empty
3. attempt PUI payment
4. payment fails (error message is also not printed as user friendly as it could be)

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -342,6 +342,43 @@ class PayUponInvoice {
 				}
 			}
 		);
+
+		add_action(
+			'woocommerce_update_options_checkout_ppcp-pay-upon-invoice-gateway',
+			function () {
+				$customer_service_instructions = filter_input( INPUT_POST, 'woocommerce_ppcp-pay-upon-invoice-gateway_customer_service_instructions', FILTER_SANITIZE_STRING );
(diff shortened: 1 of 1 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -342,6 +342,43 @@ function () {
 				}
 			}
 		);
+
+		add_action(
+			'woocommerce_update_options_checkout_ppcp-pay-upon-invoice-gateway',
+			function () {
+				$customer_service_instructions = filter_input( INPUT_POST, 'woocommerce_ppcp-pay-upon-invoice-gateway_customer_service_instructions', FILTER_SANITIZE_STRING );
+				if ( '' === $customer_service_instructions ) {
+					$gateway_settings = get_option( 'woocommerce_ppcp-pay-upon-invoice-gateway_settings' );
```
AlexP11223, line 351: Would be great to at least extract this name somewhere.

Or a simple wrapper `WcGatewaySettings` with functions like `load(): array`, `save(array)`. Or a `ContainerInterface` wrapper like [Settings](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Settings/Settings.php).

AlexP11223 (approved):

Issue #632, opened 2022-05-04
Empty payer data may cause CITY_REQUIRED error for certain checkout countries (615)


AlexP11223:
Checkout via product/cart buttons may not work for logged-out users when the country is set to specific locations like Vietnam.

### To Reproduce

1. set WooCommerce store location country to Vietnam
2. set Default customer location is to Store country/region
3. alternatively, just set Vietnam as a checkout country when testing
4. log out and navigate to a single product page
5. make sure Vietnam would be the selected country
6. click PayPal smart button on single product page while not logged in
7. API error occurs: `[UNPROCESSABLE_ENTITY] The requested action could not be performed, semantically incorrect, or failed business validation. https://developer.paypal.com/docs/api/orders/v2/#error-CITY_REQUIRED`

### Possible cause

It happens because it is a country without postal code and this check for removing shipping fails.
https://github.com/woocommerce/woocommerce-paypal-payments/blob/d18df34e9bafd21c2a63e73a36ec37cbb1c46ec8/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php#L164

It seems like this check is actually supposed to be “is this shipping address valid”, not just country + postal code. Because I doubt that it is enough for a valid address, it probably just happens so that the postal code is usually filled when the rest of the fields are filled too.

For example it also fails for logged in users with some address fields deleted in the account.

raywick:
Hi! 

This issue is still happening with Ecuador. I have the last version of the plugin, but I still receive the same error. Should I create a new issue?

Issue #624, opened 2022-05-02
Mini Cart smart buttons visible after adding subscription product to cart from "shop" page while Vaulting is disabled. (660)


Narek13:
### Describe the Bug

When Vaulting is disabled, the Smart Buttons will temporarily display in the Mini Cart anyway after adding the subscription product to the cart.
Updating the cart quantity or adding another product resolves the behavior.

### To Reproduce

1. Activate the Storefront theme.
2. Disable the Vaulting.
3. Add subscription product to cart from the shop page (not the single product page).
4. Hover over mini cart → smart buttons will appear.
5. Adding another product to the cart or updating the quantity via the Cart page removes the smart buttons from the Mini Cart

**ALSO**
6. When there is a regular product in mini-cart together with a subscription product.
7. Then removing the subscription product should make the button appear for the regular product left in the mini-cart.

### Expected Behavior

- When a subscription product is added to the Cart while Vaulting is disabled, the smart buttons should not appear in the Mini Cart.
- And then, removing the subscription product from mini cart should make the button visible again if there is a still regular product in the cart.

### Actual Behavior

- When a subscription product is added to the Cart while Vaulting is disabled, the smart buttons are still loaded.
- After removing the subscription product from mini cart when there is a regular product, the buttons are still not shown.

Issue #616, opened 2022-04-25
Add Fraud Processor Response as an order note. (647)


Narek13:
### Describe the Bug

The Fraud Processor Response should be included in order notes as following:

- Address Verification Result
   - AVS
   - Address Match (Y/N)
   - Postal / Zip Code Match (Y/N)
  
- Card Security Code Result
   - CVV2 Match (Y/N)

### To Reproduce

1. Pay for order with credit card.
2. Check the order edit screen in admin area.

### Screenshots

![364d6c09-a87f-41a0-b048-d2f80732bf30](https://user-images.githubusercontent.com/11319597/165121500-c3b5f21c-7208-4ebe-865b-8b10a7bfd94b.png)

### Expected Behavior

The Fraud Processor Response is added in order notes when paying via credit cards.

### Actual Behavior

The Fraud Processor Response is not added.

Issue #606, opened 2022-04-18
Smart buttons not loading when free product is in cart but shipping costs are available. (632)


Narek13:
### Describe the Bug

When there is a free product in the cart but shipping cost, the button will not be loaded

### To Reproduce

1. Add free product to the cart.
2. Visit checkout page.
3. If there are no shipping costs, PayPal Payments gateway will be hidden.
4. When there are shipping costs, the gateway appears but the smart button won’t load because of the free product in the cart

### Screenshots

<img width="658" alt="Screen Shot 2022-04-18 at 17 51 52" src="https://user-images.githubusercontent.com/11319597/163818045-2f41cae6-e8d9-49ec-85d9-ed3749dce2d1.png">

### Expected Behavior

Smart buttons are displayed when the Checkout total is above 0.

### Actual Behavior

Smart buttons are not displayed when the Checkout total is above 0 and there is a free product.

PR #650, opened 2022-05-20
Release 1.8.1

Dinamiko:
* Fix - Manual orders return an error for guest users when paying with PayPal Card Processing #530
* Fix - "No PayPal order found in the current WooCommerce session" error for guests on Pay for Order page #605
* Fix - WooCommerce PayPal Payments & Conditional Discounts By Orion #548
* Fix - Empty payer data may cause CITY_REQUIRED error for certain checkout countries #632
* Fix - Mini Cart smart buttons visible after adding subscription product to cart from "shop" page while Vaulting is disabled #624
* Fix - Smart buttons not loading when free product is in cart but shipping costs are available #606
* Fix - Smart button & Pay Later messaging disappear on the cart page after changing shipping method #288
* Fix - Disabling PayPal Checkout on the checkout page also removes the button from the Cart and Product Pages #577
* Fix - Partial refunds via PayPal are created twice/double in WooCommerce order #522
* Fix - Emoji in product description causing INVALID_STRING_LENGTH error #491
* Enhancement - Vaulting & Pay Later UI/UX #174
* Enhancement - Redirect after updating settings for DCC sends you to PPCP settings screen #392
* Enhancement - Add Fraud Processor Response as an order note #616
* Enhancement - Add the Paypal Fee to the Meta Custom Field for export purposes #591

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,7 @@
 = 1.8.1 - TBD =
 * Fix - Manual orders return an error for guest users when paying with PayPal Card Processing #530
 * Fix - "No PayPal order found in the current WooCommerce session" error for guests on Pay for Order page #605
-* Fix - WooCommerce PayPal Payments & Conditional Discounts By Orion #548
+* Fix - Error on order discount by third-party plugins #548
 * Fix - Empty payer data may cause CITY_REQUIRED error for certain checkout countries #632
 * Fix - Mini Cart smart buttons visible after adding subscription product to cart from "shop" page while Vaulting is disabled #624
 * Fix - Smart buttons not loading when free product is in cart but shipping costs are available #606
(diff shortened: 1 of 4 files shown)


changelog.txt
```diff
@@ -1,5 +1,21 @@
 *** Changelog ***
 
+= 1.8.1 - TBD =
+* Fix - Manual orders return an error for guest users when paying with PayPal Card Processing #530
+* Fix - "No PayPal order found in the current WooCommerce session" error for guests on Pay for Order page #605
+* Fix - WooCommerce PayPal Payments & Conditional Discounts By Orion #548
```
AlexP11223: Could be better to summarize what was actually done instead of a bit confusing titles like this?

e.g. "Fix - Error on order discount by third-party plugins"

AlexP11223 (approved):

panilo:
#605 fixed in this release

matt-22:
I've updated to V1.8.1 but the name fields aren't appearing?
What was the fix? Is name field no longer required and hence it's not appearing?

image

PR #658, opened 2022-05-27
[PUI] Remove page reload after clicking on "Onboard with Pay Upon Invoice" checkbox

Dinamiko:
During onboarding, when merchant click on "Onboard with Pay Upon Invoice" checkbox page reloads. We should prevent that action and align the behavior with other check-boxes on the landing page.

image-20220524-201453

### DIFF
--- a/modules/ppcp-onboarding/assets/js/onboarding.js
+++ b/modules/ppcp-onboarding/assets/js/onboarding.js
@@ -88,7 +88,7 @@ const ppcp_onboarding = {
                 return res.json();
             }).then((data)=>{
                 if (!data.success) {
-                    alert('Could not update signup buttons: ' + data);
+                    alert('Could not update signup buttons: ' + JSON.stringify(data));
                     return;
                 }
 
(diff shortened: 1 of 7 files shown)


modules/ppcp-onboarding/assets/js/onboarding.js
```diff
@@ -87,7 +87,19 @@ const ppcp_onboarding = {
             }).then((res)=>{
                 return res.json();
             }).then((data)=>{
-                location.reload();
+                if (!data.success) {
+                    console.error(data);
```
AlexP11223: Maybe should at least `alert`? Otherwise the user has no easy way of knowing that it failed, and no errors to report to the support.

modules/ppcp-onboarding/assets/js/onboarding.js
```diff
@@ -88,7 +88,7 @@ const ppcp_onboarding = {
                 return res.json();
             }).then((data)=>{
                 if (!data.success) {
-                    console.error(data);
+                    alert('Could not update signup buttons: ' + data);
```
AlexP11223: Displaying objects like this will not work, it will be `[object Object]`.

```
const data = {a: [1, 2]}
alert('error: ' + data)
```

You can use `JSON.stringify(data)`

Dinamiko: Done, thanks!

AlexP11223 (approved):

PR #662, opened 2022-05-30
Fix disable-funding when free trial but dcc disabled

### LINKED ISSUE #661: free trial period causing incorrerct disable-funding parameters with DCC disabled (708)
When a free trial product is in the cart and the PayPal Card Processing is disabled, the PayPal smart buttons will not load on the checkout page due to incorrect disable-funding parameters:

image

### To Reproduce

- disable PayPal Card Processing & enable Vaulting
- add a free trial product to the cart
- visit Checkout page
- the smart button won’t load

AlexP11223:
fixes #661

was sending array values instead of keys

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -922,9 +922,9 @@ class SmartButton implements SmartButtonInterface {
 		}
 
 		if ( $this->is_free_trial_cart() ) {
-			$all_sources = $this->all_funding_sources;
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):

tomrobak (commented):
That fix shows PayPal button but not solving the issue. After clicking PayPal button WooCommerce returns error: **Something went wrong. Please try again or choose another payment source.**

paypal-free-trial-console-error

AlexP11223:
@tomrobak this seems to be a different issue, not related to this one. We had reports about it and investigating with PayPal.

PR #674, opened 2022-06-02
Do not check order approved status when it does not contains physical goods

### LINKED ISSUE #673: Something went wrong error in Virtual products when using vaulted payment (399)
When user wants to pay for a virtual product using a vaulted payment he is not able to do so. Error message "Something went wrong. Please try again or choose another payment source." will be displayed. 

### To Reproduce

1. Enable vaulting.
2. Go to a virtual product and add it to the cart.
3. Click PayPal button and place order.
4. Order is processed correctly.
5. Go to virtual product page again.
6. Click vaulted PayPal button.

### Screenshots
image-2021-11-08-13-02-09-289

### Expected Behavior

User is able to purchase any product from any page with a vaulted PayPal button

### Actual Behavior

"Something went wrong. Please try again or choose another payment source." error is displayed

### Environment

 - WordPress Version: 6.0
 - WooCommerce Version: 6.5
 - Plugin Version: 1.8.1

Dinamiko:
Fixes #673.

### DIFF
--- a/modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
+++ b/modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
@@ -183,7 +183,7 @@ class ApproveOrderEndpoint implements EndpointInterface {
 				wp_send_json_success( $order );
 			}
 
-			if ( $this->order_helper->contains_physical_goods( $order ) && ! $order->status()->is( OrderStatus::APPROVED ) && ! $order->status()->is( OrderStatus::CREATED ) {
+			if ( $this->order_helper->contains_physical_goods( $order ) && ! $order->status()->is( OrderStatus::APPROVED ) && ! $order->status()->is( OrderStatus::CREATED ) ) {
 				$message = sprintf(
 				// translators: %s is the id of the order.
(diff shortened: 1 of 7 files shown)


modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
```diff
@@ -173,7 +183,7 @@ public function handle_request(): bool {
 				wp_send_json_success( $order );
 			}
 
-			if ( ! $order->status()->is( OrderStatus::APPROVED ) ) {
+			if ( $this->order_helper->contains_physical_goods( $order ) && ! $order->status()->is( OrderStatus::APPROVED ) ) {
```
AlexP11223: Why we don't need to check this status for virtual products? And why it was happening only for the 2nd order?

Dinamiko: @AlexP11223 because when using a vaulted payment for an order that does not contains physical goods, it receives the order status as `CREATED` instead of `APPROVED` here. The same happens here: https://github.com/woocommerce/woocommerce-paypal-payments/pull/674/files#diff-e1a4bb00801efe53b8a9341d53440edf1208f856bb0c68df685afca73a5de272R174

AlexP11223 (approved):

Issue #590, opened 2022-04-13
programmatically capturing/voiding authorized payments (506)


Sirvijver:
Multiple users and third-party developers are asking us how they can capture authorized orders.
Right now, it’s only possible through order actions, which is not ideal. 

(linked to thread: https://github.com/woocommerce/woocommerce-paypal-payments/issues/234#issuecomment-946717592)

JStockton-Kishmish:
I believe this is the issue that is currently blocking one of my clients from adopting this plugin - just want to confirm that I'm right and I don't have to submit a new issue.

We have some customizations that require being able to void authorizations based on a status change (through the `woocommerce_order_status_<status_transition_to>` action) as well as when an order is covered 100% by a gift card.

If I'm understanding this issue correctly - you'd provide some way of doing this (e.g. `$paypal_payments->void_authorization($order_id);`); if so then that would resolve our problem.

Let me know if I'm misunderstanding and I'll submit a new issue, otherwise thank you in advanced for your support!

P.S. I did see that the work around is to register a new module and have it do whatever we need it to so we may go that route, but obviously an official solution is preferable.

InpsydeNiklas:
Correct, the long-term goal is to provide an easy-to-use method for programmatically capturing/voiding/refunding payments since these actions currently only can be performed through the WooCommerce or PayPal GUI. As mentioned above, [the filter to create a custom module](https://github.com/woocommerce/woocommerce-paypal-payments/wiki/Actions-and-Filters#adding-modules-and-accessing-the-plugin-container) essentially allows you to modify the entire plugin behavior already. But depending on what you want to achieve, we understand that a "classic solution" would be desired for routine tasks such as handling payments. So a regular hook/filter will eventually cover these tasks.

One way is to do the same thing as the [action](https://github.com/woocommerce/woocommerce-paypal-payments/blob/094133ce90e100f1938cbf3f61cd211ea4f94568/modules/ppcp-wc-gateway/src/WCGatewayModule.php#L275-L276) is by retrieving `'wcgateway.processor.authorized-payments'` from the container.

JStockton-Kishmish:
Perfect, thanks for the confirmation and the feedback!

[...]

PR #681, opened 2022-06-13
Filter refunded status

JStockton-Kishmish:
**Issue**: #590 

---

### Description

Apply the [woocommerce_order_fully_refunded_status](http://hookr.io/filters/woocommerce_order_fully_refunded_status/) filter to allow using `wc_refund_payment()` for voiding a transaction programmatically.  The only issue that I see with doing so currently is that the order status gets changed which, in my case, I do not want.

### Steps to Test

Create 2 orders both using the Paypal gateway in authorize mode .  Refund them by calling the following script:

```
<?php
require_once('wp-load.php');

$order_1 = wc_get_order(1234); // Replace 1234 with the post_ID of the first order
$order_2 = wc_get_order(1235); // Replace 1235 with the post_ID of the second order

wc_refund_payment($order_1);
add_filter('woocommerce_order_fully_refunded_status', '__return_false');
wc_refund_payment($order_2);
```

Observe that `$order_1` is now in the refunded status whereas $order_2 remained in the status it was at in the beginning, but both authorizations were voided in Paypal.

### Changelog Entry

> Allow filtering of refunded status on voids

### DIFF
--- a/modules/ppcp-wc-gateway/src/Processor/RefundProcessor.php
+++ b/modules/ppcp-wc-gateway/src/Processor/RefundProcessor.php
@@ -143,7 +143,15 @@ class RefundProcessor {
 						$this->payments_endpoint->void( $authorization );
 					}
 
-					$wc_order->set_status( 'refunded' );
+					$wc_order->set_status(
+						apply_filter(
+							'woocommerce_order_fully_refunded_status',
+							'refunded',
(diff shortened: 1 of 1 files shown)


InpsydeNiklas:
An API to achieve this was added in #1262. Closing this PR. Thanks for the contribution!

PR #696, opened 2022-06-22
Do not enter the checkout continuation state for dcc

### LINKED ISSUE #695: DCC causes checkout continuation state after checkout validation error (479)
After the checkout with credit cards failed because of a checkout validation issue (e.g. missing zip code) with correct credit card details, then the checkout will be partially within an “express checkout” after refreshing the checkout page.

When the form field validation failed but the card details were correct, the payment is pre-authorized, all gateways remain visible, but after an AJAX reload (e.g. when entering zip code) the “Place order” button for the credit card gateway disappears.

Selecting PayPal as a gateway in this state would cause the order to be processed with the credit card as well. No smart buttons would be loaded in this state.

### To Reproduce

- enable PayPal Card Processing, add a product to cart, visit checkout
- enter billing details and leave a required field empty, e.g. zip code
- select PayPal Card Processing and enter valid CC details
- click “Place order” button
- WooCommerce order creation will fail because of the missing required field (zip code) and throw an error
- at this point, an AJAX reload in the checkout, e.g. after entering zip code will cause the “Place order” button to be hidden. 
- refresh the checkout page (F5) to make the button reappear. 
- the checkout is now in a “continuation state” where all gateways are displayed, but the payment will be completed with the previously entered credit card details when clicking the “Place order” button for any gateway

AlexP11223:
Fixes #695

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -234,7 +234,6 @@ class SmartButton implements SmartButtonInterface {
 		if (
 			$this->settings->has( 'dcc_enabled' )
 			&& $this->settings->get( 'dcc_enabled' )
-			&& ! $this->session_handler->order()
 		) {
 			add_action(
 				$this->checkout_dcc_button_renderer_hook(),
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1082,6 +1081,23 @@ private function context(): string {
 		return $context;
 	}
 
+	/**
+	 * Checks if PayPal payment was already initiated (on the product or cart pages).
+	 *
+	 * @return bool
+	 */
+	private function is_paypal_continuation(): bool {
```
AlexP11223, line 1089: Not sure if we use some other name for this state already, but if not, then "continuation" seems suitable.

Dinamiko (approved):

PR #704, opened 2022-06-27
Fix button_renderer logic & the behavior for single products

### LINKED ISSUE #703: The smart buttons are not loaded on single product page while subscription product exists in the Cart (677)
### Describe the Bug

When Vaulting is disabled and a subscription product is in the Cart, the smart buttons won’t render on the single product page for regular products because the gateway is disabled.

### To Reproduce

1. Disable Vaulting.
2. Disable Manual renewals.
3. Add subscription product to the Cart.
4. Visit single product page for a non-subscription product.
5. → smart buttons won’t render on the single-product page.

### Expected Behavior

The smart buttons should be loaded on the single product page even when a subscription product is in the cart because the express checkout from the single product page for this one product would override the existing Cart anyway.

### Actual Behavior

The smart buttons are not loaded on non-subscription product when subscription product exists in the cart.

Narek13:
**Issue**: #703

---

### Description

When Vaulting is disabled and a subscription product is in the Cart, the smart buttons won’t render on the single product page for regular products because the gateway is disabled.

the problem is we should not rely on [gateway availability](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8118da278d826fdc03a71ea3ac4f3c13c07537e7/modules/ppcp-button/src/Assets/SmartButton.php#L536) when on single product page. Besides the [button_renderer](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8118da278d826fdc03a71ea3ac4f3c13c07537e7/modules/ppcp-button/src/Assets/SmartButton.php#L518) method (the callback for rendering buttons) is not organized well. it is covering 4 scenarios: for single product page, for checkout page, for pay for order page, for cart page. And inside that method we are checking which context it is (single product, checkout … etc) to run appropriate logic. The purpose of `button_renderer`  method is **to render the buttons**, so the checks should  not be done inside this method, but for each action itself.

The PR will improve the logic of `button_renderer` method to fix the behavior for single products described above.

### Steps to Test

1. Disable Vaulting.
2. Disable Manual renewals.
3. Add subscription product to the Cart.
4. Visit single product page for a non-subscription product.
5. → smart buttons won’t render on the single-product page.

---

Closes #703 .


Dinamiko (approved):

Narek13:
@Dinamiko Can you please check the latest changes here b15c03d4181b6af88e3376e5dd93e808996d7271 ? 

I have found the problem, when cart contains the subscription we [change](https://github.com/woocommerce/woocommerce-paypal-payments/blob/49048c59d5b51c2973328e6bfa4ac8d802081e8b/modules/ppcp-button/src/Assets/SmartButton.php#L913) the intent  to authorize , so when trying to order from single product page (non-subscription) we are getting the intent mismatch error

Dinamiko (approved):

modules/ppcp-button/resources/js/button.js
```diff
@@ -112,7 +112,7 @@ const bootstrap = () => {
             PayPalCommerceGateway,
             renderer,
             messageRenderer,
-        );w
```
AlexP11223, line 115: oops :astonished: 

we really should add a linter for JS. :)

PR #711, opened 2022-07-01
PUI gateway is displayed with unsupported store currency (733)

Dinamiko:
When merchant set store currency different than Euro then PUI gateway will be displayed regardless.

### Steps to Test

1. Install plugin
2. Set store currency to any currency different then Euro.
3. Enable PUI gateway
4. Navigate to checkout with a supported product type.
5. Set billing country to Germany

Observe PUI gateway is displayed while it should not.

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -284,7 +284,7 @@ class WCGatewayModule implements ModuleInterface {
 					$methods[] = $container->get( 'wcgateway.credit-card-gateway' );
 				}
 
-				if ( 'DE' === $container->get( 'api.shop.country' ) ) {
+				if ( 'DE' === $container->get( 'api.shop.country' ) && 'EUR' === get_woocommerce_currency() ) {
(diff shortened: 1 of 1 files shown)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -284,7 +284,7 @@ static function ( $methods ) use ( $container ): array {
 					$methods[] = $container->get( 'wcgateway.credit-card-gateway' );
 				}
 
-				if ( 'DE' === $container->get( 'api.shop.country' ) ) {
+				if ( 'DE' === $container->get( 'api.shop.country' ) && 'EUR' === get_woocommerce_currency() ) {
```
AlexP11223: Maybe better to use https://github.com/woocommerce/woocommerce-paypal-payments/blob/1f3ee422ef108a45008791199989af98ccbc9e0b/modules/ppcp-api-client/services.php#L372

Dinamiko: Yup, much better, thanks.

AlexP11223 (approved):

PR #725, opened 2022-07-11
Fix phpcs warnings

AlexP11223:


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -518,7 +518,8 @@ public function register_assets(): void {
 				'ppcp-pay-upon-invoice',
 				trailingslashit( $this->module_url ) . 'assets/js/pay-upon-invoice.js',
 				array(),
-				$this->asset_version
+				$this->asset_version,
+				true
```
AlexP11223, line 522: Is there a reason why `in_footer` was not `true`?

Dinamiko: There is no reason, I forgot to pass it here.

Dinamiko (approved):

PR #731, opened 2022-07-14
Introduce the WP hook proxy functionality.

### LINKED ISSUE #730: Type check error caused by potential plugin conflict (780)
### Describe the Bug

Recently we have received a report about the following error:

```
Fatal error: Uncaught TypeError: WooCommerce\PayPalCommerce\WcGateway\Settings\SettingsRenderer::render_multiselect(): Return value must be of type string, null returned in /var/www/vhosts/stage.gaco.netz.rocks/htdocs/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Settings/SettingsRenderer.php:178
```

This is because of a conflict with other plugin which is using the same `filter/action`
A decision was made to implement the WPHookProxyTrait to be used as a proxy for WordPress defined callbacks.
The error will not break the site (If WP_DEBUG is false), instead, the fields which are invalid will not be displayed and the error info will be added to logs.

### To Reproduce

1. Put the following filter in active theme’s functions.php file

```
add_filter(
    'woocommerce_form_field',
    static function ( $field, $key, $args, $value ) {
        //var_dump($args['type']);
        if ( 'text' === $args['type'] ) {
            return null;
        }
        return $field;
    },
    1,
    4
);
```
This will trigger the similar conflict described above

2. Check the settings page and see the error

### Expected Behavior

Plugin settings page loaded correctly( missing only the conflicted ones ). The logs are added in WC logs.

### Actual Behavior

Plugin settings page is broken because of the error.

Narek13:
**Issue**: #730

---

### Description

Recently we have received a report about the following error:

```
Fatal error: Uncaught TypeError: WooCommerce\PayPalCommerce\WcGateway\Settings\SettingsRenderer::render_multiselect(): Return value must be of type string, null returned in /var/www/vhosts/stage.gaco.netz.rocks/htdocs/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Settings/SettingsRenderer.php:178
```

This is because of a conflict with other plugin which is using the same `filter/action`
A decision was made to implement the WPHookProxyTrait to be used as a proxy for WordPress defined callbacks.
The error will not break the site (If WP_DEBUG is false), instead, the fields which are invalid will not be displayed and the error info will be added to logs.

The PR will add the `WPHookProxyTrait` described above.

### Steps to Test

1. Put the following filter in active theme’s functions.php file

```
add_filter(
    'woocommerce_form_field',
    static function ( $field, $key, $args, $value ) {
        //var_dump($args['type']);
        if ( 'text' === $args['type'] ) {
            return null;
        }
        return $field;
    },
    1,
    4
);
```
This will trigger the similar conflict described above

2. Check the settings page and see the error

---

Closes #730 .

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -11,6 +11,9 @@ namespace WooCommerce\PayPalCommerce\WcGateway;
 
 use Dhii\Container\ServiceProvider;
 use Dhii\Modular\Module\ModuleInterface;
+use Exception;
+use Throwable;
+use TypeError;
 use WC_Order;
 use WooCommerce\PayPalCommerce\AdminNotices\Repository\Repository;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Capture;
@@ -37,11 +40,14 @@ use WooCommerce\PayPalCommerce\WcGateway\Settings\SettingsListener;
 use WooCommerce\PayPalCommerce\WcGateway\Settings\SettingsRenderer;
 use Interop\Container\ServiceProviderInterface;
 use Psr\Container\ContainerInterface;
+use WooCommerce\WooCommerce\Logging\Logger\WPHookProxyTrait;
 
 /**
  * Class WcGatewayModule
  */
 class WCGatewayModule implements ModuleInterface {
+
+	use WPHookProxyTrait;
 
 	/**
 	 * {@inheritDoc}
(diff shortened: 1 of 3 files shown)


Dinamiko (approved):

sun:
> This is because of a conflict with other plugin which is using the same `filter/action`

I'm part of the team that reported this bug and I'm not able to confirm this statement.  The Call Stack shows that WooCommerce (Core) is calling the payment gateway callback functions of WooCommerce PayPal Payments directly in order to generate the settings page.

The proposed solution looks way more complex than they need to be.  PHP is rightfully complaining about a diverging return data type hint.  That's all that needs correcting.

https://github.com/woocommerce/woocommerce-paypal-payments/pull/749 replaces this PR and fixes the problem in a simpler way.

InpsydeNiklas:
Throwing the fatal error is the preferred behavior, so we'll not pursue this for now.

PR #733, opened 2022-07-19
Add separate card button gateway

### LINKED ISSUE #671: Option to separate JSSDK APM payment buttons into individual WooCommerce gateways (697)
Display the Black Credit/debit Card Button separated from the PayPal option (Stand-alone Button): https://developer.paypal.com/docs/checkout/standard/customize/standalone-buttons/ 

The way the button should look like in the checkout is:
image-20220602-084810

AlexP11223:
Closes #671 

Added the ability to have the card button as a separate gateway.

Also some refactoring, moved `process_payment` to the gateways to make it simpler and easier to understand, remove gateway id checks.

Added `Send Checkout billing form data to PayPal smart card fields` checkbox in the settings, allowing to disable #480 which can cause issues. However, currently it conflicts with the new address handling for card buttons, needs clarification.


Dinamiko (approved):

AlexP11223:
0c17c6b1de9adb0ca477092e7276cb0380898503 added select with 3 options for card billing data handling.

image

By default it sets the 2nd one for Latin America shops, and the 1st one otherwise, so most  of the merchants that want the 1st mode (the current behavior) should not be affected.

PR #740, opened 2022-07-21
Add fields to the basic js validation error message

### LINKED ISSUE #739: Checkout Field Validation Message (789)
### Describe the Bug

The checkout field validation message can be improved with more clarification, so the error message tells what required fields are missing. Some themes highlight the fields (see red border), but some may not.

So, instead of the error message: “Required form fields are not filled or invalid.”, it should state the field names for example:

- First Name field is not filled or invalid.
- Street Address field is not filled or invalid.
- Please agree to Terms and Conditions.

### To Reproduce

- Add a product to cart and navigate to the checkout 
- Proceed to check out without filling some required fields

AlexP11223:
Fixes #739 

image

Uses the text from `<label>`, and for the terms and conditions checkbox there is a map `name => text` because otherwise it would the checkbox text like "I have read and agree ...".

### DIFF
--- a/modules/ppcp-button/resources/js/button.js
+++ b/modules/ppcp-button/resources/js/button.js
@@ -38,9 +38,22 @@ const bootstrap = () => {
             requiredFields.each((i, input) => {
                 jQuery(input).trigger('validate');
             });
-            if (jQuery('form.woocommerce-checkout .validate-required.woocommerce-invalid:visible').length) {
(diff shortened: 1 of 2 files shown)


Dinamiko (approved):

AlexP11223:
Made the messages more like the WC validation message.

Uses WC translations when possible, works in languages which change the order of `Billing/Shipping` and `<Field Label>`, put the label in the middle of the message etc.

PR #742, opened 2022-07-21
PUI feature request: add option for a phone number field next to the Birth Date field (799)

Dinamiko:
Most merchants prefer to keep the phone number optional in the Checkout for better conversion.

But PUI requires a phone number.

If the merchant does not provide or phone number or only has the field optional, an error may occur when the buyer attempts a payment without the phone number.

We could add an option in the PUI gateway settings to insert a (required) phone number field into the gateway like here:

image-20220712-234329

This way, if a merchant has an optional phone number field, the new field would ensure that buyers provide a phone number before clicking the “Place order” button for PUI payments.

### Following scenarios must be covered:
- When the billing_phone field is optional or hidden, display our own mandatory phone field under the birth date.
- When the billing_phone field is required and visible, do not display our own mandatory phone field under the Birth date.


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -333,6 +333,26 @@ function( string $description, string $id ): string {
...
 
+					$checkout_fields = WC()->checkout()->get_checkout_fields();
+					if ( ! array_key_exists( 'billing_phone', $checkout_fields['billing'] ) ) {
+						woocommerce_form_field(
+							'billing_phone',
+							array(
+								/**
+								 * Use translation from WooCommerce here.
+								 * phpcs:disable WordPress.WP.I18n.TextDomainMismatch
+								 */
```
AlexP11223: You can use one-line `// phpcs:ignore ...`

AlexP11223:
> Following scenarios must be covered:

So are they covered? :thinking: 

I see only the key check here.

Dinamiko:
@AlexP11223 I´ve updated the scenarios that needs to be covered in the description.

AlexP11223 (approved):

PR #749, opened 2022-07-26
Fixed PHP Fatal error: Return value must be of type string, null returned.

### LINKED ISSUE #730: Type check error caused by potential plugin conflict (780)

sun:
Fixes https://github.com/woocommerce/woocommerce-paypal-payments/issues/730

Replaces https://github.com/woocommerce/woocommerce-paypal-payments/pull/731

### Description

The following PHP Fatal Error appears when trying to set up the plugin:

```
Fatal error: Uncaught Error: WooCommerce\PayPalCommerce\WcGateway\Settings\SettingsRenderer::render_multiselect():
Return value must be of type string, null returned
in wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Settings/SettingsRenderer.php on line 178
```

Resolving this error subsequently exposes further fatal errors in the other render callbacks in the same class.

### Steps to Test

1. Download the plugin and activate it.
2. Go to `/wp-admin/admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway`

### Changelog Entry

> Fixed PHP Fatal error: Return value must be of type string, null returned.

Closes #730

### DIFF
--- a/modules/ppcp-wc-gateway/src/Settings/SettingsRenderer.php
+++ b/modules/ppcp-wc-gateway/src/Settings/SettingsRenderer.php
@@ -172,7 +172,7 @@ class SettingsRenderer {
 	 *
 	 * @return string
 	 */
-	public function render_multiselect( $field, $key, $config, $value ): string {
+	public function render_multiselect( $field, $key, $config, $value ): ?string {
 
 		if ( 'ppcp-multiselect' !== $config['type'] ) {
 			return $field;
@@ -211,7 +211,7 @@ class SettingsRenderer {
 	 *
 	 * @return string
 	 */
-	public function render_password( $field, $key, $config, $value ): string {
+	public function render_password( $field, $key, $config, $value ): ?string {
 
 		if ( 'ppcp-password' !== $config['type'] ) {
 			return $field;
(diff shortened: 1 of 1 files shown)


InpsydeNiklas:
The error is caused when the `woocommerce_form_field` filter is used incorrectly on the admin pages (like some third-party Checkout Field Editor plugins did in the past). When using this filter to customize WooCommerce checkout form fields, the code should only execute on the checkout page and not on the admin pages.

PR #762, opened 2022-08-02
Disable oxxo and card button features via constants

Dinamiko:
Due to 1.9.2 patch release we need to disable via constant OXXO and APM button features as they are already merged into trunk.

### DIFF
--- a/woocommerce-paypal-payments.php
+++ b/woocommerce-paypal-payments.php
@@ -24,8 +24,8 @@ define( 'PAYPAL_SANDBOX_API_URL', 'https://api.sandbox.paypal.com' );
 define( 'PAYPAL_INTEGRATION_DATE', '2022-04-13' );
 
 define( 'PPCP_FLAG_SUBSCRIPTION', true );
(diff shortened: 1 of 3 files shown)


woocommerce-paypal-payments.php
```diff
@@ -24,6 +24,8 @@
 define( 'PAYPAL_INTEGRATION_DATE', '2022-04-13' );
 
 define( 'PPCP_FLAG_SUBSCRIPTION', true );
+define( 'PPCP_FLAG_OXXO', apply_filters( 'woocommerce_paypal_payments_enable_oxxo', false ) );
+define( 'PPCP_FLAG_SEPARATE_APM_BUTTONS', apply_filters( 'woocommerce_paypal_payments_enable_separate_apm_buttons', false ) );
```
AlexP11223: these filters should have some more specific name, like `..._feature`. Or `allow` instead of `enable` (but it's already used for cards).

Otherwise it sounds like they enable the gateways, like the checkbox in WC settings.

AlexP11223 (approved):

PR #769, opened 2022-08-04
During PayPal express checkout all gateways are available as a payment method (850)

Dinamiko:
When customer perform express checkout, on a checkout page all gateways are available as payment options.

### Steps to Test

- Enable all gateways
- Place the order with PayPal from mini cart, product or cart page

### What should happen
When redirected to checkout page customer only should see PayPal gateway.

### What happened instead
When redirected to checkout page customer will see all gateways as an available payment method.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
+++ b/modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
@@ -106,11 +106,6 @@ class DisableGateways {
 	 * @return bool
 	 */
 	private function needs_to_disable_gateways(): bool {
-		$wc_ajax = filter_input( INPUT_GET, 'wc-ajax', FILTER_SANITIZE_STRING ) ?? '';
-		if ( $wc_ajax === 'update_order_review' && $this->session_handler->funding_source() !== 'paypal' ) {
-			return false;
-		}
-
 		$order = $this->session_handler->order();
 		if ( ! $order ) {
 			return false;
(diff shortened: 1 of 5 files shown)


modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
```diff
@@ -107,7 +107,7 @@ private function disable_all_gateways() : bool {
 	 */
 	private function needs_to_disable_gateways(): bool {
 		$wc_ajax = filter_input( INPUT_GET, 'wc-ajax', FILTER_SANITIZE_STRING ) ?? '';
-		if ( $wc_ajax === 'update_order_review' ) {
+		if ( $wc_ajax === 'update_order_review' && $this->session_handler->funding_source() !== 'paypal' ) {
```
AlexP11223: hm, will it still work correctly when using other buttons like Venmo, Pay Later?

Dinamiko: @AlexP11223 yup it wirks for other buttons, the fix was for not hidding other gateways after getting a WC validation error when using DCC payment. I used `update_order_review` to ensure we are on that context, but when coming from express checkout `update_order_review` also exist and hides the other gateways, that´s the reason I added the conditional to not execute it when is `paypal`.

Dinamiko: you´re right, it only handles PayPal button.

Dinamiko:
@AlexP11223 I´ve just noticed the original issue was fixed here: https://github.com/woocommerce/woocommerce-paypal-payments/commit/c11635343d1c780d1f1ffe7fc0b7d74a8bdfbec5#diff-0999d0e241de01c000b49cd2a486a81fb1d4a51614fa78cea314b7e7fddc7538R114-R116

So no need for checking `update_order_review` here.

AlexP11223 (approved):

PR #778, opened 2022-08-09
Error message formatting in order notes contains file path (864)

Dinamiko:
### Description

When certain API responses are triggered, like `PAYMENT_DENIED`, then the order notes may contain needed information.

### Steps to Test

mock `PAYMENT_DENIED` error with this filter:
```
        $args['headers']['PayPal-Mock-Response'] = '{"mock_application_codes": "PAYMENT_DENIED"}';
    }
    return $args;
}, 10, 2);
```
- Add product to cart, visit checkout
- Go through payment process
- Attempt will fail
- Failed order is created with the file path in the order notes

### Expected behaviour

Other note does not contain file path.


Dinamiko:
Fixed in https://github.com/woocommerce/woocommerce-paypal-payments/pull/779

PR #787, opened 2022-08-12
Refactor saved credit card payment handler

Dinamiko:
Move saved credit card payment logic into its own class under vaulting.


tests/stubs/WC_Payment_Gateway_CC.php
```diff
@@ -3,5 +3,10 @@
 
 class WC_Payment_Gateway_CC
 {
+	public function init_settings() {}
+	public function process_admin_options() {}
 
+	protected function get_return_url($wcOrder) {
+		return $wcOrder;
```
AlexP11223, line 10: returns order instead of string :thinking:

AlexP11223 (approved):

PR #793, opened 2022-08-17
Tracking API

### LINKED ISSUE #792: Tracking API (155)
### Describe the Bug

As a merchant, you can use the Add Tracking API to send shipment tracking numbers to PayPal for PayPal transactions. PayPal uses these numbers to discern and investigate fraudulent customers. If you sell tangible goods, you are encouraged to send these numbers to PayPal as soon as you ship items to your customers. This information enables you to access funds more quickly, provide seller protection, and keep customers informed.

Integration Guide: [https://developer.paypal.com/docs/tracking/integrate/](https://developer.paypal.com/docs/tracking/integrate/)

**Implementation details**
First part includes basic implementation:

- Add setting to enable tracking in “PayPal Checkout Settings” section, disabled by default. 
- For German merchants that have PUI (Pay upon Invoice) enabled make tracking automatically enabled and make the field disabled for editing.
- When tracking is enabled display a tracking information metabox in the order with the following fields: Transaction ID, Tracking Number, Status and Carriers. This metabox represents the PayPal tracking API user interface for the merchant to send tracking info to PayPal.
- The first time when no information is sent to PayPal it should display a “Create” button label, while when there is already existing information it should display “Update” button label.
- For Carriers values list see [Add Tracking API Carriers](https://developer.paypal.com/docs/tracking/reference/carriers/) There should be the GLOBAL ones plus the ones for the merchant WC store country.

### Screenshots

Screen Shot 2022-08-17 at 15 35 48

Screen Shot 2022-08-17 at 15 35 39

Narek13:
**Issue**: #792

---

### Description

As a merchant, you can use the Add Tracking API to send shipment tracking numbers to PayPal for PayPal transactions. PayPal uses these numbers to discern and investigate fraudulent customers. If you sell tangible goods, you are encouraged to send these numbers to PayPal as soon as you ship items to your customers. This information enables you to access funds more quickly, provide seller protection, and keep customers informed.

Integration Guide: [https://developer.paypal.com/docs/tracking/integrate/](https://developer.paypal.com/docs/tracking/integrate/)

PR includes:

- Add setting to enable tracking in “PayPal Checkout Settings” section, disabled by default. 
- For German merchants that have PUI (Pay upon Invoice) enabled make tracking automatically enabled and make the field disabled for editing.
- When tracking is enabled display a tracking information metabox in the order with the following fields: Transaction ID, Tracking Number, Status and Carriers. This metabox represents the PayPal tracking API user interface for the merchant to send tracking info to PayPal.
- The first time when no information is sent to PayPal it should display a “Create” button label, while when there is already existing information it should display “Update” button label.
- For Carriers values list see [Add Tracking API Carriers](https://developer.paypal.com/docs/tracking/reference/carriers/) There should be the GLOBAL ones plus the ones for the merchant WC store country.

### Screenshots

Screen Shot 2022-08-17 at 15 35 48

Screen Shot 2022-08-17 at 15 35 39

---

Closes #792 .


modules/ppcp-order-tracking/src/OrderTrackingModule.php
```diff
@@ -0,0 +1,164 @@
...
+		 * @var Settings $settings
+		 */
+		$settings = $c->get( 'wcgateway.settings' );
+
+		/**
+		 * The PUI helper.
+		 *
+		 * @var PayUponInvoiceHelper $pui_helper
+		 */
+		$pui_helper = $c->get( 'wcgateway.pay-upon-invoice-helper' );
```
AlexP11223, line 51: You can use `assert($x instanceof X)` instead of these huge phpdocs.

modules/ppcp-order-tracking/src/OrderTrackingModule.php
```diff
@@ -0,0 +1,164 @@
...
+		if ( $pui_helper->is_pui_ready_in_admin() ) {
```
AlexP11223, line 54: If this module's `run` is always executed, calling this method (with `pui_product_status->pui_is_active()` inside) may be bad for performance, because it may execute API requests (when PUI is not available).

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
```diff
@@ -78,4 +97,17 @@ private function is_valid_currency(): bool {
...
 		return false;
 	}
+
+	/**
+	 * Checks whether PUI is ready in admin screen.
+	 *
+	 * @return bool
+	 */
+	public function is_pui_ready_in_admin(): bool {
+		if ( $this->pui_product_status->pui_is_active() && $this->shop_country === 'DE' ) {
```
AlexP11223: Should check the country first for better performance.

modules/ppcp-order-tracking/src/OrderTrackingModule.php
Narek13: done. 9ea71be37bb583e2de331ec1b8ea222c28629e91

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
Narek13: Done. 4e8379af022df302d4e7ed4f3245fdd5cb3ed2b6

modules/ppcp-order-tracking/src/OrderTrackingModule.php
Narek13: Now with https://github.com/woocommerce/woocommerce-paypal-payments/commit/4e8379af022df302d4e7ed4f3245fdd5cb3ed2b6 it will first check the Country code, but I don't know what else could be done here to improve it cause on **every request** I need to know once PUI is activated for German country then automatically enable tracking. I can also check this on settings screen only but in that way you have to visit that screen to auto-enable the tracking setting. Not sure if that makes sense.

AlexP11223: One way could be to fire a hook [here when status becomes true](https://github.com/woocommerce/woocommerce-paypal-payments/blob/20880d8250542384bbdc9b3ebd8da7a39fc4d779/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceProductStatus.php#L99) to trigger the check. And also e.g. after onboarding, `woocommerce_paypal_payments_gateway_migrate` etc.

Dinamiko (approved):

PR #821, opened 2022-09-05
Connection tab improvements

Narek13:
### Description

The PR will add the following improvements:

- After connecting a PayPal account, should redirect the user to the Connection tab and not the PayPal Checkout tab

- ACDC eligibility check incorrect; must check if enabled for PayPal account

- if country/currency in WC does not match ACDC eligibility, hide the ACDC feature onboarding section (same as with PUI)

- ACDC and PUI eligibility checks should be separate for Sandbox and Live accounts

- when Live & Sandbox accounts are connected, and one gets disconnected, automatically swap to the other account (e.g. auto uncheck Sandbox checkbox when disconnecting sandbox account).

### DIFF
--- a/modules/ppcp-onboarding/assets/js/onboarding.js
+++ b/modules/ppcp-onboarding/assets/js/onboarding.js
@@ -247,6 +247,12 @@ function ppcp_onboarding_productionCallback(...args) {
             }
         );
 
+        if (sandboxSwitchElement.checked) {
+            sandboxSwitchElement.checked = false;
+        } else {
+            sandboxSwitchElement.checked = true;
+        }
+
         isDisconnecting = true;
 
         document.querySelector('.woocommerce-save-button').click();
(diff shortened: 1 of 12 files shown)


modules/ppcp-onboarding/assets/js/onboarding.js
```diff
@@ -247,6 +247,12 @@ function ppcp_onboarding_productionCallback(...args) {
             }
         );
 
+        if (sandboxSwitchElement.checked) {
+            sandboxSwitchElement.checked = false;
+        } else {
+            sandboxSwitchElement.checked = true;
+        }
```
Dinamiko: Seems we are switching the checkbox status here, if so maybe just: 
`sandboxSwitchElement.checked = ! sandboxSwitchElement.checked;`

Dinamiko (approved):
Looks good, only a little comment about refactor a conditional.

modules/ppcp-onboarding/assets/js/onboarding.js
Narek13: d7aef159f954a65ebd2b720e2d98abaaed971653 done

PR #837, opened 2022-09-09
PUI-relevant webhook not subscribed to (666)

Dinamiko:
According to PayPal docs we should listen to: `CHECKOUT.PAYMENT-APPROVAL.REVERSED`

https://developer.paypal.com/docs/checkout/apm/pay-upon-invoice/integrate-pui-merchant/#link-listentowebhooks

But this webhook is not subscribed currently.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -577,20 +577,6 @@
                 "portable",
                 "shim"
             ],
-            "funding": [
-                {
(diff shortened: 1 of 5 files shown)


Dinamiko:
Fixed in #842.

PR #842, opened 2022-09-12
PUI-relevant webhook not subscribed to (666)

Dinamiko:
According to PayPal docs we should listen to: CHECKOUT.PAYMENT-APPROVAL.REVERSED

https://developer.paypal.com/docs/checkout/apm/pay-upon-invoice/integrate-pui-merchant/#link-listentowebhooks

But this webhook is not subscribed currently.

### DIFF
--- a/modules/ppcp-webhooks/src/Handler/RequestHandlerTrait.php
+++ b/modules/ppcp-webhooks/src/Handler/RequestHandlerTrait.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * Trait which helps to handle the request.
+ *
+ * @package WooCommerce\PayPalCommerce\Webhooks\Handler
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Webhooks\Handler;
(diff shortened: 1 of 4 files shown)


modules/ppcp-webhooks/src/Handler/RequestHandlerTrait.php
```diff
@@ -0,0 +1,110 @@
...
+				return ! empty( $order_id );
+			}
+		);
+	}
+
+	/**
+	 * Get WC orders from the given custom ids.
+	 *
+	 * @param array $custom_ids The custom ids.
+	 * @return stdClass|WC_Order[]
```
AlexP11223: Looks like it returns an array or an object depending on your args (pagination). https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-order-functions.html#source-view.23

So I guess here it should be just `WC_Order[]`.

Maybe can `assert` or check with `if` before `return`.

AlexP11223 (approved):

PR #870, opened 2022-09-28
Hide PayPal Card Processing tab is not available in country or for merchant (886)

Dinamiko:
When the PayPal merchant account is not enabled for advanced card payments, the PayPal Card Processing tab should be hidden entirely:
image-20220818-185938

### DIFF
--- a/.psalm/stubs.php
+++ b/.psalm/stubs.php
@@ -5,3 +5,6 @@ if (!defined('EP_PAGES')) {
 if (!defined('MONTH_IN_SECONDS')) {
 	define('MONTH_IN_SECONDS', 30 * DAY_IN_SECONDS);
 }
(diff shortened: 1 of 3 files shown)


AlexP11223:
Maybe better to remove it here?
https://github.com/woocommerce/woocommerce-paypal-payments/blob/2355083f51c62b5a28d3e473cedd04b113adbb84/modules/ppcp-wc-gateway/services.php#L239

then this will apply (which now can be simplified then)
https://github.com/woocommerce/woocommerce-paypal-payments/blob/eb156c435fcd6e61577b4cc546a4ec8e822e6108/modules/ppcp-wc-gateway/src/WCGatewayModule.php#L327-L334

AlexP11223 (approved):

PR #873, opened 2022-09-29
PUI payment instructions printed in the refund email (899)

Dinamiko:
When refunding a Pay upon Invoice order, the Ratepay payment instructions are inserted into the refund email.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -258,7 +258,13 @@ class PayUponInvoice {
 		add_action(
 			'woocommerce_email_before_order_table',
 			function( WC_Order $order, bool $sent_to_admin ) {
-				if ( ! $sent_to_admin && PayUponInvoiceGateway::ID === $order->get_payment_method() && $order->has_status( 'processing' ) ) {
+				$action = filter_input( INPUT_POST, 'action', FILTER_SANITIZE_STRING ) ?? '';
+				if (
+					! $sent_to_admin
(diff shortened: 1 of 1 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -258,7 +258,13 @@ function ( WC_Order $wc_order, string $order_id ) {
 		add_action(
 			'woocommerce_email_before_order_table',
 			function( WC_Order $order, bool $sent_to_admin ) {
-				if ( ! $sent_to_admin && PayUponInvoiceGateway::ID === $order->get_payment_method() && $order->has_status( 'processing' ) ) {
+				$action = filter_input( INPUT_POST, 'action', FILTER_SANITIZE_STRING ) ?? '';
+				if (
+					! $sent_to_admin
+					&& PayUponInvoiceGateway::ID === $order->get_payment_method()
+					&& $order->has_status( 'processing' )
+					&& ( $action && $action !== 'woocommerce_refund_line_items' )
```
AlexP11223: I guess it should work in most cases, but probably not the most reliable solution, e.g. maybe there are other ways to trigger refund.

It seems like the 4th parameter of `woocommerce_email_before_order_table` is a `WC_Email` object with `id` property. Maybe use it to find the email types we need (`customer_processing_order`?).

https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/includes/emails

Dinamiko: Done! much better like this, thanks :)

AlexP11223 (approved):

PR #876, opened 2022-09-30
Onboard with PUI Checkbox automatically set when shop is set to Germany

Dinamiko:
The “Onboard with Pay upon Invoice” checkbox should not be checked by default.


modules/ppcp-onboarding/src/Render/OnboardingOptionsRenderer.php
```diff
@@ -80,10 +80,16 @@ public function render( bool $is_shop_supports_dcc ): string {
 	 */
 	private function render_pui_option(): string {
```
AlexP11223, line 81: `@throws` not needed.

AlexP11223 (approved):

PR #882, opened 2022-10-03
Revert vault failed value

Dinamiko:
While refactoring in #729 I accidentally removed the value of the subscription failing behaviour type, this PR reverts it back, also uses WC API for adding meta to the order instead of WP post meta API.

### DIFF
--- a/modules/ppcp-vaulting/src/PaymentTokenChecker.php
+++ b/modules/ppcp-vaulting/src/PaymentTokenChecker.php
@@ -143,7 +143,7 @@ class PaymentTokenChecker {
 
 		try {
 			$subscription_behavior_when_fails = $this->settings->get( 'subscription_behavior_when_vault_fails' );
-			$wc_order->add_meta_data( self::VAULTING_FAILED_META_KEY, $subscription_behavior_when_fails );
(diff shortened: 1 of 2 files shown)


modules/ppcp-vaulting/src/PaymentTokenChecker.php
```diff
@@ -143,6 +143,8 @@ public function check_and_update( int $order_id, int $customer_id, string $inten
 
 		try {
 			$subscription_behavior_when_fails = $this->settings->get( 'subscription_behavior_when_vault_fails' );
+			$wc_order->update_meta_data( self::VAULTING_FAILED_META_KEY, $subscription_behavior_when_fails );
+			$wc_order->save_meta_data();
 		} catch ( NotFoundException $exception ) {
 			return;
```
AlexP11223: should we just return when the value is missing? I think it can happen when updated the plugin but not re-saved the settings yet. Probably should do the old behavior in this case.

Dinamiko: Good catch! I´ve just added it, thanks :)

AlexP11223 (approved):

PR #897, opened 2022-10-10
PayPal Transaction Key meta field not populated for PUI payments (904)

Dinamiko:
The PayPal and PayPal Card Processing gateway store the PayPal fee in the PayPal Transaction Key meta field (for export purposes).

### Steps to reproduce
- create PUI order
- after webhook is retrieved, PayPal fee is patched into the WC order
- "PayPal Transaction Key" meta field does not exist

### Expected behaviour
PayPal Transaction Key meta field exists and is populated with the PayPal fee for every gateway that relates to PayPal Payments.

Possible cause

meta field must probably be created for each gateway

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -248,7 +248,8 @@ class PayUponInvoice {
 						$wc_order->save_meta_data();
 						$paypal_fee = $breakdown->paypal_fee();
 						if ( $paypal_fee ) {
-							update_post_meta( $wc_order->get_id(), 'PayPal Transaction Key', $paypal_fee->value() );
+							$wc_order->update_meta_data( 'PayPal Transaction Key', $paypal_fee->value() );
+							$wc_order->save_meta_data();
 						}
 					}
 				} catch ( RuntimeException $exception ) {
--- a/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
@@ -246,6 +246,10 @@ class PayUponInvoice {
 					if ( $breakdown ) {
 						$wc_order->update_meta_data( PayPalGateway::FEES_META_KEY, $breakdown->to_array() );
 						$wc_order->save_meta_data();
+						$paypal_fee = $breakdown->paypal_fee();
+						if ( $paypal_fee ) {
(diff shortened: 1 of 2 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -246,6 +246,10 @@ function ( WC_Order $wc_order, string $order_id ) {
 					if ( $breakdown ) {
 						$wc_order->update_meta_data( PayPalGateway::FEES_META_KEY, $breakdown->to_array() );
 						$wc_order->save_meta_data();
+						$paypal_fee = $breakdown->paypal_fee();
+						if ( $paypal_fee ) {
+							update_post_meta( $wc_order->get_id(), 'PayPal Transaction Key', $paypal_fee->value() );
```
AlexP11223: I thought we need to use WC order's `update_meta_data` instead of the WP function?

Dinamiko: @AlexP11223 yes but how could otherwise render it on WC Order custom fields?

AlexP11223: hm, I thought it depends only on the name? So if you use the same name without `_` prefix, then it will show it.

AlexP11223:
What is `PayPal Transaction Key`? 

`$paypal_fee->value()` seems to be just the fee amount, not a key.

modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
```diff
@@ -246,6 +246,11 @@ function ( WC_Order $wc_order, string $order_id ) {
 					if ( $breakdown ) {
 						$wc_order->update_meta_data( PayPalGateway::FEES_META_KEY, $breakdown->to_array() );
 						$wc_order->save_meta_data();
+						$paypal_fee = $breakdown->paypal_fee();
+						if ( $paypal_fee ) {
+							$wc_order->update_meta_data( 'PayPal Transaction Key', $paypal_fee->value() );
+							$wc_order->save_meta_data();
```
AlexP11223: You can use just one `save` after `if`.

modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php
Dinamiko: You´re rigth, I just added it here also in `woocommerce_paypal_payments_order_captured` filter, thanks.

AlexP11223 (approved):

PR #901, opened 2022-10-10
Fix the disabled tracking checkbox condition

### LINKED ISSUE #900: Tracking on Connection tab always enabled (933)
### Describe the Bug

There is a bug with tracking checkbox disabled condition. It is always showing as enabled regardless of PUI and shop country settings.

### To Reproduce

1. Go to WC->Settings -> General and set country store something else to Germany (example:Spain)
2. Go to Connection tab -> Tracking is enabled
3. Disable tracking.
4. Move to PayPal checkout tab or somewhere else
5. Go back to Connection tab
6. checkbox Enable tracking information feature on your store is enabled

### Expected Behavior

When PUI is enabled from PUI tab and Woo country is set to Germany then tracking should be always enabled and the checkbox should not be editable (disabled)
In other cases you should be able to enable/disable tracking.

### Actual Behavior

Tracking is always enabled

Narek13:
**Issue**: #900

---

### Description

There is a bug with tracking checkbox disabled condition. It is always showing as enabled regardless of PUI and shop country settings.

### Steps to Test

1. Go to WC->Settings -> General and set country store something else to Germany (example:Spain)
2. Go to Connection tab -> Tracking is enabled
3. Disable tracking.
4. Move to PayPal checkout tab or somewhere else
5. Go back to Connection tab
6. checkbox Enable tracking information feature on your store is enabled -> but should be disabled

---

Closes #900 .

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
@@ -26,22 +26,21 @@ class PayUponInvoiceHelper {
 	protected $checkout_helper;
 
 	/**
-	 * The settings.
+	 * The api shop country.
 	 *
-	 * @var Settings
+	 * @var string
 	 */
-	protected $settings;
+	protected $api_shop_country;
 
 	/**
 	 * PayUponInvoiceHelper constructor.
 	 *
 	 * @param CheckoutHelper $checkout_helper The checkout helper.
-	 * @param Settings       $settings The Settings.
+	 * @param string         $api_shop_country The api shop country.
 	 */
-	public function __construct( CheckoutHelper $checkout_helper, Settings $settings ) {
-
-		$this->checkout_helper = $checkout_helper;
-		$this->settings        = $settings;
+	public function __construct( CheckoutHelper $checkout_helper, string $api_shop_country ) {
+		$this->checkout_helper  = $checkout_helper;
+		$this->api_shop_country = $api_shop_country;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
```diff
@@ -92,12 +91,12 @@ private function is_valid_currency(): bool {
...
+	 * Checks whether PUI gateway is enabled.
 	 *
-	 * @return bool True if PUI is active, otherwise false.
-	 * @throws NotFoundException If problem when checking the settings.
+	 * @return bool True if PUI gateway is enabled, otherwise false.
 	 */
-	public function is_pui_enabled(): bool {
-		return $this->settings->has( 'products_pui_enabled' ) && $this->settings->get( 'products_pui_enabled' );
+	public function is_pui_gateway_enabled(): bool {
+		$gateway_settings = get_option( 'woocommerce_ppcp-pay-upon-invoice-gateway_settings' );
```
Dinamiko, line 99: Better use gateway `enabled` property instead of WP option here, here you have an implementation example: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php#L91-L92

Dinamiko (changes requested):

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
Narek13: @Dinamiko I have tried but when you use it too early, like in settings then the `WC()` doesn't exist yet and I am getting the `Fatal error: Uncaught Error: Call to a member function payment_gateways() on null` and I think that's why the same `get_option` is also used [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/dade07ca95d56680a329edccc62884326ce274fc/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php#L53) and in some other places

Dinamiko (approved):

Issue #513, opened 2022-02-24
improve checkout validation & order creation (448)


InpsydeNiklas:
### Describe the Bug

Following update 1.6.2, the PayPal order is created first and the WooCommerce checkout validation & order creation happens afterward. This change was done to improve plugin compatibility and prevent orders from incorrectly receiving a "Failed" status.
This change means users can leave all WooCommerce checkout fields empty and click the PayPal button. The user can log into their account, confirm the payment, and only after returning to the checkout page, the user would get an error about the checkout fields being empty. No confirmation email will be sent and it should be obvious the payment was not successful.

But users should not receive an error like this **after** confirming the payment screen. If possible, errors should always be thrown before.
Preferably, it should work like this instead:

- the user clicks the PayPal button
- validate WooCommerce checkout fields
- create draft WC order and assign new WP user if registering in the process or Pending Payment order (PayPal Standard style)
- create PayPal order
- after completing the payment in PayPal, create persisting WooCommerce order

### To Reproduce

1. visit the checkout page
2. leave required fields empty
3. click PayPal smart button
4. log in and confirm payment
5. checkout page will reload with an error about the required fields missing
6. user can either fill the required fields and start from step 3 again or reload the page, fill the fields and click the "Place order" button, as the payment has already been pre-authorized.

### Expected Behavior

Checkout form field validation should occur immediately. Clicking the smart button when not all required fields are set throws an error immediately without letting the user log in to their PayPal account.

### Actual Behavior

Checkout form field validation occurs after the PayPal order has been created, letting the user pre-authorize the payment without having provided all details. No WooCommerce order is created.

### Environment

 - WordPress Version 5.9.1
 - WooCommerce Version 6.2.1
 - Plugin Version 1.7.0-test5

### Additional Details

We can intercept the PayPal button click using its onclick parameter (like we do [already](https://github.com/woocommerce/woocommerce-paypal-payments/blob/7f7ae4c700449b9afaa40d00fde9bb1c58cf999b/modules/ppcp-button/resources/js/button.js#L17-L20)) and then calling actions.reject() like [discussed here](https://github.com/paypal/paypal-checkout-components/issues/139#issuecomment-711624526) if we need to stop the payment process (though it seems like it can be used only until this handler finishes, so e.g. sending a request to server and then rejecting will not work).

[...]

PR #942, opened 2022-10-27
Execute server-side WC validation when clicking button

AlexP11223:
Added execution of WC [`validate_checkout`](https://github.com/woocommerce/woocommerce/blob/1e94d3fc180783fd17ae096e84b87dda21dfc5e4/plugins/woocommerce/includes/class-wc-checkout.php#L849) during order creation.

This should be better than the basic JS validation #513, but it is still done in a bit non-standard/hacky way, may break depending on other plugins using the hooks, WC version (the method is not the public API, though hopefully `protected` is close enough to that and unlikely to have breaking changes) etc. There is a filter for disabling it. Also we catch/log the (non-validation) errors during the validation execution, if this happens it will be skipped and continue like before #513. The JS validation also remains for now, but disabled by default. 

Also fixed `errorHandler`s in JS, there were multiple instances of these objects creating their own error lists on the page, which was resulting in outdated messages still being present in some cases.

And it seems like after bb86f1b8e the order creation errors were replaced by the generic message (though probably not just this commit because after simply reverting it the messages were still broken), so fixed this too.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
@@ -66,7 +66,7 @@ class CheckoutActionHandler {
                         }
                     }
 
-                    throw new Error(data.data.message);
+                    throw {type: 'create-order-error', data: data.data};
                 }
                 const input = document.createElement('input');
                 input.setAttribute('type', 'hidden');
@@ -82,9 +82,15 @@ class CheckoutActionHandler {
             onCancel: () => {
                 spinner.unblock();
             },
-            onError: () => {
-                this.errorHandler.genericError();
+            onError: (err) => {
+                console.error(err);
                 spinner.unblock();
+
+                if (err && err.type === 'create-order-error') {
+                    return;
+                }
+
+                this.errorHandler.genericError();
             }
         }
     }
--- a/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
@@ -59,7 +59,9 @@ class CheckoutActionHandler {
                         );
                     } else {
                         errorHandler.clear();
-                        if (data.data.details.length > 0) {
+                        if (data.data.errors.length > 0) {
+                            errorHandler.messages(data.data.errors);
+                        } else if (data.data.details.length > 0) {
                             errorHandler.message(data.data.details.map(d => `${d.issue} ${d.description}`).join('<br/>'), true);
                         } else {
                             errorHandler.message(data.data.message, true);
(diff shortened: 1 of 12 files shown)


modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
```diff
@@ -82,9 +84,15 @@ class CheckoutActionHandler {
             onCancel: () => {
                 spinner.unblock();
             },
-            onError: () => {
-                this.errorHandler.genericError();
+            onError: (err) => {
+                console.error(err);
```
Dinamiko, line 88: Is this console error on purpose here?

modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
```diff
@@ -2,9 +2,10 @@ import CartActionHandler from '../ActionHandler/CartActionHandler';
 import ErrorHandler from '../ErrorHandler';
```
Dinamiko: Seems `ErrorHandler` is not needed.

modules/ppcp-button/resources/js/modules/ContextBootstrap/CheckoutBootstap.js
```diff
@@ -8,11 +8,12 @@ import {
 } from "../Helper/CheckoutMethodState";
 
 class CheckoutBootstap {
-    constructor(gateway, renderer, messages, spinner) {
+    constructor(gateway, renderer, messages, spinner, errorHandler) {
```
Dinamiko, line 10: Check import `ErrorHandler` on line 1, seems not needed.

modules/ppcp-button/resources/js/modules/ContextBootstrap/MiniCartBootstap.js
```diff
@@ -2,17 +2,18 @@ import ErrorHandler from '../ErrorHandler';
 import CartActionHandler from '../ActionHandler/CartActionHandler';
```
Dinamiko, line 1: Check import `ErrorHandler` on line 1, seems not needed.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
```diff
@@ -3,10 +3,11 @@ import UpdateCart from "../Helper/UpdateCart";
 import SingleProductActionHandler from "../ActionHandler/SingleProductActionHandler";
```
Dinamiko, line 2: Check import `ErrorHandler` on line 1, seems not needed.

Dinamiko (approved):
Looks good, only a few imports that seems not needed.

modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
AlexP11223: Yes, we don't show it on the page, so should show it at least in console if it comes from other sources.

PR #960, opened 2022-11-02
Enable OXXO/Standard Card button, add contact message and ensure accepts non physical products (958)

Dinamiko:
- Enables OXXO/Standard Card button by default.
- Adds contact message to payment settings.
- Ensures accepts non physical products on checkout.


modules/ppcp-wc-gateway/src/Helper/CheckoutHelper.php
```diff
@@ -35,14 +35,6 @@ public function is_checkout_amount_allowed( float $minimum, float $maximum ): bo
 			if ( $cart_total < $minimum || $cart_total > $maximum ) {
 				return false;
 			}
-
-			$items = $cart->get_cart_contents();
-			foreach ( $items as $item ) {
-				$product = wc_get_product( $item['product_id'] );
-				if ( is_a( $product, WC_Product::class ) && ! $this->is_physical_product( $product ) ) {
-					return false;
```
AlexP11223, line 43: If it was needed for another gateway, is it handled somewhere else now?

Dinamiko: @AlexP11223 good point, fixed!

AlexP11223 (approved):

PR #962, opened 2022-11-02
Create the pay later tab

### LINKED ISSUE #961: Create Pay Later tab. (915)
### Describe the Feature

It is decided to create a new Pay Later tab to host settings for Pay Later messaging and eventually also the Pay Later button.
The new Pay Later tab will include:

- All existing Pay Later Messaging settings. (with improved design).
- New settings to manage Pay Later button.
- The previews.

The tab will be displayed only when the country is eligible for Pay Later.

Narek13:
**Issue**: #961

---

### Description

Creates the new Pay Later tab to host settings for Pay Later messaging and also the Pay Later button.
The new Pay Later tab will include:

- All existing Pay Later Messaging settings. (with improved design).
- New settings to manage Pay Later button.
- The previews.

Screen Shot 2022-11-02 at 19 08 51

The tab will be displayed only when the country is eligible for Pay Later.
Also the PR includes the migration script from old settings, which will update the values in database for the new settings on plugin update just once.

### Steps to Test

1. Check the new Pay Later tab with all settings.
2. Check if conditionally display of fields works.
3. If Vault is enabled then the Pay Later button and messaging are disabled with disabled checkbox and message.
4. Pay Later tab is not displayed when the store country is not eligible for Pay Later.
5. The Pay Later button is removed/added in Pay Later button preview regarding of the Pay Later Button enabled/disabled checkbox.
6. The messaging previews are working for each location regarding to appropriate settings.
7. When plugin is upgraded the old settings are migrated to the new Pay Later tab.
8. The migration works only once.

---

Closes #961 .


modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
```diff
@@ -323,6 +326,13 @@ private function define_method_description(): string {
 			);
 		}
 
+		if ( $this->is_pay_later_tab() ) {
+			return __(
+				'Example description.',
```
Dinamiko: Do not forgot to fill this value with real description.

modules/ppcp-wc-gateway/pay-later-tab-settings.php
```diff
@@ -0,0 +1,609 @@
...
+// phpcs:disable WordPress.Security.NonceVerification.Recommended
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\WcGateway;
+
+use Psr\Container\ContainerInterface;
+use WooCommerce\PayPalCommerce\ApiClient\Helper\DccApplies;
+use WooCommerce\PayPalCommerce\Onboarding\Environment;
+use WooCommerce\PayPalCommerce\Onboarding\Render\OnboardingOptionsRenderer;
```
Dinamiko: These 3 seems not needed.

Dinamiko (approved):

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -231,7 +232,7 @@ public function __construct(
 	 * Registers the necessary action hooks to render the HTML depending on the settings.
 	 *
 	 * @return bool
-	 * @throws \WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException When a setting was not found.
+	 * @throws NotFoundException When a setting was not found.
```
AlexP11223: `NotFoundException` is not supposed to happen in functions like this because we check `has`.

modules/ppcp-onboarding/resources/js/settings.js
```diff
@@ -223,6 +162,138 @@ document.addEventListener(
...
+                return;
+            }
+
+            const toggleElementsBySelectedLocations = () => {
+                if (! stylingPerMessagingElement.checked || ! payLaterMessagingEnabled.checked) {
+                    return;
+                }
+
+                let checkedLocations = document.querySelectorAll(messagingLocationsSelector + ' :checked');
+                let selectedLocations = [...checkedLocations].map(option => option.value);
```
AlexP11223: can be `const`.

modules/ppcp-onboarding/resources/js/settings.js
```diff
@@ -1,11 +1,18 @@
 document.addEventListener(
```
AlexP11223, line 1: `settings.js` is complex and difficult to read (and not about onboarding anymore) but I am not sure if there is a simple way to improve that.

AlexP11223:
We also should not load the PayPal SDK/init preview rendering on other pages without previews. Added a commit for that.

modules/ppcp-wc-gateway/pay-later-tab-settings.php
Narek13: Removed f228b83e669ca2480e1441ff7b051b44c74bf923

modules/ppcp-button/src/Assets/SmartButton.php
Narek13: Removed 54596fc108e2939412ec61f0abc6f072b53c4a01
But PHPstorm still tells me that "PHPDoc comment doesn't contain all the necessary @throws tags"

modules/ppcp-onboarding/resources/js/settings.js
Narek13: done f3a0c0d0ce97227f8fdcd0be6db009bbc8a7bc30

modules/ppcp-onboarding/resources/js/settings.js
Narek13: yes, it became complex, but I also don't know how to improve it, maybe separate PR is needed. But let's skip for now.

modules/ppcp-button/src/Assets/SmartButton.php
AlexP11223: Yeah, it doesn't know about the relation between `has` and `get` (though maybe it can make sense for PHPStorm devs to consider such context for this inspection), and in general technically it's a correct warning because these functions may be independent from each other (but such implementation would not make sense).

But here we intent to avoid exceptions, so adding such `@throws` makes no sense and only confuses those who use these our functions.

A way to avoid the warning is to add `try catch` for these exceptions instead of `if`, but it is less convenient to read/write.

And I mean this issue is in most of the functions with such `@throws`, not just here. You don't have to fix everything now of course, maybe just don't add such `@throws` in the future.

Narek13: ok, understand, thanks.

modules/ppcp-wc-gateway/pay-later-tab-settings.php
```diff
@@ -58,7 +58,7 @@
 			'input_class'  => array( 'wc-enhanced-select' ),
 			'default'      => array(),
 			'desc_tip'     => false,
-			'description'  => __( 'Example description', 'woocommerce-paypal-payments' ),
+			'description'  => __( 'Example description 1', 'woocommerce-paypal-payments' ),
```
AlexP11223: Why the numbers make it easier? :thinking:

Narek13: For NIklas, later to provide better descriptions, like "Change "Example description 1" to ...." :)

modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
Narek13: Fixed 0e653aa3df79c209e854faef17c7eb0f293dc31a

modules/ppcp-wc-gateway/pay-later-tab-settings.php
Narek13: Fixed 0e653aa3df79c209e854faef17c7eb0f293dc31a

Narek13:
4d0adae73b6f55eb349997d09b012d4a38e942ca -> change the Client ID used for the admin previews if the shop country is eligible for messaging (otherwise will use the currently connected Client ID)
ac420a6d7e4906a25392aea1087e3e81a3fea21c -> fix the small problem with “Pay Later” button preview
4e905ae6e146986180fc90f6d07c3507117725dc ->  fix the styling for preview when the big one is used (1x4 in messaging)

PR #965, opened 2022-11-04
Improve cart subscriptions check and "All products for subscriptions" compatibility

### LINKED ISSUE #629: checkout conflict with "All products for subscriptions" plugin (456)
**The problem**
Checking out with a product that has a “subscription plan” from the "All products for subscriptions" plugin attached, then the payment will fail from the checkout page.
image

Express checkout from product pages or cart appears to be working though.
`2021-12-08T09:37:49+00:00 INFO Trying to save payment for subscription parent order #957.
2021-12-08T09:37:50+00:00 DEBUG GET https://api.sandbox.paypal.com/v2/vault/payment-tokens/?customer_id=test1
Response Debug ID: 5cb1be7671cdf
Response: Array
(
    [code] => 200
    [message] => OK
)

2021-12-08T09:37:50+00:00 ERROR Payment for subscription parent order #957 was not saved.
2021-12-08T09:37:51+00:00 DEBUG  https://api.sandbox.paypal.com/v2/checkout/orders/86C3495571185152F
Response Debug ID: bac2c9f494594
Response: Array
(
    [code] => 200
    [message] => OK
)

2021-12-08T09:37:51+00:00 DEBUG Trying to void order 86C3495571185152F, payments: {"authorizations":[],"captures":[{"id":"71655111UM6968425","status":"COMPLETED","amount":{"currency_code":"EUR","value":"1.10"},"final_capture":true,"seller_protection":{"status":"NOT_ELIGIBLE"},"invoice_id":"test957","custom_id":"957"}]}.
2021-12-08T09:37:51+00:00 ERROR Payment failed: No voidable authorizations.
2021-12-08T09:38:11+00:00 DEBUG POST https://api.sandbox.paypal.com/v1/identity/generate-token
Response Debug ID: cf09486a8025f
Response: Array
(
    [code] => 200
    [message] => OK
)`

**Steps to reproduce**
- install WooCommerce Subscriptions and this plugin called "All products for subscriptions"
- enable manual renewals in the WooCommerce Subscriptions settings
- Set up a regular simple product with a subscription plan from this plugin
- buy this product with the subscription plan from the checkout page
- if no payment method was saved within three minutes, then the order mentioned above may occur since the plugin attempts to void the already captured payment

**Expected behaviour**
A product with a subscription payment plan should be purchasable, with or without vaulting enabled.

**Possible cause**
I think this used to work up until version 1.6.1. When version 1.6.2 implemented a check to fail WooCommerce Subscriptions orders if no payment method was saved is likely when this conflicting behavior started to occur.

Before update 1.6.2, PayPal Payments did not care about whether or not a payment method was saved. It just processed the transaction and whatever happened in the future, happened in the future (sometimes “unexplained” renewal failures).

AlexP11223:
Fixes #629, now it uses authorize intent as expected, and for the product page button for now the button is hidden when the subscription mode is chosen via "All products for subscriptions".

Also fixed and refactored the button hiding. It was only hiding, not showing again on condition change (except for variations) + was using some weird old methods and for some reason part of the checks were in another class making it more difficult to understand.


Dinamiko (approved):

AlexP11223:
Fixed some remaining things and compatibility with newer versions of this plugin, and also now the cart buttons are hidden when the `intent` used during the SDK loading does not match the current cart intent (when adding/removing subscription on the cart page). Ideally we should reload the SDK in this case, but it seems to be not easy to do with our current implementation, so maybe later.

The commit about `SmartButton` is from the blocks branch. Seems like it will take some time to finish this feature, while these changes can be useful for other things, such as here.

Dinamiko (approved):

PR #966, opened 2022-11-04
PUI gateway displayed on pay for order page when mandatory billing fields are left empty or country is unsupprted (739)

Dinamiko:
If merchant sent an link to customer to pay for the order without all mandatory billing fields filled in or added unsupported one to billing details, customer will see PUI gateway and is able to place the order which will fail on the same page.

### Steps To Reproduce
- From WC orders click on Add order button
- Expand the billing section to see all the billing fields
- Fill in all mandatory billing fields except country / region. ( or leave all or some mandatory fields empty)
- Add an supported product to the order 
- Create the order
- Click on customer payment page link

Observe, PUI gateway is displayed.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
@@ -167,8 +167,9 @@ class PayUponInvoiceHelper {
 			'country',
 		);
 
-		foreach ( $address as $key => $value ) {
-			if ( in_array( $key, $required_fields, true ) && $value === '' ) {
+		foreach ( $required_fields as $key ) {
+			$value = $address[ $key ] ?? '';
+			if ( $value === '' ) {
 				return false;
 			}
 		}
--- a/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
@@ -78,6 +78,39 @@ class PayUponInvoiceHelper {
 	}
 
 	/**
+	 * Checks whether pay for order is ready for PUI.
+	 *
+	 * @return bool
+	 */
+	public function is_pay_for_order_ready_for_pui(): bool {
+		/**
+		 * Needed for WordPress `query_vars`.
+		 *
+		 * @psalm-suppress InvalidGlobal
+		 */
+		global $wp;
+
+		$order_id = isset( $wp->query_vars['order-pay'] ) ? (int) $wp->query_vars['order-pay'] : 0;
+		if ( $order_id === 0 ) {
+			return false;
+		}
+
+		$order = wc_get_order( $order_id );
+		if ( ! is_a( $order, WC_Order::class ) ) {
+			return false;
+		}
+
+		$address = $order->get_address();
+		foreach ( $address as $key => $value ) {
+			if ( $value === '' ) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	/**
 	 * Checks if currency is allowed for PUI.
 	 *
 	 * @return bool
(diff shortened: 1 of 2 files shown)


modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
```diff
@@ -77,6 +77,39 @@ public function is_checkout_ready_for_pui(): bool {
...
+		if ( $order_id === 0 ) {
+			return false;
+		}
+
+		$order = wc_get_order( $order_id );
+		if ( ! is_a( $order, WC_Order::class ) ) {
+			return false;
+		}
+
+		$address = $order->get_address();
```
AlexP11223: Does it return only required fields?

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
```diff
@@ -77,6 +77,39 @@ public function is_checkout_ready_for_pui(): bool {
...
+	 */
+	public function is_pay_for_order_ready_for_pui(): bool {
+		/**
+		 * Needed for WordPress `query_vars`.
+		 *
+		 * @psalm-suppress InvalidGlobal
+		 */
+		global $wp;
+
+		$order_id = isset( $wp->query_vars['order-pay'] ) ? (int) $wp->query_vars['order-pay'] : 0;
```
AlexP11223: I think it can be

```
(int) ($wp->query_vars['order-pay'] ?? 0)
```

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
Dinamiko: yup, `get_address` returned fields are required for creating the payment source: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PaymentSourceFactory.php#L47-L56

AlexP11223: I mean it returns something like this (for a normal successful checkout order)

```
{
  ["first_name"]=> string(4) "John"
  ["last_name"]=> string(3) "Doe"
  ["company"]=> string(0) ""
  ["address_1"]=> string(9) "1 Main St"
  ["address_2"]=> string(2) "12"
  ["city"]=> string(8) "San Jose"
  ["state"]=> string(0) ""
  ["postcode"]=> string(5) "11111"
  ["country"]=> string(2) "DE"
  ["email"]=>w string(37) "...@personal.example.com"
  ["phone"]=> string(10) "4085409578"
}
```

so there is at least empty `state` (not shown for DE in checkout I think) and `company`. Maybe other plugins can also add fields.

Seems like the `''` check will always return `false`?

Dinamiko: Fixed, thanks :)

modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
```diff
@@ -132,6 +132,50 @@ private function is_valid_product(): bool {
...
+			'email',
+			'phone',
+			'address_1',
+			'city',
+			'postcode',
+			'country',
+		);
+
+		foreach ( $address as $key => $value ) {
+			if ( in_array( $key, $required_fields, true ) && $value === '' ) {
```
AlexP11223: This should be simpler and faster.

```
		foreach ( $required_fields as $key ) {
			$value = $address[ $key ] ?? '';
			if ( $value === '' ) {
				return false;
			}
		}
```

AlexP11223 (approved):

Issue #959, opened 2022-11-02
when doing add_action( 'woocommerce_email_before_order_table' ), 4th parameter should not be typrized or nullable


larsenlarsson:
see:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Gateway/PayUponInvoice/PayUponInvoice.php#L286

plugin and theme developers may call do_action( 'woocommerce_email_before_order_table' in other context than emails.

therefore the fourth parameter should be nullable if necessary or the typing should be omitted:

```
		add_action(
			'woocommerce_email_before_order_table',
			function( WC_Order $order, bool $sent_to_admin, bool $plain_text, ?WC_Email $email ) {
```

InpsydeNiklas:
This was fixed already with the last update.

PR #1001, opened 2022-11-17
Type check error with German Market plugin (961)

Dinamiko:
User reported this error occuring while Pay upon Invoice is disabled and when attempting to check out with a 0€ product. So no payment gateway is involved, but the checkout still fails due to this error.
```
CRITICAL Uncaught TypeError: Argument 4 passed to WooCommerce\PayPalCommerce\WcGateway\Gateway\PayUponInvoice\PayUponInvoice::WooCommerce\PayPalCommerce\WcGateway\Gateway\PayUponInvoice\{closure}() must be an instance of WC_Email, bool given
```


AlexP11223:
Fixes #959 I guess.

AlexP11223 (approved):

Issue #1020, opened 2022-11-24
Error while syncing tracking data to PayPal -> Sync GZD Tracking (1015)


wolkenkrieger:
Since the update to 2.0.0 the following errors occured while saving/changing the status of an order (to "processed"):

1. Couldn’t sync tracking information: [INVALID_TRACKER_ID] Invalid data provided
2. Couldn’t sync tracking information: [Number of Updates More than Permissible] The requested resource ID was not found

I use the plugin "Germanized for Woocommerce" and the Shipping Carrier is Deutsche Post DHL. The Tracking ID is synced to PayPal (what, for me, is the opposite of error 1?!) but the the order will remain unshipped for PayPal.

This eoor blocks the sending of the "Order is shipped"-Mail to the customer (the coder of "Germanized" is informed, too).

wolkenkrieger:
[Additional]

printed $exception:

```
WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException Object
(
    [message:protected] => [Number of Updates More than Permissible] The requested resource ID was not found
    [string:Exception:private] => 
    [code:protected] => 400
    [file:protected] => /home/wc/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-order-tracking/src/Endpoint/OrderTrackingEndpoint.php
    [line:protected] => 289
    [trace:Exception:private] => Array
        (
            [0] => Array
                (
                    [file] => /home/wc/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-compat/src/CompatModule.php
                    [line] => 176
                    [function] => update_tracking_information
                    [class] => WooCommerce\PayPalCommerce\OrderTracking\Endpoint\OrderTrackingEndpoint
                    [type] => ->
                )

            [1] => Array
                (
                    [file] => /home/wc/wp-includes/class-wp-hook.php
                    [line] => 310
                    [function] => WooCommerce\PayPalCommerce\Compat\{closure}
                    [class] => WooCommerce\PayPalCommerce\Compat\CompatModule
                    [type] => ::
                )

            [2] => Array
                (
                    [file] => /home/wc/wp-includes/class-wp-hook.php
                    [line] => 332
                    [function] => apply_filters
                    [class] => WP_Hook
                    [type] => ->
                )

            [3] => Array
                (
                    [file] => /home/wc/wp-includes/plugin.php
                    [line] => 517
                    [function] => do_action
                    [class] => WP_Hook
[...]

PR #1024, opened 2022-11-25
Prevent throwing an exception during woocommerce_gzd_shipment_after_save hook

dennisnissle:
**Issue**: #1020 

---

### Description

Prevent throwing an exception during `woocommerce_gzd_shipment_after_save` hook as it prevents status transitions from happening in case updating tracking information in PayPal fails while saving a shipment. Check for `tracking_id` availability only as the `has_tracking()` method might perform additional tracking checks.

### Steps to Test

1. Make sure Germanized for WooCommerce is installed
2. Create a new order with PayPal as payment method
3. Create a shipment to the order containing tracking information (tracking id and/or label) and DHL as shipping provider
4. Make sure updating tracking information fails and the [exception](https://github.com/woocommerce/woocommerce-paypal-payments/blob/f6802dd9770c4bdb46ac0fd93072f840820a1c4a/modules/ppcp-compat/src/CompatModule.php#L180) is thrown
5. See how updating the shipment status when saving the shipment fails

### Changelog Entry

Fixes a bug while updating tracking information based on shipment data in Germanized for WooCommerce.

### DIFF
--- a/modules/ppcp-compat/src/CompatModule.php
+++ b/modules/ppcp-compat/src/CompatModule.php
@@ -176,7 +176,6 @@ class CompatModule implements ModuleInterface {
 					! $tracking_information ? $endpoint->add_tracking_information( $tracking_data, $wc_order->get_id() ) : $endpoint->update_tracking_information( $tracking_data, $wc_order->get_id() );
 				} catch ( Exception $exception ) {
 					$logger->error( "Couldn't sync tracking information: " . $exception->getMessage() );
-					$shipment->add_note( "Couldn't sync tracking information: " . $exception->getMessage() );
 				}
 			}
 		);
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):

dennisnissle:
Hi @Dinamiko,

there seems to be a very rare timing issue here - I've just [committed another fix](https://github.com/woocommerce/woocommerce-paypal-payments/pull/1024/commits/65b5b829fe296faac1226b98f58fbf3cfede9b1f), which prevents from calling add_note on the shipment. Could you merge that too?

Best,
Dennis

Dinamiko (approved):

Dinamiko:
Hi @dennisnissle I'm not able to reproduce it at the moment but your fix makes totally sense, there is no reason to throw exception there, thank you very much for the PR.

dennisnissle:
Hi @Dinamiko,

my bad: Actually the status update works fine but the status transition (e.g. status update actions) will not be triggered which lead to shipping emails not being sent etc.

Dinamiko:
> my bad: Actually the status update works fine but the status transition (e.g. status update actions) will not be triggered which lead to shipping emails not being sent etc.
 
@dennisnissle so is it all fixed now or should we create another PR for fixing status transition? sorry for the confusion :)

PR #1029, opened 2022-11-28
Do not call PayPal get order by ID if it does not exist (951)

Dinamiko:
For some reason (probably caused by a JS error that does not load the PayPal buttons) WC order meta returns an empty string in `OrderProcessor::process` and then `OrderEndpoint::order(string $id): Order` is called using that empty string.

```
GET https://api.paypal.com/v2/checkout/orders/
Response: Array
(
    [code] => 404
    [message] => Not Found
)
```
The above request displays "**Could not retrieve order**" error as no PayPal order exist in the request.

This PR ensures get order is not called when there is no id, it also updates the error message.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
+++ b/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
@@ -167,7 +167,12 @@ class OrderProcessor {
 		if ( ! $order ) {
 			$order_id = $wc_order->get_meta( PayPalGateway::ORDER_ID_META_KEY );
 			if ( ! $order_id ) {
-				$this->logger->warning( "No PayPal order ID found in order #{$wc_order->get_id()} meta." );
+				$this->logger->warning(
+					sprintf(
+						'No PayPal order ID found in order #%d meta.',
+						$wc_order->get_id()
+					)
+				);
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -160,12 +162,22 @@ public function __construct(
...
-		$order_id = $wc_order->get_meta( PayPalGateway::ORDER_ID_META_KEY );
-		$order    = $this->session_handler->order() ?? $this->order_endpoint->order( $order_id );
+	public function process( WC_Order $wc_order ): bool {
+		$order = $this->session_handler->order();
 		if ( ! $order ) {
-			$this->last_error = __( 'No PayPal order found in the current WooCommerce session.', 'woocommerce-paypal-payments' );
-			return false;
+			$order_id = $wc_order->get_meta( PayPalGateway::ORDER_ID_META_KEY );
+			if ( ! $order_id ) {
+				$this->logger->warning( "No PayPal order ID found in order #{$wc_order->get_id()} meta." );
```
AlexP11223: I think it must be `$`, not `#`?

Dinamiko: yeah, that's a typo, the idea was to render it as "in order #123 meta"

AlexP11223 (approved):

PR #1047, opened 2022-11-30
Fix shipping address validation

AlexP11223:
`WC_Checkout::get_posted_data` fills the shipping fields when needed, otherwise they can be missing in some cases, resulting in an unexpected validation error.

So now just calling it too, since we already had to set `$_POST`.


Dinamiko (approved):

Bebuk85:
Hi there, i think this update is messing with my settings, i'm using Woofunnel builder (Funnel Kit now), it creates a new checkout form, i'm not using address fields for my checkout because i don't need it, but this update screw up my checkout form, Is there anything to fix this?

no address fields for me need it

PR #1059, opened 2022-12-05
Save and display vaulted payment methods in WC Payment Token API (591)

Dinamiko:
### Reference
https://developer.woocommerce.com/2016/04/04/payment-token-api-in-2-6/
https://github.com/woocommerce/woocommerce/wiki/Payment-Token-API

### Acceptance
Note that saving and retrieving payments could take some time as they relies on receiving webhooks from PayPal.

#### Migration
- Save a couple of payments (PayPal and ACDC) using current released version.
- Edit plugin main file header and change Version value to a lower one.
- Upload this package via “Plugins” page replacing the existing one.
- Go to “Connection / Subscribed webhooks” and ensure `vault payment-token created` and `vault payment-token deleted` Tracked events exist.
- Go to “My account / Payment methods” and ensure previous saved payments exists.

#### Save Payments
Place a couple of orders (PayPal and ACDC) and ensure payments are saved in “My account / Payment methods“

#### Delete Payment
- Go to “My account / Payment methods“ and delete existing payment, ensure it is also deleted in PayPal side (Settings / Payments / Manage automatic payments).
- Go to PayPal saved payments and delete payment, ensure payment is also delete in WooCommerce side.

PR #1104, opened 2022-12-30
Remove user_meta storage of Payment Tokens

prettyboymp:
### Description

The meta key used for storing a copy of Payment Tokens for a customer isn't blog prefixed.  In a multisite setup, this means that a customer purchasing from more than one store will have the saved Payment Tokens overwritten with the tokens of another.  Fortunately, this isn't currently causing any issues because the only code that reads from the user meta, `PaymentTokenRepository::for_user_id()`, is no longer in use. 

### Steps to Test

1. Search the code for other uses of `PaymentTokenRepository::for_user_id()` to verify it is no longer in use.
2. Search to validate that there are no uses of `get_user_meta()` using the `PaymentTokenRepository::USER_META` key remaining to verify that no code is actually reading from the stored meta data.

### Changelog Entry

> Remove user_meta storage of Payment Tokens.

### DIFF
--- a/modules/ppcp-vaulting/src/PaymentTokenRepository.php
+++ b/modules/ppcp-vaulting/src/PaymentTokenRepository.php
@@ -20,8 +20,6 @@ use WooCommerce\PayPalCommerce\ApiClient\Factory\PaymentTokenFactory;
 class PaymentTokenRepository {
 
 
-	const USER_META = 'ppcp-vault-token';
-
 	/**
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):
Hi @prettyboymp thanks for the PR!

Let's merge this as part of the PR we are working about migrating custom implemetation to WC Payment Token API #1059

PR #1106, opened 2023-01-02
Do not check type in filter callback signature (994)

Dinamiko:
Fixes type check error in callback signature:
```
An error of type E_ERROR was caused in line 66 of the file [...]/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Gateway/OXXO/OXXO.php. Error message: Uncaught TypeError: Argument 1 passed to WooCommerce\PayPalCommerce\WcGateway\Gateway\OXXO\OXXO::WooCommerce\PayPalCommerce\WcGateway\Gateway\OXXO\{closure}() must be of the type array, null given, called in [...

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/OXXO/OXXO.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/OXXO/OXXO.php
@@ -63,7 +63,15 @@ class OXXO {
 
 		add_filter(
 			'woocommerce_available_payment_gateways',
-			function ( array $methods ): array {
+			/**
+			 * Param types removed to avoid third-party issues.
+			 *
+			 * @psalm-suppress MissingClosureParamType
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/ppcp-wc-gateway/src/Gateway/OXXO/OXXO.php [hunk 1]
```diff
@@ -63,7 +63,15 @@ public function init(): void {
 
 		add_filter(
 			'woocommerce_available_payment_gateways',
-			function ( array $methods ): array {
```
AlexP11223, line 66: hm, that's a strange issue, why would anyone call it with `null`?

Also we use this filter in other places besides oxxo.

AlexP11223 (approved):

PR #1177, opened 2023-02-10
`DEVICE_DATA_NOT_AVAILABLE` error message when FraudNet is enabled (1429)

Dinamiko:
When FraundNet setting is enabled and not configured correctly is displays `DEVICE_DATA_NOT_AVAILABLE` error message when trying to create the PayPal order.

We've found that `fraudnet.js` script was not loaded due to a bug, this PR fixes that and loads it correctly.

Please note that Fraudnet `fncls` script is currently rendered twice, the one we are implementing have the following format: 
```
<script type="application/json" fncls="fnparams-dede7cc5-15fd-4c75-a9f4-36c430ee3a99">
      {
          "f":"<32_character_GUID>",
          "s":"<merchant_id>_<page_id>",
          "sandbox":false
      }
  </script>
```
The other is added by PayPal automatically with the following format: 
```
... "s":"SMART_PAYMENT_BUTTONS","u":"example.com","cb1":"fnCallback"...
```

### Steps to reproduce
- Disable all payments except PayPal payment.
- Enable FraudNet setting.
- Ensure Checkout exist in Smart Button Locations, otherwise add it*.
- Add product to Cart and go to Checkout page.
- Search in page source code for `fnparams-`

Check for one** of the `fnparams-` scripts that contains `s` param with the following format `<merchant_id>_<page_id>` 

*The same applies for Single Product, Cart and Mini-Cart locations.
**Currently `fnparams` is rendered twice with different values, one from the plugin code (fraudnet.js) and the other added by PayPal automatically.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Assets/FraudNetAssets.php
+++ b/modules/ppcp-wc-gateway/src/Assets/FraudNetAssets.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 namespace WooCommerce\PayPalCommerce\WcGateway\Assets;
 
 use WooCommerce\PayPalCommerce\Onboarding\Environment;
+use WooCommerce\PayPalCommerce\Session\SessionHandler;
+use WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException;
 use WooCommerce\PayPalCommerce\WcGateway\FraudNet\FraudNet;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayUponInvoice\PayUponInvoiceGateway;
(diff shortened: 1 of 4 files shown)


modules/ppcp-wc-gateway/src/Assets/FraudNetAssets.php
```diff
@@ -168,19 +170,64 @@ protected function are_buttons_enabled_for_context() : bool {
...
-			return true;
+		return in_array( $this->context(), $button_locations, true );
+	}
+
+	/**
+	 * Get context.
+	 *
+	 * @return string
+	 */
+	protected function context(): string {
```
AlexP11223: Would be great to extract this somewhere to avoid duplication (currently here and in `SmartButton`). Maybe into a trait.

AlexP11223 (approved):

PR #1188, opened 2023-02-14
Add filter to fallback tracking_data['carrier']

popforce:
PayPal Payments always defaults to DHL_DEUTSCHE_POST.  As Germanized offers to create your own carriers we should be able to filter it, thus being able to return the correct  carrier for PayPal, which is e.g. DPD_DE, GLS_DE, OTHER


AlexP11223 (approved):

InpsydeNiklas:
@popforce Thanks for the contribution! The filter `woocommerce_paypal_payments_default_gzd_carrier` will be included in the next update 2.0.3.

PR #1193, opened 2023-02-15
PPEC compatibility layer does not take over subscriptions (923)

Dinamiko:
When I have a PPEC subscription and disable PPEC and enable PCP, the subscription goes to Manual renewal instead of PayPal (Legacy).

### Steps To Reproduce
- disable PayPal Payments
- create subscription with PayPal Checkout plugin
- disable PayPal Checkout plugin
- enable PayPal Payments
- visit subscription

### Expected behavior
It should say “PayPal (Legacy)” as a payment method and PayPal Payments should be able to renew it automatically.

### Possible cause
`should_mock_ppec_gateway` returns `false` because `get_current_screen` does not exist: [woocommerce-paypal-payments/SubscriptionsHandler.php at trunk · woocommerce/woocommerce-paypal-payments](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php#L186)

### Suggested solution
Instead of using `get_current_screen` , use the values in the request to validate the current screen.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@
 * Fix - Validate new user creation #1131
 * Fix - After Updating to 2.0.2, Site Health reports REST API error #1195
 * Fix - Do not send buyer-country for previews in live mode to avoid error #1186
+* Fix - PPEC compatibility layer does not take over subscriptions #1193
 * Enhancement - Save checkout form before free trial redirect #1135
 * Enhancement - Add filter for controlling the ditching of items/breakdown #1146
 * Enhancement - Add patch order data filter #1147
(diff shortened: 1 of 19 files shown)


AlexP11223 (approved):

meemal:
Is there a way I can apply this fix now to my affected site?  Time is critical

AlexP11223:
@meemal you can try this package https://github.com/woocommerce/woocommerce-paypal-payments/releases/tag/2.0.3-rc2

meemal:
I'm very grateful to you! Thanks so much, you have definitely improved my Monday morning by about 500%!

PR #1201, opened 2023-02-21
Redirect to connection tab after manual credentials input

### LINKED ISSUE #1196: Wrong redirection after applying keys (1456)
### Describe the Bug

After entering the necessary credentials, the user is redirected to the "Standard Payments" tab, but they see the contents of the "Connection" Tab

They need to click again on "Standard Payments" tab, and the correct content loads.

I've attached screenshots

### To Reproduce

1. Go to `PayPal > Connection` and click on `clear now`
2. Click on `Toggle to manual credential input`
3. Enter credentials and press save
4. You're redirected to the "Standard Payments" tab (highlighted) but the content of that tab is the one from "connection"
5. Click again on "Standard Payments" to enable paypal

### Screenshots

image

image

image

### Expected Behavior

Users should be redirected to the standard payments tab, and the correct content should appear. Otherwise, they don't know that PayPal is not activated and need to do something to enable it.

image

AlexP11223:
Fixes #1196 

I am not completely sure why it was showing the wrong tab content until refresh, most likely it was getting the old state somewhere because the new state (credentials) was not saved yet into our settings.
But either way, we should redirect to the connection tab after manual credential input, similarly as we do after the standard onboarding, for consistency and to avoid issues like this.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
+++ b/modules/ppcp-wc-gateway/src/Settings/SettingsListener.php
@@ -193,10 +193,7 @@ class SettingsListener {
 		 */
 		do_action( 'woocommerce_paypal_payments_onboarding_before_redirect' );
 
-		/**
-		 * The URL opened at the end of onboarding after saving the merchant ID/email.
-		 */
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):
Agreed, better redirect to Connection tab after manual credentials for consistency.

PR #1207, opened 2023-02-22
Do not call get_available_payment_gateways early

### LINKED ISSUE #1195: After Updating to 2.0.2, Site Health reports REST API error (1459)
### Describe the Bug

After Updating to 2.0.2, Site Health reports REST API error. I've tested this many times with and without the plugin. The issue only appears after updating the plugin to 2.0.2. Version 2.0.1 does not exhibit this issue.

### To Reproduce

Update plugin to 2.0.2
Run Site Health Analysis

### Screenshots

The REST API encountered an unexpected result
Performance
The REST API is one way that WordPress and other applications communicate with the server. For example, the block editor screen relies on the REST API to display and save your posts and pages.

When testing the REST API, an unexpected result was returned:

REST API Endpoint: https://mywebsitehere.com/wp-json/wp/v2/types/post?context=edit
REST API Response: (500) Internal Server Error

### Expected Behavior

No REST API error

### Actual Behavior

REST API error

### Environment

 - WordPress Version: 6.1.1
 - WooCommerce Version: 7.4.0
 - Plugin Version: 2.0.2
 - Browser: Chrome 110.0.5481.96
 - Any other plugins installed

AlexP11223:
Fixes #1195

Calling it during object construction causes issues with third-party plugins/filters when it is executed too early or in irrelevant requests (WP ajax for saving posts, ...). Now this function call is moved into a separate class to make it more difficult to accidentally execute early or when not needed.

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -36,6 +36,7 @@ use WooCommerce\PayPalCommerce\WcGateway\Endpoint\ReturnUrlEndpoint;
 use WooCommerce\PayPalCommerce\WcGateway\FundingSource\FundingSourceRenderer;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\CardButtonGateway;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\CreditCardGateway;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\GatewayRepository;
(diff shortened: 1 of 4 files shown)


Dinamiko (approved):

AlexP11223:
Also removed return type-hint in PUI filter function. It's probably a separate conflict but either way it makes no sense to expect `array` after `! is_array` check.

PR #1210, opened 2023-02-24
Add Playwright

Dinamiko:
Adds [Playwright](https://playwright.dev/) end-to-end testing framework.


.gitignore
```diff
@@ -11,3 +11,6 @@ modules/ppcp-wc-gateway/assets/css
 .env
 .env.e2e
 auth.json
+/tests/playwright/test-results/
+/tests/playwright/playwright-report/
+/tests/playwright/.cache/
```
AlexP11223: I would move it into `/tests/playwright/.gitignore`

AlexP11223 (approved):

.env.sample
```diff
@@ -0,0 +1,8 @@
+BASEURL="http://example.com"
```
AlexP11223: `.env` sounds more like something required for the whole project to run, because we already have `.env.e2e`

Maybe it should either use the above file or `.env.playwright`.

Also `.example` seems to be a more common convention than `.sample`.

The path can be specified like this https://github.com/motdotla/dotenv#path

tests/playwright/README.md
```diff
@@ -0,0 +1,14 @@
...
+Allows using environment variables inside the tests.
+
+- Duplicate `.env.sample` and rename it as `.env`, set values and add new variables if needed.
+
+## Run Tests
+```
+$ npx playwright test
+$ npx playwright test example.spec.js --headed
+$ npx playwright test example.spec.js --debug
+$ npx playwright test -g "Test name here"
```
AlexP11223, line 14: Would be great to add the first one to `package.json` similarly to `ddev:e2e-tests` (`ddev:playwright-tests`, ...), and to `ddev:test`.

AlexP11223 (changes requested):

tests/playwright/README.md
Dinamiko: The problem is that the tests that opens PayPal modal does not work without opening the browser (`--headed`), the idea is to create some basic tests that calls the PayPal API instead of opening the browser (https://playwright.dev/docs/test-api-testing#sending-api-requests-from-ui-tests) and use only these for CI.

 I would rather prefer to include those in package.json once implemented, for now the idea is just have Playwright available to help us in development by reducing manual testing.

AlexP11223: It's not for CI, CI executes the commands directly. It's to avoid having to remember/copy these commands every time during development, or installing npx on host (or executing `ddev ssh` before).

And if `--headed` is needed, then it probably should filter them in some simpler way than adding file names one by one, maybe there is a way to define groups like in PHPUnit and then execute the whole headed or non-headed group.

tests/playwright/place-order.spec.js
```diff
@@ -46,7 +46,7 @@ test('PayPal button place order from Product page', async ({page}) => {
     await expect(title).toHaveText('Order received');
 });
 
-test('Advanced Credit and Debit Card (ACDC) place order from Checkout page', async ({page}) => {
+test('Advanced Credit and Debit Card (ACDC) place order from Checkout page @ci', async ({page}) => {
```
AlexP11223, line 49: Does it work now if executed in the GH workflow?

package.json
```diff
@@ -43,7 +43,8 @@
     "ddev:composer-update": "ddev composer update && ddev composer update --lock",
     "ddev:unit-tests": "ddev exec phpunit",
     "ddev:e2e-tests": "cp -n .env.e2e.example .env.e2e && ddev php tests/e2e/PHPUnit/setup.php && ddev exec phpunit -c tests/e2e/phpunit.xml.dist",
-    "ddev:test": "yarn run ddev:unit-tests && yarn run ddev:e2e-tests",
+    "ddev:pw-tests": "npx playwright test --grep @ci",
```
AlexP11223: I would also add e.g. `"ddev:pw-tests-all": "npx playwright test --headed",`

Oh, and can it be `ddev exec npx .....`? To avoid installing on host.

.env.e2e.example
```diff
@@ -1 +1,10 @@
 PPCP_E2E_WP_DIR=${ROOT_DIR}/.ddev/wordpress
+
+BASEURL="http://example.com"
```
AlexP11223: Can it be `https://woocommerce-paypal-payments.ddev.site`?

tests/playwright/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Local Environment Variables
+Allows using environment variables inside the tests.
+
+- Duplicate `.env.e2e.example` and rename it as `.env.e2e`, set values and add new variables if needed.
+
+## Run Tests
+```
+$ npx playwright test
+$ npx playwright test --grep @ci
```
AlexP11223, line 11: Would be cool to add it in CI (e.g. the current e2e one) if it works.

AlexP11223 (approved):

PR #1217, opened 2023-02-28
Integrate PayPal Subscriptions API (991)

Dinamiko:
Integrate [PayPal Subscriptions API](https://developer.paypal.com/docs/subscriptions/integrate/).

## Acceptance Criteria

- Activate WC Subscriptions Plugin.
- In Connection tab go to Webhooks section and click “Resubscribe”.
- Go to “Standard Payments” tab and select “PayPal Subscriptions” in “Subscriptions Mode” selector.

### Create new subscription product
- Add new product as Simple subscription.
- Check “Connect to PayPal”, add a Plan name and click “Publish”.
- Ensure both Product and Plan links goes to the correct PayPal Product and Plan pages.

### Update subscription product
- Edit previous subscription product and change Product title and “Subscription price”.
- Ensure the name and the price are updated on PayPal Product page.

### Purchase Subscription from Checkout Page
- As customer add 1 subscription product to the cart and go to Checkout page.
- Fill Checkout form and click PayPal button.
- Ensure Subscription is created correctly on PayPal (click “PayPal Subscription” link in My Account / Subscription).

### Purchase Subscription from Single Product Page
- As customer go to subscription product page and click PayPal button.
- Fill Checkout form and click “Place Order” button.
- Ensure Subscription is created correctly on PayPal (click “PayPal Subscription” link in My Account / Subscription).

### Purchase Subscription from Cart Page
- As customer add 1 subscription product to the cart and go to Cart page.
- Fill Checkout form and click PayPal button.
- Ensure Subscription is created correctly on PayPal (click “PayPal Subscription” link in My Account / Subscription).

### Subscriber Suspend Subscription
- As customer go to My Account / Subscriptions and edit an active one.
- Click “Suspend” button.
- Ensure status in PayPal is “Inactive”.
- Click “Reactivate” button.
- Ensure status in PayPal is “Active”.

### Subscriber Cancel Subscription
- As customer go to My Account / Subscriptions and edit an active one.
- Click “Cancel” button.
- Ensure status in PayPal is “Inactive”.

### Renewals
- Create a subscription product with a live cycle of one day.
- As customer purchase this product and wait 24 hours to receive the first renewal in WC subscription from PayPal.


modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
```diff
@@ -120,6 +120,11 @@ public function create_token( string $description, string $return_url, string $c
 	 */
 	public function reference_transaction_enabled(): bool {
 		try {
+			$reference_transaction_enabled = get_transient( 'ppcp_reference_transaction_enabled' ) ?? '';
+			if ( $reference_transaction_enabled === true ) {
```
AlexP11223: `??` seems unneeded (and the variable too).

modules/ppcp-api-client/src/Endpoint/BillingPlans.php
```diff
@@ -0,0 +1,229 @@
...
+			'headers' => array(
+				'Authorization' => 'Bearer ' . $bearer->token(),
+				'Content-Type'  => 'application/json',
+				'Prefer'        => 'return=representation',
+			),
+		);
+
+		$response = $this->request( $url, $args );
+		if ( is_wp_error( $response ) || ! is_array( $response ) ) {
+			throw new RuntimeException( 'Not able to get product.' );
```
AlexP11223: Probably should be `plan`?

modules/ppcp-webhooks/src/Handler/BillingPlanPricingChangeActivated.php
```diff
@@ -0,0 +1,93 @@
...
+		$response = array( 'success' => false );
+		if ( is_null( $request['resource'] ) ) {
+			return new WP_REST_Response( $response );
+		}
+
+		$plan_id = wc_clean( wp_unslash( $request['resource']['id'] ?? '' ) );
+		$price   = wc_clean( wp_unslash( $request['resource']['billing_cycles'][0]['pricing_scheme']['fixed_price']['value'] ?? '' ) );
+		if ( $plan_id && $price ) {
+			$args = array(
+				'meta_key' => 'ppcp_subscription_plan',
```
AlexP11223, line 77: There are phpcs warnings in all these webhook handlers about possibly slow queries.

If there is no easy way to fix this, we should `phpcs:ignore WordPress.DB.SlowDBQuery` to avoid clutter.

modules/ppcp-button/services.php
```diff
@@ -65,25 +66,88 @@
...
+		if ( $source && $source->card() ) {
+			return false; // Ignore for DCC.
+		}
+		if ( 'card' === $session_handler->funding_source() ) {
+			return false; // Ignore for card buttons.
+		}
+
+		return true;
+	},
+	'button.context'                              => static function( ContainerInterface $container ): string {
```
AlexP11223: It may cause a merge conflict (if deleted from `ContextTrait`, or if not then it will be different in these two places), also it may not be a good idea to move this into a service because some of these checks may fail if called too early.

modules/ppcp-subscription/src/SubscriptionModule.php
```diff
@@ -135,6 +140,399 @@ function( array $data ) use ( $c ) {
...
+						$subscription_plan_name      = $product->get_meta( '_ppcp_subscription_plan_name' );
+
+						echo '<div class="options_group subscription_pricing show_if_subscription hidden">';
+						echo '<p class="form-field"><label for="_ppcp_enable_subscription_product">Connect to PayPal</label><input type="checkbox" id="ppcp_enable_subscription_product" name="_ppcp_enable_subscription_product" value="yes" ' . checked( $enable_subscription_product, 'yes', false ) . '/><span class="description">Connect Product to PayPal Subscriptions Plan</span></p>';
+
+						$subscription_product = $product->get_meta( 'ppcp_subscription_product' );
+						$subscription_plan    = $product->get_meta( 'ppcp_subscription_plan' );
+						if ( $subscription_product && $subscription_plan ) {
+							$environment = $c->get( 'onboarding.environment' );
+							$host        = $environment->current_environment_is( Environment::SANDBOX ) ? 'https://www.sandbox.paypal.com' : 'https://www.paypal.com';
```
AlexP11223: Maybe this URL check should be extracted somewhere, a service.

woocommerce-paypal-payments.php
```diff
@@ -25,7 +25,7 @@
 define( 'PAYPAL_SANDBOX_API_URL', 'https://api.sandbox.paypal.com' );
 define( 'PAYPAL_INTEGRATION_DATE', '2023-03-20' );
 
-define( 'PPCP_FLAG_SUBSCRIPTION', true );
+define( 'PPCP_FLAG_SUBSCRIPTIONS_API', apply_filters( 'ppcp_flag_subscriptions_api', false ) );
```
AlexP11223: Maybe it can also use an env var as default value instead of `false`? Then you can easily enable it in DDEV etc.

`apply_filters( 'ppcp_flag_subscriptions_api', getenv( 'PPCP_FLAG_SUBSCRIPTIONS_API' ) === '1' )`

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -228,6 +261,10 @@ public function __construct(
 		bool $basic_checkout_validation_enabled,
 		bool $early_validation_enabled,
 		array $pay_now_contexts,
+		string $intent,
+		string $context,
+		bool $can_save_vault_token,
```
AlexP11223: There is `$can_save_vault_token` field and `can_save_vault_token` function not using the field. Seems like one of them should be removed.

modules/ppcp-button/services.php
```diff
@@ -65,25 +67,86 @@
...
```
AlexP11223: This does not match `ContextTrait`.

Can you remove this service and use `ContextTrait` instead? Because putting this into a service may not be a good idea anyway, some of these checks may fail if called too early.

modules/ppcp-button/services.php
```diff
@@ -65,25 +67,86 @@
...
+	},
+	'button.intent'                               => static function( ContainerInterface $container ): string {
+		$settings           = $container->get( 'wcgateway.settings' );
+		$subscription_helper = $container->get( 'subscription.helper' );
+		assert( $subscription_helper instanceof SubscriptionHelper );
+		$context = $container->get( 'button.context' );
+
+		$intent               = ( $settings->has( 'intent' ) ) ? $settings->get( 'intent' ) : 'capture';
+		$product_intent       = $subscription_helper->current_product_is_subscription() ? 'authorize' : $intent;
+		$other_context_intent = $subscription_helper->cart_contains_subscription() ? 'authorize' : $intent;
```
AlexP11223: hm, looks like this causes `Notice: Function get_cart was called incorrectly. Get cart should not be called before the wp_loaded action.` when blocks are enabled.

`'blocks.method'` needs `SmartButton` which needs this thing now. And `'blocks.method'` is executed before `wp_loaded` in `woocommerce_blocks_payment_method_type_registration`.

AlexP11223: Seems like `button.intent` is not used anywhere except passing to `SmartButton`, so it can be moved into a function inside `SmartButton` or somewhere else. Similarly to context, it is easy to accidentally misuse such services, execute when not supposed to or too early.

AlexP11223 (approved):

PR #1223, opened 2023-03-02
Webhooks not registered when onboarding with manual credentials (1480)

Dinamiko:
When onboarding with manual credentials webhook registration does not work. 

### Possible cause

1. Not directly related but there are unnecessary calls to endpoints (ex: [woocommerce-paypal-payments/services.php at trunk · woocommerce/woocommerce-paypal-payments](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/services.php#L1288) ) when not onboarded yet causing 404 entries in the logs.
2. When onboarding with manual credentials it calls `/v1/notifications/webhooks` too early with wrong host and empty Authorization causing 404 not found error.

### Suggested solution

Execute webhook registration logic when onboarding is already completed.


AlexP11223:
We register webhooks via WP CRON in the standard onboarding, we can do the same here.

Also we check and register webhooks when visiting the connection page https://github.com/woocommerce/woocommerce-paypal-payments/blob/91c39fdef2a1abdea63bd8ee06908a9342c2fb16/modules/ppcp-webhooks/src/WebhookModule.php#L130-L140
So we can even completely remove the registration during onboardings. But for now it probably should remain, e.g. in case of onboarding via the WC setup wizard (not sure if WC still has it), or for changing credentials manually.

AlexP11223 (approved):

PR #1262, opened 2023-03-17
Add functions for capturing, refunding, voiding programmatically

### LINKED ISSUE #590: programmatically capturing/voiding authorized payments (506)

AlexP11223:
Resolves #590

Adds `ppcp_*` global functions for capturing, refunding, voiding, and also for retrieving PayPal order since it may be needed for deciding whether to perform these operations (like we do in `RefundProcessor`).

Also now there is an easier way to retrieve the app container: `PPCP::container()`. But as before, using it may not be the best approach for third-party developers because we do not guarantee the backward compatibility.

I am not sure if `ppcp` is the prefix that we should use, since it has no relation to the plugin name. It probably refers to [PayPal Commerce Platform](https://developer.paypal.com/docs/multiparty/) which seems suitable, but not mentioned in the plugin description etc. In our filters/hooks there is no consistent prefix currently.


api/order-functions.php
```diff
@@ -0,0 +1,107 @@
...
+	$authorized_payment_processor = PPCP::container()->get( 'wcgateway.processor.authorized-payments' );
+	assert( $authorized_payment_processor instanceof AuthorizedPaymentsProcessor );
+
+	if ( ! $authorized_payment_processor->capture_authorized_payment( $wc_order ) ) {
+		throw new RuntimeException( 'Capture failed.' );
+	}
+}
+
+/**
+ * Captures the PayPal order.
```
Dinamiko: Seems here Refunds instead of Captures.

AlexP11223: Fixed.

Dinamiko (approved):

AlexP11223:
The refund ID now is not added to meta during `ppcp_refund_order`, and the function returns it. 

This way the refund can be added to WC by webhooks, if the meta does not contain the ID yet.

(for normal refund on both sides there is `wc_refund_payment`)

PR #1275, opened 2023-03-22
Add content type json to all fetch ajax endpoints (1450)

Dinamiko:
This PR adds json content type header to all fetch ajax endpoints to avoid error: 
```
Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```


AlexP11223:
I think it's not all https://github.com/woocommerce/woocommerce-paypal-payments/search?l=JavaScript&q=fetch

AlexP11223 (approved):

PR #1290, opened 2023-03-31
DDEV command for cleaning up testing data

Dinamiko:
DDEV command for cleaning up data created during testing.


.ddev/commands/web/wp-cleanup
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+help-panel() {
```
AlexP11223: I would call it `show-help`

AlexP11223 (approved):

Issue #211, opened 2021-07-30
WooCommerce PayPal Payments doesn't support the new Checkout block


solstudioim:
## Describe the bug
I tried to use Checkout Block for my checkout page, but it's not displayed

## To reproduce

1. Download, activate, enable [WooCommerce PayPal Payments](https://wordpress.org/plugins/woocommerce-paypal-payments/)
2. Add to cart a product
3. Checkout and see if WooCommerce PayPal Payments not shown on the checkout page

## Expected behavior
WooCommerce PayPal Payments should be shown and compatible with Checkout Block, as the built-in PayPal Standard is hidden in the latest WooCommerce versions

## Screenshots
<img width="1601" alt="Markup 2021-07-30 at 08 06 04" src="https://user-images.githubusercontent.com/17371139/127584707-35b76e83-311c-43b8-bd7b-522ca0da8e97.png">
<img width="1451" alt="Markup 2021-07-30 at 08 09 50" src="https://user-images.githubusercontent.com/17371139/127584711-8b322072-9200-43b2-81d9-7036cb7e02be.png">

### WordPress (please complete the following information):
* Core version: 5.8
* WooCommerce version: 5.5.2
* WooCommerce Blocks version: 5.5.1
* Site language: English
* WooCommerce PayPal Payments: 1.4.0

### Desktop (please complete the following information):
* OS: Mac
* Browser: Chrome
* Version: 92.0.4515.107

## Additional context
A user on the forum [requested compatibility information between WC Block and WC PayPal Payments](https://wordpress.org/support/topic/compatibility-with-recommended-paypal-plugin-from-woocommerce/?view=all)

phillipcornwell:
another report: 33394334-hc

thelogicwizards:
another report: [4859295-zd](4859295-zd-woothemes)

KyleIrving:
Considering that Woo are advising users not to use PayPal Standard any more and there are no other PayPal add ons that work with WooCommerce Blocks what's the plan here?

dweuste:
> Considering that Woo are advising users not to use PayPal Standard any more and there are no other PayPal add ons that work with WooCommerce Blocks what's the plan here?

This exactly!

InpsydeNiklas:
[...]

PR #1346, opened 2023-04-27
Add express cart/checkout block (157)

AlexP11223:
Adds express payment method in WC blocks.

Requires latest WC and WC Blocks plugins. To enable blocks, add `add_filter( 'woocommerce_paypal_payments_blocks_enabled'
, '__return_true' );` or `PCP_BLOCKS_ENABLED=1` env var, and add new locations in `Smart Button Locations`, `Pay Later Button Locations` settings. Also pages with the blocks should be set in WC Advanced settings as checkout/cart pages.

Currently it only adds the express buttons on top of the checkout (and on the cart page), not classical payment method selection at the bottom.

image

 By default the page will refresh after payment completion in the popup, the billing/shipping info will be filled and only PayPal payment method will be shown, allowing to review and finish the order or cancel, similarly to payments from non-block product or cart pages (they also can be handled in this block). 

image

The review step can be disabled in the settings. This will also add shipping method selection in the PayPal popup (may cause slower popup loading, also may not handle some corner cases well, displaying shipping taxes).

If billing address retrieval is not enabled by PayPal for the merchant account, then it may fail to fill the address fields, and if review is disabled the checkout will fail.

The changes of this PR are mostly isolated in the block module, but some things about context handling are added in the `SmartButton` and related classes.

Also a bit of other small refactoring, and Playwright improvements.


Dinamiko (approved):

Dinamiko:
Implements #211.

PR #1354, opened 2023-05-04
Improve error message on certain exceptions (696)

### LINKED ISSUE #1281: Error message displayed to customers during checkout
### Describe the Bug

When a customer attempts to checkout, they receive an unfriendly and complex error message related to PayPal payment processing. The error message should be user-friendly and easily understood by the customer.

### To Reproduce

Go to the WooCommerce store as a customer
Add items to the cart
Proceed to the checkout page
Fill in the required information and choose PayPal as the payment method
Click on the 'Place Order' button (authorized mode)
See the error message

### Expected Behavior

The customer should see a user-friendly error message that informs them there was a problem processing the payment and to try again or contact the shop admin for assistance.

### Actual Behavior

The customer sees a complex error message that includes technical details, which can be confusing:

Failed to process the payment. Please try again or contact the shop admin. [UNPROCESSABLE_ENTITY] The requested action could not be performed, semantically incorrect, or failed business validation. https://developer.paypal.com/docs/api/orders/v2/#error-PAYER_CANNOT_PAY PayPalGateway.php:525 [UNPROCESSABLE_ENTITY] The requested action could not be performed, semantically incorrect, or failed business validation. https://developer.paypal.com/docs/api/orders/v2/#error-PAYER_CANNOT_PAY OrderEndpoint.php:406 Order status changed from Pending payment to Failed.

### Environment

WordPress Version: 5.9.5
WooCommerce Version: 6.9.3
Plugin Version: 2.0.3
Browser: N/A (assuming issue occurs in all browsers)
Other plugins installed: N/A (assuming issue occurs with or without other plugins)

### Additional Details

Update the error handling code to display a generic user-friendly message such as: "Failed to process the payment. Please try again or contact the shop admin for assistance."

mmaymo:
Some more error messages should be improved

### Steps to Test
To mock these error responses, this filter can be used:
```
        $args['headers']['PayPal-Mock-Response'] = '{"mock_application_codes": "INSTRUMENT_DECLINED"}';
    }
    return $args;
}, 10, 2);
```
Change `INSTRUMENT_DECLINED` for the error response you would like to rest. May not work with all errors and only appears to apply for the PayPal gateway and not advanced card payments.


AlexP11223:
Maybe it can be done inside `PayPalApiException`? (and ideally this message for the user should be a separate field handled by the thing outputting errors, but it probably would be too difficult to implement this now)

These errors do not seem specific to capture/authorize and may appear in other methods, like related to subscriptions (or if there is something specific to the action, then it may be better to pass some kind of context).

Also then you would not need to modify the response.

And we usually don't add the issue ID in each commit message, just in the branch, maybe also in the PR title.

And feel free to remove the irrelevant stuff from the PR template when submitting PR, or even just delete everything and write whatever you want. I think the main purpose of the PR description in our case should be describing shortly what was done and how/why, mentioning important details etc.

Dinamiko (approved):

modules/ppcp-api-client/src/Exception/PayPalApiException.php
```diff
@@ -141,4 +144,39 @@ public function get_details( string $error ): string {
...
+					if ( isset( $detail->issue ) && $detail->issue === $key ) {
+						return true;
+					}
+				}
+				return false;
+			}
+		);
+		if ( $improved_errors ) {
+			$improved_errors = array_values( $improved_errors );
+			$json->message   = $improved_keys_messages[ $improved_errors[0] ];
```
AlexP11223: Maybe it would be better to just return the message instead of changing the response?

```
if ( $improved_errors ) {
    ...
    return $new_message;
}
return $response->message;
```

```
$message           = $this->get_customer_friendly_message($response);
```

mmaymo:
@AlexP11223 I did the changes you suggested. I do not understand what do you mean by "this message for the user should be a separate field handled by the thing outputting errors" Do you mean in the exception another field, not in the message?

AlexP11223:
> Do you mean in the exception another field, not in the message?

Yeah, so that exception would have the original message (for logs) and the customer-friendly (for checkout). Because it can be difficult to debug what happened if we have e.g. `There was a problem processing this transaction` for multiple errors. Though can just log the error ID like `PAYEE_ACCOUNT_RESTRICTED`, if not already.

modules/ppcp-api-client/src/Exception/PayPalApiException.php
mmaymo: yeah, it makes more sense to change only the message, I was thinking we were sending back the object so I was trying to change it, but we send the response and the message. Now I see, thanks for the hint.

AlexP11223 (approved):

PR #1366, opened 2023-05-11
Add fraudnet script to SGO filter

mmaymo:
The filter we use to prevent SiteGround Optimizer from minifying our files currently does not include the FraudNet scripts

### Steps to Test
1. Add SGO plugin
2. Fraudnet script is minified


modules/ppcp-compat/services.php
```diff
@@ -49,6 +49,7 @@
 			'ppcp-gateway-settings',
 			'ppcp-webhooks-status-page',
 			'ppcp-tracking',
+			'ppcp-fraudnet',
```
AlexP11223, line 52: Seems like it's also missing `ppcp-gzd-compat`, `ppcp-clear-db`. (found via Ctrl-Shift-F regex mode `wp_(e|r).+_script\(`).

AlexP11223 (approved):

Dinamiko (approved):

PR #1375, opened 2023-05-15
[Fix] Undefined array key 0 when checking $retry_errors in process_payment method. (1698)

fitimvata:
### Description

Use array_values for convert to array with numeric keys before checking $retry_errors.

### Changelog Entry

> [Fix] Undefined array key 0 when checking $retry_errors in process_payment method.


Dinamiko (approved):

Dinamiko:
@fitimvata thanks and sorry for the delay in the response.

PR #1430, opened 2023-06-13
Add link to manual credential docs

mmaymo:
This PR adds a link to the manual credential docs in the settings page under the credential toggle


Dinamiko (commented):
Better extract html from the translation string like so: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/services.php#L1114-L1127

Dinamiko (approved):

PR #1433, opened 2023-06-14
Add notice when shop currency is unsupported

mmaymo:
This PR adds a notice when the currency declared in the shop is not supported by PayPal. 
The current behavior makes it a bit confusing for the merchants as in the checkout page, the PayPal button does not render and the “Place order” button is hidden.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Notice/UnsupportedCurrencyAdminNotice.php
+++ b/modules/ppcp-wc-gateway/src/Notice/UnsupportedCurrencyAdminNotice.php
@@ -64,11 +64,12 @@ class UnsupportedCurrencyAdminNotice {
 		}
 
 		$message = sprintf(
-			/* translators: %1$s the gateway name. */
+			/* translators: %1$s the shop currency, 2$s the gateway name. */
 			__(
-				'Attention: Your current WooCommerce store currency is not supported by PayPal. Please update your store currency to one that is supported by PayPal to ensure smooth transactions. Visit the <a href="%1$s">PayPal currency support page</a> for more information on supported currencies.',
+				'Attention: Your current WooCommerce store currency (%1$s) is not supported by PayPal. Please update your store currency to one that is supported by PayPal to ensure smooth transactions. Visit the <a href="%2$s">PayPal currency support page</a> for more information on supported currencies.',
 				'woocommerce-paypal-payments'
 			),
+			$this->shop_currency,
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Notice/UnsupportedCurrencyAdminNotice.php
```diff
@@ -0,0 +1,95 @@
...
+		return $this->state->current_state() === State::STATE_ONBOARDED && ! $this->currency_supported();
+	}
+
+	/**
+	 * Whether the currency is supported by PayPal.
+	 *
+	 * @return bool
+	 */
+	private function currency_supported(): bool {
+		$currency             = get_woocommerce_currency();
```
AlexP11223: If we need the store currency, then there is `'api.shop.currency'` service. `get_woocommerce_currency` can be `null` sometimes or changed by multi currency plugins.

modules/ppcp-wc-gateway/src/Notice/UnsupportedCurrencyAdminNotice.php
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	public function unsupported_currency_message() {
+		if ( ! $this->should_display() ) {
+			return null;
+		}
+
+		$message = sprintf(
+			/* translators: %1$s the gateway name. */
+			__(
+				'Attention: Your current WooCommerce store currency is not supported by PayPal. Please update your store currency to one that is supported by PayPal to ensure smooth transactions. Visit the <a href="%1$s">PayPal currency support page</a> for more information on supported currencies.',
```
AlexP11223: Would be good to include the currency.

modules/ppcp-wc-gateway/src/Notice/UnsupportedCurrencyAdminNotice.php
```diff
@@ -0,0 +1,96 @@
...
```
AlexP11223: Would be good to include the currency in this text (like "... Your current WooCommerce store currency (KPW) ..."), it can help with support/debugging.

mmaymo: right, sorry, I missed to see the first time you wrote it

AlexP11223 (approved):

PR #1441, opened 2023-06-16
Update Pay Later amount on the cart page and checkout when total changes

AlexP11223:
It seems the price update for the Pay Later message was handled only on product pages, so now added it on the cart page too. The price is retrieved from the server in the ajax call that we were already doing to check the intent change (when subscriptions added/removed), so it should not cause issues about wrong price retrieval like we had on product pages.


Dinamiko (approved):

AlexP11223:
Also now updating it in checkout in similar way.

PR #1468, opened 2023-06-26
Refresh DCC/PUI status caches on plugin update

### LINKED ISSUE #1432: Advanced Card Processing gateway becomes invisible post-plugin update unless admin pages are accessed once (1759)
### Describe the Bug

Post the update to version 2.0.4, a new issue has emerged resulting in the Advanced Card Processing gateway not loading after updating the plugin. The problem persists upon upgrading or downgrading to 2.0.4, and also affects subsequent versions (2.0.5, 2.1.0).

In order to mitigate this problem temporarily, one must visit any section within the PayPal Payments settings or the Payments tab under the WooCommerce settings. It's noteworthy that no changes need to be saved.

### To Reproduce

- Install version 2.0.3
- Enable ACDC gateway
- Visit the Checkout page
- Observe that the ACDC gateway is visible
- Update to any newer version (2.0.4, 2.0.5, 2.1.0)
- **Do not** navigate to the PayPal Payments settings page or the WooCommerce>Settings>Payments tab
- Reload the Checkout page
- Observe that the ACDC gateway is not visible
- Navigate to/reload the WooCommerce>Settings>Payments tab or any PayPal Payments settings tab (no need to save anything)
- Reload the Checkout page again
- Observe that the ACDC gateway is now visible

### Expected Behavior

Post plugin update, the Advanced Card Processing gateway should remain visible on the Checkout page without necessitating a visit to the WooCommerce>Settings>Payments tab or any PayPal Payments settings tab.

### Actual Behavior

Presently, the Advanced Card Processing gateway disappears from the Checkout page post-update unless certain admin pages are accessed at least once.

### Environment

- WordPress Version: Unspecified
- WooCommerce Version: Unspecified
- Plugin Version: From 2.0.4 onwards

AlexP11223:
Fixes #1432

Since #1273 we started clearing these caches on plugin updates. But we are updating these caches only on some admin pages
https://github.com/woocommerce/woocommerce-paypal-payments/blob/583dea8d99f60f710f54f6d689c5f57694562180/modules/ppcp-wc-gateway/src/WCGatewayModule.php#L421-L428
to avoid the API calls on every request in some cases, so DCC may become hidden if not visited these pages after update. Maybe we can do it more frequently now after 92ae16f99. But for now simply refreshing the caches during plugin update.

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -278,6 +278,15 @@ class WCGatewayModule implements ModuleInterface {
 				$settings->set( 'products_dcc_enabled', false );
 				$settings->set( 'products_pui_enabled', false );
 				$settings->persist();
+
+				// Update caches.
+				$dcc_status = $c->get( 'wcgateway.helper.dcc-product-status' );
+				assert( $dcc_status instanceof DCCProductStatus );
+				$dcc_status->dcc_is_active();
+
+				$pui_status = $c->get( 'wcgateway.pay-upon-invoice-product-status' );
+				assert( $pui_status instanceof PayUponInvoiceProductStatus );
+				$pui_status->pui_is_active();
 			}
 		);
 


Dinamiko (approved):

stevewhitmore:
Do we know when this will be released as a patch? We're experiencing the same issues with https://github.com/woocommerce/woocommerce-paypal-payments/issues/1432.

maxms:
Still happening after latest update to v2.2.

PR #1469, opened 2023-06-26
Smart buttons not greyed out/removed on single product when deselecting product variation (1607)

pedrosilva-pt:
When selecting a variation, the buttons appear but they remain when unselecting the variation:
Steps To Reproduce

1. enable buttons on single product
2. create variable product with no default variation
3. visit variable product → no smart buttons
4. select variation → smart buttons appear
5. deselect variation → smart buttons remain visible

### Expected behaviour

Smart buttons greyed out (like the add to cart button) after deselecting the variation. The initial idea was to remove the button, but it looks better and more intuitive if we let it behave the same as the other button that is present.


Dinamiko (approved):

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
```diff
@@ -10,56 +10,76 @@ class SingleProductBootstap {
...
+        return document.querySelector(this.formSelector);
+    }
 
     handleChange() {
-        const shouldRender = this.shouldRender();
-        setVisible(this.gateway.button.wrapper, shouldRender);
-        setVisible(this.gateway.messages.wrapper, shouldRender);
-        if (!shouldRender) {
+        if (!this.shouldRender()) {
+            hide(this.gateway.button.wrapper, this.formSelector);
```
AlexP11223, line 29: Seems like the second parameter was passed by mistake? :)  And below in `show`.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
```diff
@@ -93,7 +113,7 @@ class SingleProductBootstap {
         return !price || price === 0;
     }
 
-    isSubscriptionMode() {
+    isDisabledReasonExternalPlugins() {
```
AlexP11223: hm, not sure if I understand this name. Maybe a better name would be `isUnsupportedThirdPartyPluginState` if you wanted to cover all cases like this. But then you still should add another function or variable or comment inside about subscription mode to not lose this info. So maybe for now it should be simply `isWcsattSubscriptionMode()` (WCSATT is the old name of that plugin, `subscribe all the things`, apparently still used in the code as seen here). Because there are no other cases for now, and ideally this solution is not supposed to be permanent, we can somehow reload the buttons to support this mode.

modules/ppcp-button/resources/js/modules/Helper/ButtonDisabler.js
```diff
@@ -0,0 +1,53 @@
...
+
+export const setEnabled = (selectorOrElement, enable, form = null) => {
+    const element = getElement(selectorOrElement);
+
+    if (!element) {
+        return;
+    }
+
+    if (enable) {
+        jQuery(element).css({
```
AlexP11223, line 20: I think PayPal SDK supports disabling/enabling? https://developer.paypal.com/docs/checkout/standard/customize/validate-user-input/
As I understand you can keep the `actions` reference from `onInit` somewhere and call its methods when needed.

modules/ppcp-button/resources/js/modules/Renderer/MessageRenderer.js
```diff
@@ -30,17 +33,36 @@ class MessageRenderer {
...
-        paypal.Messages({
-            amount,
-            placement: this.config.placement,
-            style: this.config.style
-        }).render(this.config.wrapper);
+
+        paypal.Messages(options).render(this.config.wrapper);
+    }
+
+    isOptionsFingerprintEqual(options) {
```
AlexP11223: I would just call it something like `optionsEqual` without the fingerprint word, seems unneeded and can be confusing.

AlexP11223 (changes requested):

modules/ppcp-button/resources/js/modules/Renderer/MessageRenderer.js
pedrosilva-pt: I agree, will change.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
pedrosilva-pt: Ok, will change to isWcsattSubscriptionMode.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
pedrosilva-pt: The second parameter is to pass an optional reference to a form so we can trigger a submit in line with what the older plugin they referred used to do https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/trunk/assets/js/wc-gateway-ppec-generate-cart.js, I'll first see the built in PayPal Enable / Disable, then change this accordingly.

pedrosilva-pt: You're right, this is only applicable to disable.

modules/ppcp-button/resources/js/modules/Helper/ButtonDisabler.js
pedrosilva-pt: I will test the method you suggested..

pedrosilva-pt: By disabling with PayPal actions they aren't greyed out. I'll implement an hybrid solution.

pedrosilva-pt:
I added support for the native enable / disable feature of PayPal SmartButtons, but this doesn't make them grey so I had to keep the styling methods. Also there was a bit of gymnastics needed because SmartButtons only call the onSmartButtonsInit callback after SingleProductBootstap is initialized.

modules/ppcp-button/resources/js/button.js
```diff
@@ -121,10 +121,22 @@ const bootstrap = () => {
             return actions.reject();
         }
     };
-    const onSmartButtonsInit = () => {
+
+    let smartButtonsOptions = {
+        onInit: null,
```
AlexP11223, line 126: Maybe it would be better to make it an array. But probably not important for now.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
```diff
@@ -10,56 +10,93 @@ class SingleProductBootstap {
         this.messages = messages;
         this.errorHandler = errorHandler;
         this.mutationObserver = new MutationObserver(this.handleChange.bind(this));
+        this.formSelector = 'form.cart';
+
+        if (this.renderer && this.renderer.smartButtonsOptions) {
```
AlexP11223, line 15: This check seems unneeded? These properties are always set.

AlexP11223 (approved):

modules/ppcp-button/resources/js/button.js
pedrosilva-pt: I also thought on that, but didn't want to over engineer it prematurely. Probably this object should also have its own separate file/class.

modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
pedrosilva-pt: It's more to prevent errors on bad/incomplete constructor arguments. But in this case and since this class is so specific and isn't intended to be widely reused they could be skipped.

PR #1485, opened 2023-07-04
Filter to conditionally block the PayPal buttons (1788)

pedrosilva-pt:
## Description

It’s possible to grey out the PayPal buttons (as can be seen in #1469) 

It would be nice if we could expose some kind of function for merchants to conditionally block the buttons like this:

We often had support requests from merchants asking how they could disable the buttons in certain circumstances and such a feature could simplify this customization process.

In that case, merchants wouldn’t need to remove the buttons and instead could just block them.

To disable / enable the buttons the following filters can be used `woocommerce_paypal_payments_buttons_disabled` (for any context) and `woocommerce_paypal_payments_product_buttons_disabled` (specific for products). 
```
add_filter('woocommerce_paypal_payments_buttons_disabled', function (?bool $is_disabled, string $context): bool {
	return true;
}, 10, 2);

add_filter('woocommerce_paypal_payments_product_buttons_disabled', function (?bool $is_disabled, \WC_Product $product): bool {
	return true;
}, 10, 2);
```

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
+++ b/modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
@@ -1,6 +1,6 @@
 import CartActionHandler from '../ActionHandler/CartActionHandler';
+import BootstrapHelper from "../Helper/BootstrapHelper";
 import {setVisible} from "../Helper/Hiding";
-import {disable, enable} from "../Helper/ButtonDisabler";
 
 class CartBootstrap {
     constructor(gateway, renderer, errorHandler) {
@@ -48,15 +48,7 @@ class CartBootstrap {
     }
 
     handleButtonStatus() {
-        if (!this.shouldEnable()) {
-            this.renderer.disableSmartButtons(this.gateway.button.wrapper);
-            disable(this.gateway.button.wrapper);
-            disable(this.gateway.messages.wrapper);
-            return;
-        }
-        this.renderer.enableSmartButtons(this.gateway.button.wrapper);
-        enable(this.gateway.button.wrapper);
-        enable(this.gateway.messages.wrapper);
+        BootstrapHelper.handleButtonStatus(this);
     }
 
     shouldRender() {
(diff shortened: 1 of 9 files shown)


modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
```diff
@@ -40,10 +47,27 @@ class CartBootstrap {
         });
     }
 
+    handleButtonStatus() {
```
Dinamiko, line 50: Seems same function is used in CheckoutBootstap.js, maybe we can extract it to a helper file. Same for `shouldEnable` below.

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1350,6 +1352,51 @@ protected function product_supports_payment( WC_Product $product ): bool {
...
 
+	/**
+	 * Checks if PayPal buttons/messages should be rendered for the current page.
+	 *
+	 * @param string|null $context The context that should be checked, use default otherwise.
+	 *
+	 * @return bool
+	 */
+	protected function is_button_disabled( string $context = null ): bool {
+		if ( null === $context ) {
```
Dinamiko, line 1363: We removed yoda condition code style so no need to use it anymore, same for `if ( 'product' === $context ) {` below.

Dinamiko (approved):
LGTM, PR approved, just a couple of small adjustments in case you want to apply them.

AlexP11223 (approved):

modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
pedrosilva-pt: I did the refactor you suggested, also generalized a bit more and was able to cover all bootstraps with the helper functions. That way the code is a bit cleaner.

Also thought of having an Abstract class, but for now I think would add more confusion than benefit.

modules/ppcp-button/src/Assets/SmartButton.php
pedrosilva-pt: I don't mind, and actually I usually use it when comparing with null. Left the same.

PR #1491, opened 2023-07-05
Use payment_source + experience_context

AlexP11223:
Replacing the deprecated `application_context` fields with `payment_source` + `experience_context`. https://developer.paypal.com/docs/api/orders/v2/#orders_create

Affects mainly the orders created via `CreateOrderEndpoint`.
Oxxo remains unchanged: not sure if it needs such context for order creation (and in [`/confirm-payment-source`](https://developer.paypal.com/docs/api/orders/v2/#orders_confirm) it is not deprecated).
PUI does not seem to use `application_context` too.
We set `application_context.stored_payment_source` [here for subscriptions](https://github.com/woocommerce/woocommerce-paypal-payments/blob/17a346cfc91a4262d745a100b1ff0533c1e553c0/modules/ppcp-subscription/src/SubscriptionModule.php#L141), it is ok because this field was not deprecated.
And for now keeping the old format for ACDC because it's unclear how to create `payment_source` for it.

Also added `PAYER_ACTION_REQUIRED` in the list of order statuses because it seems to be created with this status now (but it does not affect anything). And now allowing both 200 and 201 HTTP status codes, it supposed to be 200 according to the docs but it was 201 when using old format.

For the existing plugin installations continuing to use the old format with `application_context` to avoid breaking filters like `ppcp_create_order_request_body_data`. This can be overridden via `ppcp_flag_old_application_context` filter or `PPCP_FLAG_OLD_APPLICATION_CONTEXT=0` (or `1`) env var.


modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -18,7 +18,7 @@
 use WooCommerce\PayPalCommerce\ApiClient\Entity\OrderStatus;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\PatchCollection;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Payer;
-use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentMethod;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentSource;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentToken;
```
Dinamiko, line 22: Seems `PaymentToken` not needed here.

Dinamiko (approved):

PR #1502, opened 2023-07-10
Onboarding from connection page with CSRF paramter manipulates email and merchant id fields (1753)

pedrosilva-pt:
## The problem

We fixed the CSRF issue that would allow malicious actors to invalidate existing API credentials.

But when visiting a malicious link while not onboarded and then onboarding from that link, it may result in the API credentials (Email & Merchant Id) containing the CSRF URL parameters or not being completed at all after onboarding.

Also, the fix results in the onboarding wizard being unable to correctly connect a second account.

This video illustrates how it is no longer possible to connect a second account while a first account is already connected:

## Steps To Reproduce

1. Disconnect the PayPal account from your site: Go to the Connection tab ( `{site url}/wp-admin/admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway&ppcp-tab=ppcp-connection` ) -> Remove PayPal Payments data from Database -> Click on “Clear now”
2. You’ll be redirected to the onboarding screen at `{site url}/wp-admin/admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway`
3. Add the following parameters to the URL and follow it `&merchantIdInPayPal=CSRF&merchantId=CSRF`
4. Under the “Toggle to manual credential input” section, notice that the merchant ID fields were filled
5. Click on the “Activate PayPal” button and complete the onboarding with your live account
6. Notice the `Live Email` and `Live Merchant Id fields` are displaying the strings from the CSRF URL parameters
7. Alternatively also perform these steps:
8. Check the **Sandbox** checkbox and click on the “Test payments with PayPal sandbox” button to complete the onboarding with your sandbox account
9. After onboarding, the merchant is sent to the `Standard Payments` tab instead of the `Connection` tab
10. Notice the navigation tab sections doesn’t show up

## Possible cause

It looks like 10. is the result of a "partially onboarded" state and only affects sandbox accounts.
For some reason, when onboarding a sandbox account, the `Sandbox Email address` and `Sandbox Merchant Id` fields are empty when the onboarding wizard was used while connection page URL was opened with the CSRF parameter. Fixing the credentials is only possible by manually adding this to the URL: `&ppcp-tab=ppcp-connection`
However, when connecting a live account instead of sandbox, the `Live Email address` & `Live Merchant Id` are replaced with the malicious CSRF string. In this case, the settings tabs are visible and since tabs are visible, it is easier to fix the credentials compared to the sandbox scenario.

## PR Description

Currently the `return_url` of `/v2/customer/partner-referrals` API call doesn't have validation, so a malicious actor can change it and if an admin clicks it `merchant_id` and `merchant_email` will be changed in the database.

To prevent this, this PR introduces a token that will be added to the 'return_url' and validated when opened.

This token implements several security measures like a `secret` (server side), uses `wp_hash` for the public token, and is attached to a WP user. Since the public token isn't stored on the database, it's resilient to DB hacking.

The data associated with the token is stored on the cache, segmented by user. The cache is also used for subsequent token data retrieval and validation. Since each token is only valid for one request the cache also allows to expire them on demand.

A test case for the `OnboardingToken` class was also included.


AlexP11223:
The reason why we did not use a simple WP nonce here was because the URL is cached (in WP transients), so a nonce

- can expire sooner
- will not work if it was cached by one admin but then another admin tries to use it

It seems like this solution still may have at least the second problem because it uses the user ID?

pedrosilva-pt:
> The reason why we did not use a simple WP nonce here was because the URL is cached (in WP transients), so a nonce
> 
>     * can expire sooner
> 
>     * will not work if it was cached by one admin but then another admin tries to use it
> 
> 
> It seems like this solution still may have at least the second problem because it uses the user ID?

The cache key for each token is namespaced also per user id. So each user will get it's own unique token I don't see a conflict there.

See here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/96eae6c690d67d006a97e0b2b44a65460080f214/modules/ppcp-onboarding/src/Helper/OnboardingUrl.php#L314C1-L316C3

The token could also be user agnostic, but I like the little added security and the race condition edge case of two users activating at the same time.

AlexP11223:
ah, yeah, I missed that it is stored per user.

Dinamiko (approved):
It looks great, good job!

modules/ppcp-api-client/src/Repository/PartnerReferralsData.php
```diff
@@ -127,4 +127,18 @@ public function data(): array {
...
+	 * Append the validation token to the return_url
+	 *
+	 * @param array  $data The referral data.
+	 * @param string $token The token to be appended.
+	 * @return array
+	 */
+	public function append_onboarding_token( array $data, string $token ): array {
+		$separator = strpos( $data['partner_config_override']['return_url'], '?' ) === false ? '?' : '&';
+
+		$data['partner_config_override']['return_url'] .= $separator . 'ppcpToken=' . $token;
```
AlexP11223: Maybe [add_query_arg](https://developer.wordpress.org/reference/functions/add_query_arg/)?

modules/ppcp-onboarding/src/Exception/RuntimeException.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Onboarding\Exception;
+
+/**
+ * Class RuntimeException
+ */
+class RuntimeException extends \RuntimeException {
```
AlexP11223: Is it needed? Maybe just `\RuntimeException` is enough? These exceptions with the same name in different modules can cause confusion and mistakes.

pedrosilva-pt: No strong feelings one way or the other, having it will allow for exception distinction on a higher level catch, but not having it will make it easier to generalize if necessary (and OnboardingUrl class it's totally generalizable for other use cases). I will change to \RuntimeException.

modules/ppcp-api-client/src/Repository/PartnerReferralsData.php
pedrosilva-pt: Those fancy WP functions!

PR #1511, opened 2023-07-14
Pay Later button and message get hidden when product/cart/checkout value is outside of range (1389, 1886, 1847)

pedrosilva-pt:
# PR Description

This PR implements improvements that should resolve these three issues:
* 1389 - Pay Later button and message get hidden when product/cart/checkout value is outside of range
* 1886 - PayPal buttons not reloading on WooCommerce Cart page asynchronous updates
* 1847 - Amount formatting edge cases for PayPal Messages widget in Single Product Page with Variants

It implements the following improvements:
* Ability to get the product price considering quantities, variants and custom fields on the product page.
* Ability to check status of the buttons / script on the product page.
* Ability to reload the paypal script with different script settings, for ex: like intent and funding options.

Challenges:
* When reloading the PayPal script all PayPal Widgets get destroyed, so references to their configuration need to be stored in order to rebuild them when necessary.

## Testing

The following filters were added `woocommerce_paypal_payments_product_buttons_paylater_disabled` and `woocommerce_paypal_payments_buttons_paylater_disabled`.

They can be used to hide the paylater button like this:
```
add_filter('woocommerce_paypal_payments_product_buttons_paylater_disabled', function (?bool $is_disabled, \WC_Product $product = null): bool {
	return (float) $product->get_price( 'raw' ) > 1000;
}, 10, 2);

add_filter('woocommerce_paypal_payments_buttons_paylater_disabled', function (?bool $is_disabled, string $context, ?float $cart_total = null): bool {
	return $cart_total > 1000;
}, 10, 3);

add_filter('woocommerce_paypal_payments_buttons_disabled', function (?bool $is_disabled, string $context, ?float $cart_total = null): bool {
	return $cart_total > 2000;
}, 10, 3);
```

### modules/ppcp-button/package.json
* Added the `@paypal/paypal-js` script.
* Reordered two imports to be equal to https://github.com/woocommerce/woocommerce-paypal-payments/commit/59afa81160c9ca74e8d36ffb4409cf058d0cb5e2#diff-6b19b93b03b45b4ae82e8b4b4f3dc0400fb16bd23da6f52ce8238d0577d9421eL15-L17

### modules/ppcp-button/resources/js/button.js
* The `messageRenderer.render();` method was removed in another PR. Should not be necessary to call it here.

### modules/ppcp-button/resources/js/modules/ActionHandler/SingleProductActionHandler.js
* Exposed the `getProducts` method to be call outside this class.

### modules/ppcp-button/resources/js/modules/ContextBootstrap/CartBootstap.js
* Added the ability to reload the script on `url_params` changes.
* Added enable / disable of buttons on buttons settings changes.

### modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
* Added the ability to get the price of the current product with quantities, variants and custom fields via an ajax call to the `simulate_cart` endpoint.
* The rerendering of the messages now get the simulated price instead of trying to get it via DOM.

### modules/ppcp-button/resources/js/modules/DataClientIdAttributeHandler.js
* Changed script loading to use `@paypal/paypal-js`

### modules/ppcp-button/resources/js/modules/Helper/ScriptLoading.js
* Changed script loading to use `@paypal/paypal-js`

### modules/ppcp-button/resources/js/modules/Helper/SimulateCart.js
* Object to handle an ajax request to `simulate_cart` endpoint.

### modules/ppcp-button/resources/js/modules/Helper/Utils.js
* General utility methods.
* _note: This class could be defined on a more generic place, it's on the buttons module for now._

### modules/ppcp-button/resources/js/modules/Renderer/MessageRenderer.js
* Changed to use the `WidgetBuilder` to register and render messages.

### modules/ppcp-button/resources/js/modules/Renderer/Renderer.js
* Changed to use the `WidgetBuilder` to register and render buttons.
* Added a `ppcp-reload-buttons` event listener to allow the reload buttons.
* Changed the `isAlreadyRendered` to be more generalized and in spirit with `WidgetBuilder`.
* note: The event handler can't be attached to a DOM element like `jQuery(wrapper).on('ppcp-reload-buttons')` but must be `jQuery(wrapper).on('ppcp-reload-buttons', wrapper, ...)` because sometimes WP / WC destroy and reconstruct DOM elements like on Cart page updates.

### modules/ppcp-button/resources/js/modules/Renderer/WidgetBuilder.js
* Allows the registration of PayPal widgets like Buttons and Messages.
* Renders Buttons and Messages.
* _note: Other PayPal Widgets like `HostedFields` could / should also be stored and built here, but for now it's not necessary._

### modules/ppcp-button/services.php
* Added the simulate-cart endpoint.

### modules/ppcp-button/src/Assets/SmartButton.php
* Added the `woocommerce_paypal_payments_product_buttons_paylater_disabled` and `woocommerce_paypal_payments_buttons_paylater_disabled` filters.
* Added the ajax `simulate_cart` endpoint to `script_data()`.
* Changed visibility of some filter checking methods so they can be used by the `simulate_cart` endpoint.

### modules/ppcp-button/src/ButtonModule.php
* Added the ajax `simulate_cart` endpoint.

### modules/ppcp-button/src/Endpoint/AbstractCartEndpoint.php
* Abstract class with methods shared by the cart endpoints: `ChangeCartEndpoint` and `SimulateCartEndpoint`.

### modules/ppcp-button/src/Endpoint/CartScriptParamsEndpoint.php
* Enforced the context as `is_cart()`, before was `mini-cart` and some options were incorrect to this request intended behaviour, like unwanted funding sources enabled or disabled.

### modules/ppcp-button/src/Endpoint/ChangeCartEndpoint.php
* Refactored to extend `AbstractCartEndpoint`.

### modules/ppcp-button/src/Endpoint/SimulateCartEndpoint.php
* Simulates the a WooCommerce cart via cloning to get the totals and buttons status of a given product with quantities, variants and other options included.

### modules/ppcp-button/yarn.lock
* The yarn.lock!

### modules/ppcp-onboarding/src/OnboardingModule.php
* Typo fix.

# Issues Description

## Pay Later button and message get hidden when product/cart/checkout value is outside of range (1389)
The pay later payment option is only eligible for payments between €1 to €1,000 (or £30 – £2,000, for every country different). However, the payment option and messaging stay visible if the cart/product is out of this range. 

The request here is that the Pay Later button and messaging gets hidden when the product/cart/checkout value is outside the range that these options are valid for.

Currently, the button and messaging will always show when activated:

### How could this be done?

We only need to have an override that allows you to globally turn on/off button & messaging. Then merchants can create filters with whatever condition they want (e.g. price range), to turn off the Pay Later button/messaging on demand.

Based on PayPal’s feedback, we can investigate this option, but we shouldn’t change anything in the default plugin behavior (e.g. hardcode any amounts)

* Introduce a filter that lets users programmatically set the conditions under which they want to display the Pay Later button.

### The problem

Changes in quantities or items on the WooCommerce cart page that affect the state of PayPal Buttons - such as enabling or disabling buttons, payment methods, or messages - are not currently being reflected on these buttons.

An equivalent behaviour can be seen with the filter woocommerce_paypal_payments_buttons_paylater_disabled once merged to trunk branch.

### Steps To Reproduce

Add the filter to a plugin or theme.

```
add_filter('woocommerce_paypal_payments_buttons_disabled', function (?bool $is_disabled, string $context): bool {
	return WC()->cart->cart_contents_total > 1000;

    }, 10, 2);
```

1. Add to the cart product that costs 100€.
2. Go to the Cart page.
3. See that the buttons are enabled.
4. Change the quantity to 20 and click Update Cart.
5. See that the buttons remain enabled.
6. Refresh the page.
7. See that the buttons are now disabled (greyed out)

### Possible cause

This issue arises due to asynchronous modifications on the backend that aren't propagated to the buttons and the buttons not being reloaded to reflect these changes.

### The problem

When on the Single Product Page with Variants there is an edge case where the amount passed to the PayPal Messages widget is being incorrectly formatted.

It only happens after selecting a variant in the dropdown and the variant amount is 1000 or more.

Basically it's having trouble formatting values like 8,000.00 as it converts it to 8 and display the PayPal Message related to 8.
Steps To Reproduce

1. Have a product with a variant with a price like 8000.99.
2. On the product page select the variant and see the PayPal Messages.
3. Have a product without variants and a price like 8000.99.
4. On the product page see the PayPal Messages and compare with the one of the product with variants.

### Expected behaviour

The PayPal Message should be consistent either coming from the Product Without Variant or the Product With Variants.

### Possible cause

[GitHub](https://github.com/woocommerce/woocommerce-paypal-payments/blob/583dea8d99f60f710f54f6d689c5f57694562180/modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js#L88)

### Suggested solution

Enhance number formatting capabilities on the JavaScript side. Or create an endpoint to get the variant amount from the server.


Dinamiko (dismissed):

modules/ppcp-button/resources/js/modules/Helper/Utils.js
```diff
@@ -0,0 +1,84 @@
...
+export const strRemoveWord = (str, word, separator = ',') => {
+    let arr = str.split(separator);
+    let index = arr.indexOf(word);
+    if (index !== -1) {
+        arr.splice(index, 1);
+    }
+    return arr.join(separator);
+};
+
+export const debounce = (func, wait) => {
```
AlexP11223: There is also this version here https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/resources/js/helper/debounce.js

dunno if we want to reuse it or copy it here (or vice versa). Using functions from other modules seems a bit weird and maybe should not be done for small utils.

modules/ppcp-button/services.php
```diff
@@ -124,6 +125,17 @@
 	'button.request-data'                         => static function ( ContainerInterface $container ): RequestData {
 		return new RequestData();
 	},
+	'button.endpoint.simulate-cart'               => static function ( ContainerInterface $container ): SimulateCartEndpoint {
+		if ( ! \WC()->cart ) {
+			throw new RuntimeException( 'cant initialize endpoint at this moment' );
```
AlexP11223, line 130: Should mention which endpoint.

Or maybe this check should be done when actually executing? If we want to keep the `$cart` parameter then it can be nullable and set to `WC()->cart` if null when executing.

modules/ppcp-button/resources/js/modules/Helper/Utils.js
pedrosilva-pt: I had the same dilemma, maybe have them in lib. In this case I can remove the `debounce` function as it's not being used, was to have some coherence since I was adding `throttle`.

I will remove debounce so there is less confusion.

modules/ppcp-button/services.php
pedrosilva-pt: Was just a duplication of `button.endpoint.change-cart`, wanted to avoid to change concepts too much to avoid unwanted side effects: https://github.com/woocommerce/woocommerce-paypal-payments/blob/84a362a7c5af5380fd8f6b8cfaae0c987386e10c/modules/ppcp-button/services.php#L139C1-L139C1

modules/ppcp-button/resources/js/modules/Renderer/MessageRenderer.js
```diff
@@ -28,7 +30,10 @@ class MessageRenderer {
         oldWrapper.parentElement.removeChild(oldWrapper);
         sibling.parentElement.insertBefore(newWrapper, sibling);
 
-        paypal.Messages(options).render(this.config.wrapper);
+        widgetBuilder.registerMessages(this.config.wrapper, options);
+        widgetBuilder.renderMessages(this.config.wrapper);
+
+        // paypal.Messages(options).render(this.config.wrapper);
```
Dinamiko: Better remove line if not needed.

modules/ppcp-button/resources/js/modules/Renderer/Renderer.js
```diff
@@ -68,43 +73,62 @@ class Renderer {
...
         // and on the cart page (where we also do not need to render separately), which may fully reload this part of the page.
         // Ideally we should also find a way to detect such full reloads and remove the corresponding keys from the set.
-        if (!hasEnabledSeparateGateways) {
-            return document.querySelector(wrapper).hasChildNodes();
-        }
-        return this.renderedSources.has(wrapper + fundingSource ?? '');
+
+        // if (!hasEnabledSeparateGateways) {
+        //     return document.querySelector(wrapper).hasChildNodes();
+        // }
```
Dinamiko: Remove if not needed.

Dinamiko (approved):
Looks good, just a couple of code comments to delete.

modules/ppcp-button/resources/js/modules/Renderer/MessageRenderer.js
pedrosilva-pt: Thanks removed.

modules/ppcp-button/resources/js/modules/Renderer/Renderer.js
pedrosilva-pt: Thanks removed.

PR #1529, opened 2023-07-21
High rate of auth voids on vaulted subscriptions for guest users (1899)

Dinamiko:
PayPal generated a PMT (Payment Token) with a customer ID of `someprefix-xyz456f`, and then sent a webhook (and was received), but the call to `v2/vault/payment-tokens` queried `someprefix-123` and then voided when it didn’t get any results.

When receiving `VAULT.PAYMENT-TOKEN.CREATED` webhook we generate the WC customer ID based on `$customer_id` value:
```
$wc_customer_id = (int)str_replace( $this->prefix, '', $customer_id );
```
It does work if `$customer_id` contains a valid user id after prefix (`someprefix-123`), but the customer id could also be generated as random alphanumeric string when the customer is not registered yet (ex: `someprefix-xyz456f`): 
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Repository/CustomerRepository.php#L47

This PR fixes that by first checking if a guest id (`ppcp_guest_customer_id`) value exists as user meta for a user and if so then use that user id. 

### Steps To Reproduce
- Create a brand new PayPal personal account
- As guest purchase a subscription using the new account
`VAULT.PAYMENT-TOKEN.CREATED` handler does not generate the correct WC customer ID from the received PayPal customer ID.

### DIFF
--- a/modules/ppcp-webhooks/src/Handler/VaultPaymentTokenCreated.php
+++ b/modules/ppcp-webhooks/src/Handler/VaultPaymentTokenCreated.php
@@ -107,12 +107,18 @@ class VaultPaymentTokenCreated implements RequestHandler {
 		$customer_id = null !== $request['resource'] && isset( $request['resource']['customer_id'] )
 			? $request['resource']['customer_id']
 			: '';
+
 		if ( ! $customer_id ) {
 			$message = 'No customer id was found.';
 			return $this->failure_response( $message );
 		}
 
-		$wc_customer_id = (int) str_replace( $this->prefix, '', $customer_id );
+		$wc_customer_id = $this->wc_customer_id_from( $customer_id );
+		if ( ! $wc_customer_id ) {
+			$message = "No WC customer id was found from PayPal customer id {$customer_id}";
+			return $this->failure_response( $message );
+		}
+
 		$this->authorized_payments_processor->capture_authorized_payments_for_customer( $wc_customer_id );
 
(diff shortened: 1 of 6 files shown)


modules/ppcp-webhooks/src/Handler/VaultPaymentTokenCreated.php
```diff
@@ -149,4 +155,29 @@ public function handle_request( WP_REST_Request $request ): WP_REST_Response {
...
 	}
+
+	/**
+	 * Returns WC customer id from PayPal customer id.
+	 *
+	 * @param string $customer_id The customer ID from PayPal.
+	 * @return int
+	 */
+	private function wc_customer_id_from( string $customer_id ): int {
+		$customers = get_users(
```
AlexP11223, line 166: There is a phpcs warning about possibly slow query. If there is no way to do it differently, then should `phpcs:ignore` it.

AlexP11223 (approved):

stffndtz:
@Dinamiko Does this create orders without a valid customer? I see this behaviour a lot since 2.2.0 and am looking for a possible fix.

Dinamiko:
@stffndtz it seems the issue is only happening to guest customers purchasing subscriptions, can you confirm if that is also your case? otherwise let me know the steps to reproduces, thanks.

stffndtz:
@Dinamiko Thanks - but I tracked down the issue, was unrelated to this PR.

PR #1553, opened 2023-07-28
Add ddev playwright addon

Dinamiko:
Adding https://github.com/julienloizelet/ddev-playwright addon.


pedrosilva-pt (approved):
Maybe we should update the README.md.
Also in doubt if we should add the playwright-report contents to the .gitignore as the seem to change with each test.

PR #1578, opened 2023-08-07
PayPal fee and PayPal Payout do not change on order if we do partial refund (1512)

pedrosilva-pt:
# PR Description

This PR:
* Adds reporting in the admin orders page about PayPal refunds and PayPal net payout after refunds.
* Adds order notes when PayPal refunds are processed.
* Provides a PHP API function (`\WooCommerce\PayPalCommerce\Api\ppcp_update_order_refund_fees`) to update refunds of a WC Order on demand.

This PR has a very large number of changes mainly due the `\WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint` not considering the PayPal refunds response data [https://developer.paypal.com/docs/api/orders/v2/#definition-refund](https://developer.paypal.com/docs/api/orders/v2/#definition-refund). So it was necessary to create several models and factories.

Basically the core functionality is to synchronize the refund totals for a given order when a new refund is processed for that respective order. And store the totals in the order meta data. In this process it also adds the refunds transaction ID to the order notes in case it's not present there already.

Then on the order edit page display the refund totals and PayPal payout net totals, and respective order notes.

Note: There was already a `Refund` entity that was used to capture the refund not to represent the refund response object. It was renamed to `RefundCapture` because using an alternative name to `Refund` for the response object would be much worse.

# Issue Description

## The problem
PayPal fee and PayPal Payout do not change on order if we do partial refund
Steps To Reproduce

## Create order
Go to WC->Orders
Click on created one
Scroll to button refund and do partially refund 
Observe new Net Payment but PayPal Fee and PayPal Payout do not change

### DIFF
--- a/modules/ppcp-api-client/src/Entity/Payments.php
+++ b/modules/ppcp-api-client/src/Entity/Payments.php
@@ -29,7 +29,7 @@ class Payments {
 	private $captures;
 
 	/**
-	 * The Captures.
+	 * The Refunds.
 	 *
 	 * @var Refund[]
 	 */
--- a/modules/ppcp-api-client/src/Entity/Payments.php
+++ b/modules/ppcp-api-client/src/Entity/Payments.php
@@ -42,7 +42,7 @@ class Payments {
 	 * @param array $captures The Captures.
 	 * @param array $refunds The Refunds.
 	 */
-	public function __construct( array $authorizations, array $captures, array $refunds ) {
+	public function __construct( array $authorizations, array $captures, array $refunds = array() ) {
 		foreach ( $authorizations as $key => $authorization ) {
 			if ( is_a( $authorization, Authorization::class ) ) {
 				continue;
--- a/modules/ppcp-api-client/src/Entity/Payments.php
+++ b/modules/ppcp-api-client/src/Entity/Payments.php
@@ -28,13 +28,21 @@ class Payments {
 	 */
 	private $captures;
 
+    /**
+     * The Captures.
+     *
+     * @var Refund[]
+     */
+    private $refunds;
+
 	/**
 	 * Payments constructor.
 	 *
 	 * @param array $authorizations The Authorizations.
 	 * @param array $captures The Captures.
+	 * @param array $refunds The Refunds.
 	 */
-	public function __construct( array $authorizations, array $captures ) {
+	public function __construct( array $authorizations, array $captures, array $refunds ) {
 		foreach ( $authorizations as $key => $authorization ) {
 			if ( is_a( $authorization, Authorization::class ) ) {
 				continue;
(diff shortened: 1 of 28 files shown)


modules/ppcp-api-client/src/Entity/Payments.php
```diff
@@ -47,8 +55,15 @@ public function __construct( array $authorizations, array $captures ) {
 			}
 			unset( $captures[ $key ] );
 		}
+		foreach ( $refunds as $key => $refund ) {
+			if ( is_a( $refund, Refund::class ) ) {
```
AlexP11223, line 59: Can it be something else? :thinking:

modules/ppcp-wc-gateway/src/Helper/RefundFeesUpdater.php
```diff
@@ -0,0 +1,193 @@
...
+	 * @param WC_Order $wc_order The WooCommerce order.
+	 * @return WP_Comment[]
+	 */
+	private function get_order_notes( WC_Order $wc_order ): array {
+		$notes = array();
+		$args  = array(
+			'post_id' => $wc_order->get_id(),
+		);
+
+		remove_filter( 'comments_clauses', array( 'WC_Comments', 'exclude_order_comments' ) );
```
AlexP11223, line 178: Would be great to add a comment why this is needed.

modules/ppcp-wc-gateway/src/Helper/RefundFeesUpdater.php
```diff
@@ -0,0 +1,193 @@
...
+	 *
+	 * @param WC_Order $wc_order The WooCommerce order.
+	 * @return void
+	 */
+	public function update( WC_Order $wc_order ): void {
+		$paypal_order_id = $wc_order->get_meta( PayPalGateway::ORDER_ID_META_KEY );
+
+		if ( ! $paypal_order_id ) {
+			$this->logger->error(
+				sprintf( 'Update order paypal refund fees. No PayPal order_id. [wc_order: %s]', $wc_order->get_id() )
```
AlexP11223: Probably should be something like `Failed to update ...`.

modules/ppcp-api-client/src/Entity/Payments.php
pedrosilva-pt: Probably not, but the main reason was to be consistent with the code already there.

image

modules/ppcp-wc-gateway/src/Helper/RefundFeesUpdater.php
pedrosilva-pt: The function like stated on the comments is based on https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-order.php#L1962-L1984, WooCommerce doesn't have a comment in their code. But I'll add one.

AlexP11223 (approved):

PR #1583, opened 2023-08-08
Add Subtotal mismatch behavior setting (1877, 255)

pedrosilva-pt:
# PR Description

This PR implements a `PurchaseUnitSanitizer` class that replaces the `ditch_items_when_mismatch` method.

The `PurchaseUnitSanitizer` has two different working modes, `ditch items` (in spirit with `ditch_items_when_mismatch` but trying to ditch as little data as possible) and `add extra line` which tries to handle mismatches by adding an extra line with roundings.

It also adds admin configuration options for:
* Subtotal mismatch behavior: `ditch` | `extra_line`
* Subtotal mismatch line name

A requirement was also to add an order note with telling there was a ditch or an extra line. So to pass data between a PayPal Order and a WooCommerce Order a new class `OrderTransient` was created.

# Issue Description
Internally, WC calculates line item prices and taxes out to four decimal places; however, PayPal can only handle amounts out to two decimal places (or, depending on the currency, no decimal places at all). Occasionally, this can cause discrepancies between the way WooCommerce calculates prices versus the way PayPal calculates them. If a mismatch occurs, this option controls how the order is dealt with so payment can still be taken.

I propose adding a setting called Subtotal Mismatch Behavior in the Connection tab under “General integration configuration” with these two options:

- Add another line item
- Do not send line items to PayPal

The default should be “Do not send line items to PayPal”.


lib/packages/Pattern/SingletonDecorator.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace WooCommerce\PayPalCommerce\Vendor\Pattern;
```
AlexP11223: Is this class written by you? Not sure if this dir is a good place for such things, it was just a quick fix for isolating some deps that can cause conflicts if there are other plugins with different versions of the same dep.

Maybe it should be some separate subdir (like `lib/isolated` and `lib/packages`) with some less weird namespace for non-generated packages then.

modules/ppcp-api-client/src/Helper/PurchaseUnitSanitizer.php
```diff
@@ -0,0 +1,349 @@
...
+		$this->extra_line_name = $extra_line_name;
+	}
+
+	/**
+	 * Indicates if mode is ditch.
+	 *
+	 * @return bool
+	 */
+	private function is_mode_ditch(): bool {
+		return $this->mode === self::MODE_DITCH;
```
AlexP11223: I would just check the `mode` without creating these functions for each possible value.

Also `is_..._mode` would sound better.

modules/ppcp-wc-gateway/src/Settings/Fields/connection-tab-fields.php
```diff
@@ -496,6 +497,42 @@
...
 			'gateway'      => Settings::CONNECTION_TAB_ID,
 		),
+		'subtotal_mismatch_behavior'                    => array(
+			'title'        => __( 'Subtotal mismatch behavior', 'woocommerce-paypal-payments' ),
+			'type'         => 'select',
+			'input_class'  => array( 'wc-enhanced-select' ),
+			'default'      => 'vertical',
+			'desc_tip'     => true,
+			'description'  => __(
+				'Differences between WooCommerce and PayPal roundings may give origin to a mismatch in order items subtotal calculations. If not handled these mismatches will cause the PayPal transaction to fail.',
```
AlexP11223: Maybe should add comma after `not handled`.

And maybe just `... may cause mismatch in ...`?

lib/packages/Pattern/SingletonDecorator.php
pedrosilva-pt: Yep I did it. It doesn't really fit any module it's too generic.

```
        "psr-4": {
            "WooCommerce\\PayPalCommerce\\": "src",
            "WooCommerce\\PayPalCommerce\\Vendor\\": "lib/packages/"
        },
```

Maybe have also a:
`"WooCommerce\\PayPalCommerce\\Common\\": "lib/common/"`

For these cases or `lib/isolated`

modules/ppcp-api-client/src/Helper/PurchaseUnitSanitizer.php
pedrosilva-pt: These functions are useful if there is any logic besides the constant itself, for instance if we had a filter, so no need to do refactors in the future. Also using mode before is visually more organized with autocompletes. But I can change this.

modules/ppcp-wc-gateway/src/Settings/Fields/connection-tab-fields.php
pedrosilva-pt: Ok will do.

lib/isolated/Pattern/SingletonDecorator.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace WooCommerce\PayPalCommerce\Isolated\Pattern;
```
AlexP11223: I meant the opposite, "isolating" refers to changing namespaces of some libs (dhii modularity here) to avoid conflicts with other plugins.

So I proposed to do it for example like this:

- `lib/packages/...` (where `...` is `Pattern/...`, etc.) with `WooCommerce\PayPalCommerce\Packages` namespace
- `lib/isolated/...` with `WooCommerce\PayPalCommerce\Vendor` namespace for the currently existing dhii stuff from the current `lib/packages`

Or to avoid messing with the second thing, maybe just put `Pattern` into `lib/Pattern`. I wonder if PSR-4 allows to define `WooCommerce\PayPalCommerce\Packages` for `lib/` but `WooCommerce\PayPalCommerce\Vendor` for `lib/packages`. If not, then we can change to `WooCommerce\PayPalCommerce\Packages\Vendor`, or just create some other dir or subdir for `Pattern`.

pedrosilva-pt: I tried solution 1 with `lib/packages/...` and `lib/isolated/...` but PSALM started giving errors even when switching back to `trunk`. Had to clear it's cache. 
`yarn ddev:psalm --clear-cache` `yarn ddev:psalm --clear-global-cache`.
Probably should avoid this to don't cause friction with all existing environments.

Solution 2 of having `lib/Pattern` should work, but I don't know if the confusion it will generate will compensate.

What about having a `lib/common` `WooCommerce\PayPalCommerce\Common` for `Pattern/...` and other common stuff in the future? Alternative names:
* a) `lib/common` `WooCommerce\PayPalCommerce\Common`
* b) `lib/base` `WooCommerce\PayPalCommerce\Base`
* c) `lib/shared` `WooCommerce\PayPalCommerce\Shared`
* d) `lib/library` `WooCommerce\PayPalCommerce\Library`
* ...

AlexP11223: Yeah, `lib/common` seems ok.

AlexP11223 (approved):
Maybe also update this readme to explain what's there, like adding a heading `packages`, and a heading `common` with short description (libs that do not fit into any module).
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/lib/README.md

PR #1625, opened 2023-08-25
Use package creation reusable workflow

AlexP11223:
Changed the package creation workflow to use the reusable workflow https://github.com/inpsyde/reusable-workflows/blob/main/docs/archive-creation.md

The result should be the same, only now the version is included in the zip name, and there is no prefix. The version input is still checked and the current version from the plugin file is added if numeric version is missing, to keep it more valid and to avoid error during plugin info parsing by `dhii/versions`. So you can enter `pcp-123-some-feature` and it will create `woocommerce-paypal-payments-2.2.1-pcp-123-some-feature.zip`.

Also now the excluded files are in a nicer `.distignore` file instead of the huge exclude string.


package.json
```diff
@@ -62,9 +62,7 @@
     "prebuild": "rm -rf ./vendor && find . -name 'node_modules' -type d -maxdepth 3 -exec rm -rf {} +",
     "build": "composer install --no-dev && yarn install && yarn run archive",
     "prearchive": "rm -rf $npm_package_name.zip",
-    "archive": "zip -r $npm_package_name.zip . -x **.git/\\* **node_modules/\\*",
-    "postarchive": "yarn run archive:cleanup && rm -rf $npm_package_name && unzip $npm_package_name.zip -d $npm_package_name && rm $npm_package_name.zip && zip -r $npm_package_name.zip $npm_package_name && rm -rf $npm_package_name",
-    "archive:cleanup": "zip -d $npm_package_name.zip .env* .ddev/\\* \\*.idea/\\* .editorconfig tests/\\* .github/\\* .psalm/\\* wordpress_org_assets/\\* \\*.DS_Store \\*README.md \\*.gitattributes \\*.gitignore \\*composer.json \\*composer.lock patchwork.json phpunit.xml.dist .phpunit.result.cache phpcs.xml* psalm*.xml* playwright.config.js wp-cli.yml \\*.babelrc \\*package.json \\*webpack.config.js \\*yarn.lock \\*.travis.yml\\"
+    "archive": "wp dist-archive . ./$npm_package_name.zip --plugin-dirname=$npm_package_name"
```
Biont, line 94: Is `composer-assets-compiler` automatically invoked or does it still(?) go into each module and runs `yarn build` there?

AlexP11223: assets compiler is running only in the workflow now. I guess we can replace the `install:modules:*` and `build:modules:*` scripts with it but I can't figure out how to setup it properly, e.g. if I run `ddev composer compile-assets -vvv` again it does not rebuild changes.

```
    Composer Assets Compiler
    starting...
    
    Not processing 'woocommerce/woocommerce-paypal-payments' because already processed.
    Nothing else to process.
    
    done.
```

Biont: @AlexP11223 You can either pass `--ignore-lock` and/or specify the source directories so it actually picks up file changes:

https://github.com/inpsyde/composer-asset-compiler/blob/v3.x/docs/007-Hash_and_Lock.md#custom-source-files

AlexP11223: @Biont  hm, weird, still the same with ` --ignore-lock`. 
And

```
            "src-paths": [
                "./modules/*/resources/js/*.js",
                "./modules/*/resources/css/*.scss"
            ]
```
had no effect too :thinking: 
Only deleting `.composer_compiled_assets` makes it to rebuild.

btw what was the purpose of `--env=root`? https://github.com/inpsyde/reusable-workflows/blob/1674a94f9eb5e5572278cef20e305c98440a1df6/.github/workflows/build-plugin-archive.yml#L57C22-L57C32

Biont: @AlexP11223 use `--ignore-lock=*`

AlexP11223:
Updated the PR, fixed assets building. https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/10593367194

Currently the assets are built via the existing `run-s ...` commands from `postinstall`. Assets Compiler fails to discover the modules in its normal way, maybe because we are using the merge plugin instead of adding them normally into `composer.json` like in other projects. So this could be something to refactor, but probably not in this PR.

In [PCP-3566](https://inpsyde.atlassian.net/browse/PCP-3566) it was suggested to keep the current workflow for now. But it needs to be done just before merging this PR, because otherwise it is not possible to run a new workflow not present in the default branch.

Also the reusable workflow branch (`@feature/PROD-171-rector-archive`) probably will need to be updated soon if it gets merged.

Dinamiko (approved):

PR #1646, opened 2023-08-30
Billing agreements endpoint called too frequently for Reference Transactions check (1870)

Dinamiko:
The call to the API that checks Reference Transactions availability should only happen once and then be cached for 3 months. But seemingly, this call happens more frequently, even without clearing the transient: `_transient_ppcp_reference_transaction_enabled`.

### Steps To Reproduce
- Set up PayPal Payments 2.2.0+
- Log into PayPal developer console
- Navigate WooCommerce for a while
- Observe API logs for the connected merchant account

### Possible cause
Checking transient `ppcp_reference_transaction_enabled` fails because expects a boolean `true` but receives a string `'1'`:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php#L124

### Suggested solution
As value is already saved as `'1'` for existing users, I suggest to just check for this value:
```
if ( get_transient( 'ppcp_reference_transaction_enabled' ) === '1' ) {
```

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
@@ -121,7 +121,7 @@ class BillingAgreementsEndpoint {
 	 */
 	public function reference_transaction_enabled(): bool {
 		try {
-			if ( get_transient( 'ppcp_reference_transaction_enabled' ) === true ) {
+			if ( get_transient( 'ppcp_reference_transaction_enabled' ) === '1' ) {
 				return true;
 			}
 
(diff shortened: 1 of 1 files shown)


modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
```diff
@@ -121,7 +121,7 @@ public function create_token( string $description, string $return_url, string $c
 	 */
 	public function reference_transaction_enabled(): bool {
 		try {
-			if ( get_transient( 'ppcp_reference_transaction_enabled' ) === true ) {
+			if ( get_transient( 'ppcp_reference_transaction_enabled' ) === '1' ) {
```
AlexP11223: hm, I think we save `true`, so it's not very obvious for it to become `'1'`.
Maybe it would be better to use `wc_string_to_bool`.

AlexP11223 (approved):

PR #1663, opened 2023-09-04
WC Payment Token created multiple times when webhook is received (1973)

Dinamiko:
WC Payment Token is created every time `VAULT.PAYMENT-TOKEN.CREATED` webhook is received while it should only be created once.

This PR fixes the issue by checking if WC payment token already exist before creating it.

### Steps to reproduce
- Enable vaulting and save a payment by placing and order.
- Wait one minute and check My Account / Payment methods, ensure payment is created correctly.
- Place another order using the saved payment.

### Expected behaviour
Only one payment should be saved.


modules/ppcp-vaulting/src/PaymentTokenHelper.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace WooCommerce\PayPalCommerce\Vaulting;
+
+/**
+ * Class PaymentTokenHelper
+ */
+class PaymentTokenHelper {
+
+	/**
+	 * Checks if given PayPal token exist as WC Payment Token.
```
AlexP11223: The description is a bit strange, seems like this method does not care about PayPal and simply checks if there is a token with the specified ID.

modules/ppcp-vaulting/src/PaymentTokenHelper.php
```diff
@@ -0,0 +1,33 @@
...
+	 *
+	 * @param array  $wc_tokens WC Payment Tokens.
```
AlexP11223: Maybe can specify the type? (I guess `WC_Payment_Token[]`)

AlexP11223 (approved):

PR #1699, opened 2023-09-15
Wrong PayPal subscription id on vaulted subscriptions (2016)

Dinamiko:
This PR fixes the issue when using same user session purchasing both PayPal Subscriptions API and Vaulted subscriptions products.

### Description
Subscriptions purchased with Vaulting mode shows PayPal connected metabox when using the same WC session.

### Steps To Reproduce
- Login as admin 
- Set “PayPal Subscriptions“ as Subscriptions Mode, uncheck Vaulting checkbox.
- As admin purchase a Subscriptions API connected product
- Set “PayPal Vaulting“ as Subscriptions Mode, check Vaulting checkbox.
- As admin purchase a non Subscriptions API connected product
- Check purchased subscription

### Expected behaviour
Purchased vaulted subscription should not display “PayPal Subscription“ metabox with the PayPal subscription ID. 

### Possible cause
We are not cleaning up session PayPal subscription id value when purchasing a PayPal Subscriptions connected product, so next purchase with the same user session will use the previous PayPal subscription id.

### Suggested solution
Clear PayPal subscription id from session when payment is complete.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
@@ -543,7 +543,6 @@ class PayPalGateway extends \WC_Payment_Gateway {
 				}
 
 				$wc_order->payment_complete();
-				WC()->session->set( 'ppcp_subscription_id', '' );
 
 				return $this->handle_payment_success( $wc_order );
 			}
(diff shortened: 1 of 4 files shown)


modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
```diff
@@ -543,6 +543,8 @@ function ( PaymentToken $token ): bool {
 				}
 
 				$wc_order->payment_complete();
+				WC()->session->set( 'ppcp_subscription_id', '' );
```
AlexP11223: Could it be missing somewhere else? Like in `destroy_session_data`.

Maybe this line should be moved there or at least to `handle_payment_success`, `handle_payment_failure`.

AlexP11223 (approved):

PR #1711, opened 2023-09-20
Remove payment vaulted checker functionality (2030)

Dinamiko:
When first version of vaulting was integrated in the plugin we had issues where payments were not saved in PayPal and therefore subscription renewals failed. As a workaround we added a payment saved checker that runs after some minutes calling PayPal to get the saved payments of the user and handle the order accordingly based on this check.

Nowadays `VAULT.PAYMENT-TOKEN.CREATED` webhook is very stable, we do not need to rely on custom code but in the webhook as single source of truth, this will avoid issues and reduce complexity in the code.

This PR removes this payment vaulted checker and includes a filter to enable it.

### Acceptance Criteria

To enable payment vaulted checker use this filter:
```
add_filter('woocommerce.deprecated-flags.woocommerce_paypal_payments.saved_payment_checker_enabled', '__return_true');
```

Once enabled it should work as before by adding the selector to decide what to do when saving payment fails:

Captura de pantalla 2023-09-20 a las 12 51 55

When filter to enable is not used (this is the default now) is should neither display the setting nor trigger the payment saved checker functionality.


modules.php
```diff
@@ -44,5 +44,13 @@
 		$modules[] = ( require "$modules_dir/ppcp-googlepay/module.php" )();
 	}
 
+	if ( apply_filters(
+		//phpcs:disable WordPress.NamingConventions.ValidHookName.UseUnderscores
+		'woocommerce.deprecated-flags.woocommerce_paypal_payments.saved_payment_checker_enabled',
```
AlexP11223: Is there an agreement about this naming? Why not just use `woocommerce.deprecated_flags` with `_`?

modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -194,7 +194,7 @@ public function create(
 	): Order {
 		$bearer = $this->bearer->bearer();
 		$data   = array(
-			'intent'              => ( $this->subscription_helper->cart_contains_subscription() || $this->subscription_helper->current_product_is_subscription() ) ? 'AUTHORIZE' : $this->intent,
+			'intent'              => apply_filters( 'woocommerce_paypal_payments_saved_payment_subscription_intent', $this->intent ),
```
AlexP11223: The filter name may be confusing. This function is not just about subscriptions, but someone may think that this filter can be used to modify intent of subscriptions without touching other orders.

Something like `woocommerce_paypal_payments_order_intent` would be better.

modules/ppcp-saved-payment-checker/composer.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "woocommerce/ppcp-saved-payment-checked",
```
AlexP11223: mistype? `checked` instead of `checker`.

modules/ppcp-saved-payment-checker/src/SavedPaymentCheckerModule.php
```diff
@@ -0,0 +1,142 @@
...
+			function( string $intent ) use ( $c ) {
+				$subscription_helper = $c->get( 'subscription.helper' );
+				assert( $subscription_helper instanceof SubscriptionHelper );
+
+				if ( $subscription_helper->cart_contains_subscription() || $subscription_helper->current_product_is_subscription() ) {
+					return 'AUTHORIZE';
+				}
+
+				return $intent;
+			}
```
AlexP11223, line 54: I think we always need this filter? This module runs only when the flag is enabled.

AlexP11223 (approved):

PR #1717, opened 2023-09-21
Feat/pcp 154 apple pay payment

mmaymo:
Added more details to the Apple Pay behind flag module. Solved some layout problems, validation file is taken from within, checkout form is included in the payment sheet,..


modules/ppcp-applepay/resources/js/ApplepayButton.js
```diff
@@ -160,26 +163,68 @@ console.log('[ApplePayButton] wrapper', wrapper)
...
      */
-    onButtonClick() {
+    async onButtonClick() {
         const paymentDataRequest = this.paymentDataRequest();
-        console.log('[ApplePayButton] onButtonClick: paymentDataRequest', paymentDataRequest, this.context);
-
+        // trigger woocommerce validation if we are in the checkout page
+        if (this.context === 'checkout') {
+            const checkoutFormSelector = 'form.woocommerce-checkout';
+            const errorHandler = new ErrorHandler(
```
AlexP11223, line 171: I think creating multiple ErrorHandler objects (here and old `button.js`) can sometimes cause some issues with displayed messages.

But not sure about the best way to fix it.

modules/ppcp-applepay/src/Helper/ApmApplies.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/**
+ * Properties of the ApplePay module.
```
AlexP11223: Wrong phpdoc.

AlexP11223 (approved):

PR #1719, opened 2023-09-21
Google Pay Settings improvements (2006)

pedrosilva-pt:
# PR Description
This PR is in essence the continuation of (1744 - Google Pay integration)

* It implements an `api.helper.failure-registry` to allow the monitoring of failed PayPal API requests.
* It checks the API failure registry on calls to `$this->partners_endpoint->seller_status()` is there is a fail withing a given timeframe then don't repeat the request. This fixes the multiple failed API requests experienced when the credentials are generated manually.
* Implements and uses the hook `ppcp_partner_referrals_option` to parse onboarding options. That is, each module is responsible of validating their own onboarding fields.
* Adds greying out of GooglePay configuration fields when is GooglePay elegible but not available.
* Implements GooglePay shipping callback.
* Adds button previews for GooglePay in the backend.
* Adds a GooglePay javascript boot-admin for the backend (used for button previews).
* Now by default `woocommerce_paypal_payments_googlepay_validate_product_status` is `true` (was `false` previously) this means that by default GooglePay is now validating product status.
* Adds the possibility of having a backend notice to be shown when for ex. PayPal product status doesn't have the GooglePay capability.
* Adds a generic and reusable method to add onboarding form options.
* Refactors PUI to use the generic onboarding options mechanism.
* Adds `SingletonDecorator` `wcgateway.settings` to prevent race conditions between `Settings` instances.

# Issue Description
* Add Google Pay to existing button preview (responsive to styling config)
* When manually connected merchant, add notices about eligibility check failing and grey out Google Pay settings fields


modules/ppcp-googlepay/src/Endpoint/UpdatePaymentDataEndpoint.php
```diff
@@ -0,0 +1,227 @@
...
+
+			// Validate nonce.
+			if (
+				! isset( $data['nonce'] )
+				|| ! wp_verify_nonce( $data['nonce'], self::nonce() )
+			) {
+				throw new RuntimeException(
+					__( 'Could not validate nonce.', 'woocommerce-paypal-payments' )
+				);
+			}
```
Dinamiko: Why validate nonce here again? it's already validated here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Endpoint/RequestData.php#L45-L53

Dinamiko (approved):

modules/ppcp-googlepay/src/Endpoint/UpdatePaymentDataEndpoint.php
pedrosilva-pt: You're right. Added `$this->request_data->read_request( $this->nonce() );` after the nonce validation and didn't notice it was already validating it. Removed the duplicate.

PR #1742, opened 2023-09-30
A solution to the "Call to undefined method WP_Post::get_meta()" issue that occurs when updating a subscription via WP-Admin.

0xBeycan:
**Issue**: #

"Call to undefined method WP_Post::get_meta()" error when updating a subscription via WP-Admin.

---

### Description

The $id parameter sent in the hook matches the current subscription id. However, the sent $subscription object, which is the secondary parameter, is derived from the WP_Post class and it gives an error because there is no get_meta() method in this class. I replaced the WP_Post class with the WC_Subscription class.

### Steps to Test

Since there is a get_meta() method in the WC_Subscription class, it no longer gives this error and moves forward.

### Documentation

No need

### Changelog Entry

The critical error "Call to undefined method WP_Post::get_meta()" that occurred during manual subscription update in WP-Admin has been fixed.

Closes # .

### DIFF
--- a/modules/ppcp-subscription/src/SubscriptionModule.php
+++ b/modules/ppcp-subscription/src/SubscriptionModule.php
@@ -593,7 +593,10 @@ class SubscriptionModule implements ModuleInterface {
 			 * @psalm-suppress MissingClosureParamType
 			 */
 			function( $id, $subscription ) use ( $c ) {
+				// Get the $subscription variable WC_Subscription again.
+				$subscription = wcs_get_subscription( $id );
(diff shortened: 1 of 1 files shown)


Sirvijver:
Hello @BeycanDeveloper 

this issue should be resolved with [this PR](https://github.com/woocommerce/woocommerce-paypal-payments/pull/1748). Please [engage us privately](https://paypal.inpsyde.com/docs/request-support/) in case you want to get access to a test package.

Kind regards, 
Joost

PR #1759, opened 2023-10-11
[Fix] Avoid calling  get_available_payment_gateways in every request.

fitimvata:
Right now every request is loading available_gateways which is very slow call around 86ms (screenshot bellow).

Screenshot 2023-10-11 at 1 02 33 PM

### DIFF
--- a/modules/ppcp-button/src/Assets/SmartButton.php
+++ b/modules/ppcp-button/src/Assets/SmartButton.php
@@ -488,38 +488,47 @@ class SmartButton implements SmartButtonInterface {
 			);
 		}
 
-		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
 
-		if ( isset( $available_gateways['ppcp-gateway'] ) ) {
-			add_action(
-				$this->pay_order_renderer_hook(),
-				function (): void {
+		add_action(
+			$this->pay_order_renderer_hook(),
+			function (): void {
+				$available_gateways = WC()->payment_gateways()->get_available_payment_gateways();
+
+				if ( isset( $available_gateways['ppcp-gateway'] ) ) {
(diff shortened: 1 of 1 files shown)


InpsydeNiklas:
Indeed loading `WC()->payment_gateways` is slow, more specifically, the `WC_Payment_Gateways::init()` method. In our tests it took `0.02 seconds` to execute.
But this only happens once per page load when instancing `WC()->payment_gateways`, so multiple calls to `WC()->payment_gateways->get_available_payment_gateways();` will in practice take the same time as doing only one call.

This pull request tries to delay the call of `WC()->payment_gateways->get_available_payment_gateways();` to only when it's needed, but only does this in one location of the code. There are multiple places on the plugin where this method is invoked (ex: [https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Gateway/GatewayRepository.php#L38](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Gateway/GatewayRepository.php#L38 "https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Gateway/GatewayRepository.php#L38")), and some of them on every page load.

**Our assessment is:**

- The performance issue does exist as `0.02 seconds` is considered slow
    
- The slow method is a WooCommerce implementation `WC_Payment_Gateways::init()`. Ideally the optimization should come directly from WooCommerce 
    
- The PR does not address all places where this method is invoked, so it does not fix the issue
    
We will close this PR for now but we will investigate additional performance optimizations after the modularity system [mentioned here](https://github.com/woocommerce/woocommerce-paypal-payments/issues/1760#issuecomment-1843435729) was introduced.

PR #1767, opened 2023-10-16
Apple Pay transaction improvements

mmaymo:
New improvements for the Apple Pay feature. 
We are still behind flag, waiting for the other PR to be merged in trunk


modules.php
pedrosilva-pt, line 1: This will enable apple pay module by default. Just to be sure it's intended :).

modules/ppcp-applepay/src/Assets/ApplePayDataObjectHttp.php
```diff
@@ -217,17 +217,11 @@ public function update_method_data(): void {
...
-			$nonce,
-			'woocommerce-process_checkout'
-		);
-		if ( ! $is_nonce_valid ) {
+
+		if ( ! $this->is_nonce_valid() ) {
 			return;
 		}
+		//phpcs:disable WordPress.Security.NonceVerification
 		$data = filter_var_array( $_POST, FILTER_SANITIZE_SPECIAL_CHARS );
```
pedrosilva-pt, line 225: It's ok, no need to change, but in this plugin the mostly used method is to send the request body as JSON [https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/resources/js/Helper/UpdatePaymentData.js#L18](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/resources/js/Helper/UpdatePaymentData.js#L18) then use `->read_request()` which already handles the nonce [https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/src/Endpoint/UpdatePaymentDataEndpoint.php#L70](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/src/Endpoint/UpdatePaymentDataEndpoint.php#L70)

pedrosilva-pt (approved):

PR #1771, opened 2023-10-17
Fix phpdoc

AlexP11223:
Updates `@since` from 2563d41d83


modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1176,7 +1176,7 @@ private function components(): array {
 		 * Filter to add further components from the extensions.
 		 *
 		 * @internal Matches filter name in APM extension.
-		 * @since TODO
+		 * @since 2.3.0
```
Dinamiko: I would rather prefer to remove it.

Dinamiko (changes requested):

Dinamiko (approved):

PR #1779, opened 2023-10-18
Save payment methods (Vault v3) integration (1393)

Dinamiko:
This PR adds Vault v3 (Save payment methods) to [Standard and Advanced Checkouts](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/). 

It also refactors subscriptions logic by moving each one (WC Subscriptions plugin, PayPal Subscriptions API and Vaulting) into its own modules.

To enable it please use the following filters:
```
add_filter('woocommerce.feature-flags.woocommerce_paypal_payments.save_payment_methods_enabled', '__return_true');
add_filter('woocommerce.feature-flags.woocommerce_paypal_payments.card_fields_enabled', '__return_true');
add_filter('woocommerce_paypal_payments_enable_cardholder_name_field', '__return_true');
```

### Done
- [x] [Save PayPal with the JavaScript SDK](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/during-purchase/js-sdk/paypal/)
- [x] [Save PayPal for purchase later with the JavaScript SDK](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/purchase-later/js-sdk/paypal/)
- [x] Purchase and renewal WC Subscriptions product with PayPal payment
- [x] [Save cards with the JavaScript SDK](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/during-purchase/js-sdk/cards/)

### Pending
- [ ] [Save cards for purchase later with the JavaScript SDK](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/purchase-later/js-sdk/cards/)
- [ ] Purchase and renewal WC Subscriptions product with Advanced Card payment


modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
```diff
@@ -830,122 +356,352 @@ function( $query, $query_vars ): array {
...
+
+				$screen_id = wc_get_page_screen_id( 'shop_subscription' );
+				remove_meta_box( 'woocommerce-subscription-schedule', $screen_id, 'side' );
+
+				$environment = $c->get( 'onboarding.environment' );
+				add_meta_box(
+					'ppcp_paypal_subscription',
+					__( 'PayPal Subscription', 'woocommerce-paypal-payments' ),
+					function() use ( $subscription_id, $environment ) {
+						$host = $environment->current_environment_is( Environment::SANDBOX ) ? 'https://www.sandbox.paypal.com' : 'https://www.paypal.com';
```
AlexP11223: can use `api.paypal-website-url` instead.

modules/ppcp-api-client/src/Entity/Order.php
```diff
@@ -237,7 +229,7 @@ static function ( PurchaseUnit $unit ): array {
 			$order['application_context'] = $this->application_context()->to_array();
 		}
 		if ( $this->payment_source() ) {
-			$order['payment_source'] = $this->payment_source()->to_array();
+			$order['payment_source'] = $this->payment_source();
```
AlexP11223: :thinking: I think the result of `to_array` is not supposed to contain objects like this?

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
```diff
@@ -0,0 +1,75 @@
...
+
+const init = () => {
+    setVisible(ORDER_BUTTON_SELECTOR, getCurrentPaymentMethod() !== PaymentMethods.PAYPAL);
+    setVisible(`#ppc-button-${PaymentMethods.PAYPAL}-save-payment-method`, getCurrentPaymentMethod() === PaymentMethods.PAYPAL);
+}
+
+document.addEventListener(
+    'DOMContentLoaded',
+    () => {
+        jQuery(document.body).on('click init_add_payment_method', '.payment_methods input.input-radio', function () {
```
AlexP11223, line 69: Maybe it would be simpler to use `updated_checkout payment_method_selected` like in `CheckoutBootstrap`?

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
```diff
@@ -0,0 +1,75 @@
...
+            console.error(error)
+        }
+    },
+    `#ppc-button-${PaymentMethods.PAYPAL}-save-payment-method`
+);
+
+
+
+const init = () => {
+    setVisible(ORDER_BUTTON_SELECTOR, getCurrentPaymentMethod() !== PaymentMethods.PAYPAL);
```
AlexP11223: Probably should be `setVisibleByClass` like in `CheckoutBootstrap` to avoid issues with some plugins/themes.

modules/ppcp-api-client/src/Entity/Order.php
pedrosilva-pt: I agree and don't understand the reason for this change. I'll leave for @Dinamiko to review.

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
pedrosilva-pt: I don't foresee the side effects of this change. I'll leave to @Dinamiko.

modules/ppcp-api-client/src/Entity/Order.php
pedrosilva-pt: `PaymentSource` was refactored it seems, the new implementations does not have `to_array()` but I think it should.

AlexP11223: It could be that `payment_source` is not needed here. So it either needs to be converted into an array or removed from this output.

It seems like `Order::to_array` is only used in two places where we output it in our ajax response or JS data. So we don't send it to the API and don't notice that this value is weird. Should be ok to remove this line then.

modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
```diff
@@ -144,10 +145,13 @@ public function handle_request(): bool {
 
 			$order = $this->api_endpoint->order( $data['order_id'] );
 
-			if ( $order->payment_source() && $order->payment_source()->card() ) {
+			$payment_source = $order->payment_source();
+			assert( $payment_source instanceof PaymentSource );
```
pedrosilva-pt: `payment_source()` is initialized conditionally on `OrderFactory`. Could there be any scenario where the order data doesn't have a `payment_source` it and the assert fail?

pedrosilva-pt (approved):

modules/ppcp-api-client/src/Entity/Order.php
Dinamiko: @pedrosilva-pt @AlexP11223 the reason I refactored `PaymentSource` is that `payment_source` on PayPal could have more types than [the ones that were added originally](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Entity/PaymentSource.php#L38-L39).

Even the same type could have different properties on it depending on the context, so that the reason I made it more flexible by allowing multiple types and an arbitrary number of properties.

AlexP11223: Yeah, I know, in this kind of abandoned branch I did it like this https://github.com/woocommerce/woocommerce-paypal-payments/tree/PCP-1801-experience-context/modules/ppcp-api-client/src/Entity/PaymentSource

But it needs to be either converted into an array or removed from `to_array` since a class instance automatically converted into JSON would not be useful anyway (I think it will be just `{}` by default).

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
Dinamiko: @AlexP11223 I tried changing from `click init_add_payment_method` to `updated_checkout payment_method_selected` but it does not work. 

I added the same as WC is using for switching saved payment methods: https://github.com/woocommerce/woocommerce/blob/2f7bd83ae229a999456e3e713af4d4a98e2bd28d/plugins/woocommerce/client/legacy/js/frontend/add-payment-method.js#L12

modules/ppcp-api-client/src/Entity/Order.php
```diff
@@ -189,16 +189,6 @@ public function application_context() {
 		return $this->application_context;
 	}
 
-	/**
-	 * Returns the payment source.
-	 *
-	 * @return PaymentSource|null
-	 */
-	public function payment_source() {
```
AlexP11223: This probably should remain? e.g. it's used in the code commented by Pedro above.

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
AlexP11223: it should work like this, without the 2nd parameter: `jQuery(document.body).on('updated_checkout payment_method_selected', () => { .......`

But I guess it should be fine too if WC is using it.

modules/ppcp-api-client/src/Entity/Order.php
Dinamiko: yeah, I just noticed it now, it is used in `SavePaymentMethodsModule.php`. I'm going to just remove the `to_array` part.

modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
```diff
@@ -145,10 +145,8 @@ public function handle_request(): bool {
 
 			$order = $this->api_endpoint->order( $data['order_id'] );
 
-			$payment_source = $order->payment_source();
-			assert( $payment_source instanceof PaymentSource );
-
-			if ( $payment_source->name() === 'card' ) {
+			$payment_source = $order->payment_source() ?? null;
```
AlexP11223: is `?? null` needed? I think it already returns `PaymentSource|null`. And afaik `??` checks only for null, so it's like "assign null if it is null".

AlexP11223 (approved):

modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
Dinamiko: yeah, no need for checking `null` here, thanks!

PR #1822, opened 2023-11-03
Error "PayPal order ID not found in meta" prevents automations from triggering when buying subscription via third-party payment gateway (2184)

Dinamiko:
Some users are seeing this error on non-PayPal orders: **Error during status transition. PayPal order ID not found in meta.**
The merchant informed that when they have PP and ACDC active and selling a subscription via Visa, MC or Amex via Woo Payments an Exception is thrown stating:
```
Error during status transition. PayPal order ID not found in meta.
```
This PR checks if the payment method  is allowed before saving the payment token.


pedrosilva-pt (approved):

nirkons:
I am still experiencing this issue even with the latest plugin version.
I'm using paypal payment gateway and stripe payment gateway with woocommerce and woocommerce subscriptions. unfotuantely this bug is still going on.

PR #1840, opened 2023-11-15
Allow redirect to PayPal with "Place order" button if smart buttons failed to load (2049)

AlexP11223:
Now  the standard "Place order" button will redirect to `paypal.com/checkoutnow?token=<order_id>` if it is not replaced by the smart buttons.

There is `woocommerce_paypal_payments_use_place_order_button` filter for enabling this mode, and `woocommerce_paypal_payments_place_order_button_text` for replacing the button text.

The replacement is done via JS, I think there is no other way to replace it just for one method. If JS is disabled, the button will work but the default text will remain.

It also works in the block checkout, adding the method to the list instead of the express buttons. The button text is not replaced because afaik there is still no reliable way to do that in blocks.

This mode also can be used as a fallback when smart button loading fails, but currently it only works if the failure happens early, before we hide the button. Later we can investigate how to detect other failures.

Also added `ppcp_create_paypal_order_for_wc_order` function in our API. The redirection URL can be generated via `PPCP::container()->get( 'api.factory.paypal-checkout-url' )( paypal_order_id )`


AlexP11223:
> The replacement is done via JS, I think there is no other way to replace it just for one method.

oh, apparently there is `order_button_text` property in `WC_Gateway`.

Dinamiko (approved):

AlexP11223:
Also there was a problem that it was not possible to repeat after cancelling because it was trying to capture a not approved order.

One related thing is #678, it seems like considering an order ready for processing just by `CREATED` status is not completely correct. Maybe it should somehow check that it is vaulting, not just `CREATED`.

But for this PR it should be enough to just check the status in the return endpoint, and avoid adding meta before processing.

AlexP11223:
Oh, looks in blocks it is also possible to replace the text via `placeOrderButtonLabel`.

PR #1843, opened 2023-11-16
Upgrade to new Hosted Card Fields for Advanced Card Processing (2219)

Dinamiko:
This PR adds Card Fields to [Checkout Advanced](https://developer.paypal.com/docs/checkout/advanced/integrate/).

### Acceptance Criteria
https://developer.paypal.com/tools/sandbox/card-testing/#link-testintegration

Enable the module by using this filter, please note at the moment the module only works when **WC merchant country is US**:
```
add_filter('woocommerce.feature-flags.woocommerce_paypal_payments.card_fields_enabled', '__return_true');
```

Enable Cardholder name field by using this filter:
```
add_filter('woocommerce_paypal_payments_enable_cardholder_name_field', '__return_true');
```


modules/ppcp-button/resources/js/modules/Renderer/CreditCardRenderer.js
```diff
@@ -25,141 +24,270 @@ class CreditCardRenderer {
...
+            )
 
-                const className = this._cardNumberFiledCLassNameByCardType(event.cards[0].type);
-                this._recreateElementClassAttribute(cardNumber, cardNumberField.className);
-                if (event.cards.length === 1) {
-                    cardNumber.classList.add(className);
+            return;
+        }
+
+        if (typeof paypal.CardFields !== 'undefined') {
```
AlexP11223: Should it be in the same file with the old ACDC? Seems like nothing is reused.

An easy way to separate could be just to create a new class with the same interface (`render` function) and decide which one to instantiate in `button.js`

AlexP11223 (approved):

Mohamed-1005:
Hello, this new hosted fields cannot be edited using css on the website like the old fields also there is a bug that Google pay not showing also when click on the PayPal as a payment method then return back to the credit cards option and again click on PayPal after that the PayPal button completely disappear.

Hope you fix all the bugs i mentioned, thanks.

titibleh:
Hello, it could be great if the new Hosted Cards Fields could change the language of placeholder fields to suite international e-commerce. Maybe you can catch the `html lang="xx-XX"` or the navigator language to show the right placeholder language. For starting you can use international meaning like "1234 1234 1234 1234" instead of "Card number"... We are not all based in USA. Thanks!

Mohamed-1005:
Any updates about how we style the new card fields? the CSS is not taking effect at all like before

PR #1888, opened 2023-12-05
Order does not contain intent on 3DS ACDC renewal (2007)

Dinamiko:
This PR fixes #1506 which did not entirely fixed the problem as incorrect payment source is still sent when subscriptions are renewed via action scheduler.

It should now send the correct payment source for both manual and automatic subscription renewals like so:
```
"payment_source": {
    "card": {
      "vault_id": "abc123",
      "stored_credential": {
        "payment_initiator": "MERCHANT",
        "payment_type": "RECURRING",
        "usage": "SUBSEQUENT",
        "previous_transaction_reference": "XYZ456"
 ```

### DIFF
--- a/modules/ppcp-wc-subscriptions/src/RenewalHandler.php
+++ b/modules/ppcp-wc-subscriptions/src/RenewalHandler.php
@@ -12,6 +12,7 @@ namespace WooCommerce\PayPalCommerce\WcSubscriptions;
 use WC_Subscription;
 use WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentSource;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentToken;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: modules/ppcp-wc-subscriptions/src/RenewalHandler.php [hunk 1]
```diff
@@ -193,6 +196,59 @@ private function process_order( \WC_Order $wc_order ): void {
...
+				foreach ( $subscriptions as $post_id => $subscription ) {
+					$previous_transaction_reference = $subscription->get_meta( 'ppcp_previous_transaction_reference' );
+					if ( $previous_transaction_reference ) {
+						$stored_credentials['previous_transaction_reference'] = $previous_transaction_reference;
+						break;
+					}
+				}
+
+				/**
+				 * Suppress ArgumentTypeCoercion
```
AlexP11223: Is it because of the `stdClass` function parameter type?
I think it can be fixed either by putting `object` instead of `stdClass` in the function phpdoc, or doing that and also using the `object` type in the function declaration itself, I think it's PHP 7.2+.

AlexP11223 (approved):

AlexP11223 (approved):

PR #1897, opened 2023-12-07
Add Pay Later Messaging block

AlexP11223:
Adds a block for embedding Pay Later messages.

It should work on any page (the scripts will load when the block is present).

The style can be configured via the block settings, it is saved into the `data-pp-` attributes of the `<div>` rendered on the frontend, so it is automatically applied by the PayPal SDK.

Using the [react-paypal-js](https://github.com/paypal/react-paypal-js) lib for rendering preview inside the block editor.

image


Dinamiko (approved):

AlexP11223:
Added the ability to use it inside the WC cart/checkout blocks, and added rendering on these pages. Also now using `50.00` amount when in editor and there is a cart/checkout block.

And re-implemented the message component without the `react-paypal-js` lib to avoid conflict when our express buttons load the SDK too.

PR #1900, opened 2023-12-11
Uniform Smart Button Display for PayPal, Google Pay, and Apple Pay (2177)

pedrosilva-pt:
**This PR also resolves:**
Do not render empty Apple Pay button div in frontend when buyer is not eligible (2411)

# PR Description
This PR refactors Google Pay and Apple Pay so that they follow the PayPal button layout, namely button `heights` (responsive and fixed) and `margins` between buttons.

PayPal takes advantage of it's buttons being rendered inside an `iframe` to add responsive behaviour via media queries.

To emulate this behaviour and add responsiveness to Google Pay and Apple Pay this PR adds `ApmButtons` helper class.
The helper calculates button margins and adds `CSS` classes to the button containers to allow for responsiveness.

This PR also ensures that Apple Pay is only rendered on eligible devices.

It also adds `apm-button.scss` mixin to share CSS rules between modules (`buttons`, `blocks`, `admin`) and between APM buttons (`googlepay`, `applepay`).

# Issue Description

_As a website owner,
I want the PayPal, Google Pay, and Apple Pay smart buttons to have a consistent appearance,
So that my website provides a cohesive and visually appealing payment experience to users._

Acceptance Criteria:

**1. PayPal Smart Button:**
* The width is already limited to 750px.
* The PayPal button shape should apply to the other buttons
* The PayPal button height should apply to the other buttons

**2. Google Pay Smart Button:**
* The width should be limited to 750px.
* The button height should be consistent with the PayPal buttons
* margins above/below should match PayPal buttons

**3. Apple Pay Smart Button:**
* The width should be limited to 750px.
* The button height should be consistent with the PayPal buttons
* margins above/below should match PayPal buttons
* margins around the button should not be present when the button cannot be displayed (e.g. when not eligible)


modules/ppcp-blocks/src/BlocksModule.php
```diff
@@ -88,6 +88,28 @@ static function () use ( $c ) {
 				$endpoint->handle_request();
 			}
 		);
+
+		// Enqueue frontend scripts.
+		add_action(
+			'wp_enqueue_scripts',
+			static function () use ( $c ) {
+				if ( ! has_block( 'woocommerce/checkout' ) && ! has_block( 'woocommerce/cart' ) ) {
```
AlexP11223, line 96: This may not load in editor (if block not added yet).

If we want to fix that, here is how I check it in the messaging branch https://github.com/woocommerce/woocommerce-paypal-payments/pull/1897/commits/4e34332c0bb476866106998dd097619c59692cf2

modules/ppcp-applepay/resources/js/ApplepayButton.js
```diff
@@ -179,13 +186,15 @@ class ApplepayButton {
...
+        $wrapper.addClass('ppcp-button-' + ppcpStyle.shape);
 
         if (ppcpStyle.height) {
-            jQuery('#' + wrapper).css('--apple-pay-button-height', `${ppcpStyle.height}px`)
+            $wrapper.css('--apple-pay-button-height', `${ppcpStyle.height}px`)
+            $wrapper.css('height', `${ppcpStyle.height}px`)
         }
 
-        jQuery(wrapper).append(appleContainer);
+        jQuery(wrapper).append(appleContainer); // ToDo: this selector doesnt seem valid.
```
AlexP11223: Yes, it seems to be doing nothing. `<apple-pay-button>` is already added because we set `innerHTML` above.

modules/ppcp-blocks/src/BlocksModule.php
pedrosilva-pt: I don't think we should load in the block editor right now as the preview there currently isn't working, maybe address it in the context of that fix:
image

AlexP11223 (approved):

Dinamiko (approved):

PR #1904, opened 2023-12-13
Block Checkout reloads when submitting order with empty fields (2389)

pedrosilva-pt:
# PR Description
To prevent the checkout page from reloading the `gotoContinuation` status variable was added. It's dynamically set only on scenarios where we want to redirect.

# Issue Description

When submitting a payment on the Block Checkout page while some fields are empty, the field error messages will only briefly appear as the page reloads

## Steps To Reproduce
- add product to cart
- navigate to Block Checkout
- leave some required fields empty
- click “Proceed to PayPal” button
- observe required field error briefly shows
- then page reloads

## Expected behaviour
page does not reload when WC validation failed

### DIFF
--- a/modules/ppcp-blocks/resources/js/checkout-block.js
+++ b/modules/ppcp-blocks/resources/js/checkout-block.js
@@ -31,7 +31,7 @@ const PayPalComponent = ({
     const {responseTypes} = emitResponse;
 
     const [paypalOrder, setPaypalOrder] = useState(null);
-    const [gotoContinuation, setGotoContinuation] = useState(false);
+    const [gotoContinuationOnError, setGotoContinuationOnError] = useState(false);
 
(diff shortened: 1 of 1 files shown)


modules/ppcp-blocks/resources/js/checkout-block.js
```diff
@@ -152,6 +153,7 @@ const PayPalComponent = ({
             if (config.finalReviewEnabled) {
                 location.href = getCheckoutRedirectUrl();
             } else {
+                setGotoContinuation(true);
```
AlexP11223: Maybe a better name could be something like `gotoContinuationOnError` or `submittingWithoutReview`.
Right now it looks a bit confusing because normally submitting in the no-review mode should not result in continuation.

Dinamiko (approved):

PR #1925, opened 2023-12-22
Implement early WC validation for Hosted Card Fields (2487)

pedrosilva-pt:
# PR Description
The PR adds support for both `Basic Checkout Validation` and `Early Checkout Validation` to the `card fields`.
It tries to maximize the reuse of existing validation logic.

A side effect is the `Early Checkout Validation` will be done even earlier also for the standard PayPal buttons.

# Issue Description
The new hosted card field implementation should trigger the early WC validation when clicking the “Place order” button.

Ideally, it should also be compatible with the basic validation (which is disabled by default).

## Acceptance Criteria
Given the new card fields are active
When filling card fields while leaving the WC form fields empty
Then the missing forms fields are immediately validated
And the 3D Secure popup does not trigger

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
@@ -48,7 +48,7 @@ class CheckoutActionHandler {
 
     configuration() {
         const spinner = this.spinner;
-        const createOrder = async (data, actions) => {
+        const createOrder = (data, actions) => {
             const payer = payerData();
             const bnCode = typeof this.config.bn_codes[this.config.context] !== 'undefined' ?
                 this.config.bn_codes[this.config.context] : '';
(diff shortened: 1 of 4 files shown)


modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
```diff
@@ -48,7 +48,13 @@ class CheckoutActionHandler {
 
     configuration() {
         const spinner = this.spinner;
-        const createOrder = (data, actions) => {
+        const createOrder = async (data, actions) => {
+            try {
+                await validateCheckoutForm(this.config);
```
AlexP11223: hm, is it not working for ACDC when doing it like before during the request?

I think this handler is also executed on the pay for order pages while before we were doing this validation only in checkout. Does it cause any issues?

pedrosilva-pt: Yes, it had the side effect of Pay for Order wanting to validate non present fields.

Previously the `created order` request wasn't validating ACDC and created the PayPal Order. But eventually the `checkout` redirect would detect the missing fields in the form and show the errors and skip the WooCommerce order creation.

I reverted back to validating only on the `create order` endpoint and adding support for credit card validation on that request.

AlexP11223 (approved):

PR #1926, opened 2023-12-22
make all hosted card fields strings translatable (2488)

pedrosilva-pt:
# PR Description

- Some of the `Card Fields` texts are already translatable via WooCommerce native translations. https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/gateways/class-wc-payment-gateway-cc.php#L60-L74
- For the `placeholders` of the `Card Fields` to be translated the `placeholder` Javascript configuration attribute needs to be passed to each field.
- The native WooCommerce `Card Number` field `placeholder` isn't translatable, this PR also adds the possibility of translating it, and applies the `woocommerce_paypal_payments_card_fields_translate_card_number` filter in case we want to disable this translation.

# Issue Description
Currently, only the Cardholder Name field can be translated with WordPress translation files.

The Card number, Expiry, and CVV fields currently are provided by the PayPal SDK but they should also be translatable with WordPress translation files.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/Renderer/CardFieldsRenderer.js
+++ b/modules/ppcp-button/resources/js/modules/Renderer/CardFieldsRenderer.js
@@ -55,28 +55,52 @@ class CardFieldsRenderer {
             const nameField = document.getElementById('ppcp-credit-card-gateway-card-name');
             if (nameField) {
                 let styles = cardFieldStyles(nameField);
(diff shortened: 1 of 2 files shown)


AlexP11223 (approved):

Mohamed-1005:
Any updates about how we style the new card fields? the CSS is not taking effect at all like before

PR #1944, opened 2024-01-05
Modularity module migration (2102)

pedrosilva-pt:
This PR migrates the modular system in the plugin to `inpsyde/modularity` 
More details about modularity: https://inpsyde.github.io/modularity/


modules/ppcp-admin-notices/src/AdminNotices.php
```diff
@@ -11,30 +11,36 @@
...
 
 	/**
 	 * {@inheritDoc}
 	 */
-	public function setup(): ServiceProviderInterface {
-		return new ServiceProvider(
-			require __DIR__ . '/../services.php',
-			require __DIR__ . '/../extensions.php'
-		);
+	public function services(): array {
```
AlexP11223, line 31: There is a possibility that simply migrating to the Syde Modularity `services` can cause issues because caching was not setup correctly in this project and the services were behaving more like `factories`.

But hopefully it works and improves performance. :pray:

Dinamiko (approved):

PR #1972, opened 2024-01-18
Fix warning undefined funding_source.

pedrosilva-pt:
This PR fixes a scenario with debug enabled where if `funding_source` is not defined it causes a warning ruining the respective JSON response.

### DIFF
--- a/modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
+++ b/modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
@@ -106,8 +106,9 @@ class SavePaymentMethodsModule implements ModuleInterface {
 				}
 
 				if ( $payment_method === PayPalGateway::ID ) {
+					$funding_source = $request_data['funding_source'] ?? null;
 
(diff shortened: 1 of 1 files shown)


modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
```diff
@@ -106,8 +106,9 @@ function( array $data, string $payment_method, array $request_data ): array {
 				}
 
 				if ( $payment_method === PayPalGateway::ID ) {
+					$funding_source = $request_data['funding_source'] ?? null;
 
-					if ( $request_data['funding_source'] === 'venmo' ) {
+					if ( $funding_source === 'venmo' ) {
```
Dinamiko: I would ensure `$funding_source` exist before checking its contents, so:
```
if ( $funding_source && $funding_source === 'venmo' ) {
```

Dinamiko (changes requested):

Dinamiko (approved):

PR #1984, opened 2024-01-24
NOT_ENABLED_TO_VAULT_PAYMENT_SOURCE error for merchants without reference transactions

Dinamiko:
In 2.5.0 release which includes Vault v3 integration, merchants without Reference Transactions enabled are getting `NOT_ENABLED_TO_VAULT_PAYMENT_SOURCE` error as the plugin is sending `vault` attribute despite it being disabled.

The plugin is sending the following vault attribute when Vaulting setting is disabled:
```
"paypal": {
            "attributes": {
                "vault": {
                    "store_in_vault": "xxxxxx",
                    "usage_type": "MERCHANT"
                }
            }
        }
```
This is causing merchants that do not have Reference Transactions enabled to have the order fail.

This PR ensures that `vault` property is not sent when Vaulting setting is disabled.

### Steps To Reproduce
Despite not being able to fully reproduce the error locally as Reference Transactions could not be disabled in Sandbox (at least we didn't found a way to do so at the moment), we need to ensure that the plugin does not send the `vault` attribute when creates the PayPal order if Vaulting setting is disabled.

- Disable Vaulting for both PayPal and Advanced Card payment payment methods.
- Enable Logging in Connection tab
- Purchase a product using one of the above payment methods
- Check log entries and ensure the `vault` attribute is not send:
```
  POST https://api-m.sandbox.paypal.com/v2/checkout/orders
  "payment_source": {
    "paypal": {
      "attributes": {
        "vault": {
          "store_in_vault": "ON_SUCCESS",
          "usage_type": "MERCHANT"
        }
```

### DIFF
--- a/modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
+++ b/modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
@@ -78,8 +78,15 @@ class SavePaymentMethodsModule implements ModuleInterface {
 		// Adds attributes needed to save payment method.
 		add_filter(
 			'ppcp_create_order_request_body_data',
-			function( array $data, string $payment_method, array $request_data ): array {
+			function( array $data, string $payment_method, array $request_data ) use ( $c ): array {
+				$settings = $c->get( 'wcgateway.settings' );
+				assert( $settings instanceof Settings );
+
 				if ( $payment_method === CreditCardGateway::ID ) {
+					if ( ! $settings->has( 'vault_enabled_dcc' ) || ! $settings->get( 'vault_enabled_dcc' ) ) {
+						return $data;
+					}
+
 					$save_payment_method = $request_data['save_payment_method'] ?? false;
 					if ( $save_payment_method ) {
 						$data['payment_source'] = array(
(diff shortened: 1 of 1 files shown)


pedrosilva-pt (approved):
My main doubt is if we should also disable other hooks or even the module as a whole if either of the vault modes is not enabled.

For example `woocommerce_paypal_payments_after_order_processor` or even `wp_enqueue_script`.

I'm not sure of the side effects of for ex adding this https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-save-payment-methods/resources/js/add-payment-method.js#L53 (although it seems to be only added on the add payment method page)

Dinamiko:
@pedrosilva-pt we could not disable the module completely at the moment as it is possible for a merchant to have Vaulting enabled only for one payment (ex. ACDC) and disabled for the other (PayPal), we have two Vaulting settings, one for each payment that can be enabled/disabled individually.

pedrosilva-pt:
> @pedrosilva-pt we could not disable the module completely at the moment as it is possible for a merchant to have Vaulting enabled only for one payment (ex. ACDC) and disabled for the other (PayPal), we have two Vaulting settings, one for each payment that can be enabled/disabled individually.

And if both vaulting modes are disabled simultaneously could we disable the module as a whole? Keeping the more specific settings checks (ACDC/PayPal) when relevant. That way if neither vaulting mode is enabled we would be certain the module would not interfere with the default behaviour. But I don't know the side effects of disabling the module.

Dinamiko:
> And if both vaulting modes are disabled simultaneously could we disable the module as a whole?

@pedrosilva-pt in this case it makes totally sense to disable the module completely, there is no reason to continue execution if Vaulting is disabled for both payments, I'll add the conditional, thanks!

PR #2079, opened 2024-03-05
Do not handle VAULT.PAYMENT-TOKEN.CREATED webhook for Vault v3 (2727)

Dinamiko:
In [Vault v3](https://developer.paypal.com/docs/multiparty/checkout/save-payment-methods/) (since version 2.5.0) there is no need to wait for and handle `VAULT.PAYMENT-TOKEN.CREATED` webhook as now the related info that needs to be stored (customer id, vault id…) is included in the response.

However the webhook is not handled correctly and a customer id not found error is thrown:
```
{
    "success": false,
    "message": "No customer id was found."
}
```

To fix the problem we could check for `resource_version` value and skip handling when version is not 2.0.

### DIFF
--- a/modules/ppcp-webhooks/src/Handler/VaultPaymentTokenCreated.php
+++ b/modules/ppcp-webhooks/src/Handler/VaultPaymentTokenCreated.php
@@ -115,6 +115,11 @@ class VaultPaymentTokenCreated implements RequestHandler {
 	 * @return WP_REST_Response
 	 */
 	public function handle_request( WP_REST_Request $request ): WP_REST_Response {
+		$resource_version = $request['resource_version'] ?? '';
+		if ( $resource_version && $resource_version !== '2.0' ) {
+			return $this->success_response();
+		}
+
 		$customer_id = null !== $request['resource'] && isset( $request['resource']['customer_id'] )
 			? $request['resource']['customer_id']
 			: '';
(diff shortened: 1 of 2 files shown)


manzurahammed:
Could you please apply the same logic in this method? Otherwise, it will return a failed response.

```json
{
    "success": false,
    "message": "Could not find handler for request type BILLING_AGREEMENTS.AGREEMENT.CREATED"
}
```

[https://github.com/woocommerce/woocommerce-paypal
-payments/blob/da6d57b77b22de420d7ca6bc0322225b914c5df5/modules/ppcp-webhooks/src/IncomingWebhookEndpoint.php#L251-L255](https://github.com/woocommerce/woocommerce-paypal-payments/blob/da6d57b77b22de420d7ca6bc0322225b914c5df5/modules/ppcp-webhooks/src/IncomingWebhookEndpoint.php#L251-L255)

Dinamiko:
@manzurahammed we are not subscribed to `BILLING_AGREEMENTS.AGREEMENT.CREATED` event, also I can not find it the [PayPal documentation](https://developer.paypal.com/api/rest/webhooks/event-names/).

This is probably coming from [deprecated billing agreements](https://developer.paypal.com/docs/api/payments.billing-agreements/v1/).

What about adding a check before the error handler and return success response if it exist? something like this:
```
		$event_type = $request['event_type'] ?: '';
		if ( in_array( $event_type, array( 'BILLING_AGREEMENTS.AGREEMENT.CREATED' ), true ) ) {
			return $this->success_response();
		}

		$message = sprintf(
			'Could not find handler for request type %s',
			$event_type
		);
		return $this->failure_response( $message );
```

Please let me know, thanks!

pedrosilva-pt (approved):

manzurahammed:
> What about adding a check before the error handler and return success response if it exist?

Looks good to me.

PR #2131, opened 2024-03-29
Fix issue #2114; Changed network_home_url() to home_url() to avoid wr…

### LINKED ISSUE #2114: Paypal Sub Directory Multisite returns network_home_url  (2829)
Clarification from WP support thread comment.
[wordpress support thread](https://wordpress.org/support/topic/return-url-not-correct/)

The return uri is correct when using express paypal checkout - the issue for us only appears when checkout is completed using woocommerce proceed to paypal button 
- Unvaulted , no Saved token for ppcp-gateway
- cancel and return url is correct

we changed `network_home_url()` to `home_url()` which successfully redirected - but we've only tasted a smol slice of pie.
let us know if additional info is requrest

paypal-checkout

--
https://github.com/woocommerce/woocommerce-paypal-payments/blob/09937fc991762c85af7afaf57d681ab253c17ea9/modules/ppcp-api-client/src/Repository/ApplicationContextRepository.php#L57

floek:
Changed network_home_url() to home_url().

**Issue**: #2114

---

### Description

Changed network_home_url() to home_url() to avoid wrong return url in multisite setup.

### Steps to Test

1. Setup multisite network, woocommerce on sub-page
2. Purchase an item and pay with PayPal, use redirects to paypal page.
3. After payment, the redirect back to the page hits the multisite main page and not the sub-page.

### Documentation

### Changelog Entry

Avoid wrong return url in multisite setup.

Closes #2114 .

### DIFF
--- a/modules/ppcp-api-client/src/Repository/ApplicationContextRepository.php
+++ b/modules/ppcp-api-client/src/Repository/ApplicationContextRepository.php
@@ -54,7 +54,7 @@ class ApplicationContextRepository {
 		$payment_preference = $this->settings->has( 'payee_preferred' ) && $this->settings->get( 'payee_preferred' ) ?
 			ApplicationContext::PAYMENT_METHOD_IMMEDIATE_PAYMENT_REQUIRED : ApplicationContext::PAYMENT_METHOD_UNRESTRICTED;
 		$context            = new ApplicationContext(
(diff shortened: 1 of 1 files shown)


InpsydeNiklas:
Thanks for the PR! Updated tests were merged with #2157

PR #2157, opened 2024-04-15
Fix the wrong return_url in multisite setup when using subdomains (2829)

Narek13:
# PR Description
Replace the `network_home_url()` with `home_url()` to avoid wrong return url in multisite setup.

# Issue Description
When using a multisite with subdomains or top-level domains, the return URL generated by PayPal Payments always points to the main site of the network.

# Steps to Reproduce

- Setup multisite network, with subdomains (top level domains)
- The store is on sub site.
- Purchase an item and pay with PayPal.
- After payment, the redirect back to the page hits the multisite main page and not the sub-page.


Dinamiko (approved):

darkopetreski:
I am experiencing the same error with 2.9.0 version.
The return url is domain.com/shop/shop/?.... and should be domain.com/shop?....
The problem occurs when I disable the smart button and do it with redirect:

```
add_filter('woocommerce_paypal_payments_use_place_order_button', '__return_true');
add_filter('woocommerce_paypal_payments_place_order_button_text', function() {
    return __( 'Place order', 'woocommerce' );
});
```

Edit: I have created a new issue: https://github.com/woocommerce/woocommerce-paypal-payments/issues/2604

PR #2182, opened 2024-04-26
Remove asterisk and dot to avoid invalid domain error

Dinamiko:
To fix AXO sdk client token invalid domain error, we must remove `*.` from [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-api-client/src/Authentication/SdkClientToken.php#L77).


danieldudzic (approved):
LGTM!

PR #1731, opened 2023-09-26
Fatal error when saving product while WooCommerce Subscriptions plugin is not active (2079)

Dinamiko:
In 2.3.0, products cannot be edited while the WooCommerce Subscriptions plugin is not active:
```
Fatal error: Uncaught Error: Class "WC_Subscriptions_Product" not found
in /var/www/html/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-subscription/src/SubscriptionModule.php on line 197
```

### Steps To Reproduce
- Install 2.3.0
- ensure WooCommerce Subscriptions plugin is not active
- edit an existing product
- observe fatal error

PR #2228, opened 2024-05-14
Fix fatal errors when subscriptions classes are not available

wjrosa:
### Description

This PR fixes a fatal error thrown when subscription classes are not available. Despite checking for the existence of some of the subscriptions classes, the class used in the `accept_only_automatic_payment_gateways` method is not verified. We introduce that check here. Without it, the following error happens:
```
2023-04-03T01:51:06+00:00 CRITICAL Uncaught Error: Class "WC_Subscriptions_Admin" not found in wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-subscription/src/Helper/SubscriptionHelper.php:91
Stack trace:
#0 wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Assets/SmartButton.php(763): WooCommerce\PayPalCommerce\Subscription\Helper\SubscriptionHelper->accept_only_automatic_payment_gateways()
#1 wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Assets/SmartButton.php(242): WooCommerce\PayPalCommerce\Button\Assets\SmartButton->has_subscriptions()
#2 wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/ButtonModule.php(57): WooCommerce\PayPalCommerce\Button\Assets\SmartButton->render_wrapper()
#3 wp-includes/class-wp-hook.php(308): WooCommerce\PayPalCommerce\Button\ButtonModule::WooCommerce\PayPalCommerce\Button\{closure}(Object(WP))
```

### Steps to Test

- Install the extension together with WooCommerce Subscriptions
- Include an `exit` before the `WC_Subscriptions` class declaration
- Add a Subscription Product to the cart
- Visit the checkout page
- Confirm that with this change no fatal errors are thrown

### DIFF
--- a/modules/ppcp-wc-subscriptions/src/Helper/SubscriptionHelper.php
+++ b/modules/ppcp-wc-subscriptions/src/Helper/SubscriptionHelper.php
@@ -89,7 +89,7 @@ class SubscriptionHelper {
 	 */
 	public function accept_only_automatic_payment_gateways(): bool {
 
-		if ( ! $this->plugin_is_active() ) {
+		if ( ! $this->plugin_is_active() || ! class_exists( \WC_Subscriptions_Admin::class ) ) {
 			return false;
 		}
 		$accept_manual_renewals = 'no' !== get_option(


mattallan (commented):
Hey @wjrosa, I'm having trouble testing this PR as I cannot reproduce the issue on `trunk`.

Looking at `plugin_is_active()`, this function is already checking for the existence of two subscriptions classes:
- `WC_Subscriptions` (the main plugin class): this typically tells us the plugin is installed/activated, and
- `WC_Subscriptions_Product` (a core subscriptions class during init)


Looking at [subscriptions-core code](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wc-subscriptions-core-plugin.php#L119-L120), the `WC_Subscriptions_Product` is loaded just before `WC_Subscriptions_Admin` so I'm not sure how a store would get into a situation where `WC_Subscriptions_Product` would exist, but `WC_Subscriptions_Admin` doesn't exist 🤔 

Inside `woocommerce-subscriptions.php` where are you putting the `exit`?

I tried putting it the line above `class WC_Subscriptions {` but that didn't work.

---

Adding a simple class exists and being defensive is always a good idea and so I'm fine to approve this PR, but I'd like to figure out how to reproduce this because the current class exists checks should be enough to assume other subscription classes exists without having to check everything 😄

wjrosa:
> Hey @wjrosa, I'm having trouble testing this PR as I cannot reproduce the issue on `trunk`.
> 
> Looking at `plugin_is_active()`, this function is already checking for the existence of two subscriptions classes:
> 
> * `WC_Subscriptions` (the main plugin class): this typically tells us the plugin is installed/activated, and
> * `WC_Subscriptions_Product` (a core subscriptions class during init)
> 
> Looking at [subscriptions-core code](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wc-subscriptions-core-plugin.php#L119-L120), the `WC_Subscriptions_Product` is loaded just before `WC_Subscriptions_Admin` so I'm not sure how a store would get into a situation where `WC_Subscriptions_Product` would exist, but `WC_Subscriptions_Admin` doesn't exist 🤔
> 
> Inside `woocommerce-subscriptions.php` where are you putting the `exit`?
> 
> I tried putting it the line above `class WC_Subscriptions {` but that didn't work.
> 
> Adding a simple class exists and being defensive is always a good idea and so I'm fine to approve this PR, but I'd like to figure out how to reproduce this because the current class exists checks should be enough to assume other subscription classes exists without having to check everything 😄

Thanks for reviewing! Yeah, you're right! I also cannot reproduce the issue here normally. I took a look at the git history and can see that the check for `WC_Subscriptions_Product` was added on September 26, 2023, [here](https://github.com/woocommerce/woocommerce-paypal-payments/pull/1731) (after the private issue was created). So no need for this change anymore. Sorry for the use of your time on this one.

PR #2236, opened 2024-05-17
Allow shipping callback and skipping confirmation page from any express button (2815)

Narek13:
# PR Description

- The PR will cover all the points mentioned below.
- If subscription plugin is active the "Final Confirmation" checkbox will be force checked and grayed out(disabled), means the shipping callback functionality will not work
- Fix the bugs found on block based pages (cart, checkout)

# Issue Description

Currently, shipping callback is disabled by default, and when enabled, it is only active on the block Cart/Checkout pages.
The callback does not work on the other buttons.

**We need to**

- Enable the shipping callback functionality by default for new stores
- Make the shipping callback functional with all buttons except the Classic Checkout
   - Single Product
   - Classic Cart
   - Mini Cart
- if PayPal returns no billing data for the buyer, ensure the shipping data is copied into the billing fields on the Classic Checkout & block Checkout
- if PayPal returns complete billing data, use the returned data for billing details
- If the merchant has a terms & conditions page configured in the WooCommerce Advanced settings, then the buyer should be forwarded to the Checkout page in continuation mode instead of completing the payment immediately instead of auto-completion, 
   - display a warning text in the settings about the missing content 
   - we likely can’t check if there are any other custom required checkout fields (e.g. a custom required birthdate field)


modules/ppcp-blocks/resources/js/checkout-block.js
```diff
@@ -82,7 +82,7 @@ const PayPalComponent = ({
         window.ppcpContinuationFilled = true;
     }, [])
 
-    const createOrder = async () => {
+    const createOrder = async (data, actions) => {
```
Dinamiko, line 85: `actions` seems not used in the body of the function.

modules/ppcp-blocks/resources/js/checkout-block.js
```diff
@@ -296,10 +297,15 @@ const PayPalComponent = ({
         onClick();
     };
 
+    const isVenmoAndVaultingEnabled = () => {
```
Dinamiko, line 300: Seem 'isVenmoAndVaultingEnabled` function is not used.

modules/ppcp-button/resources/js/modules/Renderer/Renderer.js
```diff
@@ -64,6 +68,15 @@ class Renderer {
         }
     }
 
+    shouldHandleShippingInPaypal = (venmoButtonClicked) => {
+        if (!this.defaultSettings.should_handle_shipping_in_paypal) {
+            console.log('no')
```
Dinamiko, line 73: Better not using console log in production code.

Dinamiko (approved):

PR #2246, opened 2024-05-21
Add block Checkout compatibility to Advanced Card Processing (2722)

### LINKED ISSUE #1462: Advanced Card Processing w/ Checkout Blocks support
#211 was closed, but I think this gives the wrong idea around compatibility between this plugin and the WooCommerce block-driven checkout experience.

You can, as of 2.1.0, add PayPal buttons in the block checkout. You _cannot_ take cards directly on your site (Advanced Card Processing), which was my primary want from this ticket. I've always found that sending folks to a third-party site for payment feels cheap.

Opening this ticket as a place to centralize updates from the dev team around this feature, and so that folks looking for more information on this issue get some kind of explicit information about the current status of that feature rather than sifting through the somewhat misleading information provided when #211 was closed.

Dinamiko:
Advanced Card Processing needs to be integrated with the block Checkout.

### Features
- Integrate only the new card fields (not the hosted fields)
- Add Vaulting support
- Gateway should declare support for block Checkout for WC not to display a warning when editing the block Checkout
- When ACDC is enabled, it is active for both the Classic and block Checkout

#### Customer purchase product
- As customer add product to the cart and go to checkout page
- Select ACDC payment, add credit card data and click “Place order”
    - Order processed correctly

#### Customer save payment during purchase
- As admin enable vaulting in Advanced Card Processing payment settings
- As logged in customer add product to the cart and go to checkout page
- Select ACDC payment, check “Save payment” checkbox and place order
    - Order processed correctly
    - Payment saved correctly in both PayPal and WC

#### Customer purchase WC Subscriptions product
- As customer add subscription product to the cart and go to checkout page
    - Ensure “Save card” checkbox is checked and disabled 
- Select ACDC payment and place order
    - Order processed correctly

#### 3D Secure
- As merchant go to “Advanced Card Processing” and set 3D Secure contingency to “Always trigger 3D secure”
- As customer add product to the cart and go to checkout page
- Select ACDC payment, use a card number from [this list](https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/test/#link-dsecuretestscenarios) and place order
    - Order processed correctly
    - Ensure WC order contains an order note with 3DS Authentication Result info

#### Card Icons
- As admin go to Payments / Advanced Card Processing
- In “Show logo of the following credit cards” select some card brands and save
- As customer add product to the cart and go to checkout page
    - Ensure card logos are displayed in the payment gateway


danieldudzic:
@Dinamiko  I have some questions regarding a few issues I have run into:

1. I have noticed the errors are less descriptive/apparent on the block checkout:

Note: The Block Checkout doesn't have any error on top in this scenario (invalid card number), so the payment fails and no feedback is provided to the customer.

Aside: No idea why the error on the classic checkout has a green color 🙈

|classic checkout|block checkout|
|-|-|
|classic_checkout_checkot_invalid|block_checkot_invalid|

2. The Express Credit Card payment never loads for me, and is stuck on:
Page__Checkout

3. I couldn't get the `Save your card` option to work with the generated card details. Do I need to use a specific number to test?

modules/ppcp-blocks/src/AdvancedCardPaymentMethod.php
```diff
@@ -0,0 +1,141 @@
...
+			'is_vaulting_enabled' => $this->settings->has( 'vault_enabled_dcc' ) && $this->settings->get( 'vault_enabled_dcc' ),
+		);
+	}
+
+	/**
+	 * The smart button.
+	 *
+	 * @return SmartButtonInterface
+	 */
+	private function smart_button(): SmartButtonInterface {
```
stracker-phil: Personally, I'd rename the method to `smart_button_inst()` to avoid confusion.

On first sight, the following pattern looks strange/wrong (though it's actually syntactically correct):
```php
function smart_button() {
    ...
    if ( is_callable( $this->smart_button ) ) {
       ...
    }
    ...
   return $this->smart_button;
}
```

modules/ppcp-blocks/resources/js/Components/card-fields.js
```diff
@@ -0,0 +1,81 @@
...
+        return new Promise((resolve) => {
+            console.log('start...')
+            setTimeout(() => {
+                console.log('resolve')
+                resolve()
+            }, milliseconds)
+        })
+    };
+
+    useEffect(
```
stracker-phil, line 29: Idea 1:

Maybe you can add `onApprove` as a dependency to the effect + introduce a new "isReady" state that you use for rendering the form. Somewhat like this:

```js
// New state to track if the component is ready to be rendered.
const [isReady, setIsReady] = useState(false);

// Effect that triggers when onApprove is available
useEffect(() => {
    if (!onApprove && !onPaymentSetup) {
        return;
    }
    
    setIsReady(true);
    return onPaymentSetup( ... )    
},  [onPaymentSetup, cardFieldsForm, onApprove]); // Add "onApprove" dependency here

return (
    <>
        {isReady ? // Conditionally render the PayPal component, when all dependencies are ready
            (
            <PayPalScriptProvider options={{...}}>
                <PayPalCardFieldsProvider
                    createOrder={createOrder}
                    onApprove={onApprove}
                    onError={(err) => {
                        console.error(err);
                    }}>
                    <PayPalCardFieldsForm/>
                    ...
                </PayPalCardFieldsProvider>
            </PayPalScriptProvider>
            ) : // When not ready, display "..." (or something else/nothing)
            '...'
        }
    </>
);
```

modules/ppcp-blocks/resources/js/card-fields-config.js
```diff
@@ -0,0 +1,44 @@
...
+        .then((order) => {
+            return order.data.id;
+        })
+        .catch((err) => {
+            console.error(err);
+        });
+}
+
+export function onApprove(data) {
+    fetch(config.scriptData.ajax.approve_order.endpoint, {
```
stracker-phil: Idea 2 (possibly better):

Maybe just returning a promise here will solve the problem.

```js
export function async onApprove(data) {
    return fetch(...)
}
```

The sample implementation of `onApprove` in the PayPal repository suggests that the `<PayPalCardFieldsProvider>` can accept a Promise as `onApprove` property:
https://github.com/paypal/react-paypal-js/blob/672a48de49ddaea40b5db0f7f0d9c0670d8e7f88/src/stories/utils.ts#L100

Dinamiko: @stracker-phil just checked and it works like a charm returning a promise, thanks!

stracker-phil (approved):

PR #2263, opened 2024-05-27
AXO: Send JS logs to WooCommerce logger (3172)

Dinamiko:
As most of the AXO flows happens on the frontend it would make sense to add logic for sending JS log entries to WooCommerce logger so we could debug what happened, for example which kind of user started the flow or which data was sent and received from PayPal.


modules/ppcp-axo/src/FrontendLoggerEndpoint.php
```diff
@@ -0,0 +1,80 @@
...
+		return self::ENDPOINT;
+	}
+
+	/**
+	 * Handles the request.
+	 *
+	 * @return bool
+	 * @throws Exception On Error.
+	 */
+	public function handle_request(): bool {
```
danieldudzic, line 64: Even though this is just a logger, I'd recommend we implement nonce verification for all our endpoints. Like:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/ed3a48e520780df5500b989d4eba3af7ccbe72b1/modules/ppcp-paylater-configurator/src/Endpoint/GetConfig.php#L82

modules/ppcp-axo/resources/js/AxoManager.js
```diff
@@ -584,6 +586,7 @@ class AxoManager {
             const authResponse = await this.fastlane.identity.triggerAuthenticationFlow(lookupResponse.customerContextId);
 
             log('AuthResponse', authResponse);
+            await this.log(`triggerAuthenticationFlow: ${JSON.stringify(authResponse)}`);
```
danieldudzic: I don't like the fact that now we have both `log()` and `this.log` in the code - it's confusing.

Can we perhaps combine these, to keep the code cleaner, but still keep the `console.log` functionality and also add the WooCommerce logger.

danieldudzic (changes requested):
Good job on this @Dinamiko!

I have tested the PR and the logging functionality seems to be working correctly.

logger



The code looks legit, I only have a few minor suggestions and concerns :)

I am a bit wary of including so much sensitive data (customer billing details) in the logs. Perhaps we can obfuscate/sanitize the response?

Dinamiko:
> Good job on this @Dinamiko!
> 
> I have tested the PR and the logging functionality seems to be working correctly.
> 
> logger
> 
> The code looks legit, I only have a few minor suggestions and concerns :)
> 
> I am a bit wary of including so much sensitive data (customer billing details) in the logs. Perhaps we can obfuscate/sanitize the response?

WC Logs are accessible to admins only, customer billing details can also be seen by admins in the WC order. We are already logging the requests to PayPal which includes the payer information in the request body.

Being said that, right now I'm not 100% sure which data should be logged... the idea is log the data that could be useful when debugging issues, so the idea for now is to have the frontend logging mechanism in place and then we'll decide which log entries do we really need.

modules/ppcp-axo/src/FrontendLoggerEndpoint.php
Dinamiko: > Even though this is just a logger, I'd recommend we implement nonce verification for all our endpoints. Like:
> 
> https://github.com/woocommerce/woocommerce-paypal-payments/blob/ed3a48e520780df5500b989d4eba3af7ccbe72b1/modules/ppcp-paylater-configurator/src/Endpoint/GetConfig.php#L82

[Nonce verification](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Endpoint/RequestData.php#L47) and [input sanitization](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Endpoint/RequestData.php#L61) is already implemented in `RequestData::read_request`.

What we could do to avoid confusion and code duplication is to extract that logic out of buttons module to a shared one, or at least move it to the api client module.

modules/ppcp-axo/resources/js/AxoManager.js
Dinamiko: Yeah, I don't like it either... as mentioned in a previous comment I don't know exactly which information should we log so for now I just ignored the existing console log to avoid noise if the WC logs, but yes the idea is to consolidate both in a single one.

danieldudzic (approved):

PR #2265, opened 2024-05-27
Make PayPal Subscription products unique in cart (2556)

burdulixda:
### Description

If there is a PayPal Subscription product in the cart, that can be the _only_ product in the cart.

### Steps to Test

1. Add a PayPal Subscription product in the cart;
2. Try to add another product in the cart;
3. The error notice should appear prompting the user that PayPal subscription should be a unique item in the cart;
image

### DIFF
--- a/modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
+++ b/modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
@@ -87,29 +87,35 @@ class PayPalSubscriptionsModule implements ModuleInterface {
 			12
 		);
 
-		add_filter( 'woocommerce_add_to_cart_validation', function
-		(
-			$passed_validation,
-			$product_id
-		) {
-			if ( WC()->cart->is_empty() ) {
+		add_filter(
+			'woocommerce_add_to_cart_validation',
+			/**
+			 * Param types removed to avoid third-party issues.
+			 *
+			 * @psalm-suppress MissingClosureParamType
+			 */
+			function ( $passed_validation, $product_id ) {
+				if ( WC()->cart->is_empty() ) {
+					return $passed_validation;
+				}
+
+				foreach ( WC()->cart->get_cart() as $cart_item_key => $cart_item ) {
+					if ( get_post_meta( $cart_item['product_id'], 'ppcp_subscription_product', true ) ) {
+						wc_add_notice( __( 'You can only have one subscription product in your cart.', 'woocommerce-paypal-payments' ), 'error' );
+						return false;
+					}
+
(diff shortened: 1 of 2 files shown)


modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
```diff
@@ -87,6 +87,36 @@ function( $product_id ) use ( $c ) {
...
+			 *
+			 * @psalm-suppress MissingClosureParamType
+			 */
+			function ( $passed_validation, $product_id ) {
+				if ( WC()->cart->is_empty() ) {
+					return $passed_validation;
+				}
+
+				foreach ( WC()->cart->get_cart() as $cart_item_key => $cart_item ) {
+					if ( get_post_meta( $cart_item['product_id'], 'ppcp_subscription_product', true ) ) {
```
Dinamiko: Better use [WC Data API for product meta](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-paypal-subscriptions/src/SubscriptionsApiHandler.php#L141), it's more future proof, if WC decides to change how it works internally in the future we don't care as we follow the API contract.

Dinamiko (changes requested):

modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
```diff
@@ -660,6 +693,11 @@ private function update_subscription_product_meta( WC_Product $product, Subscrip
 			if ( $product->meta_exists( 'ppcp_subscription_product' ) && $product->meta_exists( 'ppcp_subscription_plan' ) ) {
 				$subscriptions_api_handler->update_product( $product );
 				$subscriptions_api_handler->update_plan( $product );
+
+				if ( $product->get_meta( '_ppcp_enable_subscription_product', true ) === 'yes' ) {
+					update_metadata( 'post', $product->get_id(), '_sold_individually', 'yes', 'no' );
```
Dinamiko: Better use WC Product API to [set sold individually](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L1013-L1015):

Captura de pantalla 2024-05-29 a las 11 28 10

Dinamiko (changes requested):

Dinamiko (approved):

burdulixda (commented):
@Dinamiko I had to also add "set_sold_individually()" function during add to cart validation because some users may never update their products manually

Dinamiko (approved):

PR #2281, opened 2024-05-29
Disable Pay Upon Invoice if billing/shipping country not set (737)

burdulixda:
### Description

PUI was only being disabled if the user selected billing/shipping country was anything other than Germany. However, the PUI was not being disabled if the billing/shipping country was not set at all.

### Steps to Test

1. Set default customer location to 'No location by default' (WooCommerce -> Settings -> General);
2. Use incognito tab (logged out) to add product to cart;
3. Navigate to checkout page;
4. Notice the absence of PUI method if billing/shipping country is not set, or is set to any other country except Germany.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PayUponInvoiceHelper.php
@@ -56,13 +56,13 @@ class PayUponInvoiceHelper {
 		}
 
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):

burdulixda (commented):
Previous method of country retrieval was returning an empty string during some point after placing the order, causing the `is_checkout_ready_for_pui()` function to return `false`. WC Customer API always returns correct billing/shipping country, even for logged out users.

Dinamiko (approved):

PR #2308, opened 2024-06-03
Add WPML compatibility to the settings (223)

### LINKED ISSUE #2119: [WPML] Not possible to translate the "Pay via PayPal" text (223)
### Describe the Bug
Hi there, this is Diego from the WPML Compatibility team.

We got several tickets in our support forum from clients trying to translate the "Pay via Paypal"string from the plugin, which is not working as expected.

Some examples:
- https://wpml.org/forums/topic/cannot-translate-woocommerce-paypal-payments-title-and-description/
- https://wpml.org/forums/topic/i-want-to-translate-woocommerce-paypal-payments-strings/
- https://wpml.org/forums/topic/pay-via-paypal-not-translate-woocommerce-paypal-payments/

We tried to reproduce in a clean install but it was not possible, so it seems that this issue only happens in specific scenarios.

### Workaround
We created the following errata with the details and a workaround: https://wpml.org/errata/woocommerce-paypal-payments-pay-via-paypal-translation-not-working/

These are the steps:
- Go the "Translate texts in admin screens"page and register/translate the title and description strings from the "admin_texts_woocommerce-ppcp-settings" textdomain
- Open the app/public/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Settings/Settings.php file
- Look for the following snippet, around line 134
```php
foreach ( $defaults as $key => $value ) {
			if ( isset( $this->settings[ $key ] ) ) {
				continue;
			}
			$this->settings[ $key ] = $value;
		}
```
Replace with:
```php
foreach ( $defaults as $key => $value ) {
			if ( isset( $this->settings[ $key ] ) ) {
				// WPML workaround for compsupp-7085
				$this->settings[$key] = apply_filters('wpml_translate_single_string', $this->settings[$key] , 'admin_texts_woocommerce-ppcp-settings', '[woocommerce-ppcp-settings]'.$key);
				continue;
			}
			$this->settings[ $key ] = $value;
		}
```
- Note that we are just applying the wpml_translate_single_string filter there. See https://wpml.org/wpml-hook/wpml_translate_single_string/

### Solution Idea
Maybe it would be possible to add a generic filter there? This way, users could use this filter to translate the string without having to modify the original code. Something like this:
```php
foreach ( $defaults as $key => $value ) {
			if ( isset( $this->settings[ $key ] ) ) {
				// WPML workaround for compsupp-7085
				$this->settings[$key] = apply_filters('woocommerce_paypal_payments_add_to_cart_text', $this->settings[$key]);
				continue;
			}
			$this->settings[ $key ] = $value;
		}
```

This could be useful for other plugins as well.

danieldudzic:
### Description

This PR adds a filter `woocommerce_paypal_payments_settings_value` that allows for filtering of plugin settings value.

### Note

I wasn't able to create a compat filter from the compat modules, as the compat module always executes later.

### MU plugin to test

Save the following code to a file, and add it to the `wp-content/mu-plugins` folder:
```
<?php
/*
 * Plugin Name: WooCommerce PayPal Payments Settings Filter
 * Description: Adds a filter to translate WooCommerce PayPal Payments settings before other plugins are loaded.
 * Author: Danny
 * Version: 1.0
 */

add_filter(
    'woocommerce_paypal_payments_settings_value',
    function( $value, $key ) {
        return ! is_array( $value ) ? apply_filters( 'wpml_translate_single_string', $value, 'admin_texts_woocommerce-ppcp-settings', '[woocommerce-ppcp-settings]' . $key ) : $value;
    },
    10,
    2
);
```

### Steps to Test

1. Install WPML, WooCommerce WPML and String Translations.
2. Set it up.
3. Go to the String Translation page in the WPML settings.
4. Search for `Pay via PayPal.`
5. Click on `Can't find the strings you're looking to translate?` and then on the `Choose texts for translation` button.
6. Add the `Pay via PayPal.` to translation and actually translate it.
7. Go to the Checkout page with the translated string.
8. Verify that the translated string appears correctly.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Settings/Settings.php
+++ b/modules/ppcp-wc-gateway/src/Settings/Settings.php
@@ -154,8 +154,10 @@ class Settings implements ContainerInterface {
 				'woocommerce-paypal-payments'
 			),
 		);
+
 		foreach ( $defaults as $key => $value ) {
 			if ( isset( $this->settings[ $key ] ) ) {
+				$this->settings[ $key ] = apply_filters( 'woocommerce_paypal_payments_settings_value', $this->settings[ $key ], $key );
 				continue;
 			}
 			$this->settings[ $key ] = $value;


stracker-phil:
Nice workaround, and works as expected. I understand, it's more of a temporary solution, until implementing PCP-1241.

@danieldudzic is there a reason for the mu-plugin, instead of adding that filter directly into the plugin?

From my perspective, adding that code into the `Settings.php` file would be a better choice (e.g. in the top of the `load()` method). That way, we can adapt, or remove, the logic directly when needed, vs. having a mu-plugin file on client websites that is not required anymore

danieldudzic:
> Nice workaround, and works as expected. I understand, it's more of a temporary solution, until implementing PCP-1241.

Correct, although I'm don't think PCP-1241 will be happening anytime soon.

> @danieldudzic is there a reason for the mu-plugin, instead of adding that filter directly into the plugin?

It's basically a test work-around to be able to hook into the filter before it gets executed.

> From my perspective, adding that code into the Settings.php file would be a better choice (e.g. in the top of the load() method). That way, we can adapt, or remove, the logic directly when needed, vs. having a mu-plugin file on client websites that is not required anymore.

I would prefer not to pollute the core plugin code with 3rd party compatibility code. All 3rd party compatibility code should live exclusively in the `ppcp-compat` module - and unfortunately, I haven't found a good solution to load the `add_filter` before it gets called on the **Settings** with the current order of loading.

I believe this still is an improvement, as WPML customers willing to overwrite settings won't need to modify plugin files, but will be able to add a "fix" externally.

Dinamiko (approved):

PR #2325, opened 2024-06-07
Apple Pay & Google Pay buttons no longer visible in Standard Payments button previews after moving them to Advanced Card Processing tab (3179)

stracker-phil:
## Problem
After moving Google Pay and Apple Pay settings to the "**Advanced Card Processing**" tab, the button preview was not available anymore - neither in the "Standard Payments" nor the "Advanced Card Processing" tab.

## Background
1. Advanced Card Processing: Did not render the preview because relevant JS files were not enqueued
2. Standard Payments: Did not render the preview because it was tightly coupled to the presence of certain DOM elements (eg the presence of an "is-enabled" checkbox)

## Solution
1. Refactor preview button code to decouple it from DOM elements
2. Enqueue relevant JS files in both tabs

### Preview button logic

- Adjust the preview button generator to render an individual button (before this, the button preview always showed a block of all PayPal payment buttons)
- Layout of the single preview button should be as close to other "Button Styling Preview" blocks as possible
- On the "Standard Payments" tab, the APM buttons should be static and not change colors/labels when changing any settings

#### `ppcp-button/.../PreviewButton.js` 
- Contains the common logic to render a single APM button preview
- "Abstract" base class, some methods must be implemented in APM module

#### `ppcp-button/.../PreviewButtonManager.js`
- Extracted common logic to manage _one or multiple_ APM `PreviewButton`s
- Bootstrap logic that loads remote configuration from PayPal
- "Abstract" base class, some methods must be implemented in APM module

#### Both `boot-admin.js`
- Implement child classes of `PreviewButton` and `PreviewButtonManager`

#### `ppcp-wc-gateway/services.php`
- Condition now loads the APM's `boot-admin.js` in both tabs ("Standard Payments" and "Advanced Card Processing")

#### Both `extensions.php`
- APM edit form includes a button preview (at the end)
- Custom attributes for edit-fields, to trigger an update of the button preview

#### `gateway-settings.js`
- New light-weight form-observer that interacts with the changes in `extensions.php`.
- The new observer code only renders & refreshes a single APM button

### Screenshots
Fixed preview of the full button block, in the "Standard Payments" tab:
sp-button-block-preview

New preview of the single payment button, in the "Advanced Card Processing" tab:
acp-single-button-preview


modules/ppcp-button/resources/js/modules/Renderer/PreviewButton.js
```diff
@@ -0,0 +1,135 @@
...
+        return this;
+    }
+
+    /**
+     * Updates the button configuration with current details from the form.
+     *
+     * @return {this} Reference to self, for chaining.
+     */
+    setPpcpConfig(config) {
+        this.ppcpConfig = merge({}, config);
```
Dinamiko, line 71: Probably I'm missing something here... but why merge an empty object here? how is this different from:
```
this.ppcpConfig = config;
```

modules/ppcp-button/resources/js/modules/Renderer/PreviewButton.js
```diff
@@ -0,0 +1,135 @@
...
+        /*
+         * previewButtonConfig.button.wrapper must be different from this.ppcpConfig.button.wrapper!
+         * If both selectors point to the same element, an infinite loop is triggered.
+         */
+        const buttonWrapper = previewButtonConfig.button.wrapper.replace(/^#/, '')
+        const ppcpWrapper = this.ppcpConfig.button.wrapper.replace(/^#/, '')
+
+if (buttonWrapper === ppcpWrapper) {
+throw new Error(`[APM Preview Button] Infinite loop detected. Provide different selectors for the button/ppcp wrapper elements! Selector: "#${buttonWrapper}"`);
+}
```
Dinamiko: We do not have the linter installed yet... but just indent this a bit for now.

modules/ppcp-button/resources/js/modules/Renderer/PreviewButtonManager.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Protected method that needs to be implemented by the derived class.
+     * This method is responsible for creating a new PreviewButton instance and returning it.
+     *
+     * @param {string} wrapperId - CSS ID of the wrapper element.
+     * @return {PreviewButton}
+     */
+    createButtonInst(wrapperId) {
```
Dinamiko: Better use complete words: `createButtonInstance`.

Dinamiko (approved):
It looks good, PR is approved, I've added a couple of minor adjustments.

Regarding jQuery I’ve noticed that most of the code was already there before you started the implementation, we'll leave it there for now but just for you to know that the idea would be to progressively refactor to vanilla JS, so not in this PR but from now on each opportunity found to refactor it, just do it.

modules/ppcp-button/resources/js/modules/Renderer/PreviewButton.js
stracker-phil: Good catch 👍 The idea of this part is, to create a local copy of the config object.

Since the config object can be modified outside the class, we want to keep a static version of the provided configuration.

```js
this.button[id].setPpcpConfig(config)
...
config.color = 'black'
```

Without the "merge", this would cause some unexpected behavior; with "merge" changing the config object outside the class does not impact the preview, until explicitly calling `setPpcpConfig()` again.

Dinamiko (approved):

PR #2354, opened 2024-06-18
POC: Receive button properties from the Checkout Block

Dinamiko:
This is some exploratory work to explore how we can synchronise the express payment button styles better to create more cohesiveness in the Express Checkout block. This PR is a part of [POC: Pass properties to Express Checkout buttons](https://github.com/woocommerce/woocommerce/pull/47899). It updates the express payment method to receive `buttonAttributes` property from the checkout block, and override the `height` property if available.

### Testing instructions

- Run woocommerce locally and check out the `try/poc-express-checkout-button-styles` branch
- Make sure the environment is right for one of the express checkout buttons to be visible (either Apple or Google Pay)
- Add some items to the cart and visit the cart page (with the cart block)
- Ensure the express payment method is visible there
- Click "Proceed to checkout"
- Ensure the express payment method is visible there too
- In a new tab, open the checkout page in the editor.
- Click on the express checkout area and change the button size to "Small"
- Refresh the checkout page in the front-end
- You should notice the button change size to a height of 40px

Please note that is not possible to set "Large" size as PayPal only allows values between [25px and 55px](https://developer.paypal.com/docs/multiparty/checkout/standard/customize/buttons-style-guide/#size):
```
Error: Expected style.height to be between 25px and 55px - got 56px
```


Dinamiko:
Hi @alexflorisca, just to let you know that we have created this PR for working on [#47899](https://github.com/woocommerce/woocommerce/pull/47899) POC integration, for now it only checks if `buttonAttributes` exist and if so it gets the button height value from there.

I noticed that is not currently possible to set "Large" size because there is a button size limitation on PayPal.

cc. @InpsydeNiklas

danieldudzic:
@Dinamiko I messed up this branch it seems 😓 so created a new one here:

https://github.com/woocommerce/woocommerce-paypal-payments/pull/2448

PR #2381, opened 2024-07-01
Remove BCDC button from block Express Checkout area (3319)

Dinamiko:
In some sites we are seeing how the express card button is displayed when ACDC is enabled, that should not be the case as it could cause issues when both are enabled at the same time. The expected result should be the express card button not rendered when ACDC is enabled.

image-20240625-171527

The problem is happening in block checkout pages where in WC Settings the “Checkout page” is configured to use a block checkout page,[ the code responsible](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-button/src/Assets/SmartButton.php#L1407-L1409) to add `card`  to `disable-funding` uses `is_checkout` which will return `true` or `false` depending on the above configuration.

In this case, `is_checkout` returns `true` and therefore `card` is not added to `disable-funding` making the button visible.

### Steps to reproduce
- Go to WC settings / Advanced
- In “Page setup” set a block checkout page for “Checkout page”
- Ensure ACDC is enabled
- Add product to cart and visit a block checkout page
    - express card button is visible

### Changes introduced
- Extracts the logic responsible of creating the list of disabled funding sources into a new class.
- Add unit tests to cover code parts related to the issue.
- Add new condition to ensure `card` is added to disabled funding source when context is either block checkout or block cart.

### Acceptance Criteria
**GIVEN** the Advanced Card Processing is enabled
**WHEN** navigating to the block Checkout
**THEN** the BCDC button does not appear in the Express Checkout area

### DIFF
--- a/modules/ppcp-button/services.php
+++ b/modules/ppcp-button/services.php
@@ -16,6 +16,7 @@ use WooCommerce\PayPalCommerce\Button\Helper\CartProductsHelper;
 use WooCommerce\PayPalCommerce\Button\Helper\CheckoutFormSaver;
 use WooCommerce\PayPalCommerce\Button\Endpoint\SaveCheckoutFormEndpoint;
 use WooCommerce\PayPalCommerce\Button\Helper\ContextTrait;
+use WooCommerce\PayPalCommerce\Button\Helper\DisabledFundingSources;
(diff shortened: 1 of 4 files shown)


stracker-phil (approved):
Love the new unit test (and the related refactoring) here 👍

PR #2393, opened 2024-07-05
Advanced Card Processing payments get stuck on the first attempt when using CheckoutWC (2689)

stracker-phil:
### Problem

When using [CheckoutWC](https://wordpress.org/plugins/checkoutwc-lite/), payments via Advanced Card Processing did not work.

Two different problems were observed:

----

### Issue 1

_Symptom: The "Debit & Credit Cards" fields were not displayed during checkout_

#### Background
The CheckoutWC plugin creates a custom checkout page, ignoring block editor contents. If the checkout page previously used block checkout, our plugin responded with incorrect JS configuration, leading to issues with payment elements initialization.

#### Solution
Implemented a compatibility filter to check for the CheckoutWC template. Adjusted JS settings accordingly to ensure proper initialization of the classic checkout logic.

Commit: https://github.com/woocommerce/woocommerce-paypal-payments/pull/2393/commits/a349c871828dd395328877e52ee27e1fc1271e86

----

### Issue 2

_Symptom: The card payment resulted in a "pending" mode with an endless loading animation_

#### Background
Sometimes, the CheckoutWC would invoke the `CardFieldRenderer::render()` JS method multiple times. This seems to be related to caching or JS loading order, and usually only happens once per session.

The render method is responsible for injecting the credit card form into the checkout form and linking the fields to the submit button.

Our code was laid out to render exactly once. On the second render call, the original DOM fields were not updated again, but new submit listeners were created. Short: The DOM and the JS events got out of sync.

#### Solution
A small refactoring of the `CardFieldRenderer` class now allows calling the `render()` method multiple times without any side effects.

The logic to prepare field-options (styling, placeholders) is extracted to a JS getter, which parses the DOM once on first access. All subsequent render calls will use those initial field options to re-render the input fields.

Commit: https://github.com/woocommerce/woocommerce-paypal-payments/pull/2393/commits/c685a088a561ec2ea984f93454837545cf89a242


Dinamiko (approved):

Dinamiko (approved):

InpsydeNiklas:
We won't address this issue. It needs to be resolved by **CheckoutWC** directly, so I will close this PR.

PR #2404, opened 2024-07-10
Prevent enabling Standard Card Button when ACDC is enabled (2803)

burdulixda:
### Description

"IMHO hiding the tab will be too confusing for users. Because when **Create gateway for Standard Card Button** is enabled but the tab simply doesn’t appear, users may consider the plugin to be broken.

It’s more complex but for best possible UX, I propose force-unchecking and greying out the `Enable Standard Card Button gateway` checkbox when ACDC is enabled while **Create gateway for Standard Card Button** is checked. 

image

It should also be impossible to enable this method from the **Payments** tab (it should automatically disable itself after trying to enable it with the toggle).

Additionally, a notice should be displayed on the **PayPal Payments settings** and on the Payments tab if the user checked the **Create gateway for Standard Card Button** setting while **Advanced Card Processing** is active.

> Note: We have a custom notice placement between the PayPal logo and our settings tabs, like [these notices](https://github.com/woocommerce/woocommerce-paypal-payments/blob/fa76f1a73cdbdb21dd3b652ab460c03f68a3e766/modules/ppcp-applepay/src/Helper/AvailabilityNotice.php#L167) (not like the old one in the screenshot below, as that doesn’t yet use the “new” notice placement.

So I think that in the context of this issue, it’s reasonable to refactor the notice class to handle this additional scenario: [woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Notice/GatewayWithoutPayPalAdminNotice.php at fa76f1a73cdbdb21dd3b652ab460c03f68a3e766 · woocommerce/woocommerce-paypal-payments](https://github.com/woocommerce/woocommerce-paypal-payments/blob/fa76f1a73cdbdb21dd3b652ab460c03f68a3e766/modules/ppcp-wc-gateway/src/Notice/GatewayWithoutPayPalAdminNotice.php#L99)

- (old) Can’t use **Standard Card Button** or **Advanced Card Processing** while `Enable PayPal features for your store` is unchecked
- (old) Can’t use **Standard Card Button** while `Classic Checkout `location for smart buttons is disabled
- (new) Can’t use **Standard Card Button** while **Advanced Card Processing** is enabled 

So you can update the notice strings for the old scenario and give them the custom placement mentioned above (between the PayPal logo and our settings tabs).

About the notices, here is my **string proposal** to update the code in the existing place:

| **Old strings**  | **New Strings** | **Linked URL** |
| ------------------------ | ------------------------ | -------------------------|
| `'%1$s cannot be used without the PayPal gateway. <a href="%2$s">Enable the PayPal gateway</a>.',`  | `'%1$s cannot be used while PayPal features are disabled for your store. <a href="%2$s">Enable PayPal features for your store here</a>.',`  | `admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway`
| `'%1$s cannot be used without enabling the Checkout location for the PayPal gateway. <a href="%2$s">Enable the Checkout location</a>.',` | `'%1$s cannot be used without the Classic Checkout enabled in the Smart Button Locations. <a href="%2$s">Enable the Classic Checkout in the Smart Button Locations here</a>.'` | `admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway#field-button_style_heading` |
| **NOT PRESENT** | `'The <a href="%1$s">Standard Card Button</a> cannot be used while <a href="%2$s">Advanced Card Processing</a> is enabled.',` |`%1$s = 'admin.php?page=wc-settings&tab=checkout&section=ppcp-card-button-gateway'` <br> `%2$s = 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway&ppcp-tab=ppcp-credit-card-gateway'` |

### Steps to Test

1. Go to WooCommerce PayPal Payments plugin settings;
2. Standard Payments -> check "Create gateway for Standard Card Button";
3. Advanced Card Processing -> check "Enable Advanced Card Processing";
4. Go to Standard Card Button, observe the notice and grayed out enable button.


modules/ppcp-wc-gateway/src/Notice/GatewayWithoutPayPalAdminNotice.php
```diff
@@ -125,11 +138,20 @@ public function message(): ?Message {
...
+				$url2
+			);
+		} else {
+			$message = sprintf(
+				$text,
+				$name,
+				admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway' )
+			);
+		}
+
```
Dinamiko, line 154: `else` could be removed here, in general try to avoid using `else` unless it's strictly necessary:
```
		$message = sprintf(
			$text,
			$name,
			admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway' )
		);
		
		if ( $notice_type === self::NOTICE_DISABLED_CARD_BUTTON ) {
			$message = sprintf(
				$text,
				$url1,
				$url2
			);
		}
```

Dinamiko (changes requested):

Dinamiko (approved):

PR #2414, opened 2024-07-12
Google Pay: Fix the incorrect popup triggering (2645)

danieldudzic:
### Description

This PR fixes the incorrect Google Pay behavior upon user button click. Our async transaction info fetching was being interpreted by browsers as 'non-user initiated' resulting in a "`To make this purchase using Google Pay, open this website in a browser app, such as Chrome or Safari`" error.

This PR adds transaction info preloading, resolving the issue.

### Steps to Test

1. Activate Google Pay.
2. Add a product to the cart.
3. Try to pay with Google Pay, make sure you get the correct screen, and confirm that you can complete the order.

### Screenshots

|Before|After|
|-|-|
|Google Pay 2|Google_Pay|


modules/ppcp-googlepay/resources/js/GooglepayButton.js
```diff
@@ -218,15 +212,15 @@ class GooglepayButton {
     async onButtonClick() {
```
Dinamiko: Maybe remove also this `async` as it seems not using `await` inside.

Dinamiko (approved):

modules/ppcp-googlepay/resources/js/GooglepayButton.js
danieldudzic: Good catch!

PR #2422, opened 2024-07-16
Add inline notice to inform users about ACDC block Checkout support if the store uses a Classic Checkout setup (3283)

burdulixda:
### Description

To advertise for the new ACDC blocks integration, we need an inline notice in the ACDC settings to inform users with legacy store configurations that ACDC supports blocks. To avoid confusion, the notice should only appear when Fastlane is **NOT** enabled.

### Solution

image


modules/ppcp-wc-gateway/services.php
```diff
@@ -292,6 +293,34 @@ static function ( ContainerInterface $container ): Settings {
...
+						'woocommerce-paypal-payments'
+					),
+					esc_url( $checkout_page_link ),
+					esc_url( $instructions_link )
+				);
+			} else {
+				return '';
+			}
+
+			return '<div class="ppcp-notice ppcp-notice-success"><p>' . $notice_content . '</p></div>';
```
Dinamiko: I would remove the `else` here by inverting the condition:
```
			if ( CartCheckoutDetector::has_block_checkout() ) {
				return '';
			}

			$notice_content = sprintf(
			/* translators: %1$s: URL to the Checkout edit page. %2$s: URL to the WooCommerce Checkout instructions. */
				__(
					'<span class="highlight">Info:</span> The <a href="%1$s">Checkout page</a> of your store currently uses a classic checkout layout or a custom checkout widget. Advanced Card Processing supports the new <code>Checkout</code> block which improves conversion rates. See <a href="%2$s">this page</a> for instructions on how to upgrade to the new Checkout layout.',
					'woocommerce-paypal-payments'
				),
				esc_url( $checkout_page_link ),
				esc_url( $instructions_link )
			);

			return '<div class="ppcp-notice ppcp-notice-success"><p>' . $notice_content . '</p></div>';
```

Dinamiko (changes requested):

Dinamiko (approved):

PR #2491, opened 2024-08-06
Use client credentials for user tokens (3518)

Dinamiko:
This PR adds client credentials as authorization when requesting tokens, it also caches the tokens to avoid request rate limitations.

PR #2502, opened 2024-08-09
Saved PayPal account isn't displayed on checkout (3520)

Dinamiko:
In #2491 we introduced cache for access tokens used for adding `data-user-id-token` to the  JSSDK , this is causing issues when reusing the cached access token for example not displaying the saved payment token in the PayPal button.

This PR removes cache logic for access tokens used in `data-user-id-token`. It also ensures `data-user-id-token` is only added to guest users with WC Subscriptions product in the cart.

### Steps to reproduce
- Navigate to Shop → My account → Payment methods
- Save a PayPal account
- Add product to the cart
- Navigate to checkout

Actual result: Saved PayPal account isn’t displayed on checkout.
Expected Result: Saved PayPal account should be displayed on checkout.

### DIFF
--- a/modules/ppcp-api-client/services.php
+++ b/modules/ppcp-api-client/services.php
@@ -1669,8 +1669,7 @@ return array(
 		return new UserIdToken(
 			$container->get( 'api.host' ),
 			$container->get( 'woocommerce.logger.woocommerce' ),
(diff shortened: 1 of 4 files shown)


stracker-phil (approved):

danieldudzic (approved):
LGTM

PR #2509, opened 2024-08-13
Use SVG for APM gateway icons (3503)

burdulixda:
### Description

**Problem:**

PNG icons are scaled by the storefront CSS (theme). When using a different theme, the custom gateway icons for Apple Pay and Google Pay are way too large out-of-the-box

**Solution:**

Switch the icons from PNGs to SVG icons with a defined width and height - i.e., apply the same logic as we use for the credit card icons


modules/ppcp-googlepay/resources/css/styles.scss
```diff
@@ -17,3 +17,9 @@
 #ppc-button-ppcp-googlepay {
 	display: none;
 }
+
+label[for='payment_method_ppcp-googlepay'] {
+	img {
+		max-height: 25px;
+	}
+}
```
stracker-phil: We could consider using a plugin-wide CSS rule for all payment icons, instead of using different CSS rules for each gateway.

**Suggestion**

Undo this change, and instead modify `hosted-fields.scss` and change the first rule to:
```scss
#payment ul.payment_methods [class*=" payment_method_ppcp-"] label img {
  max-height: 25px;
  display: inline-block;
}
```

stracker-phil (approved):

PR #2556, opened 2024-09-03
Require PHP 7.4+, WP 6.3+, WC 6.9+ (1773)

AlexP11223:
Bumped the versions in the readmes, plugin file, DDEV, composer.json, removed PHP 7.2-7.3 in GHA.

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -1,7 +1,7 @@
 name: woocommerce-paypal-payments
 type: php
 docroot: .ddev/wordpress
-php_version: "7.2"
+php_version: "7.4"
 webserver_type: apache-fpm
 router_http_port: "80"
 router_https_port: "443"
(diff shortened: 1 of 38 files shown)


stracker-phil:
Not sure, if those values are actually used, but two references to WP 5.3 and PHP 7.2 found here:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/0b74af10612b11a72ae27831ad2c6babcaf3a1ce/phpcs.xml.dist#L6-L7

NB: In my tests, those values were not used by local `phpcs`, and I could not find a reference in a GitHub action.

stracker-phil:
Should we also update the default values in the `FilePathPluginFactory` class, just for completeness?

https://github.com/woocommerce/woocommerce-paypal-payments/blob/0b74af10612b11a72ae27831ad2c6babcaf3a1ce/src/FilePathPluginFactory.php#L85-L86

I think it's optional, but since we declare defaults, it makes sense to set them to the current requirements

stracker-phil (approved):
Looks good, works well 👏

PR #2566, opened 2024-09-04
Add title, description and gatewayId to the express payment method

alexflorisca:
This PR ads `title`, `description` and `gatewayId` props to express payment methods to be displayed in the editor. These are used to display the available express payment methods to a merchant, once they click on the express checkout block.

This is parallel work with https://github.com/woocommerce/woocommerce/pull/50791 and the `feature/express-payment-improvements` branch should be used to test. However this should also work with the current release of WC, and can be safely merged and released before the work in Core is released.

If these props are no provided, the `name` is used instead and the list will look like this:

Screenshot 2024-09-04 at 16 31 45

After this PR, the list should look like this:

Screenshot 2024-09-04 at 16 29 37

The link for the PayPal payment method should take you to the PayPal settings.

#### Testing instructions

1. Visit the checkout page in the editor
2. Click on the express payment block
3. See on the right hand sidebar that the list of Express Payment methods shows, and that the PayPal method has a title, description and the link goes to the PayPal settings page.
4. Smoke test with current WC version and make sure the express method still works as expected.

### Changelog Entry

> Add additional `title`, `description` and `gatewayId` props to the express payment method for displaying in the editor


alexflorisca:
Hey folks, any movement on this? Should be a simple one to test and merge, be great to see it in the next release as the express checkout improvements will land in WC 9.4 in a months time

package-lock.json
AlexP11223, line 1: Seems to be committed accidentally? I think this project uses `yarn`.

Dinamiko (approved):

InpsydeNiklas:
Hey @alexflorisca, this will be included in the 2.9.1 update which is scheduled for the next week: https://github.com/woocommerce/woocommerce-paypal-payments/releases

PR #2448, opened 2024-07-22
Receive button properties from the Checkout Block (3202)

danieldudzic:
This is some exploratory work to explore how we can synchronise the express payment button styles better to create more cohesiveness in the Express Checkout block. This PR is a part of https://github.com/woocommerce/woocommerce/pull/47899. It updates the express payment method to receive `buttonAttributes` property from the checkout block, and override the `height` property if available.

### Testing instructions

- Run woocommerce locally and check out the `try/poc-express-checkout-button-styles` branch
- Make sure the environment is right for one of the express checkout buttons to be visible (either Apple or Google Pay)
- Add some items to the cart and visit the cart page (with the cart block)
- Ensure the express payment method is visible there
- Click "Proceed to checkout"
- Ensure the express payment method is visible there too
- In a new tab, open the checkout page in the editor.
- Click on the express checkout area and change the button size to "Small"
- Refresh the checkout page in the front-end
- You should notice the button change size to a height of 40px

Please note that is not possible to set "Large" size as PayPal only allows values between [25px and 55px](https://developer.paypal.com/docs/multiparty/checkout/standard/customize/buttons-style-guide/#size):
```
Error: Expected style.height to be between 25px and 55px - got 56px
```

### Comment from @Dinamiko 

Hi @alexflorisca, just to let you know that we have created this PR for working on https://github.com/woocommerce/woocommerce/pull/47899 POC integration, for now it only checks if `buttonAttributes` exist and if so it gets the button height value from there.

I noticed that is not currently possible to set "Large" size because there is a button size limitation on PayPal.

cc. @InpsydeNiklas

PR #2570, opened 2024-09-05
Do not add pay later button in editor (3535)

### LINKED ISSUE #2515: Multiple PayPal buttons are rendered in one express payment container (3535)
### Describe the Bug

In the express checkout section of the Checkout block, PayPal renders 2 buttons instead of 1, if both PayPal and PayLater are enabled. Only 1 button should be rendered with each call to `registerExpressPaymentMethod`.

### To Reproduce

1. Enable both PayPal and Pay Later buttions
2. Go to the Checkout block in the editor
3. See that 2 buttons are rendered vertically in one grid item, breaking the layout. 

### Screenshots

Screenshot 2024-08-14 at 10 00 09

### Expected Behavior

I would expect only 1 button to be rendered per slot.

### Actual Behavior

2 (or possible more?) buttons are rendered in the same grid slot

### Environment

WC 9.2 - but test with latest WC trunk if you can, it includes some updates to the express checkout area that make this bug more prominent.

AlexP11223:
In 4e34332c0bb47 we removed all APMs in editor because it is difficult to determine which APMs are available when registering buttons separately in editor, `canMakePayment` does not work there.

But `fundingSource` was not set to `paypal`, so it could still render at least Pay Later button (in the same express payment method cell, breaking layout when there are other methods).

Fixed that, now only one PayPal button can be shown in editor.


stracker-phil (approved):

alexflorisca:
Hey there, just checking up on this PR as it was ready to merge a month ago? Is there a reason it hasn't yet? \cc @InpsydeNiklas

InpsydeNiklas:
@alexflorisca We found some remaining edge-case issues where incorrect buttons were displayed in the editor, which seemed more difficult to resolve. However, we merged this now and will include it with #2448 in the upcoming 2.9.4 update.

alexflorisca:
Thanks @InpsydeNiklas. Just to be annoying, I think the best UX would be to render the PayLater button as a separate button, if it's enabled in the backend. Otherwise the merchant and shopper UX will be different!

Editor: 
Screenshot 2024-11-01 at 16 40 37

Frontend:
Screenshot 2024-11-01 at 16 40 10

(You can see similar discrepancies with WooPayments too).

It would be great to address this too.

InpsydeNiklas:
@alexflorisca I completely agree! Disabling the non-PayPal buttons (which also includes Venmo) is more of a temporary workaround to prevent the layout in the editor from breaking. 
The problem is that the buttons rendered in the editor are using the `PayPalButtons` component from the PayPal React library. However, as far as I am aware, this component currently does not allow us to split the buttons into individual ones, like how it works in the frontend. We continue to look for ways to make it work, but it won't be ready for this release cycle, so we settled on this workaround for now.

AlexP11223:
Splitting may be possible (or we can try again to use the same library as in frontend), but the main problem is determining which buttons are available (e.g. for the merchant/shop country). 

In frontend we use `canMakePayment` with async promises, but in editor it is not called for some reason. https://github.com/woocommerce/woocommerce-paypal-payments/blob/0b6e04d0688e063327127b1d010500c15faff860/modules/ppcp-blocks/resources/js/checkout-block.js#L826-L845

So unless there is a way to get something like `canMakePayment` in editor, we would need to figure out some other way to determine the buttons availability earlier (before the express methods registration, e.g. on the server side) without the PayPal JS SDK.

alexflorisca:
@AlexP11223 I see, we don't currently run the `canMakePayment` function in the editor, but I think it might be wise to do so if we want to create parity with the front-end in terms of which buttons are displayed. I've brought it up with my team for discussion and I'll let you know what we decide.

alexflorisca:
After talking it through with a colleague, we think you should be able to do this outside of the `canMakePayment` method. You can pass a different component as the `edit` prop when you initialise the express payment method, so this should be able to handle the process of initialising the buttons without `canMakePayment` being called. 

It's also  worth noting that in the editor, we have preview data, which is always a US based store with USD as the currency. So this may or may not affect what you render. The ideal situations here is to match the settings configured by the merchant ( e.g. number of buttons configured in the PayPal settings) and display these in the editor, ignoring shop location/country and bypassing any other conditional logic for wether the buttons load or not.

stracker-phil:
Hi @alexflorisca thanks for those inputs. Those sound like reasonable changes to me, so I've gone ahead and created an internal issue for your suggestion, and we'll discuss it in the next planning session.

PR #2573, opened 2024-09-09
Check that get_the_ID is valid before using

AlexP11223:
There was a report that sometimes `get_the_ID` returns an invalid/string value, so added an `is_numeric` check and cast to avoid type errors.


Narek13 (approved):
LGTM, but also couldn't reproduce

PR #2574, opened 2024-09-09
Improve card fields hiding

### LINKED ISSUE #2567: Duplicated Credit Card Fields (3657)
### Describe the Bug

On website www.waterwayroutes.co.uk

After upgrade from 2.8.3 to 2.9 the Credit Card number, expiry and CVV fields are duplicated and a customer has to enter their information in the second of each field.

Reverting to 2.8.3 solves the problem.  No other changes were made to the website at the time.

### To Reproduce

Add a product to the shopping cart
Go through the checkout process
Choose Card Payment and not PayPal payment
The card fields (Number, Expiry and CVV)  are shown duplicated, one above the other.

### Expected Behavior

The card fields should not be duplicated

### Environment

 - WordPress = 6.6.1 running Hueman theme
 - WooCommerce Version = 9.2.3
 - Plugin Version 2.9 is faulty, version 2.8.3 works fine
 - Browser [e.g. Chrome, Safari] and Version  Chrome
 - Any other plugins installed = WooCommerce Extended Coupon Features FREE and WooCommerce Shipping & Tax

### Additional Details

  ```
  ```

</details>

AlexP11223:
Hiding the fields with `!important` instead of just setting `hidden`, this should fix #2567, at least the case from the first comment.

### DIFF
--- a/modules/ppcp-card-fields/resources/js/Render.js
+++ b/modules/ppcp-card-fields/resources/js/Render.js
@@ -1,6 +1,20 @@
 import { cardFieldStyles } from './CardFieldsHelper';
 import { hide } from '../../../ppcp-button/resources/js/modules/Helper/Hiding';
 
+function renderField( cardField, inputField ) {
+	if ( ! inputField || inputField.hidden || ! cardField ) {
+		return;
+	}
+
+	// Insert the PayPal card field after the original input field.
+	const styles = cardFieldStyles( inputField );
+	cardField( { style: { input: styles } } ).render( inputField.parentNode );
+
+	// Hide the original input field.
+	hide( inputField, true );
+	inputField.hidden = true;
+}
+
 export function renderFields( cardFields ) {
 	const nameField = document.getElementById(
 		'ppcp-credit-card-gateway-card-name'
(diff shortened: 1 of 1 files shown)


modules/ppcp-card-fields/resources/js/Render.js
```diff
@@ -9,7 +10,7 @@ export function renderFields( cardFields ) {
 		cardFields
 			.NameField( { style: { input: styles } } )
 			.render( nameField.parentNode );
-		nameField.hidden = true;
```
stracker-phil, line 12: We should keep this line, so the if-condition works as expected: Each field should be hidden once, and not trigger a `field.render()` call again:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/5237ebc0ed39997fd2abb27e95e97c2dbe78906f/modules/ppcp-card-fields/resources/js/Render.js#L7

stracker-phil (changes requested):
Using `hide()` should hopefully solve the display problem, but it must be called additionally, not instead of using `hidden = true` to preserve the module logic

stracker-phil (approved):

harmansingh03:
Hey there,
I have see where was the problem and solved it so can i have some credits for resolving the issue?
Thanks!!

PR #2577, opened 2024-09-10
Set email when creating order for express payment

AlexP11223:
Looks like the email was not set in `WooCommerceOrderCreator`, so for guest orders via express checkout with disabled confirmation it was remaining empty (for logged in customers it was setting email to the WP user email if empty during saving).

So now setting the order email to the email from `WC()->customer` (retrieving `email` instead of `billing_email` because it seems to be the WC behavior) , or to the PayPal payer email if empty.

### Steps to Test

GIVEN the PayPal shipping callback is enabled
AND the customer is a guest user
WHEN completing a PayPal express payment
THEN the WooCommerce order displays the buyer's PayPal email address in both the regular Email and PayPal Email fields

GIVEN the PayPal shipping callback is enabled
AND the customer is a logged-in user with a saved email address in their profile
WHEN completing a PayPal express payment
THEN the WooCommerce order displays the email address from the buyer's WooCommerce profile in the regular Email field
AND the PayPal account email address in the PayPal Email field

### DIFF
--- a/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
+++ b/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
@@ -12,6 +12,7 @@ namespace WooCommerce\PayPalCommerce\Button\Helper;
 use Exception;
 use RuntimeException;
 use WC_Cart;
+use WC_Customer;
 use WC_Data_Exception;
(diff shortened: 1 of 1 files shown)


Narek13 (approved):
Thanks

PR #2613, opened 2024-09-17
Fix the step counter in block checkout in the initial Fastlane stage (3693)

stracker-phil:
### Description

This PR adds logic to recalculate the step numbering in the block checkout during the initial Fastlane stage.

In that phase, only two steps are visible: The email address and the payment methods. In the default step order, those steps are numbered "1" and "5" (when shipping & shipping options are available), which looks weird.

### Solution

The step numbers are already implemented by WooCommerce using CSS counter-attributes. What is missing is a recalculation of the CSS counter value after hiding or showing steps.

A new JS function briefly adds a CSS class to the body, which changes a counter-attribute using `!important` - this triggers the desired recalculation of the counter values.

### Screenshots

| Before | After |
|---|---|
|  |  |

### DIFF
--- a/modules/ppcp-axo-block/resources/css/gateway.scss
+++ b/modules/ppcp-axo-block/resources/css/gateway.scss
@@ -278,3 +278,10 @@ a.wc-block-axo-change-link {
 #shipping-fields .wc-block-components-checkout-step__heading {
 	display: flex;
 }
+
+// 11. Fix Step Numbering
+body.ppcp-recalculate {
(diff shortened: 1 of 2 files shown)


danieldudzic:
@stracker-phil Tested and this works!

I have experimented a bit, and I believe I was able to fix the issue with a more light-weight (non-JS) approach.  https://github.com/woocommerce/woocommerce-paypal-payments/pull/2617 - please give it a try, and let me know if it fixes the issue for you too.

PR #2638, opened 2024-09-24
Axo Blocks: Localize strings

danieldudzic:
### Description

This PR adds string localization to the block checkout Fastlane related code.

### Solution

The strings are being localized on the frontend via `{ __ } from '@wordpress/i18n'`


Narek13 (approved):
LGTM, thanks

AlexP11223 (approved):

PR #2647, opened 2024-09-26
Axo: Load classic and block integration modules by default (3738)

danieldudzic:
### Description

This PR adds default loading of Axo modules.

### Testing instructions

1. Make sure Fastlane loads both for Classic Checkout and Block Checkout without the `PCP_AXO_ENABLED` feature flag being set.

### DIFF
--- a/modules.php
+++ b/modules.php
@@ -84,7 +84,7 @@ return function ( string $root_dir ): iterable {
 
 	if ( apply_filters(
 		'woocommerce.feature-flags.woocommerce_paypal_payments.axo_enabled',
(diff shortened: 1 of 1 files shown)


stracker-phil:
Looks solid
1. If `PCP_AXO_ENABLED` is not set: `getenv()` returns `false`, the _not-'0'_ condition evaluates to true, AXO is enabled.
2. If `PCP_AXO_ENABLED` is set to '1': The condition will evaluate to true, enabling AXO.
3. If `PCP_AXO_ENABLED` is set to '0': The condition will evaluate to false, disabling AXO.

stracker-phil (approved):

PR #2648, opened 2024-09-26
Axo Block: Load ACDC for logged in users in the Block Checkout (3739)

danieldudzic:
### Description

This PR adds ACDC loading for logged in users in the Block Checkout.

### Testing instructions

1. Enable ACDC.
2. Navigate to block Checkout as a logged in user.
3. Confirm that the ACDC gateway is available.

### DIFF
--- a/modules/ppcp-blocks/src/BlocksModule.php
+++ b/modules/ppcp-blocks/src/BlocksModule.php
@@ -69,8 +69,8 @@ class BlocksModule implements ServiceModule, ExtendingModule, ExecutableModule {
 			function( PaymentMethodRegistry $payment_method_registry ) use ( $c ): void {
 				$payment_method_registry->register( $c->get( 'blocks.method' ) );
 
-				// Include ACDC in the Block Checkout only in case Axo doesn't exist or is not available.
-				if ( ! $c->has( 'axoblock.available' ) || ! $c->get( 'axoblock.available' ) ) {
+				// Include ACDC in the Block Checkout only in case Axo doesn't exist or is not available or the user is logged in.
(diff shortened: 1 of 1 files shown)


modules/ppcp-blocks/src/BlocksModule.php
```diff
@@ -69,8 +69,8 @@ function () {
 			function( PaymentMethodRegistry $payment_method_registry ) use ( $c ): void {
 				$payment_method_registry->register( $c->get( 'blocks.method' ) );
 
-				// Include ACDC in the Block Checkout only in case Axo doesn't exist or is not available.
-				if ( ! $c->has( 'axoblock.available' ) || ! $c->get( 'axoblock.available' ) ) {
+				// Include ACDC in the Block Checkout only in case Axo doesn't exist or is not available or the user is logged in.
+				if ( ! $c->has( 'axoblock.available' ) || ! $c->get( 'axoblock.available' ) || is_user_logged_in() ) {
```
stracker-phil, line 73: I think it makes more sense to move the `is_user_logged_in()` check to the `'axoblock.available'` service:

instead of hard-coded return true, it could look like this:

```php
	// If AXO Block is configured, onboarded and available for the current user.
	'axoblock.available' => static function ( ContainerInterface $container ) : bool {
		return ! is_user_logged_in();
	},
```

stracker-phil: @danieldudzic and me came to the conclusion that the `<gateway>.available` service always checks for a merchant-level-availability. Performing user-specific checks should be done outside that service, that's why we'll go with this proposed condition

stracker-phil (approved):

PR #2653, opened 2024-09-27
Fastlane card preview UI for classic checkout

stracker-phil:
### Description

This PR improves the payment-method styling for the saved Fastlane card in Ryan flow.

**Details**
- The payment method details are styled to look like a real credit card
- The preview is responsive
- Only affects the classic checkout

### Screenshots

| Before | After |
|---|---|
|  |  |

### DIFF
--- a/modules/ppcp-axo/resources/css/styles.scss
+++ b/modules/ppcp-axo/resources/css/styles.scss
@@ -93,6 +93,11 @@
 		float: right;
 	}
 
+	.ppcp-card-icon {
+		width: auto;
+		height: 25px;
+	}
+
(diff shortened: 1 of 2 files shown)


danieldudzic (changes requested):
We should sync the design for the card across classic and block checkout.

Which one should be chosen I don't know 😅 We should focus on making it as compatible as possible across different themes.

Would be good to receive some guidance from a designer.

| Block Checkout | Classic Checkout |
|---|---|
|block|classic|

cc: @InpsydeNiklas

danieldudzic (approved):

PR #2661, opened 2024-10-01
Fix Google Pay preview config and style

AlexP11223:
Looks like something went wrong with merging somewhere and these changes ended up only in the unused class https://github.com/woocommerce/woocommerce-paypal-payments/commit/722289a54514e6f1266b6f13c4bab39601af4d49

Also there were some changes in the configuration function, the preview in the settings was broken because `apiConfig` was not applied.

Removed the class and moved the changes to `Preview/GooglePayPreviewButton.js`


stracker-phil (commented):
The code changes are great, thanks for cleaning this up.

Only thing to adjust is the code-style - `eslint --fix` should do the trick (tabs for indentation + linebreak between `{ /**`)

stracker-phil (approved):

PR #2667, opened 2024-10-03
Fix multi-currency support

AlexP11223:
After migrating to modularity the services are properly cached, which causes problems with plugins overriding the WC currency because we cache it too early (both the `api.shop.currency` service and the classes using it, such as `ItemFactory`).

So I extracted it into the `CurrencyGetter` class, and changed the code to call its `get()` as late as possible.


Dinamiko (approved):

davidhealey:
Hi,

I was just to about to open a bug report because I noticed in releases after 2.9.0 the wrong currency is passed to PayPal.

This is with 2.9.0 - the value is correct
Screenshot_2024-10-04_16-20-35

This is with 2.9.2 - the value should be $102
Screenshot_2024-10-04_16-19-00

Do I need to open a bug report or is this PR dealing with the issue I described?

PR #2675, opened 2024-10-08
Add a new namespaced script loader (3771)

danieldudzic:
### Description

This PR attempts to remedy various issues stemming from multiple PayPal scripts conflicting.

The new loader, `PayPalScriptLoading.js`, requires passing a namespace when loading an instance of the PayPal script. It also tries to separate the processing of the config options and performing the actual script loading.

It also migrates the following to the new loader:
- Axo and Axo Block
- Express payment buttons
- Google Pay

### Steps to test
1. Test Fastlane in the Classic and Block checkouts. Ensure the script loads correctly and doesn't cause conflicts with the smart buttons. (reload the page a couple of times)
2. Ensure Google Pay loads correctly in all contexts. 

### Screenshots

N/A


Dinamiko (approved):

stracker-phil (approved):

modules/ppcp-button/resources/js/modules/Helper/PayPalScriptLoading.js
```diff
@@ -0,0 +1,101 @@
+import { loadScript } from '@paypal/paypal-js';
+import dataClientIdAttributeHandler from '../DataClientIdAttributeHandler';
+import widgetBuilder from '../Renderer/WidgetBuilder';
+import { processConfig } from './ConfigProcessor';
+
+const loadedScripts = new Map();
+const scriptPromises = new Map();
```
AlexP11223, line 7: I think this will not be fully global. We import this file in other modules and then webpack builds separate JS assets. So maybe e.g. putting it into `window.` would be more correct. But not sure if matters currently.

PR #2691, opened 2024-10-11
Buttons loading on single product page when the subscription support is not declared  (1815)

inpsyde-maticluznar:
### Description
This PR adds a condition that checks if vaulting is enabled before rendering buttons.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
+++ b/modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
@@ -234,6 +234,11 @@ class SingleProductBootstap {
 			this.errorHandler
 		);
 
+        if(!this.gateway.vaultingEnabled){
+            return;
+        }
+
+
 		if (
 			PayPalCommerceGateway.data_client_id.has_subscriptions &&
 			PayPalCommerceGateway.data_client_id.paypal_subscriptions_enabled
--- a/modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
+++ b/modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
@@ -233,11 +233,15 @@ class SingleProductBootstap {
 			this.form(),
 			this.errorHandler
 		);
-
-        if(!this.gateway.vaultingEnabled){
-            return;
-        }
-
+		if (
+			! this.gateway.vaultingEnabled &&
+			[ 'subscription', 'variable-subscription' ].includes(
+				this.gateway.productType
+			) &&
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/resources/js/modules/ContextBootstrap/SingleProductBootstap.js
```diff
@@ -234,6 +234,11 @@ class SingleProductBootstap {
 			this.errorHandler
 		);
 
+        if(!this.gateway.vaultingEnabled){
```
AlexP11223: I think the buttons need to be removed only for subscription products, not all products?

Also not sure if just checking `vaultingEnabled` is correct since the subscription settings became more complex.

inpsyde-maticluznar:
After following the provided suggestions, I created a new commit. 

When displaying or not displaying buttons we consider the following properties:
- Product type `'subscription', 'variable-subscription'`
- Subscriptions Mode in PayPal plugin settings
- Manual Renewal Payments in WooCommerce subscription settings

Dinamiko (approved):

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1318,8 +1318,22 @@ public function script_data(): array {
 			'should_handle_shipping_in_paypal'        => $this->should_handle_shipping_in_paypal && ! $this->is_checkout(),
 			'needShipping'                            => $this->need_shipping(),
 			'vaultingEnabled'                         => $this->settings->has( 'vault_enabled' ) && $this->settings->get( 'vault_enabled' ),
+			'productType'                             => null,
+			'manualRenewalEnabled'                    => '',
```
AlexP11223: Maybe should be `false`. It may be converted into string when passing to JS, but I think it would be better to keep it as a normal boolean here (like all other values) to avoid confusion.

modules/ppcp-button/src/Assets/SmartButton.php
```diff
@@ -1318,8 +1318,22 @@ public function script_data(): array {
...
 
+		if ( is_product() ) {
+			$product = wc_get_product( get_the_ID() );
+			if ( is_a( $product, \WC_Product::class ) ) {
+				$localize['productType'] = $product->get_type();
+			}
+		}
+
+		if ( class_exists( '\WCS_Manual_Renewal_Manager' ) ) {
+			/** @psalm-suppress UndefinedClass */
```
AlexP11223: Also there is the annoying WP phpcs error about missing description :)

Another way could be to add the class stub into https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/.psalm/wcs.php

PR #2697, opened 2024-10-14
Ensure setting exist before getting it (3780)

Dinamiko:
New plugin installation throw error because setting `axo_enabled` does not exist.

### Steps to reproduce
- Install plugin on a fresh WordPress installation (or delete `woocommerce-ppcp-settings` option from database)
- Error is thrown
```
Fatal error: Uncaught WooCommerce\PayPalCommerce\WcGateway\Exception\NotFoundException in /var/www/html/wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Settings/Settings.php:89
```

This PR ensures setting exist before using it.


danielhuesken (approved):
Looks good

PR #2699, opened 2024-10-14
Subscriptions migrated from PPEC fail to renew with PCP after Vault V3 implementation in version 2.5.0 (2778)

Dinamiko:
As of 2.5.0 subscription renewals that were migrated from the PayPal Checkout Gateway are not processing successfully. 

### Steps To Reproduce
- Enable the PayPal Checkout (PPEC) plugin
- Ensure HPOS is disabled
- Create a subscription using the PPEC plugin
- Disable PPEC
- As admin try to renewal the subscription manually 
    - Observe that the status remains "Pending Payment" without any log indicators for the renewal process.

This PR fixes the issue by using `{$wpdb->prefix}wc_orders` instead of `$wpdb->wc_orders` which does not exist for in non HPOS configurations. It also adds `ppec_paypal` payment to vault v2 order renewal handler.


modules/ppcp-compat/src/CompatModule.php
```diff
@@ -79,6 +79,13 @@ public function run( ContainerInterface $c ): bool {
 			$this->initialize_wc_bookings_compat_layer( $logger );
 		}
 
+		add_action(
+			'woocommerce_paypal_payments_gateway_migrate',
+			function() {
+				delete_transient( 'ppcp_has_ppec_subscriptions' );
+			}
+		);
```
Narek13: This can be simplified and additionally add the `static` for the callback

```suggestion
add_action( 'woocommerce_paypal_payments_gateway_migrate', static fn() => delete_transient( 'ppcp_has_ppec_subscriptions' ) );
```

modules/ppcp-wc-subscriptions/src/RenewalHandler.php
```diff
@@ -406,7 +406,7 @@ private function process_order( \WC_Order $wc_order ): void {
 				return;
 			}
 
-			if ( $wc_order->get_payment_method() === PayPalGateway::ID ) {
+			if ( $wc_order->get_payment_method() === PayPalGateway::ID || $wc_order->get_payment_method() === 'ppec_paypal' ) {
```
Narek13: We call the `get_payment_method()` 6 times inside the same `process_order` method, so does it make sense to store it inside variable on line 259 for example `$payment_method = $wc_order->get_payment_method();` and then use the `$payment_method` variable in all these `if` checks?

Narek13 (approved):
Thanks @Dinamiko. Left a couple of minor suggestions which don't prevent me to approve the PR

Dinamiko:
@Narek13 thanks for the feedback!

inpsyde-maticluznar (approved):

PR #2701, opened 2024-10-15
Fix displaying of HTML tags in product title on choosing a product for tracking (2801)

danielhuesken:
Fix the display of HTML tags in product titles when choosing a product to track at an order.


Narek13 (approved):
Thanks

PR #2702, opened 2024-10-16
Payment with OXXO cause continuation state for next payment (2395)

inpsyde-maticluznar:
### Description

Clear session before redirecting OXXO.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
@@ -112,6 +112,7 @@ class ReturnUrlEndpoint {
 		}
 
 		if ( $wc_order->get_payment_method() === OXXOGateway::ID ) {
+			$this->session_handler->destroy_session_data();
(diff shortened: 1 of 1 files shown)


AlexP11223:
I think it will break https://github.com/woocommerce/woocommerce-paypal-payments/pull/1840#issuecomment-1813986944 so it needs to be solved somehow differently.

A quick fix could be to check if it was OXXO, though I am not sure that I understand the issue, why the session is not destroyed correctly for OXXO like for other methods?

inpsyde-maticluznar:
> I think it will break [#1840 (comment)](https://github.com/woocommerce/woocommerce-paypal-payments/pull/1840#issuecomment-1813986944) so it needs to be solved somehow differently.
> 
> A quick fix could be to check if it was OXXO, though I am not sure that I understand the issue, why the session is not destroyed correctly for OXXO like for other methods?

Thank you @AlexP11223 . I will try to compare the process with other payment methods and see if there is a better solution than a quick fix you mentioned. Until then, the PR is in draft.

Dinamiko (approved):

PR #2710, opened 2024-10-18
Set "Sold individually" only for subscription connected to PayPal (3795)

AlexP11223:
"Sold individually" was set on saving for all products, not just subscriptions connected to PayPal. Added checking of `$_POST['_ppcp_enable_subscription_product']`, seems to fix that.


AlexP11223:
ah, did not notice that it was not converted to `bool`, added `wc_string_to_bool` and early return to avoid such mistakes here in the future.

Dinamiko (approved):

PR #2711, opened 2024-10-18
Axo: Integrate shipping restrictions (3715)

danieldudzic:
### Description

This PR integrates the WooCommerce shipping configuration into the Fastlane checkout (both Classic and Block-based).

### Steps to test

1. Configure a Shipping Zone with a shipping method just for one US state, for example, Alabama.
2. Add a shipping method to the `Rest of the world` zone.
3. Test with the shipping method of the `Rest of the world` zone disabled.
4. In classic and block checkouts with Fastlane, test using `test@awesome.com` and ensure only the Alabama address is accepted.
5. Re-enable the shipping method for the `Rest of the world` zone. Make sure all addresses are valid.

### Screenshots

|Settings|Block Checkout|
|-|-|
|WooCommerce_settings_‹_WooCommerce_PayPal_Payments_—_WordPress|Block_Checkout_–_WooCommerce_PayPal_Payments|

### DIFF
--- a/modules/ppcp-axo-block/resources/js/hooks/useAllowedLocations.js
+++ b/modules/ppcp-axo-block/resources/js/hooks/useAllowedLocations.js
@@ -1,0 +1,21 @@
+import { useMemo } from '@wordpress/element';
+
+/**
+ * Custom hook returning the allowed shipping locations based on configuration.
+ *
+ * @param {Object}          axoConfig                            - The AXO configuration object.
+ * @param {Array|undefined} axoConfig.enabled_shipping_locations - The list of enabled shipping locations.
+ * @return {Array} The final list of allowed shipping locations.
+ */
+const useAllowedLocations = ( axoConfig ) => {
+	return useMemo( () => {
+		const enabledShippingLocations =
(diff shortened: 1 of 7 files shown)


modules/ppcp-axo-block/resources/js/hooks/useAllowedLocations.js
```diff
@@ -0,0 +1,21 @@
+import { useMemo } from '@wordpress/element';
+
+/**
+ * Custom hook returning the allowed shipping locations based on configuration.
+ *
+ * @param {Object}          axoConfig                            - The AXO configuration object.
+ * @param {Array|undefined} axoConfig.enabled_shipping_locations - The list of enabled shipping locations.
+ * @return {Array} The final list of allowed shipping locations.
+ */
+const useAllowedLocations = ( axoConfig ) => {
```
Dinamiko, line 10: why not just passing `axoConfig.enabled_shipping_locations` as parameter here? it seems the whole `axoConfig` is not used inside this method.

Also not sure about the second `@param` in the docblock, while it is used inside the method, it's not ipassed as parameter.

Dinamiko (approved):

PR #2713, opened 2024-10-18
Create entrypoint to render basic react app instead of regular plugin settings (3792)

Dinamiko:
This PR adds a new module (`ppcp-settings`) for the new settings React application. It also introduces [@wordpress/scripts](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/) for building the assets.

The module is disabled by default and can be enabled by setting the value of `PCP_SETTINGS_ENABLED` constant, for example via a filter:
```
add_filter('woocommerce.feature-flags.woocommerce_paypal_payments.settings_enabled', '__return_true');
```
Once enabled it replaces the PayPal settings page gateway with the React application:

Captura de pantalla 2024-10-18 a las 17 24 26


inpsyde-maticluznar (commented):
@Dinamiko can you check if this is a better way of registering pages? I found this in WooCommerce Payments plugin.

- Create page components, example: https://github.com/Automattic/woocommerce-payments/blob/develop/client/connect-account-page/index.tsx#L73

- Use the following filter: `woocommerce_admin_pages_list`, example: https://github.com/Automattic/woocommerce-payments/blob/develop/client/index.js#L37

- Push pages to the pages list, example: https://github.com/Automattic/woocommerce-payments/blob/develop/client/index.js#L48-L57

- Register pages in the PHP, example:
https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-payments-admin.php

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -0,0 +1,95 @@
...
+					$style_asset_file['version']
+				);
+
+				wp_enqueue_style( 'ppcp-admin-settings' );
+			}
+		);
+
+		add_action(
+			'woocommerce_paypal_payments_gateway_admin_options_wrapper',
+			static function( PayPalGateway $gateway ) {
```
Narek13: The param is not used.

```suggestion
			static function(): void {
```

inpsyde-maticluznar (approved):

Narek13 (approved):
Thanks

PR #2714, opened 2024-10-18
Use basic redirect gateway when checkout smart buttons disabled (3347)

AlexP11223:
Now when the smart buttons are disabled for checkout the basic redirect gateway is added instead (like the `woocommerce_paypal_payments_use_place_order_button` filter).

In block checkout it was already added by default unless disabled via the `woocommerce_paypal_payments_blocks_add_place_order_method` filter.

Also disabling this gateway when there is a subscription since it does not properly.

When the smart buttons are disabled for both classic checkout and block express checkout, the settings about subscriptions are hidden.

In `settings.js` I tried to refactor the `toggleInputsBySelectedLocations` function a bit since it is too big and difficult to understand.


Narek13 (approved):
Thanks, LGTM

PR #2721, opened 2024-10-21
Prevent plugin use for "Send only" countries (3790)

inpsyde-maticluznar:
### Description

This will show an error message to merchants configuring the WC store to send-only country. The message is displayed before onboarding as an inline notice and after onboarding as a classic notice.


Dinamiko (approved):

inpsyde-maticluznar:
@Dinamiko I put `SendOnlyCountryNotice` in the array whose items are checked against `instanceof GatewayWithoutPayPalAdminNotice`. This caused an issue that I fixed in the latest commit. Please take a look: https://github.com/woocommerce/woocommerce-paypal-payments/pull/2721/commits/47b7c66f53d17682c9f5a11670ce95af606aae15

Thanks!

PR #2722, opened 2024-10-21
Apple Pay button disappears on Classic Checkout (3644)

stracker-phil:
### Description

This PR refactors the Apple Pay gateway to use the latest code patterns, which are already implemented for Google Pay.

The only visible difference of this PR should address the following bug: When using Apple Pay as a separate gateway (classic checkout), the payment button disappeared after editing the billing address. It re-appeared after toggling to a different gateway and then selecting Apple Pay again.

### Changes

The main change is extending the `ApplePayButton` class from `PaymentButton`. This base class was introduced for Google Pay to combine reusable gateway logic.

To make the changes easier to review, the migration was done in 40+ commits -- all marked with the 🚧 icon.


modules/ppcp-applepay/resources/js/boot-block.js
```diff
@@ -12,26 +13,22 @@ const ppcpConfig = ppcpData.scriptData;
 
 const buttonData = wc.wcSettings.getSetting( 'ppcp-applepay_data' );
 const buttonConfig = buttonData.scriptData;
-const namespace = 'ppcpBlocksPaypalApplepay';
+const namespace = 'ppcpBlocksPaypalAppglepay';
```
danieldudzic: Typo found :) 
```suggestion
const namespace = 'ppcpBlocksPaypalApplepay';
```

danieldudzic (approved):
Great effort, @stracker-phil.

Only found one minor typo in the namespace, otherwise code is legit.

I'm pre-approving.

PR #2747, opened 2024-10-31
Handle manual connection in new onboarding

AlexP11223:
Added a basic handler for the manual connection mode. For now using a dummy endpoint not actually performing the API requests. And not saving the returned merchant ID/email because as I understand it will be implemented separately too. (for all onboarding modes)

Also fixed the psalm config, adding exclusions for `node_modules` because one of the libs (`flatted`) contains PHP files. Added `.gitkeep` because for some reason psalm fails if the pattern is not matching anything.


modules/ppcp-settings/src/Endpoint/ConnectManualRestEndpoint.php
```diff
@@ -0,0 +1,92 @@
...
+			return rest_ensure_response(
+				array(
+					'success' => false,
+				)
+			);
+		}
+
+		$result = array(
+			'merchantId' => 'bt_us@woocommerce.com',
+			'email'      => 'AT45V2DGMKLRY',
```
Dinamiko, line 86: These two needs to be dynamic values.

Dinamiko (changes requested):

Dinamiko (approved):

PR #2752, opened 2024-11-04
added ApplePay and GooglePay to refund payment methods

danielhuesken:
This fixes that the "Void authorization" Butten in Orders will not be shown.


modules/ppcp-wc-gateway/src/Processor/RefundProcessor.php
```diff
@@ -119,7 +109,8 @@ public function __construct(
 	 */
 	public function process( WC_Order $wc_order, float $amount = null, string $reason = '' ) : bool {
 		try {
-			if ( ! in_array( $wc_order->get_payment_method(), $this->allowed_refund_payment_methods, true ) ) {
+			$payment_gateways = WC()->payment_gateways()->payment_gateways();
+			if ( ! isset( $payment_gateways[ $wc_order->get_payment_method() ] ) || ! $payment_gateways[ $wc_order->get_payment_method() ]->supports( 'refunds' ) ) {
```
AlexP11223, line 113: Seems unnecessary, I think WC is not supposed to call this method when refunds are not supported? Or at least it would not show the refund button.

But I guess additional check will not hurt.

AlexP11223 (approved):

PR #2777, opened 2024-11-08
Apple Pay: Only render the button markup when Apple Pay is available (3861)

danieldudzic:
### Description

This PR adds availability checking for the Apple Pay button to only render the button markup when Apple Pay is available.

### Steps to test

1. Enable Apply Pay and test using Safari.
2. Open the block checkout (on the page) and verify that the Apple Pay button displays correctly.
3. Open the block checkout with a different browser. Ensure the button doesn't render and the rest of the buttons display in a correct layout.
 
### Screenshots

|Before|After|
|-|-|
|Block_Checkout_–_WooCommerce_PayPal_Payments|Block_Checkout_–_WooCommerce_PayPal_Payments|


stracker-phil:
Note: This PR is not current anymore; we use the more generic approach from #2779

PR #2779, opened 2024-11-08
Hide payment button wrapper for ineligible buttons (3861)

stracker-phil:
### Description

This PR hides the entire payment button wrapper from the DOM for ineligible payment buttons.

It affects ApplePay and GooglePay, by removing the empty "placeholder" div when one of those buttons is not eligible.

The default value of `#isEligible` is set to null to trigger a button `refresh()` when the initial isEligible flag is set (before this PR, the refresh() was only triggered when a button became eligible, but not when it stayed ineligible)


Narek13 (approved):
LGTM

PR #2790, opened 2024-11-13
Screen 5: Add the last onboarding step (Complete Setup) (3909)

Dinamiko:
Create screen number 5:
complete setup


modules/ppcp-settings/node_modules/.gitkeep
stracker-phil, line 1: We need to keep this file to use CI actions

node_modules/.gitkeep
stracker-phil, line 1: We need to keep this file to use CI actions

stracker-phil (approved):

PR #2791, opened 2024-11-13
Store Manual Connection details in DB (3891)

danielhuesken:
Add saving of manual credentials and payee information


modules/ppcp-settings/src/Endpoint/ConnectManualRestEndpoint.php
```diff
@@ -134,6 +135,18 @@ public function connect_manual( WP_REST_Request $request ) : WP_REST_Response {
...
+		if ( $use_sandbox ) {
+			$settings->set_is_sandbox( true );
+			$settings->set_sandbox_client_id( $client_id );
+			$settings->set_sandbox_client_secret( $client_secret );
+		} else {
+			$settings->set_is_sandbox( false );
+			$settings->set_live_client_id( $client_id );
+			$settings->set_live_client_secret( $client_secret );
+		}
+		$settings->save();
```
stracker-phil: Let's move all `$settigs->set_...` calls to the bottom (line 162+, after successful payee-request)

**Reason**
If the user is connected to one account, and then enters a different secret/ID with invalid values, the DB state would be corrupt: It would store the new (invalid) secret/ID, while also holding the previous (valid) merchant-ID and email.
For that reason, I suggest to only update the settings once we have confirmed the client ID and secret are valid.

modules/ppcp-settings/src/Endpoint/ConnectManualRestEndpoint.php
```diff
@@ -134,6 +135,18 @@ public function connect_manual( WP_REST_Request $request ) : WP_REST_Response {
 			);
 		}
 
+		$settings = new GeneralSettings();
```
stracker-phil: We want to reduce dependencies to make the classes more reusable, i.e., please use dependency injection for this (the plugin uses Modularity; services are defined in the `services.php` file).

stracker-phil (approved):
Thanks 👍

PR #2794, opened 2024-11-14
Disable "Add payment method" button while adding card

danielhuesken:
Disable "Add payment method" button while adding card so that it will not be added multiple times.

### DIFF
--- a/modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
+++ b/modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
@@ -82,9 +82,8 @@ import {
 					renderFields( cardFields );
 				}
 
-				document
-					.querySelector( '#place_order' )
-					?.addEventListener( 'click', ( event ) => {
+                const placeOrderButton = document.querySelector( '#place_order' );
+				placeOrderButton?.addEventListener( 'click', ( event ) => {
 						const cardPaymentToken = document.querySelector(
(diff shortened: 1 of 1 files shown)


modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
```diff
@@ -95,11 +95,11 @@ import {
 						) {
 							return;
 						}
-
+                        document.getElementById('place_order').disabled = true;
```
Dinamiko: Better check is not null before accessing property:
```
 document.getElementById('place_order')?.disabled = true;
```

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
```diff
@@ -95,11 +95,11 @@ import {
...
 						event.preventDefault();
-
 						cardFields.submit().catch( ( error ) => {
 							console.error( error );
+                            document.getElementById('place_order').disabled = false;
```
Dinamiko: Same as above here.

Dinamiko (changes requested):

Dinamiko (approved):

modules/ppcp-save-payment-methods/resources/js/add-payment-method.js
danielhuesken: The element must exist now because we also use it for `addEventListener`. and call it inside of it.

PR #2819, opened 2024-11-21
Convert existing payment methods screen into the new design (3946)

inpsyde-maticluznar:
Apply the latest design to admin payment method cards.


Narek13 (approved):
Thanks, LGTM

PR #2834, opened 2024-11-22
Fastlane - Block checkout static meta comment is missing (3757)

stracker-phil:
### Description

This PR moves the feature-detection code into a separate hook that is always fired on checkout, cart and shop pages.

Before this PR, the flag was only output if AXO was eligible, the cart did not contain a subscription and the session was not in continuation mode:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/7d0504e1816bf69406eb50541020a2a940f7e131/modules/ppcp-axo/src/AxoModule.php#L197-L203


danieldudzic:
So this will fire each time ACDC and Fastlane are enabled, right?

danieldudzic (approved):

danieldudzic:
Seems that these changes have already made their way into `trunk`. I am closing.

PR #2855, opened 2024-11-28
New SEPA Direct Debit integration (3698)

Dinamiko:
Implement [SEPA direct debit](https://developer.paypal.com/limited-release/direct-banking/sepa-direct-debit/).

# SEPA direct debit features
- Supported transaction types: One-off or Recurring Payments.
- Settlement: Instant or Delayed.
- Vaulting: Supported
- Transaction Currency: EUR only
- Refunds: Full, Partial and Multiple Partial refunds supported.
- Integration Type: PayPal Orders v2 API and Braintree SDKs.

SEPA Direct Debit allows you to access funds immediately if you opt for Instant Settlements. Delayed settlement option will make funds available in the standard SEPA settlement timeframe of 2-3 days.


modules/ppcp-sepa/src/SEPAGateway.php
```diff
@@ -0,0 +1,297 @@
...
+	use ProcessPaymentTrait;
+
+	const ID = 'ppcp-sepa';
+
+	/**
+	 * PayPal Orders endpoint.
+	 *
+	 * @var Orders
+	 */
+	private $orders_endpoint;
```
Narek13, line 40: I would suggest to always add typehints since we are already on PHP 7.4 +
```suggestion
	private Orders $orders_endpoint;
```

modules/ppcp-sepa/src/SEPAGateway.php
```diff
@@ -0,0 +1,297 @@
...
+use WooCommerce\PayPalCommerce\WcGateway\Processor\RefundProcessor;
+
+/**
+ * Class SEPAGateway
+ */
+class SEPAGateway extends WC_Payment_Gateway {
+
+	use ProcessPaymentTrait;
+
+	const ID = 'ppcp-sepa';
```
Narek13, line 33:
```suggestion
	public const ID = 'ppcp-sepa';
```

modules/ppcp-sepa/src/SEPAPaymentMethod.php
```diff
@@ -0,0 +1,97 @@
...
+ * Class SEPAPaymentMethod
+ */
+class SEPAPaymentMethod extends AbstractPaymentMethodType {
+
+	/**
+	 * The URL of this module.
+	 *
+	 * @var string
+	 */
+	private $module_url;
```
Narek13, line 24: Same here, missing typehints for the class

modules/ppcp-sepa/src/SEPAGateway.php
stracker-phil: I agree with this suggestion; when possible, we should use the type hints.

Just want to point out that this is only possible for custom properties, and we cannot add a type hint for properties that are defined without a type in the base class

PR #2863, opened 2024-11-29
New Settings UI: Validate advanced options client ID (3958)

hmouhtar:
### Description

Add extra validation logic to make sure that the advanced options client ID field matches the provided regex, otherwise show an inline error message.

### Steps to Test

1. Enable the new Settings UI.
2. Start the PayPal onboarding process.
3. Inside the "Welcome to PayPal Payments" step, scroll down and click on "See advanced options".
4. Verify that the "Connect Account" button is disabled by default.
5. Verify that the "Connect Account" button remains disabled until both "Live Client ID" and "Live Secret Key" have correct values.
6. Verify that "Live Client ID" will display an inline error message that says "Please enter a valid Client ID" when the entered value doesn't match the provided regex.
7. Verify that the error message goes away when entering a correct client ID, e.g. "AQ85ce4QdFoaNqh5iG9kQMYBFwDzOhqkPcbpq1S7Te5ImQPkQpotd4-AeB_T6aISKJ0YOIgQU8kzusIN"

### Screenshots

WooCommerce-settings-‹-WooCommerce-PayPal-Payments-—-WordPress-11-29-2024_11_50_AM

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/AdvancedOptionsForm.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/AdvancedOptionsForm.js
@@ -30,7 +30,7 @@ const AdvancedOptionsForm = ( { setCompleted } ) => {
 	const refClientSecret = useRef( null );
 
 	const isValidClientId = useMemo( () => {
-		return clientId ? /^A[\w-]{79}$/.test( clientId ) : true;
+		return /^A[\w-]{79}$/.test( clientId );
 	}, [ clientId ] );
 
 	const isFormValid = useMemo( () => {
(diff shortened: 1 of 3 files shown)


modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/AdvancedOptionsForm.js
```diff
@@ -29,38 +29,13 @@ const AdvancedOptionsForm = ( { setCompleted } ) => {
...
-			if ( value ) {
-				continue;
-			}
-
-			ref?.current?.focus();
-			createErrorNotice( errorMessage );
-			return false;
-		}
+	const isValidClientId = useMemo( () => {
+		return clientId ? /^A[\w-]{79}$/.test( clientId ) : true;
```
Dinamiko: It seems to me that this returns `true` if `clientId` does not exist... if so, why?

Dinamiko (changes requested):

modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/AdvancedOptionsForm.js
hmouhtar: @Dinamiko It was to prevent showing validation errors when clientId is empty, but we are also checking that on the field itself at the moment of assigning the class, so this is redundant. I have changed it on the latest commit 👍🏻

Dinamiko (approved):

PR #2868, opened 2024-12-02
Check if the "extended" section exists before adding items

Narek13:
# Issue Description

ERROR Failed to create a task in the 'Things to do next' section of WC

# PR Description

The message Task list ID does not exist is fired from [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-wc-gateway/src/Settings/WcTasks/Registrar/TaskRegistrar.php#L26) , when WC tries to add the task to an “extended“ section, which is actually the “What To Do“ next section. So it looks like the “extended“ section doesn’t exist for the user somehow.

The PR adds a simple check to see if the “extended“ list exists before adding our items

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace WooCommerce\PayPalCommerce\WcGateway;
 
+use Automattic\WooCommerce\Admin\Features\OnboardingTasks\TaskLists;
 use Exception;
 use Psr\Log\LoggerInterface;
 use Throwable;
@@ -867,10 +868,16 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 				$logger = $container->get( 'woocommerce.logger.woocommerce' );
 				assert( $logger instanceof LoggerInterface );
 				try {
+					$task_lists = TaskLists::get_lists();
+					if ( ! isset( $task_lists['extended'] ) ) {
+						return;
+					}
+
 					$simple_redirect_tasks = $container->get( 'wcgateway.settings.wc-tasks.simple-redirect-tasks' );
 					if ( empty( $simple_redirect_tasks ) ) {
 						return;
 					}
+
 					$task_registrar = $container->get( 'wcgateway.settings.wc-tasks.task-registrar' );
 					assert( $task_registrar instanceof TaskRegistrarInterface );
 
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -9,7 +9,6 @@ declare(strict_types=1);
 
 namespace WooCommerce\PayPalCommerce\WcGateway;
 
-use Automattic\WooCommerce\Admin\Features\OnboardingTasks\TaskLists;
 use Exception;
 use Psr\Log\LoggerInterface;
 use Throwable;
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -867,10 +868,16 @@ static function () use ( $container ): void {
 				$logger = $container->get( 'woocommerce.logger.woocommerce' );
 				assert( $logger instanceof LoggerInterface );
 				try {
+					$task_lists = TaskLists::get_lists();
+					if ( ! isset( $task_lists['extended'] ) ) {
+						return;
```
stracker-phil: Maybe it's beneficial to add this condition directly into the `TaskRegistrar` class where the 'extended' attribute is used, or would that cause other problems?

https://github.com/woocommerce/woocommerce-paypal-payments/blob/4484386e498c09bdaf24df73ad28307de747e6f5/modules/ppcp-wc-gateway/src/Settings/WcTasks/Registrar/TaskRegistrar.php#L24-L31

Narek13: Ideally, the purpose of the `TaskRegistrar` is "To Register The Task", so I think it shouldn't care about any condition. It should add or throw an error on failure. And also with early return, the rest of other code is not running.
So why do you think it is beneficial to add it inside?

Edit: What can be improved here is to maybe pass the "extended" as a dependency through constructor, or as an additional argument, I haven't done that, because we are only going to use the "extended" which is actually the "Things To Do Next" section.

stracker-phil: I have two concerns about the current PR: 

1. If a different place in our plugin wants to register a "things to do next" task, that code would also need to manually check for the existence of the "extended" list before registering a task, i.e. this is a pattern that requires code duplication if we keep it
1. Since the service only asserts that the task-registrar implements a certain interface (`assert( $task_registrar instanceof TaskRegistrarInterface )`) we cannot guarantee in the service-file that the task-registrar actually uses TaskLists, or a different way to register tasks.


Short: I think checking the TaskLists::get_lists() response should be a responsibility of the Registrar class that also adds the task to that object.

Two options that, I think, would be better:
- Either add a new `$task_registrar->can_add_task()` before trying to register.
- Or, simply keep the error (nofix), since it's not a bug, but documents missing WooCommerce configuration which is outside our plugin's control (**my preference**)

Narek13: Sorry, but I don't see a reason for other part of the code to try to register a task in a different way?

- If you want to create "Simple redirect tasks" let's say from other module, you just need to extend [this](https://github.com/woocommerce/woocommerce-paypal-payments/blob/4484386e498c09bdaf24df73ad28307de747e6f5/modules/ppcp-wc-gateway/services.php#L2036) configuration
- if you want to add "new" type of tasks you can create your own config, get that config [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/4484386e498c09bdaf24df73ad28307de747e6f5/modules/ppcp-wc-gateway/src/WCGatewayModule.php#L885) and merge with the rest of the tasks before passing to `TaskRegistrar`

The whole idea is to have a central place/way to register the tasks. 

> Or, simply keep the error (nofix), since it's not a bug, but documents missing WooCommerce configuration which is outside our plugin's control

I think if you don’t have that section, you shouldn’t add any tasks to it, so for me it’s ok to check if section exists before trying to add anything to it.

What do you think about this @Dinamiko ?
cc: @InpsydeNiklas

Narek13: Update: I am passing the `$list_id` to the `register` method and adding the check there now: 10a95b0f2e797b484d7c67896bcf1aea6b4dad42

Dinamiko (approved):

PR #2870, opened 2024-12-02
New Settings UI: Map storeCountry and currency to WooCommerce settings (3962)

hmouhtar:
### Description

Replace the hardcoded country and currency values on the Welcome and Optional Payment Methods steps with data from WooCommerce.

### Steps to Test

1. Enable the new Settings UI.
2. Start the PayPal onboarding process.
3. Verify that the Welcome step matches the expected content based on the WooCommerce store country and currency.
4. Verify that the Optional Payment Methods step matches the expected content based on the WooCommerce store country and currency,.

Note: The payment flow is still hardcoded to ACDC.

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/StepPaymentMethods.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/StepPaymentMethods.js
@@ -3,7 +3,7 @@ import { __, sprintf } from '@wordpress/i18n';
 import OnboardingHeader from '../../ReusableComponents/OnboardingHeader';
 import SelectBoxWrapper from '../../ReusableComponents/SelectBoxWrapper';
 import SelectBox from '../../ReusableComponents/SelectBox';
-import { OnboardingHooks } from '../../../data';
+import { CommonHooks, OnboardingHooks } from '../../../data';
 import OptionalPaymentMethods from '../../ReusableComponents/OptionalPaymentMethods/OptionalPaymentMethods';
 
 const OPM_RADIO_GROUP_NAME = 'optional-payment-methods';
@@ -13,6 +13,9 @@ const StepPaymentMethods = ( {} ) => {
 		areOptionalPaymentMethodsEnabled,
 		setAreOptionalPaymentMethodsEnabled,
 	} = OnboardingHooks.useOptionalPaymentMethods();
+
+	const { storeCountry, storeCurrency } = CommonHooks.useWooSettings();
+
 	const pricesBasedDescription = sprintf(
 		// translators: %s: Link to PayPal REST application guide
 		__(
@@ -42,8 +45,8 @@ const StepPaymentMethods = ( {} ) => {
 								useAcdc={ true }
 								isFastlane={ true }
 								isPayLater={ true }
-								storeCountry={ 'us' }
-								storeCurrency={ 'usd' }
+								storeCountry={ storeCountry }
+								storeCurrency={ storeCurrency }
 							/>
 						}
 						name={ OPM_RADIO_GROUP_NAME }
--- a/modules/ppcp-settings/resources/js/data/common/hooks.js
+++ b/modules/ppcp-settings/resources/js/data/common/hooks.js
@@ -43,6 +43,7 @@ const useHooks = () => {
 	const clientSecret = usePersistent( 'clientSecret' );
 	const isSandboxMode = usePersistent( 'useSandbox' );
 	const isManualConnectionMode = usePersistent( 'useManualConnection' );
+	const flags = usePersistent( 'flags' );
 
 	const savePersistent = async ( setter, value ) => {
 		setter( value );
@@ -69,6 +70,7 @@ const useHooks = () => {
 		},
 		connectViaSandbox,
 		connectViaIdAndSecret,
+		flags,
 	};
 };
 
@@ -108,4 +110,10 @@ export const useManualConnection = () => {
 		setClientSecret,
 		connectViaIdAndSecret,
 	};
+};
+
+export const useWooSettings = () => {
+	const { flags = {} } = useHooks();
+	const { country, currency } = flags;
+	return { storeCountry: country, storeCurrency: currency };
 };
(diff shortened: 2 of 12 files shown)


modules/ppcp-settings/src/Endpoint/CommonRestEndpoint.php
```diff
@@ -60,6 +60,20 @@ class CommonRestEndpoint extends RestEndpoint {
...
+	 * Map the internal flags to JS names.
+	 *
+	 * @var array
+	 */
+	private array $flag_map = array(
+		'country'  => array(
+			'js_name' => 'country',
+		),
+		'currency' => array(
+			'js_name' => 'currency',
```
stracker-phil: Instead of renaming the variables in the hook, we can directly use the js_name `storeCountry` and `storeCurrency` here and change the hook to 

```js
export const useWooSettings = () => {
	const { flags } = useHooks();
	return flags;
}
```

modules/ppcp-settings/src/Endpoint/CommonRestEndpoint.php
```diff
@@ -103,13 +117,23 @@ public function register_routes() {
...
-		return $this->return_success( $js_data );
+		$js_flags = $this->sanitize_for_javascript(
+			$this->settings->get_flags(),
+			$this->flag_map
+		);
+
+		return $this->return_success(
+			$js_data,
+			array(
+				'flags' => $js_flags,
```
stracker-phil: Possibly, the attribute name `flags` could be a bit misleading. Let's change this to something like `wooSettings` (or `woo`, etc) to make it clearer what data the object contains.

I.e. rename all `flags`, `$flags`, `get_flags`, and so on, in all places in this PR to a more semantic term

stracker-phil (changes requested):
The code is nice, and the changes are almost correct 👍

Only two minor things:
1. I'd appreciate it if we can update the namings, and move from the "flags" to a more semantic attribute name
2. The Redux store handles the "flags"/"wooSettings" incorrectly. They must be extracted to a new selector, and should not be part of the persistent data collection

modules/ppcp-settings/resources/js/data/common/hooks.js
```diff
@@ -43,6 +43,7 @@ const useHooks = () => {
 	const clientSecret = usePersistent( 'clientSecret' );
 	const isSandboxMode = usePersistent( 'useSandbox' );
 	const isManualConnectionMode = usePersistent( 'useManualConnection' );
+	const flags = usePersistent( 'flags' );
```
stracker-phil: We should create a new selector for this, since the "flags" (or "wooSettings") should not be part of Transient or Persistent response. They are read-only

```js
// Read-only values.
const wooSettings = useSelect( ( select ) => select( STORE_NAME ).wooSettings(), [] );
```

The selectors would look similar to this:

```js
// ...
export const transientData = ( state ) => {
	const { data, wooSettings, ...transientState } = getState( state ); // ← extract the wooSettings, to ensure they are not part of the "transientState" collection.
	return transientState || EMPTY_OBJ;
};

export const wooSettings = ( state ) => {
	return getState( state ).wooSettings || EMPTY_OBJ;
};
```

modules/ppcp-settings/resources/js/data/common/reducer.js
```diff
@@ -39,7 +43,10 @@ const commonReducer = createReducer( defaultTransient, defaultPersistent, {
 		setPersistent( state, action ),
 
 	[ ACTION_TYPES.HYDRATE ]: ( state, payload ) =>
-		setPersistent( state, payload.data ),
+		setPersistent( state, {
+			...payload.data,
+			flags: payload.flags,
+		} ),
```
stracker-phil: This logic is not entirely correct.

The "flags" (or "wooSettings") should not be part of the Persistent object: The "setPersistent" method should only store data that is input by the user (form fields) and that should be saved/changed on the server.

The country/currency are read-only flags and must be manually added tot the new state object, like this:

```js
// ...
	[ ACTION_TYPES.HYDRATE ]: ( state, payload ) => {
		const newState = setPersistent( state, payload.data ); // this sets "persistant" items to the state.data object

		// Here we set read-only data to the state.wooSettings object. This object should be only accessible via a new selector
		if ( payload.wooSettings ) {
			newState.wooSettings = { ...newState.wooSettings, ...payload.wooSettings };
		}

		return newState;
	},
// ...
```

modules/ppcp-settings/src/Data/CommonSettings.php
```diff
@@ -116,4 +135,13 @@ public function get_client_secret() : string {
 	public function set_client_secret( string $client_secret ) : void {
 		$this->data['client_secret'] = sanitize_text_field( $client_secret );
 	}
+
+	/**
+	 * Returns the list of read-only customization flags
```
Narek13:
```suggestion
	 * Returns the list of read-only customization flags.
```

modules/ppcp-settings/resources/js/data/common/reducer.js
hmouhtar: Hello @stracker-phil, I've applied all your suggested changes in the latest commit. Thanks 👍🏻

modules/ppcp-settings/resources/js/Components/Screens/Onboarding/StepPaymentMethods.js
```diff
@@ -13,6 +13,11 @@ const StepPaymentMethods = ( {} ) => {
 		areOptionalPaymentMethodsEnabled,
 		setAreOptionalPaymentMethodsEnabled,
 	} = OnboardingHooks.useOptionalPaymentMethods();
+
+	const { storeCountry, storeCurrency } = CommonHooks.useWooSettings();
+
+    console.log(storeCountry, storeCurrency);
```
Narek13: leftover?

stracker-phil: That line is not even part of the latest branch... Strange that it shows up in the PR 🤔

stracker-phil (approved):

PR #2875, opened 2024-12-04
Create the missing APM cards (3985)

Narek13:
# Issue Description

The Alternative Payment methods currently are only listed with:

- Bancontact
- iDEAL
- eps
- BLIK

We need to add more cards in the Alternative payment methods section the remaining payment methods:

- MyBank
- Przelewy24
- Trustly
- Multibanco
- Pay upon Invoice (only visible when store country is DE and store currency is EUR) 
- OXXO  (only visible when store country is MX and store currency is MXN) 

The descriptions for these Alternative payment methods should be taken from their existing descriptions in the old UI.

# PR Description

- Adds the APM cards for:
  - MyBank
  - Przelewy24
  - Trustly
  - Multibanco
  - Pay upon Invoice (only visible when store country is DE and store currency is EUR) 
  - OXXO  (only visible when store country is MX and store currency is MXN)
- Fixes descriptions


stracker-phil (approved):
Looks good  👍

PR #2876, opened 2024-12-05
Unsuccessfully payment with Fastlane, error on console and broken env (3996)

Dinamiko:
Unsuccessfully payment with Fastlane, error on console and broken env.

### Steps To Reproduce
- Navigate to shop as guest
- Add product to cart
- Open console
- Navigate to block checkout page
    - Observe error on console

Screenshot 2024-12-03 at 09 44 51

This PR fixes the error by getting `PAYMENT_STORE_KEY` from global `wc` variable instead of importing it.

### DIFF
--- a/modules/ppcp-axo-block/resources/js/plugins/PayPalInsightsLoader.js
+++ b/modules/ppcp-axo-block/resources/js/plugins/PayPalInsightsLoader.js
@@ -1,7 +1,6 @@
 import { registerPlugin } from '@wordpress/plugins';
 import { useEffect, useCallback, useState, useRef } from '@wordpress/element';
 import { useSelect } from '@wordpress/data';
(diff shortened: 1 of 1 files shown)


danieldudzic (approved):

stracker-phil (approved):
Should fix the problem ✔︎

Note: To be absolutely sure the `wcBlockData` object is present, you could add a new dependency to the register_script call:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/4484386e498c09bdaf24df73ad28307de747e6f5/modules/ppcp-axo-block/src/AxoBlockModule.php#L198-L204

New depedency:

```php
array( 'wp-plugins', 'wp-data', 'wp-element', 'wc-blocks-registry', 
    'wc-blocks-data-store' // ensures the wcBlockData object is present
),
```

PR #2878, opened 2024-12-05
Automatically expand bcdc form fields upon selection when gateway is separated

danielhuesken:
Add the expanding option to the buttons when separated.
Removed card GW from early validation to get the fields shown.


Dinamiko (approved):

InpsydeNiklas:
This is not needed anymore.

PR #2891, opened 2024-12-09
Changing subscription payment method to ACDC results in an error (3949)

Dinamiko:
When an active subscription from another payment method, e.g. Stripe, is switched to ACDC from the my-account section, an error will be displayed in the frontend.

### Steps To Reproduce
- buy a subscription with a third-party payment method like Stripe
- ensure ACDC Vaulting is enabled
- in my-account > Payment methods section, click “Add payment method”
- add an ACDC card, like 2223000048400011
- navigate to the subscription in my-account > Subscriptions  > Change payment
- change the payment method to the saved ACDC token
   - observe an error in the frontend that the payment method could not be changed
- in subscription admin, observe that the payment method was changed despite the error
- upon triggering a renewal, observe error `PAYMENT_SOURCE_MISMATCH` error

This PR ensures that previous transaction logic is not executed when is not a payment from the plugin.


modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php
```diff
@@ -496,4 +527,41 @@ function ( array $supports ) use ( $c ): array {
...
 	}
+
+	/**
+	 * Adds the given WC payment token into the given WC Order.
+	 *
+	 * @param WC_Order $wc_order
+	 * @param int $wc_payment_token_id
+	 * @param string $return_url
+	 * @param SessionHandler $session_handler
+	 * @return array
```
Narek13: I think this is always `string[]`
```suggestion
	 * @return array{result: string, redirect: string, errorMessage?: string}
```

Narek13 (approved):
Thanks. Just a small suggestion which you can ignore

danielhuesken (approved):

Narek13 (approved):

PR #2892, opened 2024-12-09
Hide the price notice when no badges are visible (4000)

Dinamiko:
Step 1 (welcome) and 4 (optional payment methods) contains badges with country & currency relevant pricing details. When using a store country that is not defined, the badges are hidden.

When those badges are not displayed, the footer remark must be hidden as well:
image-20241204-110333

### Steps to reproduce
- Set the WooCommerce store country to something like Algeria, Brazil, or China
- Check that in steps 1 and 4 price notice is hidden

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/ReusableComponents/WelcomeDocs/pricesBasedDescription.js
+++ b/modules/ppcp-settings/resources/js/Components/ReusableComponents/WelcomeDocs/pricesBasedDescription.js
@@ -1,0 +1,10 @@
+import { __, sprintf } from '@wordpress/i18n';
+
+export const pricesBasedDescription = sprintf(
+	// translators: %s: Link to PayPal REST application guide
+	__(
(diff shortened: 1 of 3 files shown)


modules/ppcp-settings/resources/js/Components/ReusableComponents/WelcomeDocs/pricesBasedDescription.js
```diff
@@ -0,0 +1,10 @@
+import { __, sprintf } from '@wordpress/i18n';
+
+export const pricesBasedDescription = sprintf(
+	// translators: %s: Link to PayPal REST application guide
+	__(
+		'<sup>1</sup>Prices based on domestic transactions as of October 25th, 2024. <a target="_blank" href="%s">Click here</a> for full pricing details.',
+		'woocommerce-paypal-payments'
+	),
+	'https://woocommerce.com/document/woocommerce-paypal-payments/#manual-credential-input '
```
Narek13, line 9: BTW we are still missing these links

Narek13 (approved):

PR #2896, opened 2024-12-10
plugin queries variations when button/messaging is disabled on single product page

danielhuesken:
Removed the additional query over variations when "variable product" is in stock.

- The variable product is in stock when one variation is in stock.
- The check on all variations must only run when sock management is active on variations and variable product with out of stock

So it will improve the performance because it only queries all variations on edge cases.

The HTML div problem was already solved.


Narek13 (approved):
Thanks

Dinamiko (approved):

PR #2905, opened 2024-12-11
Change hardcoded address match and postal match to matched details on cvv response

danielhuesken:
The order node displays now a detailed message from card verification. 
2. Order node removed and included in the other.

### DIFF
--- a/modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
+++ b/modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
@@ -108,12 +108,12 @@ class FraudProcessorResponse {
 			'3' => __( '3: The merchant did not provide AVS information. Not processed.', 'woocommerce-paypal-payments' ),
 			'4' => __( '4: Address not checked, or acquirer had no response. Service not available.', 'woocommerce-paypal-payments' ),
 		);
+
 		/**
-		 * Translators: %s is fraud AVS code
-		 *
 		 * @psalm-suppress PossiblyNullArrayOffset
 		 * @psalm-suppress PossiblyNullArgument
 		 */
+		/* translators: %s is fraud AVS code */
 		return $messages[ $this->avs_code() ] ?? sprintf( __( '%s: Error', 'woocommerce-paypal-payments' ), $this->avs_code() );
 	}
 
@@ -145,12 +145,12 @@ class FraudProcessorResponse {
 			'3' => __( '3: Merchant has indicated that CVV2 is not present on card', 'woocommerce-paypal-payments' ),
 			'4' => __( '4: Service not available', 'woocommerce-paypal-payments' ),
 		);
+
 		/**
-		 * Translators: %s is fraud CVV2 code
-		 *
 		 * @psalm-suppress PossiblyNullArrayOffset
 		 * @psalm-suppress PossiblyNullArgument
 		 */
+		/* translators: %s is fraud CVV2 code */
 		return $messages[ $this->cvv_code() ] ?? sprintf( __( '%s: Error', 'woocommerce-paypal-payments' ), $this->cvv_code() );
 	}
 }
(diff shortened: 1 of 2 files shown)


modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
```diff
@@ -51,10 +51,10 @@ public function avs_code(): ?string {
 	/**
 	 * Returns the CVV response code.
 	 *
-	 * @return string|null
+	 * @return string
 	 */
 	public function cvv_code(): ?string {
```
Narek13: I think this can't be `null` now, because you have `$this->cvv2_code = (string) $cvv2_code;` in constructor and `(string) null` will return an empty string, so the return value of this method is always a string like in Dockblock comment

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
```diff
@@ -17,32 +17,32 @@ class FraudProcessorResponse {
...
+		$this->cvv2_code = (string) $cvv2_code;
 	}
 
 	/**
 	 * Returns the AVS response code.
 	 *
-	 * @return string|null
+	 * @return string
 	 */
 	public function avs_code(): ?string {
```
Narek13: Same here
```suggestion
	public function avs_code(): string {
```

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
```diff
@@ -64,11 +64,99 @@ public function cvv_code(): ?string {
...
+			'4' => '4: Address not checked, or acquirer had no response. Service not available.',
+		);
+
+		/**
+		 * Psalm suppress
+		 *
+		 * @psalm-suppress PossiblyNullArrayOffset
+		 * @psalm-suppress PossiblyNullArgument
+		 */
+		return $messages[ $this->avs_code() ] ?? sprintf( '%s: Error', $this->avs_code() );
```
Narek13, line 122: if the constructor param `?string $avs_code` is `'null'` or `''` then `$this->avs_code()` here will return `''` and the error message will be like this: `string(7) ": Error"`

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
```diff
@@ -64,11 +64,99 @@ public function cvv_code(): ?string {
...
+			'4' => '4: Service not available',
+		);
+
+		/**
+		 * Psalm suppress
+		 *
+		 * @psalm-suppress PossiblyNullArrayOffset
+		 * @psalm-suppress PossiblyNullArgument
+		 */
+		return $messages[ $this->cvv_code() ] ?? sprintf( '%s: Error', $this->cvv_code() );
```
Narek13, line 160: same here

Narek13 (approved):
Thanks for working on this, just couple of suggestions to consider

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
danielhuesken: That is what I would like to have. Then I must not handle different return types.

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
danielhuesken: That will be checked in Line 85.

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
danielhuesken: That will be checked in Line 133.

modules/ppcp-api-client/src/Entity/FraudProcessorResponse.php
Narek13: Yes but since you have cast in constructor already
```
$this->avs_code  = (string) $avs_code;
$this->cvv2_code = (string) $cvv2_code;
```

you have already forced it to be only a string. Sorry, feel free please to ignore

danielhuesken: Ah now I have seen what you mean ;) I will change it. Thanks!

Narek13 (approved):

Dinamiko (approved):

PR #2909, opened 2024-12-11
Fix PHP notices related to settings-mapping (4028)

stracker-phil:
### Description

Several improvements for the `SettingsMapHelper` 

- Prevent PHP notice about accessing undefined array-keys
- Add caching mechanism to avoid repeat loops - which becomes more important as we map more settings

### DIFF
--- a/modules/ppcp-compat/src/SettingsMapHelper.php
+++ b/modules/ppcp-compat/src/SettingsMapHelper.php
@@ -9,6 +9,8 @@ declare( strict_types = 1 );
 
 namespace WooCommerce\PayPalCommerce\Compat;
 
+use RuntimeException;
+
 /**
  * A helper class to manage the transition between legacy and new settings.
  *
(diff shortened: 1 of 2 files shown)


modules/ppcp-compat/src/SettingsMapHelper.php
```diff
@@ -5,66 +5,155 @@
...
+	 */
+	protected array $model_cache = array();
+
 	/**
 	 * Constructor.
 	 *
-	 * @param SettingsMap[] $settings_map A list of mapped settings.
+	 * @throws RuntimeException When an old key has multiple mappings.
+	 *
+	 * @param SettingsMap[] $settings_map A list of settings maps containing key definitions.
```
Narek13:
```suggestion
	 * @param SettingsMap[] $settings_map A list of settings maps containing key definitions.
	 *
	 * @throws RuntimeException When an old key has multiple mappings.
```

I think the `throws` tag should go after the `param` and `return` tags

Narek13 (approved):
Thanks for taking care of this. Just a minor suggestion which doesn't prevent me from approving

modules/ppcp-compat/src/SettingsMapHelper.php
stracker-phil: PhpStorm arranged those headers, but that's just my personal configuration 🙈

In most of the code the @throws tag comes even later, after the @return, so I've updated my PhpStorm settings accordingly

Narek13 (approved):

PR #2910, opened 2024-12-12
PayPal Subscription Creation Fails for Multi-Year Plans

danielhuesken:
Add logic that prevents from not being allowed periods in PayPal subscriptions.


Narek13 (approved):
Thanks for working on this. Haven't tested, but LGTM

PR #2912, opened 2024-12-12
Review and fix JS warnings in "badgeBoxUtils.js" (4029)

stracker-phil:
### Description

Refactors the price details in the onboarding wizard to address two issues:

1. Resolve a console warning (possibly a debug-output)
2. Prepare the code for currency-specific fixed fees and correct

The entire country- and currency-logic is encapsulated inside `<PricingDescription>` and `<PricingTitleBadge>` components to guarantee that all components display the correct amount for the current country and currency


Narek13 (approved):
Thanks. LGTM

PR #2919, opened 2024-12-16
Remove currency requirement for Vault v3

danielhuesken:
Only use counties for allowing saved payments.
- updated county list
- removed currencies
- add depreciation message for the old filter


modules/ppcp-save-payment-methods/src/Helper/SavePaymentMethodsApplies.php
```diff
@@ -9,60 +9,46 @@
...
 	 */
-	private $allowed_country_currency_matrix;
-
-	/**
-	 * The getter of the 3-letter currency code of the shop.
-	 *
-	 * @var CurrencyGetter
-	 */
-	private CurrencyGetter $currency;
+	private array $allowed_country_currencies;
```
Dinamiko: Why currencies in the name? shouldn't this be just `$allowed_countries`?

modules/ppcp-save-payment-methods/services.php
```diff
@@ -20,75 +20,56 @@
...
-			'PHP',
-			'PLN',
-			'SEK',
-			'THB',
-			'TWD',
-			'USD',
-		);
+	'save-payment-methods.supported-countries'           => static function ( ContainerInterface $container ) : array {
+		if ( has_filter( 'woocommerce_paypal_payments_save_payment_methods_supported_country_currency_matrix' ) ) {
+			_deprecated_hook( 'woocommerce_paypal_payments_save_payment_methods_supported_country_currency_matrix', '3.0.0', 'woocommerce_paypal_payments_save_payment_methods_supported_countries', esc_attr__( 'Please use the new Hook to filer countries for saved payments in PayPal Payments.', 'woocommerce-paypal-payments' ) );
```
Dinamiko: There is a typo in "Please use the new Hook to **filer**", it should be **filter**.

Dinamiko (dismissed):

stracker-phil (approved):

PR #2934, opened 2024-12-18
String translations not applied to Card Fields on Checkout block

danielhuesken:
Script translations must be configured additional when not used directly in a block.


Narek13 (approved):
I have tested locally and everything works fine for me. Thanks

PR #2937, opened 2024-12-19
New Settings UI: Disable settings in send-only countries (4027)

hmouhtar:
### Description

This PR updates the settings screen to hide all settings if the store's current country is classified as a send-only country. Instead, a message explaining this limitation will be displayed.

### Steps to Test

1. Enable the new settings UI
2. Set the WooCommerce store country to one that is within the send-only list. This list can be found in [PCP-3970](https://inpsyde.atlassian.net/browse/PCP-3790).
3. Verify that no settings appear on the PayPal settings screen. Instead, a message explaining this limitation should be displayed.

### Screenshots

WooCommerce-settings-‹-WooCommerce-PayPal-Payments-—-WordPress-12-19-2024_02_41_PM


hmouhtar:
@stracker-phil I wasn’t sure if there is a specific style we need to use for notices, so I haven’t added any styling for the notice yet, just plain text.

modules/ppcp-settings/resources/js/Components/Screens/SendOnlyMessage.js
stracker-phil, line 1: We should translate this message via `__()` or inject the string from server-side using `wp_localize_script()`

stracker-phil (changes requested):
Localization is missing 😉

modules/ppcp-settings/resources/js/Components/Screens/SendOnlyMessage.js
hmouhtar: Hi @stracker-phil, I added localization to the message in my latest commit 👍🏻

modules/ppcp-paylater-configurator/src/PayLaterConfiguratorModule.php
```diff
@@ -68,8 +69,9 @@ static function () use ( $c ) {
 				$current_page_id     = $c->get( 'wcgateway.current-ppcp-settings-page-id' );
 				$is_wc_settings_page = $c->get( 'wcgateway.is-wc-settings-page' );
 				$messaging_locations = $c->get( 'paylater-configurator.messaging-locations' );
+				$onboarding_state    = $c->get( 'onboarding.state' );
```
stracker-phil: This uses the legacy onboarding status.

To check if the onboarding in the new UI is completed, you can use this logic:

```php
$onboarding_profile = $c->get( 'settings.data.onboarding' );
$onboarding_profile->get_completed(); // true = onboarding is done, false = still in the onboarding wizard
```

stracker-phil (changes requested):

modules/ppcp-paylater-configurator/src/PayLaterConfiguratorModule.php
hmouhtar: Hi @stracker-phil, I've switched to the new onboarding profile class on my latest commit 👍🏻

stracker-phil (approved):

PR #2940, opened 2024-12-20
Make the webhook resubscribe/simulate logic usable in React application (3930)

inpsyde-maticluznar:
This PR integrates webhook settings in the new UI. Here are important things to mention: 

- 2 new endpoints:
  - webhook_settings:
    - GET returns subscribed webhooks and an URL
    - POST runs webhook resubscription
  - webhook_simulate
    - GET checks simulations status
    - POST starts simulation 
- `@wordpress/notices` is used for notifications. There are no visual differences between notification types, so I added icons:
  - Success icon: ✔️
  - Error icon: ❌
  - No icon when info notice
- I wanted to separate webhooks from other settings. This is a consequence : [link to the code](https://github.com/woocommerce/woocommerce-paypal-payments/pull/2940/files#diff-0e5a8daa058e5b341626153d55a2cc5b6d2a14240dede808e37d42cc58ed93e6R24-R26) Is this fine, or could we improve it?
- This code does not handle the situation when there are no subscribed events or a webhook URL. What should we do in this situation? 
- The `actionProps.isBusy` parameter is added to the button settings block, and then passed to the WP Button component. 
- The old UI displayed a visual error after checking the webhook simulation for the 15th time. This error has now been removed. The user will see the message Waiting for the webhook to arrive…, and the button will display a busy animation that remains active throughout all 30 attempts to check the simulation status. Nothing else will occur during this time.


Narek13 (approved):

PR #2942, opened 2024-12-20
Do not validate ACDC card brand keys locally but via Confirm Payment Source API

danielhuesken:
Streamline the call off confirm_payment_source() and reference implementation for future investigation.


InpsydeNiklas:
This is no longer needed.

PR #2956, opened 2024-12-30
New Settings UI: Remove section links from PayPal settings page (4049)

hmouhtar:
### Description

Some plugins, such as [WooPayments](https://es.wordpress.org/plugins/woocommerce-payments/), may add extra sections to the checkout tab . By default, WooCommerce outputs the links to these sections at the beginning of the tab. 

To prevent these section links from displaying on the PayPal settings page, this PR takes care of removing the action responsible of echoing the links. 

### Steps to Test

1. Enable the new PayPal settings UI
2. [Install the WooCommerce Payments plugin](https://es.wordpress.org/plugins/woocommerce-payments/)
3. Go to the Payments tab inside the WooCommerce settings page.
4. Verify that there are 2 links at the top: "WooPayments" and "All payment methods".
5. Go to the PayPal settings section.
6. Verify that the 2 links are not displaying.

It might be worth moving the logic for removing an action based on its name to a separate function, which could be useful in future cases like this one, where we don't have access to a class instance to remove an action or filter directly.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -227,7 +227,6 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 			}
 		);
 
-
 		/**
 		 * Removes the default WooCommerce action that displays extra section links in the checkout tab.
 		 * Prevents unwanted section links (e.g., from WooPayments) from appearing on the PayPal settings page.
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -227,6 +227,36 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 			}
 		);
 
+		add_action(
+			'admin_init',
+			function() use ( $container ) {
+				global $wp_filter;
+
+				if ( ! $container->has( 'wcgateway.is-ppcp-settings-page' ) ||
+					$container->get( 'wcgateway.is-ppcp-settings-page' ) === false ) {
+					return;
+				}
+
+				if ( empty( $wp_filter['woocommerce_sections_checkout']->callbacks ) ) {
+					return;
+				}
+
+				foreach ( $wp_filter['woocommerce_sections_checkout']->callbacks as $priority => $callbacks ) {
+					foreach ( $callbacks as $callback_key => $callback_data ) {
+						$function = $callback_data['function'];
+
+						if (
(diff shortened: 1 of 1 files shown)


stracker-phil (changes requested):
Nice solution, technically it looks good.

Just 2 small things:
1. Since it's such a technical bit, I would be much happier if there was a short comment above the add_action() giving a short hint on why we're removing items from the global `$wp_filter` array here 😉
2. Psalm complains about line 228. Maybe a different check works better? I'd try this: `$function[0] instanceof WC_Settings_Payment_Gateways`

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -203,6 +203,36 @@ static function () use ( $container ) : void {
...
+				global $wp_filter;
+
+				if ( ! $container->has( 'wcgateway.is-ppcp-settings-page' ) ||
+					$container->get( 'wcgateway.is-ppcp-settings-page' ) === false ) {
+					return;
+				}
+
+				if ( empty( $wp_filter['woocommerce_sections_checkout']->callbacks ) ) {
+					return;
+				}
```
Narek13: Why not having like this? 

```suggestion
				if ( 
					! $container->has( 'wcgateway.is-ppcp-settings-page' )
					|| ! $container->get( 'wcgateway.is-ppcp-settings-page' )
					|| empty( $wp_filter['woocommerce_sections_checkout']->callbacks )
				) {
					return;
				}
```

Otherwise makes sense to set the `global $wp_filter;` after the first check

Narek13 (approved):
Thanks for working on this. Just a minor suggestion which you can ignore

modules/ppcp-settings/src/SettingsModule.php
hmouhtar: @Narek13  I've moved the declaration of `$wp_filter` after the first check as suggested. Thanks 🙂

Narek13 (approved):

stracker-phil (approved):

stracker-phil (approved):

stracker-phil:
Though this solution works, I think a **CSS-only solution** aligns better with the current setup and keeps the codebase cleaner (all UI adjustments are currently done via CSS, this would be the first UI cleanup that's done via PHP)

I've added a new PR that provides the relevant CSS changes to hide the extra menu items: #3008

stracker-phil:
Closing this in favor of #3008

PR #2966, opened 2025-01-08
Create the component for live account connection (4078)

Narek13:
# Issue Description

The “Sandbox” block currently is displayed regardless of whether a sandbox or live account is connected. This block currently is only suitable for the scenario where the plugin is already connected to a live account.

Similar to this block, a block is needed that allows a user with an already connected sandbox account to switch to a live account

# PR Description

The PR will adjust the functionality as following:

- When sandbox is connected
  - Show sandbox block with credentials
  - Show new live account block
  - The button will take to the onboarding screen

Screenshot 2025-01-08 at 13 51 48

- When sandbox is not connected and live connected
  - Show live account block with credentials
  - Show sandbox connection block

Screenshot 2025-01-08 at 13 41 13


modules/ppcp-settings/resources/js/Components/Screens/Overview/TabSettingsElements/Blocks/LiveAccount.js
```diff
@@ -0,0 +1,206 @@
...
+	TITLE_BADGE_POSITIVE,
+} from '../../../../ReusableComponents/TitleBadge';
+import ConnectionInfo, {
+	connectionStatusDataDefault,
+} from '../../../../ReusableComponents/ConnectionInfo';
+
+const LiveAccount = ( { settings, updateFormValue } ) => {
+	const className = settings.sandboxConnected
+		? 'ppcp-r-settings-block--sandbox-connected'
+		: 'ppcp-r-settings-block--sandbox-disconnected';
```
stracker-phil: wrong class name?

modules/ppcp-settings/resources/js/Components/Screens/Overview/TabSettingsElements/Blocks/LiveAccount.js
```diff
@@ -0,0 +1,206 @@
...
+
+	return (
+		<AccordionSettingsBlock
+			title={ __( 'Live Payments', 'woocommerce-paypal-payments' ) }
+			className={ className }
+			description={
+				settings.sandboxConnected
```
stracker-phil: `sandboxConnected` looks wrong. Is this really the correct prop? If yes, then we should rename it to `isConnected`

stracker-phil (changes requested):
I really think we should not create a full copy of the sandbox section, but use one component that has a "sandbox" prop, to adjust the labels and relevant details

Like this:
```jsx
<Content>
    <ConnectionDetails
        sandbox={ true } // ← Sandbox
        settings={ settings }
        updateFormValue={ updateFormValue }
    />
</Content>
<Content>
    <ConnectionDetails
        sandbox={ false } // ← Production
        settings={ settings }
        updateFormValue={ updateFormValue }
    />
</Content>
```

Narek13:
Thanks @stracker-phil. about [this one](https://github.com/woocommerce/woocommerce-paypal-payments/pull/2966#pullrequestreview-2537754334):

> I really think we should not create a full copy of the sandbox section, but use one component that has a "sandbox" prop, to adjust the labels and relevant details ...

I think that will not work. The problem is that when let's say is sandbox we need to add a `<content> {sandbox is on}</content>` for sandbox and `<content>{sandbox is on}</content>` for production. Same when sandbox is off. for each case we have to add 2 `<content>`s, 1 for production and 1 for sandbox.

So we can have something like this

```
<Content>
    <ConnectionDetails
        sandbox={ true } // ← Sandbox
        for = 'sandbox'
        settings={ settings }
        updateFormValue={ updateFormValue }
    />
</Content>
<Content>
    <ConnectionDetails
        sandbox={ true } // ← Production
        for = 'production'
        settings={ settings }
        updateFormValue={ updateFormValue }
    />
</Content>
```

or we have to put the `<Content>`s inside the `ConnectionDetails`
plus all the string changes.
At the end to not overcomplicate the `ConnectionDetails` I have decided to create additional component.

maybe let's discuss this tomorrow in a call?

Edit: as discussed in slack channel we will keep only 1 block so we can have a single `<ConnectionDetails>` component @stracker-phil

stracker-phil (approved):

PR #2971, opened 2025-01-09
Success and error messages for "Refresh" button

mmaymo:
This Pr adds is isBusy property to the feature button, so that when refresh is clicked all buttons wait and show a loading animation. It also adds notices similar to the webhook resubscribe implementation.


danieldudzic (approved):
Works great! Just added a minor fix.

PR #2993, opened 2025-01-15
Integration with PHP-Scoper (968)

Narek13:
# Issue Description

We have received some reports about the fatal errors when running the plugin alongside the `psr/log` package, so we need to use the PHP-Scoper to scope the package

# PR Description

The PR will integrate PHP-Scoper with everything excluded except of the `psr/log` package.

### DIFF
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -9,35 +9,16 @@ declare(strict_types=1);
 
 use Isolated\Symfony\Component\Finder\Finder;
 
-$finders = array(
-	Finder::create()
-		->files()
-		->ignoreVCS( true )
-		->ignoreDotFiles( false ) // We need to keep .distignore around.
-		->exclude(
-			array(
-				'.github',
-				'.ddev',
-				'.idea',
-				'modules',
-				'tests',
-				'api',
-				'lib',
-				'src',
-				'vendor',
-			)
-		)
-		->in( '.' ),
-);
-
 return array(
 	'prefix'                  => 'WooCommerce\\PayPalCommerce\\Vendor',
-	'finders'                 => $finders,
-	'patchers'                => array(),
-	'exclude-namespaces'      => array(),
+	'finders'                 => array(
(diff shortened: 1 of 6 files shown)


scoper.inc.php
```diff
@@ -0,0 +1,51 @@
...
+		->in( '.' ),
+);
+
+return array(
+	'prefix'                  => 'WooCommerce\\PayPalCommerce\\Vendor',
+	'finders'                 => $finders,
+	'patchers'                => array(),
+	'exclude-files'           => array(), // list<string>.
+	'exclude-namespaces'      => array(
+		'/^(?!Psr).*/',
```
Biont: I'd suggest to add a comment about our intention here.
So far this is the only reason we're doing this and it may not be clear to the casual viewer that we are _only prefixing `Psr\*`_ here - nothing else.

scoper.inc.php
```diff
@@ -0,0 +1,51 @@
...
+	), // list<string|regex>.
+	'exclude-constants'       => array(), // list<string|regex>.
+	'exclude-classes'         => array(),     // list<string|regex>.
+	'exclude-functions'       => array( '/.*/' ), // list<string|regex>.
```
Biont: I'm intrigued by this. Are there any core functions within PSR packages that necessitate this?

Biont (changes requested):
Really just a small inquiry to add a comment.

On a larger scale, this change begs the question why we're even using PSR contracts - if not to stay compatible with a larger ecosystem.

I know this still allows us to use generic modules internally (and simply not expose a standard API after the build process), but if we have any intention of allowing third-party devs to interface with our code, we might want to consider `PsrCompatWrapper` objects of some sort. But that's just a remark.

scoper.inc.php
Narek13: Thanks. Was just a leftover.

Biont (approved):

Dinamiko (approved):

PR #3008, opened 2025-01-20
Remove WooPayments menu item on PayPal Payments settings pages (4049)

stracker-phil:
### Description

> Alternative to #2956 

Hides submenus in the Payments tab, when the new PayPal Settings UI is loaded.

Some plugins, such as [WooPayments](https://es.wordpress.org/plugins/woocommerce-payments/), may add extra sections to the checkout tab. Those extra sections are output as submenu-links at the beginning of the Payments tab.

This solution uses CSS to hide the submenu item


Narek13 (approved):
Thanks

PR #3043, opened 2025-02-03
Show new settings UI to new merchants only (3904)

Dinamiko:
Leverage the existing [feature flag](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules.php#L93-L98) (currently defaulted to “off”) to conditionally display either the new or old settings UI based on whether a merchant has existing settings data. Specifically:

**New Merchants (No Existing Settings):**

Should see the new UI (welcome screen) by default when the feature flag is on.

Clearing or resetting settings has no effect if there are no preexisting settings; they remain on the new UI.

**Existing Merchants (Has Existing Settings):**

Should continue to see the old UI, even if they use the “clear/reset settings” feature, as long as the plugin data persists.

If they completely remove/uninstall the plugin (incl. erasing any stored data) and then reinstall, they should be treated as a new merchant and see the new UI upon reactivation.

### Acceptance Criteria

**Important**: This package set new Settings UI feature enabled by default, so no external plugin is needed to enable the feature.

**New User Detection:**

If no settings data is detected (e.g., fresh install), the new UI is shown.

Confirm that clearing or resetting does not switch an existing user to the new UI unless all stored plugin data is genuinely removed.

**Existing User Detection:**

If any existing plugin settings data is present, the old UI is shown.

The button to switch to the new UI should not yet be displayed (only in a later update)

Confirm that simply resetting settings does not override this logic unless the plugin is fully uninstalled and reinstalled (no leftover data).

**Welcome Screen for New Merchants:**

Verify that the first-time (new) user flow displays the welcome screen for the new UI.

**Testing & Validation:**

Ensure QA covers both fresh install scenarios (triggers new UI) and upgrade scenarios from previous versions (retains old UI).

Include tests for the reset/clear settings flow to confirm correct UI behavior.

**Implementation Notes:**

The primary logic should hinge on detecting the presence (or absence) of existing settings data, alongside the state of the feature flag.

Be sure to handle edge cases where partial settings may remain or be reintroduced after reinstallation.

Maintain backward compatibility for existing merchants and ensure feature-flag toggling is respected in all scenarios.


Narek13 (approved):
@Dinamiko please check the CS, other then that LGTM. Thanks.

woocommerce-paypal-payments.php
```diff
@@ -229,4 +229,11 @@ function is_woocommerce_activated(): bool {
 		return class_exists( 'woocommerce' );
 	}
 
+	// Set new merchant flag on plugin install.
+	add_action( 'woocommerce_paypal_payments_gateway_migrate', function() {
```
stracker-phil: We could also use the `$installed_version` parameter to detect first installation:

```php
add_action( 'woocommerce_paypal_payments_gateway_migrate', static function( $version ) {
  if ( false !== $version ) {
    return;
  }

  update_option( 'woocommerce-ppcp-is-new-merchant', '1' );
});
```

woocommerce-paypal-payments.php
```diff
@@ -229,4 +229,11 @@ function is_woocommerce_activated(): bool {
...
+		$legacy_settings = get_option(Settings::KEY);
+		if(! $legacy_settings) {
+			update_option('woocommerce-ppcp-is-new-merchant', true);
```
stracker-phil: Suggestion: For clarity, I would set the same value here as the value that's used for the condition:

```php
update_option('woocommerce-ppcp-is-new-merchant', '1');  // '1' instead of true
```

stracker-phil (approved):
Approved, but have left 2 comments on possible changes

PR #3051, opened 2025-02-04
Sync `activePanel` with the URL on Initial Load (4149)

Narek13:
## Issue Description

Currently when reloading the settings screen, the “Overview” tab is automatically opened while the previously open tab should stay visible.

## PR Description

The issue is that while updating the URL with the active panel (`updateQueryString`), we're not initializing the `activePanel` state from the URL when the app loads. Since `activePanel` is initialized to 'overview' in `App.js`, it doesn't check the URL on refresh.

The PR will solve the problem by syncing the `activePanel` with the URL on reload

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/App.js
+++ b/modules/ppcp-settings/resources/js/Components/App.js
@@ -6,6 +6,7 @@ import SpinnerOverlay from './ReusableComponents/SpinnerOverlay';
 import SendOnlyMessage from './Screens/SendOnlyMessage';
 import OnboardingScreen from './Screens/Onboarding';
 import SettingsScreen from './Screens/Settings';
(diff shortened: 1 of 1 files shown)


danielhuesken (approved):
LGTM

stracker-phil (commented):
Instead of initializing the state and then instantly updating it, it would be a bit nicer to directly initialize the state with the valid option:

```js
// Instead of:
const [ activePanel, setActivePanel ] = useState( defaultPanel );
useEffect( () => setActivePanel( getQuery().panel || defaultPanel ), [] );

// This is a bit nicer:
const [ activePanel, setActivePanel ] = useState( getQuery().panel || defaultPanel );
```

stracker-phil (approved):

PR #3061, opened 2025-02-05
Limit length of Soft Descriptor (4165)

mmaymo:
This PR limits the length of the `soft descriptor` option in ExpertSettings->PayPal Settings to 22 characters, both in frontend and backend

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/ReusableComponents/Controls/SoftdescriptorInput.js
+++ b/modules/ppcp-settings/resources/js/Components/ReusableComponents/Controls/SoftdescriptorInput.js
@@ -1,0 +1,19 @@
+import ControlTextInput from './ControlTextInput';
+
+const SoftDescriptorInput = ( { value, onChange, placeholder } ) => {
+	const handleChange = ( newValue ) => {
+		if ( newValue.length <= 22 ) {
+			onChange( newValue );
+		}
+	};
+
+	return (
+		<ControlTextInput
+			value={ value }
+			onChange={ handleChange }
+			placeholder={ placeholder }
+		/>
+	);
+};
+
+export default SoftDescriptorInput;
--- a/modules/ppcp-settings/src/Data/SettingsModel.php
+++ b/modules/ppcp-settings/src/Data/SettingsModel.php
@@ -139,7 +139,8 @@ class SettingsModel extends AbstractDataModel {
 	 * @param string $descriptor The soft descriptor to set.
 	 */
 	public function set_soft_descriptor( string $descriptor ) : void {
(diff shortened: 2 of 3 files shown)


modules/ppcp-settings/resources/js/Components/ReusableComponents/Controls/SoftdescriptorInput.js
```diff
@@ -0,0 +1,19 @@
+import ControlTextInput from './ControlTextInput';
+
+const SoftDescriptorInput = ( { value, onChange, placeholder } ) => {
+	const handleChange = ( newValue ) => {
+		if ( newValue.length <= 22 ) {
+			onChange( newValue );
+		}
+	};
```
stracker-phil, line 8: As this is a unique field, I think it's not too useful to create a reusable component for it.

Suggestion 1: Move this method into the `PaypalSettings.js` file instead of creating a new component.
Suggestion 2: Wrap the method in a `useCallback()` hook.

Why useCallback:
This makes the function "stable", which means, the function is cached by React, and does not change when any of the other props change. In this specific case, it has very little impact, but it's a good practice to get used to, as it avoids unnecessarily re-rendering components.

```js
const handleChange = useCallback(( newValue ) => {
	if ( newValue.length <= 22 ) {
		onChange( newValue );
	}
}, [ onChange ]); // ← only if "onChange" prop changes, a new copy of the "handleChange" function is generated
```

modules/ppcp-settings/src/Data/SettingsModel.php
```diff
@@ -139,7 +139,8 @@ public function get_soft_descriptor() : string {
 	 * @param string $descriptor The soft descriptor to set.
 	 */
 	public function set_soft_descriptor( string $descriptor ) : void {
-		$this->data['soft_descriptor'] = $this->sanitizer->sanitize_text( $descriptor );
+		$trimmed_descriptor = substr( $descriptor, 0, 22 );
+		$this->data['soft_descriptor'] = $this->sanitizer->sanitize_text( $trimmed_descriptor );
```
stracker-phil: Practically, this is working well, but generally, we should:

- First: Sanitize the value
- Second: Modify/use the sanitized value

i.e.
```php
// First sanitize, to remove invalid characters/text (like <script> tags)
$sanitized_descriptor = $this->sanitizer->sanitize_text( ... );

// Second, trim and store the sanitized value
$this->data['soft_descriptor'] = substr( $sanitized_descriptor, 0, 22 );
```

stracker-phil (changes requested):
We need to implement an additional requirement as well - only allow those chars:
- Alphanumeric characters
- Dashes
- Asterisks
- Periods (.)
- Spaces

stracker-phil (approved):

PR #3068, opened 2025-02-06
Brand name: Default to WP site title (4164)

mmaymo:
In the legacy UI, the initial default value for the “brand name” was the WP site title.
Apply this default to the new UI - if the brand name is left empty, it should default to the site title


stracker-phil (approved):
Great work!

PR #3082, opened 2025-02-10
Visual test - advanced options enable sandbox mode

sedatbsr:
**Issue**: #

**Ticket**:

**Slack Thread**:

---

### Description

**1. linting script updated as it was not working 
2. visual test for Req ID PCP-3908 and PCP-4000 added.**

### Steps to Test

1. Go to '…'
2. Click on '…'
3. Scroll down to …'

### Documentation

### Changelog Entry

Closes # .

### DIFF
--- a/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
@@ -83,6 +83,7 @@ test.describe.serial( () => {
 		await pcpOnboarding.visit();
 		await pcpOnboarding.openAdvancedOptions();
 		await pcpOnboarding.enableSandboxMode();
+		await pcpOnboarding.page.waitForLoadState();
 		await percy.takeSnapshot( testInfo.title, percyConfig );
 	} );
 } );
--- a/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
@@ -75,6 +75,16 @@ test.describe.serial( () => {
 		await pcpOnboarding.page.waitForLoadState();
 		await percy.takeSnapshot( testInfo.title, percyConfig );
 	} );
+
+	test( 'PCP-0000 | Settings - Oboarding - Enable Sandbox mode - Default UI @percy', async ( {
+		pcpOnboarding,
+		percy,
+	}, testInfo ) => {
+		await pcpOnboarding.visit();
+		await pcpOnboarding.openAdvancedOptions();
+		await pcpOnboarding.enableSandboxMode();
+		await percy.takeSnapshot( testInfo.title, percyConfig );
+	} );
 } );
 
 test( 'PCP-0000 | Settings - Badge values per country @percy', async ( {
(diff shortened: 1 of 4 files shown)


tests/qa/utils/admin/pcp-onboarding.ts
```diff
@@ -43,9 +43,10 @@ export class PcpOnboarding extends PcpAdminPage {
 	connectToPayPalButton = () =>
 		this.page.getByRole( 'button', { name: 'Connect to PayPal' } );
 
-	enableSandboxModeLabel = () => this.page.getByText( 'Enable Sandbox Mode' );
+	enableConnectionTypeLabel = ( label: string ) =>
```
mishautkin: IMO, in this case, it would be more straightforward if locator is created specifically for "Enable Sandbox Mode" rather then parameterized. I'd keep `enableSandboxModeLabel` locator and `manuallyConnectLabel` can be added separately when needed/.

tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
```diff
@@ -83,6 +83,7 @@ test.describe.serial( () => {
 		await pcpOnboarding.visit();
 		await pcpOnboarding.openAdvancedOptions();
 		await pcpOnboarding.enableSandboxMode();
+		await pcpOnboarding.page.waitForLoadState();
```
mishautkin: 1. `page.waitForLoadState()` waits for `load`. I haven't checked but probably "networkidle" may be required to wait for requests in the background?
2. Maybe it's better to include this "wait" into the `enableSandboxMode` method? Since it'll be required to wait for it every time this method's called.

mishautkin (commented):
Looks good, but left a couple of comments.

Also, for the future, fixes of script in package.json is not relevant to the topic of current PR and can be pushed as a separate PR.

tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
sedatbsr: I found out that a small loading mask appears in front of the content. It might be because of that so will try with `networkidle`.

tests/qa/utils/admin/pcp-onboarding.ts
sedatbsr: Reverted to its initial state.

tests/qa/utils/admin/pcp-onboarding.ts
```diff
@@ -84,10 +83,9 @@ export class PcpOnboarding extends PcpAdminPage {
 		);
 		const isToggleChecked = isChecked.includes( 'is-checked' );
 		if ( ! isToggleChecked ) {
-			await this.enableConnectionTypeLabel(
-				'Enable Sandbox Mode'
-			).click();
+			await this.enableSandboxModeLabel().click();
 		}
+		this.page.waitForLoadState( 'networkidle' );
```
mishautkin: I guess this "wait" needs
1. `await`
2. to be moved inside of `if` condition

mishautkin (changes requested):

tests/qa/utils/admin/pcp-onboarding.ts
sedatbsr: Within the block, in case the toggle is already checked, the function returns immediately without waiting, I think that's you mean, right?
I can make this change. But I have a question. Don't we want all the page loaded including background activities before having a proper screenshot? Please shed some light on my path :)

mishautkin (approved):

PR #3088, opened 2025-02-11
No ACDC eligibility: Hide "Online Card Payments" (4197, 4148)

Dinamiko:
When signed in with a personal account, the “Online Card Payments” section should be completely hidden.

CleanShot 2025-02-07 at 17 43 41@2x-20250207-164439

This PR uses custom hook `useMerchant` to get the type of merchant account, then conditionally renders the component based on that value.

It also ensures correct visibility for BCDC, ACDC, Apple and Google Pay based on store country.


Narek13 (approved):
Thanks

PR #3099, opened 2025-02-12
Force enabled the styling per location option (4214)

Narek13:
## Issue Description

The styling tab data isn't showing on frontend when the “styling per location“ old setting is unchecked because the Key with “general“ prefix is returned, therefor the mapping doesn’t work.

## PR Description

Maps the “styling per location“ value and force it to always return `true` when the new settings are used

### DIFF
--- a/modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
+++ b/modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
@@ -40,11 +40,12 @@ class StylingSettingsMapHelper {
 	public function map(): array {
 
 		$mapped_settings = array(
-			'smart_button_locations'     => '',
-			'pay_later_button_locations' => '',
-			'disable_funding'            => '',
-			'googlepay_button_enabled'   => '',
-			'applepay_button_enabled'    => '',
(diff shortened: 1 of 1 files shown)


modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
```diff
@@ -70,6 +71,9 @@ public function mapped_value( string $old_key, array $styling_models ) {
 			case 'smart_button_locations':
 				return $this->mapped_smart_button_locations_value( $styling_models );
 
+			case 'smart_button_enable_styling_per_location':
+				return true;
```
stracker-phil, line 75: Does it make sense to add a filter here to allow programmatically disabling this to use a single styling option for all locations?

stracker-phil (approved):

modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
Narek13: @stracker-phil I think it doesn't make sense for the new settings, as it doesn't have any option for using a single styling

PR #3103, opened 2025-02-13
Fix non wc gateway items null enabled key

Dinamiko:
Fix broken payment methods save when no `enabled` key is provided for non wc gateway items (ex. venmo).


Narek13 (approved):
LGTM

PR #3114, opened 2025-02-17
qa - uk payinthree instead of paylater

sedatbsr:
**Issue**: #

**Ticket**:

**Slack Thread**:

---

### Description

**Test for the first onboarding screen for the UK added. Verifies that the first screen displays Pay in 3 instead of Pay Later**

### Steps to Test

1. Go to '…'
2. Click on '…'
3. Scroll down to …'

### Documentation

### Changelog Entry

Closes # .

### DIFF
--- a/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
@@ -101,13 +101,13 @@ test.describe.serial( () => {
 		percy,
 		wooCommerceApi,
 	}, testInfo ) => {
-		await pcpOnboarding.visit();
-		await pcpOnboarding.gotoInitialOnboardingPage();
 		await wooCommerceApi.updateGeneralSettings( {
 			woocommerce_default_country: 'GB',
 			woocommerce_currency: 'GBP',
 		} );
-		await pcpOnboarding.page.reload();
+		await pcpOnboarding.visit();
+		await pcpOnboarding.gotoInitialOnboardingPage();
+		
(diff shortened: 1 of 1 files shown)


tests/qa/tests/02-dashboard-ui/visual-default-ui.spec.ts
```diff
@@ -95,6 +95,21 @@ test.describe.serial( () => {
...
 	} );
+
+	test.only( 'PCP-0000 | Settings - UK - Onboarding - Default UI (acdc, paylater) - Default UI @percy', async ( {
+		pcpOnboarding,
+		percy,
+		wooCommerceApi,
+	}, testInfo ) => {
+		await pcpOnboarding.visit();
+		await pcpOnboarding.gotoInitialOnboardingPage();
+		await wooCommerceApi.updateGeneralSettings( {
```
mishautkin, line 104: Minor comment:
if you call `wooCommerceApi.updateGeneralSettings` before `pcpOnboarding.visit` - no need to reload the page. This would speed up test for a few seconds. In this case there's only one `visit` and no loop. I guess you inherited this from my examle for badges tests.

sedatbsr: Actually, the reason why I visited page first was to observe the badges before the doing the country change via api. After request is fulfilled, I reload the page and expect the change is done. But since there is no screenshot taken at the first visit (and there's actually no need), it will be better to make request then visit the page and take the screenshot. I am changing it.

mishautkin (approved):

PR #3115, opened 2025-02-17
Features: refactor to use rest endpoints (4210)

mmaymo:
This PR addresses: To conditionally output Features in the Overview tab (and to keep logic in the frontend component as minimal as possible), we should move the Features data from [a static file](https://github.com/woocommerce/woocommerce-paypal-payments/blob/69a9090e4e1f4cc7742d85db5955df32abec8b17/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/Overview/features-config.js) to a REST endpoint.


danieldudzic (changes requested):
Not sure we need the `Data/FeaturesSettings.php` It doesn't seem like it's storing any persistent state. I think it can be removed.

modules/ppcp-settings/resources/js/data/features/actions.js
```diff
@@ -0,0 +1,135 @@
...
+		selectTab( TAB_IDS.PAYMENT_METHODS, 'ppcp-card-payments-card' ).then(
+			() => {
+				setActiveModal( 'ppcp-applepay' );
+			}
+		);
+	},
+	pay_later: () => {
+		selectTab( TAB_IDS.PAY_LATER_MESSAGING );
+	},
+};
```
danieldudzic: I believe we can move this to the definition file and handle it like todos. More info in: PCP-4174

danieldudzic (commented):
Very good job overall, I'd just add some minor tweaks and it will be ready.

stracker-phil (approved):
Well done

danieldudzic (approved):

PR #3118, opened 2025-02-17
Check the i18n translations in React app (4038)

mmaymo:
The JS translations are not autoloaded for custom JS (only for Blocks) 
This PR adds `wp_set_script_translations` to address that.


danielhuesken (approved):

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -198,11 +206,14 @@ static function ( $hook_suffix ) use ( $container ) {
 					wp_enqueue_script(
 						'ppcp-paylater-configurator-lib',
 						'https://www.paypalobjects.com/merchant-library/merchant-configurator.js',
-						array(),
+						['wp-i18n'],
```
stracker-phil: The short array syntax is not allowed - we need to use `array()`

stracker-phil (changes requested):
Looks good, content-wise

But code format needs some adjustments (some spaces are missing, and we need to use `array()` instead of `[]`)

stracker-phil (approved):

PR #3124, opened 2025-02-18
Store and use the Merchant country from PayPal (4155)

Dinamiko:
This PR gets the `country` parameter from `v1.customer.partners.*.merchant-integrations.*` response and set its value to `country_code` flag for new merchants. 

### Acceptance Criteria
- After onboarding, check the available features and payment methods
- Update the WooCommerce country to something else (e.g. Zimbabwe) and confirm that all Features and payment methods are still available.

### DIFF
--- a/modules/ppcp-api-client/src/Entity/SellerStatus.php
+++ b/modules/ppcp-api-client/src/Entity/SellerStatus.php
@@ -81,8 +81,12 @@ class SellerStatus {
 		return $this->capabilities;
 	}
 
+	public function country() : string {
+		return $this->country;
+	}
+
 	/**
-	 * Returns the enitity as array.
+	 * Returns the entity as array.
 	 *
 	 * @return array
 	 */
--- a/modules/ppcp-api-client/src/Entity/SellerStatus.php
+++ b/modules/ppcp-api-client/src/Entity/SellerStatus.php
@@ -29,14 +29,22 @@ class SellerStatus {
 	private $capabilities;
 
 	/**
+	 * Merchant country on PayPal.
+	 *
+	 * @var string
+	 */
+	private $country;
+
+	/**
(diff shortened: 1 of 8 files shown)


modules/ppcp-api-client/src/Entity/SellerStatus.php
```diff
@@ -28,15 +28,23 @@ class SellerStatus {
 	 */
 	private $capabilities;
 
+	/**
+	 * Merchant country on PayPal.
+	 *
+	 * @var string
+	 */
+	private $country;
```
Narek13:
```suggestion
	private string $country;
```

modules/ppcp-settings/src/Service/AuthenticationManager.php
```diff
@@ -267,10 +286,17 @@ public function authenticate_via_oauth( bool $use_sandbox, string $shared_id, st
...
+		try {
+			$seller_status = $this->partners_endpoint->seller_status();
+		} catch ( PayPalApiException $exception ) {
+			$seller_status = null;
+		}
+
+		$connection->is_sandbox       = $use_sandbox;
+		$connection->client_id        = $credentials['client_id'];
+		$connection->client_secret    = $credentials['client_secret'];
+		$connection->merchant_id      = $credentials['merchant_id'];
```
Narek13, line 298: Are we sure these keys exist in `$credentials` or should we use it like `$credentials['client_id'] ?? ''`

stracker-phil: Re @Narek13: Those three items are **always** included in the return value of `get_credentials()`. It's a private method in the same class, so we have full control over the return value and can trust it will stay compatible.

stracker-phil (approved):

PR #3148, opened 2025-02-25
Qa fixes

mishautkin:
### Description

Fix minor inconsistencies in playwright project (`/tests/qa`).

### Changelog Entry

- Fix settings fixtures URLs
- Remove testIgnore from project 'all' in playwright.config
- Create installAndActivatePcp outside of configurePcp in utils.ts


tests/qa/utils/urls.ts
sedatbsr, line 1: Misha, I connected with business account and saw that URLs do not match with the updated ones. There is still `ppcp-tab=ppcp-connection` part in URL. But it also works when I move the that part. In case we need to assert URL, wont it fail?

mishautkin: @Sedat, yes I noticed. I think that this might be reported because on one hand it's not harming the functioning but on the other hand the pages aren't declared inambiguously... I would claritfy this with devs.

sedatbsr (approved):

sedatbsr (approved):

PR #3158, opened 2025-02-26
Qa links learn more and link for fees in footer

sedatbsr:
### Description

Test case for Settings - _**Country**_ - Onboarding - Links Learn more and link for fees in footer added.
* Data file created under _.test-data_` learn-more-links.data.ts`
* A new .spec file has been created under _dashboard-ui_ `learn-more-links.spec.ts`
* Unused imports removed from .spec files
* A locator and functions have been to _pcp-onboarding_ POM file for the new test case.

### DIFF
--- a/tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts
@@ -11,35 +11,40 @@ test.describe( () => {
 	const countries = Object.keys( learnMoreLinksByCountry );
 
 	for ( const country of countries ) {
-		test.only( learnMoreLinksByCountry[country].testTitle, async ( {
-			pcpOnboarding,
-			wooCommerceApi,
-		} ) => {
-			const expectedLinks = learnMoreLinksByCountry[ country ].links;
+		test(
+			learnMoreLinksByCountry[ country ].testTitle,
+			async ( { pcpOnboarding, wooCommerceApi } ) => {
+				const expectedLinks = learnMoreLinksByCountry[ country ].links;
 
-			await wooCommerceApi.updateGeneralSettings( {
-				woocommerce_default_country: country,
-				woocommerce_currency: 'USD'
-			} );
+				await wooCommerceApi.updateGeneralSettings( {
+					woocommerce_default_country: country,
+					woocommerce_currency: 'USD',
+				} );
 
-			await pcpOnboarding.visit();
-			await pcpOnboarding.page.waitForLoadState( 'networkidle' );
-			for ( const { url, title } of expectedLinks ) {
-				const link = await pcpOnboarding.page.locator( `a[href="${ url }"]` );
-				await expect.soft( link ).toBeVisible();
-				
-				if (await link.isVisible()) {
-				const newContext = await pcpOnboarding.page.context().browser()?.newContext();
-					const newPage = await newContext?.newPage();
-					await newPage?.goto( url );
+				await pcpOnboarding.visit();
+				await pcpOnboarding.page.waitForLoadState( 'networkidle' );
+				for ( const { url, title } of expectedLinks ) {
+					const link = await pcpOnboarding.page.locator(
+						`a[href="${ url }"]`
+					);
+					await expect.soft( link ).toBeVisible();
 
-				expect.soft( newPage.url() ).toBe( url );
-				expect.soft( await newPage.title() ).toContain( title );
-				
-				await newPage.close();
-				await newContext.close();
+					if ( await link.isVisible() ) {
+						const newContext = await pcpOnboarding.page
+							.context()
+							.browser()
+							?.newContext();
+						const newPage = await newContext?.newPage();
+						await newPage?.goto( url );
+
+						expect.soft( newPage.url() ).toBe( url );
+						expect.soft( await newPage.title() ).toContain( title );
+
+						await newPage.close();
+						await newContext.close();
+					}
 				}
 			}
-		} );
+		);
 	}
 } );
--- a/tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts
@@ -1,0 +1,45 @@
+/**
+ * External dependencies
+ */
+/**
+ * Internal dependencies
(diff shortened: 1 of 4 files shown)


tests/qa/utils/admin/pcp-onboarding.ts
```diff
@@ -108,5 +110,49 @@ export class PcpOnboarding extends PcpAdminPage {
 		}
 	};
 
+	getLearnMoreLinks = async (): Promise<
+		{ url: string; title: string }[]
+	> => {
+		await this.page.waitForSelector( '.ppcp-r-welcome-docs a', {
```
mishautkin: I see that you have the locator created. Please add locator and use `.waitFor()`:
```ts
wait this.page.learnMoreLinks().waitFor( { state: 'visible' } );
```

tests/qa/utils/admin/pcp-onboarding.ts
```diff
@@ -108,5 +110,49 @@ export class PcpOnboarding extends PcpAdminPage {
...
+					? href
+					: new URL( href, 'https://www.paypal.com' ).href;
+
+				extractedLinks.push( {
+					url: fullUrl,
+					title: await link.innerText(),
+				} );
+			}
+		}
+		console.log( 'extractedLinks:', extractedLinks );
```
mishautkin: Remove `console.log`

tests/qa/utils/admin/pcp-onboarding.ts
```diff
@@ -108,5 +110,49 @@ export class PcpOnboarding extends PcpAdminPage {
...
+		const [ newPage ] = await Promise.all( [
+			this.page.context().waitForEvent( 'page' ),
+			link.first().click(),
+		] );
+
+		await newPage.waitForLoadState( 'domcontentloaded' );
+		return newPage;
+	};
+
+	getPageTitle = async ( newPage = this.page ) => {
```
mishautkin: This is just a one-line method. May be you may directly use `await newPage.title();` in a test?

tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+			wooCommerceApi,
+		} ) => {
+			await wooCommerceApi.updateGeneralSettings( {
+				woocommerce_default_country: country,
+				woocommerce_currency: 'USD'
+			} );
+
+			await pcpOnboarding.visit();
+			await pcpOnboarding.page.waitForLoadState( 'networkidle' );
+			const pageLinks = await pcpOnboarding.getLearnMoreLinks();
```
mishautkin: I hope that I understood the purpose of `getLearnMoreLinks` correctly. I would suggest to use soft assertions instead.

In this case you don't need to use links array comparison and can avoid additional methods in `pcpOnboarding`, unless there are more tests to use same functionality. I give example of my suggested test version in one code block (but it can be split into methods if to be reused):
```ts
await pcpOnboarding.visit();
await pcpOnboarding.waitForLoadingMaskRemoved();
for ( const { url, title } of expectedLinks ) {
    const link = this.pcpOnboarding.learnMoreLinks().locator( `[href="${ url }"]` );
    // Assert that expected link is visible but do not stop test if not
    await expect.soft( link ).toBeVisible();
    // If link is visible
    if (await link.isVisible()) {
        // Open link in a new context to not revisit Onboarding again
	const newContext = await this.page.context().browser()?.newContext();
        const newPage = await newContext?.newPage();
        await newPage?.goto(link);
	// Assertions
	expect.soft( newPage.url() ).toBe( url );
	expect.soft( await newPage.title() ).toContain( title );
        // Close new page and context
	await newPage.close();
	await newContext.close();
    }
}

File: tests/qa/utils/admin/pcp-onboarding.ts (hunk 1 above)
Comment (sedatbsr) [reply tests/qa/utils/admin/pcp-onboarding.ts]: I tried this but because of the PW strict mode, it fails. Because it returns more than one element. I can remove the locator and use it directly inside the function.

Comment (sedatbsr) [reply tests/qa/utils/admin/pcp-onboarding.ts]: Alright. If I wait for the `.first()` one, it works.

File: tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts (hunk 4 above)
Comment (sedatbsr) [reply tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts]: So you mean that there is no need to create a function and directly write the code test block, right? I didn't think of soft assertion's usage which sounds reasonable.

Comment (mishautkin) [reply tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts]: I'd say, Having these methods in `pcpOnboarding` is reasonable if you're going to reuse them more then once across several specs.

If it's only used in one test - you can place code in a test directly or create a function in a spec file, if the code is bulky (or used in more than one test of a spec).

If "Learn more" links are going to be tested across several fixtures then methods could be placed in `pcp-admin-page.ts`. This would allow all pcp-fixtures to inherit them.

Comment (sedatbsr) [reply tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts]: It didn't work no matter how much I tried. It fails when it comes to `await expect.soft( link ).toBeVisible();` 
Here is what I did:
* I checked the locators and printed in the console to be sure what it returns.
* Tried a the waits including `waitForLoadingMaskRemoved`. 
* Used timeout in locator.
* waited before soft assertion
* Checked if the hrefs are dynamically loaded

File: tests/qa/tests/02-dashboard-ui/learn-more-links.spec.ts [hunk 5]
```diff
@@ -11,34 +11,34 @@ test.describe( () => {
 	const countries = Object.keys( learnMoreLinksByCountry );
 
 	for ( const country of countries ) {
-		test( `PCP-0000 | Settings -  ${ country } - Onboarding - Links Learn more and link for fees in footer`, async ( {
+		test.only( learnMoreLinksByCountry[country].testTitle, async ( {
```
mishautkin: Remove `.only`, please.

mishautkin (commented):
A minor comment. Also I'm not sure if the initial methods are present in `pcpOnboarding`. You may stash them locally but please check if they're not in the commit if they're not used.

mishautkin (approved):

PR #3160, opened 2025-02-26
Phone Number not saved in WC order when using Pay Now experience (4257)

danielhuesken:
- Add phone number from Payer to block checkout
- Add phone number from Payer to check out without confirmation


Narek13 (approved):
LGTM

Dinamiko (approved):

PR #3163, opened 2025-02-26
Implement 3D secure check for Google Pay (4156)

stracker-phil:
### Description

Adjust the Google Pay logic to accept more test cards by removing the "PAYMENT_AUTHORIZATION" intent

### Screenshots

The new Google Pay sheet with additional test cards:
CleanShot 2025-02-26 at 18 25 10


Dinamiko (approved):

modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
```diff
@@ -271,10 +268,79 @@ public function handle_request(): bool {
...
+		 * Note: PayPal is currently investigating this case.
+		 * Maybe the order status is wrong and should be ACCEPTED, in that case,
+		 * we could drop the condition and always run proceed_with_order().
+		 */
+		if ( $order_status->is( OrderStatus::PAYER_ACTION_REQUIRED ) ) {
+			$proceed = $this->threed_secure->proceed_with_order( $order );
+		} elseif ( 'card' === $source_name ) {
+			// For credit cards, we also check the 3DS response.
+			$proceed = $this->threed_secure->proceed_with_order( $order );
+		}
```
Narek13, line 315: Why instead of the `if/else` we can't just use

```
if (
    $order_status->is( OrderStatus::PAYER_ACTION_REQUIRED )
    || 'card' === $source_name
) {
    $proceed = $this->threed_secure->proceed_with_order( $order );
}
```

modules/ppcp-button/src/Endpoint/ApproveOrderEndpoint.php
```diff
@@ -271,10 +268,79 @@ public function handle_request(): bool {
...
+
+			case ThreeDSecure::REJECT:
+				// Rejection case 2, payment was rejected.
+				throw new RuntimeException(
+					__(
+						'Unfortunately, we can\'t accept your card. Please choose a different payment method.',
+						'woocommerce-paypal-payments'
+					)
+				);
+		}
```
Narek13, line 344: Are we certain that $proceed will always be one of the expected ThreeDSecure::* constants? If not, we should add a default case to handle unexpected values and avoid silent failures.

Also, since both `PROCEED` and `NO_DECISION` lead to the same outcome (return), we can group them to simplify the switch.


```
switch ( $proceed ) {
	case ThreeDSecure::PROCEED:
	case ThreeDSecure::NO_DECISION:
		// Payment can proceed.
		return;

	case ThreeDSecure::RETRY:
		// Rejection case 1, verification can be retried.
		throw new RuntimeException(
			__(
				'Something went wrong. Please try again.',
				'woocommerce-paypal-payments'
			)
		);

	case ThreeDSecure::REJECT:
		// Rejection case 2, payment was rejected.
		throw new RuntimeException(
			__(
				'Unfortunately, we can\'t accept your card. Please choose a different payment method.',
				'woocommerce-paypal-payments'
			)
		);

	default:
		// Failsafe: unknown status
		throw new RuntimeException(
			__(
				'Unexpected 3DS decision. Please try again or contact support.',
				'woocommerce-paypal-payments'
			)
		);
}

```

modules/ppcp-button/src/Helper/ThreeDSecure.php
```diff
@@ -118,42 +134,40 @@ public function return_decision( int $decision, Order $order ) {
...
+			}
+
+			if ( $enrollment === AuthResult::ENROLLMENT_STATUS_UNAVAILABLE ) {
+				return self::PROCEED;
+			}
+
+			if ( $enrollment === AuthResult::ENROLLMENT_STATUS_NO ) {
+				return self::PROCEED;
+			}
 		}
```
Narek13, line 153: This can be also simplified.

```suggestion
		if ( ! $authentication && in_array( $enrollment, [
			AuthResult::ENROLLMENT_STATUS_BYPASS,
			AuthResult::ENROLLMENT_STATUS_UNAVAILABLE,
			AuthResult::ENROLLMENT_STATUS_NO,
		], true ) ) {
			return self::PROCEED;
		}
```

Narek13 (approved):
Thank you for your work on this. I’ve left a few suggestions, but none are blocking approval. 

There are some `if/else` combinations which can be avoided/simplified but if I understood correctly they are added to have a better documentation

PR #3174, opened 2025-02-28
Error during initialization: TypeError: window[this.namespace].Applepay is not a function (4275)

Dinamiko:
There is this error in frontend (ex. in Single Product page) even if Apple Pay is disabled in Payment Methods screen:

Captura de pantalla 2025-02-28 a las 16 13 09

This PR ensures that scripts are loaded only if the payment method is enabled.


Narek13 (approved):
LGTM

PR #3211, opened 2025-03-13
Hide the gateways if they are not enabled (4303)

Narek13:
## Issue Description

Currently, when installing PayPal Payments and before connecting a PayPal account, these additional payment methods may be displayed:

image-20250311-172500

## PR Description
- ensure no payment methods other than PayPal are displayed on Payments tab before connecting a PayPal account
- hide all payment methods on the Payments tab when they are disabled or not available in the Payment Methods tab in PayPal Payments settings

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -467,7 +467,6 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 					);
 				}
 
-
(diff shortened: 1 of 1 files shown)


stracker-phil (approved):
I don't really like to add another `get_option( "woocommerce_{$gateway_name}_settings", array() );` here, but we can refactor this in a future PR - currently we have this exact same function in 3 places

PR #3225, opened 2025-03-17
Pay Later messaging component active with Configure button when "Save PayPal and Venmo" is active (4351)

Dinamiko:
Despite using PayPal Vaulting, in the Overview tab the Pay Later Messaging component is visible and says Active.

This PR removes Active badge and Configure button form Pay Later feature item when setting Save PayPal and Venmo is enabled.


Narek13 (approved):
LGTM. Thanks

PR #3244, opened 2025-03-19
PHP: Create branded-experience detection (4390)

Dinamiko:
This PR adds detection of “branded experience” which happens during plugin activation. It also persist the value into the data store and provides a new custom hook to get this data in components.

### Requirements

- The flag must not change after it was set once
- When the merchant uses “Disconnect + Start Over” remove the branding
- When uninstalling the plugin, remove the branding
    - “*Remove the branding*” … set it to `DIRECT` or delete the setting
- Validation, existing merchants
    - When accessing the flag, it must be validated to guarantee a correct response
    - If the flag has no value or is invalid (e.g. for existing merchants), set it to `DIRECT`
        - Branding can only happen during plugin installation for new merchants

### How to test

1 “Direct” installation path, this is the default value when no WooCommerce path is detected:

- On a fresh WP install install and activate the plugin
    - Or reset database, activate plugin, manually change plugin version to 3.0.1 in plugin header and then reload the page
- Ensure installation path is “direct”
    - Go to plugin settings page, in the browser console type the following:

```
wp.data.select('wc/paypal/common').persistentData()
```

It should return an object including the following key and value:

```
installationPath: "direct"
```

2 “WC profiler“ path, when the plugin was attached via the WC profiler path:

- TODO

3 “Payments settings“ path, when the plugin was attached via the Payments settings page:

- TODO


modules/ppcp-settings/src/Service/BrandedExperience/ActivationDetector.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace WooCommerce\PayPalCommerce\Settings\Service\BrandedExperience;
+
+/**
+ * Class that includes detection logic for Branded Experience.
+ */
+class ActivationDetector {
+	const CORE_PROFILER    = 'core-profiler';
+	const PAYMENT_SETTINGS = 'payment-settings';
+	const DIRECT           = 'direct';
```
Narek13: missing the visibilities for these constants.

modules/ppcp-settings/src/Data/GeneralSettings.php
```diff
@@ -257,4 +260,24 @@ public function get_merchant_country() : string {
...
+	 * @param string $installation_path The branded experience installation path.
+	 *
+	 * @return void
+	 */
+	public function set_installation_path( string $installation_path ) : void {
+		$this->data['installation_path'] = $installation_path;
+	}
+
+	/**
+	 * Gets branded experience installation path.
```
Narek13:
```suggestion
	 * Retrieves the installation path for the branded experience.
```

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -664,4 +683,17 @@ protected function is_gateway_enabled( string $gateway_name ): bool {
 
 		return $gateway_enabled === 'yes';
 	}
+
+	/**
+	 * Get and persist the branded experience path.
```
Narek13:
```suggestion
	 * Detects and persists the branded experience installation path.
```

Narek13:
Thanks @Dinamiko for working on this. Just a small suggestions. Would be nice to set the visibilities of the installation path constants to `public` so I can use them when creating the service for BN Codes mapping for different installation paths.

```
'api.bn-codes'                                   => static function (): array {
		return array(
			'core-profiler'    => 'WooPPCP_Ecom_PS_CoreProfiler',
			'payment-settings' => 'WooPPCP_Ecom_PS_PaymentSettings',
		);
	},
```

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -664,4 +683,17 @@ protected function is_gateway_enabled( string $gateway_name ): bool {
...
+	 *
+	 * @param ActivationDetector $detector The branded experience path detector.
+	 * @param GeneralSettings    $general_settings The general settings data model.
+	 *
+	 * @return void
+	 */
+	private function get_and_persist_branded_experience_path( ActivationDetector $detector, GeneralSettings $general_settings ): void {
+		$general_settings->set_installation_path( $detector->detect_activation_path() );
```
stracker-phil: Is this logic correct?

It will update the installation path on every plugin update, and always use the WooCommerce option to define the value.

This does not allow merchants to opt-out of the branded-experience by uninstalling + reinstalling the plugin from the wp.org repo... Also, if the woo-option is deleted or modified, it could happen that after a plugin update a "branded-only" merchant suddenly can see all credit card payment options.

Suggestion: Only set the value if it's empty to make it more stable.
Also, we need a way to unset the path when a merchant re-installs the plugin from the wp.org plugin repo to "unbrand" it

stracker-phil (approved):

PR #3246, opened 2025-03-20
Integration tests setup improvements

Dinamiko:
This PR ensures ddev activates WC plugin before PayPal plugin. It also adds `skip-ci` PHPUnit group to skip tests in CI.


Narek13 (approved):
Thanks for this

PR #3250, opened 2025-03-20
Fix post onboarding payment methods frontend/backend sync (4340)

danieldudzic:
### Description

This PR will address an issue where payment methods are not properly synchronized between the frontend and backend after completing the onboarding process. When a merchant completes the onboarding flow, certain payment methods that should be enabled based on their selections (like ACDC) are not being activated in the new Settings UI.

The core issues are:
1. ACDC as a gateway gets registered too late for us to toggle it during page load.
2. No mechanism to refresh payment gateway data after onboarding flags get set. This causes issues hiding payment gateways based on country/currency/misc exclusion rules.

## Changes

- This PR will add state tracking flags (`gateways_synced` and `gateways_refreshed`) to the onboarding profile
- <strike>Create new REST endpoints for gateway synchronization and refreshing</strike>
- Implement custom hooks `usePaymentGatewaySync` and `usePaymentGatewayRefresh` to handle the synchronization process
- Add proper Redux actions and state management for tracking gateway sync status
- Modify the payment method card component to wait for gateway refresh before rendering
- Move the payment gateway synchronization to be initialized from the Overview tab after onboarding

## Steps to Test

1. Complete a fresh PayPal onboarding process
2. Ensure you selected to accept card payments during onboarding 
3. Navigate to PayPal settings > Payment methods tab
4. Verify that all eligible payment methods (based on onboarding selections) are properly enabled
5. Ensure the payment method card UI correctly reflects the enabled/disabled state of each method
6. Disconnect and repeat the process and verify that everything is working as expected

### DIFF
--- a/modules/ppcp-settings/resources/js/data/onboarding/actions.js
+++ b/modules/ppcp-settings/resources/js/data/onboarding/actions.js
@@ -127,7 +127,7 @@ export const updateGatewaysRefreshed = ( refreshed = true ) =>
  */
 export function syncGateways() {
 	return async ( { dispatch } ) => {
-		dispatch( { type: ACTION_TYPES.SYNC_GATEWAYS } );
+		dispatch( setPersistent( 'gatewaysSynced', true ) );
 		await dispatch( persist() );
 		return { success: true };
 	};
@@ -135,7 +135,7 @@ export function syncGateways() {
 
 export function refreshGateways() {
 	return async ( { dispatch } ) => {
-		dispatch( { type: ACTION_TYPES.REFRESH_GATEWAYS } );
+		dispatch( setPersistent( 'gatewaysRefreshed', true ) );
 		await dispatch( persist() );
 		return { success: true };
 	};
--- a/modules/ppcp-settings/resources/js/data/onboarding/actions.js
+++ b/modules/ppcp-settings/resources/js/data/onboarding/actions.js
@@ -100,3 +100,43 @@ export function refresh() {
 		select.persistentData();
 	};
 }
+
+/**
+ * Persistent. Updates the gateway synced status.
+ *
+ * @param {boolean} synced The sync status to set
+ * @return {Action} The action.
+ */
+export const updateGatewaysSynced = ( synced = true ) =>
+	setPersistent( 'gatewaysSynced', synced );
+
+/**
+ * Persistent. Updates the gateway refreshed status.
+ *
+ * @param {boolean} refreshed The refreshed status to set
+ * @return {Action} The action.
+ */
+export const updateGatewaysRefreshed = ( refreshed = true ) =>
+	setPersistent( 'gatewaysRefreshed', refreshed );
+
+/**
+ * Action creator to sync payment gateways.
+ * This will both update the state and persist it.
+ *
+ * @return {Function} The thunk function.
+ */
+export function syncGateways() {
+	return async ( { dispatch } ) => {
+		dispatch( { type: ACTION_TYPES.SYNC_GATEWAYS } );
+		await dispatch( persist() );
+		return { success: true };
+	};
+}
+
+export function refreshGateways() {
+	return async ( { dispatch } ) => {
+		dispatch( { type: ACTION_TYPES.REFRESH_GATEWAYS } );
+		await dispatch( persist() );
+		return { success: true };
+	};
+}
--- a/modules/ppcp-settings/resources/js/data/onboarding/actions.js
+++ b/modules/ppcp-settings/resources/js/data/onboarding/actions.js
@@ -128,7 +128,7 @@ export const updateGatewaysRefreshed = ( refreshed = true ) =>
 export function syncGateways() {
 	return async ( { dispatch } ) => {
 		dispatch( setPersistent( 'gatewaysSynced', true ) );
-		await dispatch( persist() );
+		await dispatch.persist();
 		return { success: true };
 	};
 }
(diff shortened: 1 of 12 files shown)


Dinamiko (approved):

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -608,19 +610,28 @@ static function ( $disable_funding ) use ( $container ) {
...
+		);
 
-				$payment_settings->save();
+		// Toggle payment gateways after onboarding based on flags.
+		add_action(
+			'init',
+			function() use ( $container ) {
+				$settings_data_manager = $container->get( 'settings.service.data-manager' );
+				assert( $settings_data_manager instanceof SettingsDataManager );
+				$settings_data_manager->init_hooks();
```
stracker-phil: Do we need the 'init' action here?

Also, I think that the ideal place to register the action callback is actually inside this module, and not in the `SettingsDataManager` (a service should be "doing" things and not "controlling" the flow; i.e. it's perfect to do_action() but the add_action() is more a responsibility of the `SettingsModule`).

My suggestion for this hook, to keep things more concise:

```php
add_action(
  'woocommerce_paypal_payments_sync_gateways',
  static function() use ( $container) {
    $settings_data_manager = $container->get( 'settings.service.data-manager' );
    assert( $settings_data_manager instanceof SettingsDataManager );
    $settings_data_manager->sync_gateway_settings();
  }
);
```

modules/ppcp-settings/src/Service/SettingsDataManager.php
```diff
@@ -206,14 +203,40 @@ protected function apply_configuration( ConfigurationFlagsDTO $flags ) : void {
...
+
 	/**
 	 * Enables or disables payment gateways depending on the provided
 	 * configuration flags.
 	 *
 	 * @param ConfigurationFlagsDTO $flags Shop configuration flags.
 	 * @return void
 	 */
-	protected function toggle_payment_gateways( ConfigurationFlagsDTO $flags ) : void {
+	public function toggle_payment_gateways( ConfigurationFlagsDTO $flags ) : void {
```
stracker-phil: Possibly, a left over from some earlier tests 😉

I would recommend making the method protected/private again as it's an internal API

modules/ppcp-settings/src/Endpoint/OnboardingRestEndpoint.php
```diff
@@ -178,4 +212,49 @@ public function update_details( WP_REST_Request $request ) : WP_REST_Response {
...
+	 * @return WP_REST_Response The response.
+	 */
+	public function sync_gateways( WP_REST_Request $request ) : WP_REST_Response {
+		// Execute the sync action
+		do_action( 'woocommerce_paypal_payments_sync_gateways' );
+
+		// Update gateways_synced in the profile.
+		$data                    = $this->profile->to_array();
+		$data['gateways_synced'] = true;
+		$this->profile->from_array( $data );
```
stracker-phil: Is there a reason for using `to_array()` and `from_array()` conversions? After all, we have implemented setters which make the process more type-safe and avoid assignment of incorrect values:

```php
$this->profile->set_gateway_synced( true );
$this->profile->save();
```

modules/ppcp-settings/src/SettingsModule.php
danieldudzic: Very good point 🙏

modules/ppcp-settings/resources/js/data/onboarding/actions.js
```diff
@@ -100,3 +100,43 @@ export function refresh() {
...
+ * This will both update the state and persist it.
+ *
+ * @return {Function} The thunk function.
+ */
+export function syncGateways() {
+	return async ( { dispatch } ) => {
+		dispatch( { type: ACTION_TYPES.SYNC_GATEWAYS } );
+		await dispatch( persist() );
+		return { success: true };
+	};
```
stracker-phil: While this is correct, our Redux stores generally use the (badly named) helpers `setPersistent()` or `setTransient()`

A more aligned way for this action would be:

```js
export function syncGateways() {
  return async ( { dispatch } ) => {
    setPersistent( 'gatewaysSynced', true );
    await dispatch.persist(); // note that dispatch is not a function but an object
  }
}
```

stracker-phil: But it looks like this redux action is doing almost the same thing as the REST endpoint `POST /wp-json/wc/v3/wc_paypal/onboarding/sync-gateways`:

it changes the `gatewaysSynced` flag in the Redux store and calls persist() to save the change to the WP database, effective triggering another REST call to the endpoint `POST /wp-json/wc/v3/wc_paypal/onboarding`

So: I think it would be cleaner to remove the two new REST endpoints and leverage the existing method `OnboardingRestEndpoint::update_details()` ([here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/e992f45c06f1b2a75b6623e7191e12cfbec09900/modules/ppcp-settings/src/Endpoint/OnboardingRestEndpoint.php#L204))

And integrate the update logic in the `OnbardingProfile` itself, specifically this setter:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/e992f45c06f1b2a75b6623e7191e12cfbec09900/modules/ppcp-settings/src/Data/OnboardingProfile.php#L223

```php
public function set_gateways_synced( bool $synced ): void {
  if ($this->data['gateways_synced'] ) {
    return; // The flag cannot be reverted to false!
  }
  if ( $synced ) {
    // When enabling the flag, call the action.
    do_action( 'woocommerce_paypal_payments_sync_gateways' );
  }

  $this->data['gateways_synced'] = $synced;
}
```

stracker-phil (approved):

PR #3251, opened 2025-03-20
Introduce the `PartnerAttribution` helper for handling the BN codes (4391)

Narek13:
## PR Description

- Implement the `PartnerAttribution` helper, which allows to persist the BN code based on the installation path via `initialize_bn_code()` which only inserts a new wp-option when no value is present in the DB.
- Refactor the code to use the new helper class for retrieving the BN code
  - `WooPPCP_Ecom_PS_CoreProfiler` if the installation path is `core-profiler`
  - `Woo_PPCP` -> the old existing value if the installation path is anything else
- Filters the request arguments to add the `'PayPal-Partner-Attribution-Id'` header. This ensures that all API requests include the appropriate BN code retrieved from the `PartnerAttribution` helper. Using this approach avoids the need or extensive refactoring of existing classes that use the `RequestTrait`.


modules/ppcp-api-client/services.php
```diff
@@ -963,4 +964,21 @@ static function( ContainerInterface $container ): PurchaseUnitSanitizer {
...
 		return CONNECT_WOO_URL;
 	},
+	/**
+	 * BN Codes mapping for different installation paths.
+	 *
+	 * @returns array<string, string> The map of installation paths to BN Codes.
+	 */
+	'api.bn-codes'                                   => static function (): array {
+		return array(
+			'core-profiler' => 'WooPPCP_Ecom_PS_CoreProfiler',
```
stracker-phil: I think this was a misunderstanding:

We currently have 2 paths in our detection logic, but should map both to the same BN code. We are not sure if/which path will be removed, so we should keep both mappings until those details are clarified. Also, please use the defined const values instead of "magic strings"

```php 
return array(
  ActivationDetector::CORE_PROFILER    => 'WooPPCP_Ecom_PS_CoreProfiler',
  ActivationDetector::PAYMENT_SETTINGS => 'WooPPCP_Ecom_PS_CoreProfiler',
)
```

stracker-phil: Also, I recommend dropping this service and moving the array directly into the `"api.helper.partner-attribution"` service.

```php
'api.helper.partner-attribution' => static function ( ContainerInterface $container ) : PartnerAttribution {
    $bnCodes = array(
        ActivationDetector::CORE_PROFILER    => 'WooPPCP_Ecom_PS_CoreProfiler',
        ActivationDetector::PAYMENT_SETTINGS => 'WooPPCP_Ecom_PS_CoreProfiler',
    );
    
    return new PartnerAttribution(
        'ppcp_bn_code',
        $bnCodes,
        PPCP_PAYPAL_BN_CODE
    );
},
```

Reason:
The BN mapping is only used by the helper service, and keeping the array in that helper makes the connection between the array and the `PartnerAttribution` class clearer. As long as the map is only used by a single service there's no reason to split it into a second service.

Narek13: > Also, I recommend dropping this service and moving the array directly into the "api.helper.partner-attribution" service.

I was thinking it is a configuration which can be used by other services as well.

Narek13: However I have applied both suggestions. Thanks

modules/ppcp-api-client/src/ApiModule.php
```diff
@@ -133,6 +134,34 @@ static function () use ( $c ) {
...
+			'ppcp_request_args',
+			static function ( array $args ) use ( $c ) {
+				if ( isset( $args['headers']['PayPal-Partner-Attribution-Id'] ) ) {
+					return $args;
+				}
+
+				$partner_attribution = $c->get( 'api.helper.partner-attribution' );
+				assert( $partner_attribution instanceof PartnerAttribution );
+
+				$args['headers']['PayPal-Partner-Attribution-Id'] = $partner_attribution->get_bn_code();
```
stracker-phil, line 163: Possibly not needed, but I would add a check to see if `'header'` actually exists before assigning the BN code:

```php
if (!isset($args['headers']) || !is_array($args['headers'])) {
    $args['headers'] = array();
}

$args['headers']['PayPal-Partner-Attribution-Id'] = $partner_attribution->get_bn_code();
```

tests/PHPUnit/ApiClient/Helper/PartnerAttributionTest.php
```diff
@@ -0,0 +1,121 @@
...
+	 */
+	private string $bn_code_option_name = 'ppcp_bn_code';
+
+	/**
+	 * BN codes mapping.
+	 *
+	 * @var array
+	 */
+	private array $bn_codes = [
+		'core-profiler'    => 'WooPPCP_Ecom_PS_CoreProfiler',
```
stracker-phil: I would also use the `ActivationDetector` const values for the array keys, but keep the string values for BN codes, as we must only test if the BN code changes, and do not really care if/when the internal const value changes:

```php
private array $bn_codes = [
  ActivationDetector::CORE_PROFILER    => 'WooPPCP_Ecom_PS_CoreProfiler',
  ActivationDetector::PAYMENT_SETTINGS => 'WooPPCP_Ecom_PS_CoreProfiler',
);
```

Narek13: Already done :)

stracker-phil (approved):

Dinamiko (approved):

stracker-phil (approved):

PR #3255, opened 2025-03-21
Exclude new scripts from cache plugins (4412)

Dinamiko:
Any scripts & styles from our new settings UI should be excluded from caching plugins, such as cache (we already have features to support multiple cache plugins: siteground, litespeed, w3 total cache, litespeed etc).

This PR renames `index.js` output file to `ppcp-admin-settings.js` and add it to the caching compatibility services.


danielhuesken (approved):

Narek13 (approved):
Thanks

InpsydeNiklas:
This is no longer needed.

PR #3263, opened 2025-03-24
Simplify the new UI check for the onboarding.js file (4384)

danieldudzic:
### Description
This PR will simplify the conditional check in the `OnboardingModule.php` file for when to enqueue scripts.

This will remove the `Uncaught TypeError: Cannot set properties of null (setting 'checked')` error from the console when the new UI is loaded.

### DIFF
--- a/modules/ppcp-onboarding/src/OnboardingModule.php
+++ b/modules/ppcp-onboarding/src/OnboardingModule.php
@@ -55,9 +55,9 @@ class OnboardingModule implements ServiceModule, ExtendingModule, ExecutableModu
 						'1' === get_option( 'woocommerce-ppcp-is-new-merchant' )
 						|| getenv( 'PCP_SETTINGS_ENABLED' ) === '1'
 					)
-				)
+				) {
 					return;
-
+				}
 
 				$asset_loader = $c->get( 'onboarding.assets' );
 				assert( $asset_loader instanceof OnboardingAssets );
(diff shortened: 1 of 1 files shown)


modules/ppcp-onboarding/src/OnboardingModule.php
```diff
@@ -48,13 +48,7 @@ public function run( ContainerInterface $c ): bool {
...
 			'admin_enqueue_scripts',
 			function() use ( $c ) {
-				if (
-					apply_filters(
-					// phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
-						'woocommerce.feature-flags.woocommerce_paypal_payments.settings_enabled',
-						getenv( 'PCP_SETTINGS_ENABLED' ) === '1'
-					) && ! SettingsModule::should_use_the_old_ui()
-				) {
+				if ( ! SettingsModule::should_use_the_old_ui() ) {
```
Dinamiko: I think it would be better to rely on the filter than on the static method because the new settings module is not added to the container for existing merchants.

So in the conditional I suggest to remove the call to the static method and use just the filter:
```php
if(
   apply_filters(
      'woocommerce.feature-flags.woocommerce_paypal_payments.settings_enabled',
      '1' === get_option( 'woocommerce-ppcp-is-new-merchant' ) 
      || getenv( 'PCP_SETTINGS_ENABLED' ) === '1'
	)
)
```

Dinamiko (changes requested):

Dinamiko (approved):

PR #3268, opened 2025-03-25
Fix the branded-only detection (4427)

stracker-phil:
### Description

Switch the branded-only detection logic to use the new `woocommerce_paypal_branded` option


Narek13 (approved):
Thanks. LGTM

Narek13:
BTW. The Unit test should be fixed as well @stracker-phil

danieldudzic (approved):
LGTM 👍

PR #3278, opened 2025-03-27
Check the branded-only flag when settings-UI is loaded the first time (4439)

stracker-phil:
### Description

Fixes the branded-mode detection logic by checking for the installation path at a later stage of the plugin lifecycle


danieldudzic (approved):
LGTM 💚

Dinamiko (approved):

PR #3288, opened 2025-03-31
Enabling Save PayPal does not disable Pay Later messaging (4450)

Dinamiko:
When enabling “Save PayPal and Venmo” in Settings it disables Pay Later button correctly, however it does not disable Pay Later messaging in frontend pages.

This PR adds a new `woocommerce_paypal_payments_should_render_pay_later_messaging` filter that allows skip the render of Pay Later messaging. Then the filter is used in the new Settings module which does not render Pay Later messaging if "Save PayPal and Venmo" setting is enabled.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -639,10 +639,10 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 		assert( $gateway_redirect_service instanceof GatewayRedirectService );
 		$gateway_redirect_service->register();
 
-		// Do not render Pay Later messaging if "Save PayPal and Venmo" setting is enabled.
+		// Do not render Pay Later messaging if the "Save PayPal and Venmo" setting is enabled.
 		add_filter(
 			'woocommerce_paypal_payments_should_render_pay_later_messaging',
-			function() use ( $container ) {
+			static function() use ( $container ): bool {
 				$settings_model = $container->get( 'settings.data.settings' );
 				assert( $settings_model instanceof SettingsModel );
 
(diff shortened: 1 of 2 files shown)


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -638,6 +639,21 @@ static function () use ( $container ) {
 		assert( $gateway_redirect_service instanceof GatewayRedirectService );
 		$gateway_redirect_service->register();
 
+		// Do not render Pay Later messaging if "Save PayPal and Venmo" setting is enabled.
+		add_filter(
+			'woocommerce_paypal_payments_should_render_pay_later_messaging',
+			function() use ( $container ) {
```
Narek13:
```suggestion
			static function() use ( $container ): bool {
```

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -638,6 +639,21 @@ static function () use ( $container ) {
 		assert( $gateway_redirect_service instanceof GatewayRedirectService );
 		$gateway_redirect_service->register();
 
+		// Do not render Pay Later messaging if "Save PayPal and Venmo" setting is enabled.
```
Narek13:
```suggestion
		// Do not render Pay Later messaging if the "Save PayPal and Venmo" setting is enabled.
```

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -638,6 +639,21 @@ static function () use ( $container ) {
...
+			'woocommerce_paypal_payments_should_render_pay_later_messaging',
+			function() use ( $container ) {
+				$settings_model = $container->get( 'settings.data.settings' );
+				assert( $settings_model instanceof SettingsModel );
+
+				if ( $settings_model->get_save_paypal_and_venmo() ) {
+					return false;
+				}
+
+				return true;
```
Narek13: This can be simplified like this: `return ! $settings_model->get_save_paypal_and_venmo();`

Narek13 (approved):
LGTM. Thanks! Just a couple of suggestions, but they are not blocking approval.

PR #3293, opened 2025-04-01
Fastlane gateway visible on Pay for Order page (4428)

mmaymo:
We need to remove the Fastlane gateway from the list later and then the `order-pay` endpoint returns true when we are in that page.

### DIFF
--- a/modules/ppcp-axo/src/AxoModule.php
+++ b/modules/ppcp-axo/src/AxoModule.php
@@ -349,25 +349,25 @@ class AxoModule implements ServiceModule, ExtendingModule, ExecutableModule {
 			}
 		);
 
-	// Remove Fastlane on the Pay for Order page.
-	add_filter(
-    		'woocommerce_available_payment_gateways',
+		// Remove Fastlane on the Pay for Order page.
+		add_filter(
+			'woocommerce_available_payment_gateways',
 			/**
 			 * Param types removed to avoid third-party issues.
 			 *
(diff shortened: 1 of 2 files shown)


modules/ppcp-axo/src/AxoModule.php
```diff
@@ -349,6 +349,36 @@ function () use ( $c ) {
...
+					if ( $method && $method->id === 'ppcp-axo-gateway' ) {
+						unset( $methods[ $key ] );
+						break;
+					}
+				}
+
+
+				return $methods;
+			}
+		);
```
danieldudzic: Logic is solid, 💚  I'm so happy there's a relatively easy fix to this issue! 

If I were to nit-pick I'd suggest we simplify this code a bit. Something like:

```suggestion
	// Remove Fastlane on the Pay for Order page.
	add_filter(
    		'woocommerce_available_payment_gateways',
			/**
			 * Param types removed to avoid third-party issues.
			 *
			 * @psalm-suppress MissingClosureParamType
			 */
    		static function ( $methods ) {
        		if ( ! is_array( $methods ) || ! is_wc_endpoint_url( 'order-pay' ) ) {
            		return $methods;
        	}
        
        	// Remove Fastlane if present.
        	unset( $methods[AxoGateway::ID] );
        
        	return $methods;
    	}
	);
```

danieldudzic (approved):
Looking good! I have requested a minor change to keep the code simple. 💚

PR #3295, opened 2025-04-01
Remove branded-only flag when uninstalling PayPal Payments (4454)

stracker-phil:
### Description

During plugin uninstallation two flags are cleared from the DB:

1. Our plugin's `installation_path`: This flag effectively triggers the branded-only-experience.
2. WooCommerce's `woocommerce_paypal_branded`: The "installation_path" is defined based on this flag. If this flag would stay in the DB, the plugin would always enter the branded-only mode during next installation, regardless of the actual installation path. Setting this flag on every relevant installation action is a requirement WooCommerce must implement.

### DIFF
--- a/modules/ppcp-settings/src/Data/GeneralSettings.php
+++ b/modules/ppcp-settings/src/Data/GeneralSettings.php
@@ -41,6 +41,13 @@ class GeneralSettings extends AbstractDataModel {
 	protected array $woo_settings = array();
 
 	/**
+	 * Contexts in which the installation path can be reset.
+	 */
+	private const ALLOWED_RESET_REASONS = array(
+		'plugin_uninstall',
+	);
+
+	/**
 	 * Constructor.
 	 *
 	 * @param string $country              WooCommerce store country.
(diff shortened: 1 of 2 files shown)


modules/ppcp-settings/src/Data/GeneralSettings.php
```diff
@@ -40,6 +40,13 @@ class GeneralSettings extends AbstractDataModel {
 	 */
 	protected array $woo_settings = array();
 
+	/**
+	 * Contexts in which the installation path can be reset.
+	 */
+	private const ALLOWED_RESET_REASONS = array(
+		'plugin_uninstall',
+	);
+
```
stracker-phil, line 49: This `ALLOWED_RESET_CONTEXTS` constant provides a whitelist of valid reasons for resetting the installation path. This isn't meant as a security measure (direct DB access can always bypass it), but as a guardrail for devs to ensure **intentional usage**.

By requiring a specific reason that matches a predefined value, we ensure the `reset_installation_path()` method is only called with clear intent, preventing accidental resets through normal plugin operation.

Dinamiko (approved):

Narek13 (approved):
LGTM. Thanks

PR #3300, opened 2025-04-02
Add capture pre-conditions for card payment source (4461)

Dinamiko:
For non-3DS ACDC transactions, after a successful JSSDK call to confirm-payment-source endpoint, further GET calls register order status as `APPROVED`. It does look like that in the case of 3DS, the order status remains as `CREATED`.

In order to prevent capturing orders with the wrong status or without 3DS approval this PR adds the following pre-conditions:

- When `payment_source` = `card`.
    - .capture call only if status = `APPROVED`
    - OR `authentication_result` object is present and passes 3DS approval logic.

If these conditions doesn’t match then we’ll throw an exception which would prevent the `.capture` endpoint to be called.

### What this PR does
- Add new `CardCaptureValidator` class that implements the validation logic
- Add new action `woocommerce_paypal_payments_before_capture_order` to validate orders before capture
- Forces deletion of WC order on failed capture with a filter to disable it
- Add unit tests to ensure capture card validation works correctly across different scenarios

### DIFF
--- a/modules/ppcp-card-fields/src/Service/CardCaptureValidator.php
+++ b/modules/ppcp-card-fields/src/Service/CardCaptureValidator.php
@@ -31,18 +31,20 @@ class CardCaptureValidator {
 		}
 
 		$payment_source = $order->payment_source();
-		if ( $payment_source && $payment_source->name() !== 'card' ) {
-			return true;
-		}
+		if ( $payment_source ) {
+			if ( $payment_source->name() !== 'card' ) {
+				return true;
+			}
 
-		/**
-		 * The LiabilityShift response determines how you might proceed with authentication.
-		 *
-		 * @link https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/response-parameters/
-		 */
-		$liability_shift = $payment_source->properties()->authentication_result->liability_shift ?? '';
-		if ( in_array( $liability_shift, array( 'POSSIBLE', 'YES' ), true ) ) {
-			return true;
+			/**
+			 * The LiabilityShift response determines how you might proceed with authentication.
+			 *
+			 * @link https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/response-parameters/
+			 */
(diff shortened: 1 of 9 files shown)


modules/ppcp-card-fields/src/Service/CardCaptureValidator.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\CardFields\Service;
+
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\OrderStatus;
+
+/**
+ * CardCaptureValidator constructor.
```
Narek13: I think this is a leftover, I mean the "constructor"

modules/ppcp-card-fields/src/Service/CardCaptureValidator.php
```diff
@@ -0,0 +1,52 @@
...
+			/**
+			 * LiabilityShift determines how to proceed with authentication.
+			 *
+			 * @link https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/response-parameters/
+			 */
+			$liability_shift = $payment_source->properties()->authentication_result->liability_shift ?? '';
+			if ( in_array( $liability_shift, array( 'POSSIBLE', 'YES' ), true ) ) {
+				return true;
+			}
+		}
```
Narek13: Could be simplified with early returns

```suggestion
		$payment_source = $order->payment_source();
		if ( ! $payment_source ) {
		    return false;
		}
		
		if ( $payment_source->name() !== 'card' ) {
		    return true;
		}
		
		$liability_shift = $payment_source->properties()->authentication_result->liability_shift ?? '';
		return in_array( $liability_shift, array( 'POSSIBLE', 'YES' ), true );
```

Narek13 (approved):
Thanks for making the changes. LGTM, despite the CI fails

PR #3304, opened 2025-04-03
Improve branded-only upgrade path (4463)

Dinamiko:
This PR ensures general settings option is deleted on plugin uninstall.


Narek13 (approved):
LGTM

uninstall.php
```diff
@@ -96,8 +96,10 @@ function clear_plugin_branding( ContainerInterface $container ) : void {
 	 */
 	delete_option( 'woocommerce_paypal_branded' );
 
+	delete_option( 'woocommerce-ppcp-data-common' );
```
stracker-phil: We should not delete the whole settings entry here...

PR #3308, opened 2025-04-04
Apple Pay displayed in Things to do next when in branded-only mode (4468)

stracker-phil:
### Description

Adds some additional checks to prevent Card payments (ACDC, AXO) or wallets (Apple Pay, Google Pay) to appear on the overview page of the new settings UI when in branded only mode.


danieldudzic (approved):
LGTM 💚

PR #3309, opened 2025-04-04
option with BN code not created in branded-only mode (4470)

Dinamiko:
The option that controls the BN code in branded-only mode is not created.

### Steps To Reproduce
- set up plugin in branded-only mode
- navigate to any (classic) page with PayPal button
- type `PayPalCommerceGateway` in the browser console

Observe BN code is Woo_PPCP:

image-20250404-073228


stracker-phil (approved):

Narek13 (approved):
Thanks

PR #3319, opened 2025-04-14
Component cleanup: Icons, paths, links, code style (3928)

mmaymo:
This PR fixes js lint in the settings module. Some rules were suppressed to avoid changing implementation


modules/ppcp-settings/resources/js/Components/ReusableComponents/PricingTitleBadge.js
stracker-phil, line 1: Nice 👍

modules/ppcp-settings/resources/js/Components/ReusableComponents/PricingTitleBadge.js
```diff
@@ -36,6 +36,7 @@ const PricingTitleBadge = ( { item } ) => {
 	);
 
 	const label = sprintf(
+		// Translators: %1$s is the percentage, %2$s is the fixed amount.
```
stracker-phil, line 39: Nice 👍

modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Steps/index.js
```diff
@@ -57,7 +57,9 @@ const filterSteps = ( steps, conditions ) => {
 };
 
 export const getSteps = ( flags ) => {
+	// eslint-disable-next-line react-hooks/rules-of-hooks
```
stracker-phil: Instead of ignoring the "rules-of-hooks" check:
- My initial idea was to pass all conditions as arguments, like `getSteps( flags, skipPaymentsScreen )`, ie keep it a plain helper function without dependencies (would also make it more testable)
- An alternative would be to change the helper method to become a real hook -- change `getSteps(flags)` to `useSteps()` or similar, then we can drop the flags argument and collect all details inside the hook

The current mix of receiving input arguments and collecting other details via hooks is a bit messy

modules/ppcp-settings/resources/js/hooks/useHandleConnections.js
```diff
@@ -29,8 +29,10 @@ const ACTIVITIES = {
 
 export const useHandleOnboardingButton = ( isSandbox ) => {
 	const { onboardingUrl } = isSandbox
-		? CommonHooks.useSandbox()
-		: CommonHooks.useProduction();
+		? // eslint-disable-next-line react-hooks/rules-of-hooks
+		  CommonHooks.useSandbox()
+		: // eslint-disable-next-line react-hooks/rules-of-hooks
+		  CommonHooks.useProduction();
```
stracker-phil: Instead of ignoring these warnings, I think it would be better to correct the code to call hooks unconditionally:

```js
  const sandboxData = CommonHooks.useSandbox();
  const productionData = CommonHooks.useProduction();
  
  const { onboardingUrl } = isSandbox ? sandboxData : productionData;
  ```

File: modules/ppcp-settings/resources/js/hooks/usePaymentGatewaySync.js [hunk 4]
```diff
@@ -19,8 +19,8 @@ export const usePaymentGatewaySync = () => {
 	const { isReady: merchantIsReady } = CommonHooks.useStore();
 
 	const [ isSyncing, setIsSyncing ] = useState( false );
-	const [ syncCompleted, setSyncCompleted ] = useState( false );
-	const [ syncError, setSyncError ] = useState( null );
+	const [ , setSyncCompleted ] = useState( false );
+	const [ , setSyncError ] = useState( null );
```
stracker-phil: Hm... looks like we do not use those two states. There's no point in setting the state but not using the value anywhere.

To clean this up, the two states should be completely removed from the code.

Possibly, those states are a left over from a previous change

stracker-phil:
As we're working on this, can you also remove this utility file?

https://github.com/woocommerce/woocommerce-paypal-payments/blob/800bf1b7660d802b4358643e2d15f4b039abad2b/modules/ppcp-settings/resources/js/utils/data.js

This file contains a deprecated helper that's only used in a single component:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/800bf1b7660d802b4358643e2d15f4b039abad2b/modules/ppcp-settings/resources/js/Components/ReusableComponents/PaymentMethodIcons.js

I would really like to delete data.js, as we should not use the getImage() helper anywhere, except inside the PaymentMethodIcon component - here's an idea for a much cleaner solution:

```js
import { Icon } from '@wordpress/components';

const imageUrl( name ) {
  const filename = `icon-button-${ name }.svg`;
  
  // Move the logic into this file, as a private helper method.
}

const PaymentMethodIcon = ( { type } ) => (
  <Icon
    icon={ imageUrl( type ) }
    className="ppcp--method-icon"
  />
);

export default PaymentMethodIcon;
```

mmaymo:
@stracker-phil sorry for the delay

modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/Payment/PaymentDependencyMessage.js
```diff
@@ -22,6 +22,7 @@ const PaymentDependencyMessage = ( { parentId, parentName } ) => {
 		{
 			methodLink: (
 				<strong>
+					{ /* eslint-disable-next-line jsx-a11y/anchor-is-valid */ }
```
Narek13, line 25: We have a ticket to fix the accessibility issues, but not sure if this should be done here, instead of disabling the lint or in a separate issue.

cc: @InpsydeNiklas

InpsydeNiklas:
This is no longer needed and will be dealt with in a separate context.

PR #3340, opened 2025-04-23
Move 3D Secure settings from Payment Methods to Settings tab (4521)

danieldudzic:
### Description

This PR will relocate the `3D Secure` setting from the `Advanced Credit and Debit Card Payments` modal (in `Payment Methods`) tab to the `Settings` tab.

### Screenshots

|Before|After|
|-|-|
|acdc_moda_l_before|acdxc_modal_after|

|Before|After|
|-|-|
|settings_before|settings_after|

### Testing

- Verify that the 3D Secure settings are present and properly displayed
- Select a different 3D Secure option and save settings
- Refresh the page and verify the selected setting persists
- Navigate to the Payment Methods tab and verify the 3D Secure settings have been removed from the "`Advanced Credit and Debit Card Payments` modal

### DIFF
--- a/modules/ppcp-compat/src/CompatModule.php
+++ b/modules/ppcp-compat/src/CompatModule.php
@@ -282,8 +282,8 @@ class CompatModule implements ServiceModule, ExtendingModule, ExecutableModule {
 	 * The migration will be done on plugin update if it hasn't already done.
 	 */
 	protected function migrate_three_d_secure_setting(): void {
-		$payment_settings = get_option('woocommerce-ppcp-data-payment') ?: array();
-		$data_settings = get_option('woocommerce-ppcp-data-settings') ?: array();
+		$payment_settings = get_option( 'woocommerce-ppcp-data-payment' ) ?: array();
+		$data_settings    = get_option( 'woocommerce-ppcp-data-settings' ) ?: array();
 
 		// Skip if payment settings don't have the setting but data settings do.
 		if ( ! isset( $payment_settings['three_d_secure'] ) && isset( $data_settings['three_d_secure'] ) ) {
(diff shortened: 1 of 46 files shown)


AlexP11223 (approved):

modules/ppcp-compat/src/CompatModule.php
```diff
@@ -274,6 +275,38 @@ function () use ( $c, $is_smart_button_settings_migrated_option_name ) {
...
+
+				// Move the setting.
+				$data_settings['three_d_secure'] = $payment_settings['three_d_secure'];
+				unset( $payment_settings['three_d_secure'] );
+
+				// Save both.
+				update_option( 'woocommerce-ppcp-data-settings', $data_settings );
+				update_option( 'woocommerce-ppcp-data-payment', $payment_settings );
+			}
+		);
```
stracker-phil: I think it would be best to either have the full logic inside the action-handler (my recommendation), or instantly migrate without waiting for the "migrate_on_update" hook.

modules/ppcp-compat/src/CompatModule.php
```diff
@@ -274,6 +275,38 @@ function () use ( $c, $is_smart_button_settings_migrated_option_name ) {
...
+	 * The migration will be done on plugin update if it hasn't already done.
+	 */
+	protected function migrate_three_d_secure_setting(): void {
+		$payment_settings = get_option( 'woocommerce-ppcp-data-payment' ) ?: array();
+		$data_settings    = get_option( 'woocommerce-ppcp-data-settings' ) ?: array();
+
+		// Skip if payment settings don't have the setting but data settings do.
+		if ( ! isset( $payment_settings['three_d_secure'] ) && isset( $data_settings['three_d_secure'] ) ) {
+			return;
+		}
```
Dinamiko: I see there are more like this in other methods of this class but I suggest to encapsulate this code inside the hook callback, the reason is that all code inside the `run` method is executed multiple times every request whereas code inside hooks is only executed when needed.

Dinamiko (changes requested):

Dinamiko (approved):

PR #3355, opened 2025-05-02
ACDC payments for Subscription are failing at checkout for new users (4549)

Dinamiko:
The problem seems caused by Pay Later enabled together with Vaulting at the same time in the JSSDK:

Captura de pantalla 2025-04-30 a las 17 35 59

### How to reproduce
- Reset environment
- Activate WC, configure country as US
- Activate WC Subscriptions
- Activate PayPal plugin
- Onboard using US business merchant account
- Select “Business”, “Subscriptions” and “Available with additional application” when onboarding
- Once onboarded visit Payment Methods tab, notice how Pay Later is disabled but checkbox is checked:

Captura de pantalla 2025-04-30 a las 17 49 43

This PR fixes `use_subscriptions` flag as it was using `SUBSCRIPTIONS` instead of `subscriptions` and therefore the condition always returned `false`.

Then uses `use_subscriptions` flag value to enable Pay Later when is not `true`.


Narek13 (approved):
Haven't tested but LGTM. Thanks

PR #3362, opened 2025-05-06
APMs automatically enabled when selecting "No, ..." during onboarding (4467)

mmaymo:
This PR fixes the behaviour where APM's were enabled even if we selected "No" during onboarding. 
The toggle $is_enabled now is conditioned to the flag "use_card_payments"


modules/ppcp-settings/src/Service/SettingsDataManager.php
```diff
@@ -165,7 +165,7 @@ public function reset_all_settings() : void {
 	 */
 	public function set_defaults_for_new_merchant( ConfigurationFlagsDTO $flags ) : void {
 		if ( $this->onboarding_profile->is_setup_done() ) {
-			return;
+			//return;
```
Dinamiko: Check and fix this commented line.

Dinamiko (changes requested):

modules/ppcp-settings/src/Service/SettingsDataManager.php
mmaymo: Forgot that sorry, fixed

Dinamiko (approved):

PR #3369, opened 2025-05-07
Auto-enable logging during onboarding (4591)

stracker-phil:
### Description

Changes the default state of the logging-module to be enabled while the merchant did not complete the onboarding wizard.

- This change affect the default state for legacy, and the new UI
- Since the merchant has no access to the plugin settings, the logger can only be disabled via the WP filter
- The filter `woocommerce_paypal_payments_is_logging_enabled` still has the final decision over the logger state

### Steps to Test

In the new UI:

1. Disconnect using the "Start over" toggle to completely reset all settings
    - This step will disable the logger, prior to this PR
2. On the first page of the onboarding wizard, expand the "Advanced options" and enable the "Sandbox" toggle
    - This action triggers an API call to the PayPal servers, which is logged
3. Navigate to "WooCommerce → Settings → Logs" and verify the PayPal API call is logged

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -1501,16 +1501,26 @@ return array(
 			$container->get( 'woocommerce.logger.woocommerce' )
 		);
 	},
-	'wcgateway.logging.is-enabled'                         => function ( ContainerInterface $container ) : bool {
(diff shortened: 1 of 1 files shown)


Dinamiko (approved):

Narek13 (approved):
LGTM. Thanks

PR #3371, opened 2025-05-09
Block checkout - Address form is missing after payment on Product and Cart pages (4557)

Dinamiko:
Due to a infinite updates error in React, the address component is not rendered, this PR fixes that by using the proper `useEffect` dependecy.

### Steps To Reproduce
- In the dashboard → WooCommerce → Settings → Advanced select Block cart and Block checkout pages as default.
- On frontend navigate to the product and pay for it with PayPal or Pay Later.
- Check the address fields on Block checkout page.


danieldudzic (approved):
Looks good to me!

PR #3373, opened 2025-05-09
Fix #3372 by adding digital goods rounding item

### LINKED ISSUE #3372: Rounding differences potentially lead to order failure (4975)
### Describe the Bug

Rounding differences can change the basket to contain physical goods and lead to additional errors such as order failure.

### To Reproduce

1. Create an order with only digital goods which introduces rounding differences (I guess because PayPal is receiving net values; here, this happens with a VAT of 19% and two items of 9.50 gross value each).
2. An extra_line item `Subtotal mismatch` will be added by default by `PurchaseUnitSanitizer`. This item is a physical product.
3. An error can occur at a later point, which would otherwise not happen, such as in `OrderProcessor::process`:
```
if ($this->order_helper->contains_physical_goods($order) && !$this->order_is_ready_for_process($order)) {
    throw new Exception(__('The payment is not ready for processing yet.', 'woocommerce-paypal-payments'));
}
```
This is because `$this->order_helper->contains_physical_goods($order)` now returns `true`.

### Expected Behavior

I expect the order to go through.

### Actual Behavior

An exception is sometimes (always?) thrown after the order is completed and the order fails (even though it has already been paid) at the mentioned lines in `OrderProcessor::process`.

### Environment

 - WordPress Version 6.8.1
 - WooCommerce Version 9.8.4
 - Plugin Version 3.0.5

### Additional Details

My workaround is to patch `PurchaseUnitSanitizer::sanitize_item_amount_mismatch`:

```
$this->purchase_unit['items'][] = (new Item($line_name, $roundings_money, 1, category: Item::DIGITAL_GOODS))->to_array();
```

Problems with that implementation: This will always add digital goods and is only PHP 8.0+ (not 7.4).

luzat:
Rounding differences were previously added as physical goods, which changed the properties of the cart. Now, if the cart does not contain physical items, the extra line item will be added as a digital goods item.

**Issue**: #3372

---

### Description

Rounding differences can change the basket to contain physical goods and lead to additional errors such as order failure. This PR add the extra line item for rounding errors as a digital goods item by default, except if there are already some physical goods in the basket.

It may be viable to always add the rounding difference as a digital goods item, but I decided to change as little as possible compared to the current handling.

### Steps to Test

1. Create an order with only digital goods which introduces rounding differences (I guess because PayPal is receiving net values; here, this happens with a VAT of 19% and two items of 9.50 gross value each).
2. An extra_line item `Subtotal mismatch` will be added by default by `PurchaseUnitSanitizer`. This item is now a digital product.
3. Redo step 1 with at least some physical goods.
4. A physical goods extra_line item `Subtotal mismatch` will be added by default by `PurchaseUnitSanitizer`.

### Changelog Entry

Fix - Rounding items are digital goods unless the basket contains physical goods #3372


danieldudzic (approved):
Thank you for this PR @luzat!

I am approving. With the only minor suggestion to perhaps extract this to a separate method to keep the `sanitize_item_amount_mismatch` cleaner.

```
	/**
	 * Determines the appropriate category for rounding items based on existing items.
	 * 
	 * @return string The category (Item::DIGITAL_GOODS or Item::PHYSICAL_GOODS)
	 */
	private function determine_rounding_item_category(): string {
		// Check if all items are digital goods.
		foreach ( $this->items() as $item ) {
			$category = $item['category'] ?? Item::PHYSICAL_GOODS;
			if ( $category !== Item::DIGITAL_GOODS ) {
				return Item::PHYSICAL_GOODS;
			}
		}
		
		// All items are digital goods.
		return Item::DIGITAL_GOODS;
	}
```

luzat:
@danieldudzic I have updated the PR.

Dinamiko (approved):

PR #3384, opened 2025-05-14
Qa add test pcp 4357

sedatbsr:
Test added:
PCP-4357

### DIFF
--- a/tests/qa/tests/02-dashboard-ui/subscriptions.spec.ts
+++ b/tests/qa/tests/02-dashboard-ui/subscriptions.spec.ts
@@ -2,7 +2,8 @@
  * Internal dependencies
  */
 import { test, expect } from '../../utils';
-import { storeConfigDefault, subscriptionsPlugin } from '../../resources';
+import { storeConfigDefault, subscriptionsPlugin, merchants, Pcp } from '../../resources';
+import { time } from 'console';
 
 test.beforeAll( async ( { utils, pcpApi } ) => {
 	await utils.configureStore( storeConfigDefault );
@@ -13,6 +14,10 @@ test.beforeAll( async ( { utils, pcpApi } ) => {
 test.describe( () => {
 	test.beforeAll( async ( { requestUtils } ) => {
 		await requestUtils.activatePlugin( subscriptionsPlugin.slug );
+	} );
+
+	test.afterAll( async ( { requestUtils } ) => {
+		await requestUtils.deactivatePlugin( subscriptionsPlugin.slug );
 	} );
 
 	test( 'PCP-4356 | Subscription - Settings - US - Onboarding - Connect with personal account - Subscription type of product not allowed', async ( {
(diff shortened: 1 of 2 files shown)


tests/qa/tests/02-dashboard-ui/subscriptions.spec.ts
```diff
@@ -42,4 +47,69 @@ test.describe( () => {
...
+		await pcpOnboarding.activatePayPalPaymentsButton().click();
+	
+		await pcpOnboarding.businessRadio().click();
+		await pcpOnboarding.continueButton().click();
+	
+		await pcpOnboarding.physicalGoodsCheckbox().check();
+		await pcpOnboarding.subscriptionsCheckbox().check();
+		await pcpOnboarding.continueButton().click();
+	
+		await pcpOnboarding.chooseOptionalPaymentMethods( true );
```
mishautkin, line 68: This is a really minor thing (in current case) and I also often fall into such methods creation (which is why I need to refactor code afterwards...). But it can become important in more complex scenarios. It has been outlined to me by Christian. Consider simply leaving 2 actions here:
```ts
await pcpOnboarding.enableOptionalPaymentMethodsRadio().click();
await pcpOnboarding.continueButton().click();
```
Because:
- Name of `chooseOptionalPaymentMethods` doesn't tell about clicking Continue button.
- It would be logical to keep `continueButton().click()` exposed also because it's present on lines 65 and 69. But then on line 72 - the scenario actually navigates back to the initial page.
- `chooseOptionalPaymentMethods( false )` - sounds a bit confusing.

Would apprciate your point of view ;)

sedatbsr: I wrapped those actions in that function because there are 2 or 3 test cases that shares same scenario. So I thought parameterization would be fine here. 

About `continueButton()`, you are right, it is logical to keep it out of the function. So this case most probably subject to change. 

`chooseOptionalPaymentMethods` function can be parameterized in different way maybe so it can be more meaningful. Such as;

.test-data:
```
{
	subscriptionsEnabled: true,
	optionalPaymentsEnabled: true,
	title: 'PCP-0000 | Subscriptions ENABLED + Optional Payments ENABLED', (maybe different)
},
{
	subscriptionsEnabled: true,
	optionalPaymentsEnabled: false,
	title: 'PCP-0000 | Subscriptions ENABLED + Optional Payments DISABLED',
}
```

and in test, it can be used like:

.spec file:
```
...
if (subscriptionsEnabled) {
	await pcpOnboarding.subscriptionsCheckbox().check();
	} else {
	await pcpOnboarding.subscriptionsCheckbox().uncheck();
	}
	await pcpOnboarding.continueButton().click();

	await pcpOnboarding.chooseOptionalPaymentMethods(optionalPaymentsEnabled);
	await pcpOnboarding.continueButton().click();
...
```

mishautkin: Ok, got your point!

mishautkin (approved):

PR #3385, opened 2025-05-14
OAuth login often fails in new UI (4582)

stracker-phil:
### Description

This PR refactores the OAuth process in two ways:

#### I) Solve timing issues between REST and redirect

The initial REST call only stores the auth_code and ID, without attempting to resolve them directly. This ensures that the REST call finishes very fast.

The following page reload fetches the code/id from the DB and performs the API requests which were initially handled in the Ajax call. This consolidates the full authentication logic to the same HTTP request.

#### II) Handling concurrent page loads

After finishing the login popup, PayPal often starts TWO page reloads. While the first (detached) request handles the authentication correctly, the second request (which happens in the browser) would previously run into a “token already used” situation and display the login screen again.

A new state was introduced to mark the token as “currently processing” (max lifetime 10 seconds). When any follow-up request encounter this “currently processing” state, they will pause for 1 second and do a simple page reload, waiting for the authentication request to complete.

### DIFF
--- a/modules/ppcp-settings/src/Handler/ConnectionListener.php
+++ b/modules/ppcp-settings/src/Handler/ConnectionListener.php
@@ -156,7 +156,7 @@ class ConnectionListener {
 		$this->logger->info( 'Found OAuth merchant data in request', $data );
 
 		try {
-			$this->authentication_manager->finish_oauth_authentication( $data );
+			$this->authentication_manager->handle_oauth_authentication( $data );
 			$this->mark_token_as_processed( $token );
 		} catch ( \Exception $e ) {
 			$this->logger->error( 'Failed to complete authentication: ' . $e->getMessage() );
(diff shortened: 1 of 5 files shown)


modules/ppcp-settings/src/Handler/ConnectionListener.php
```diff
@@ -133,34 +144,65 @@ public function process( int $user_id, array $request ) : void {
...
+			/*
+			 * Authentication token is currently processed (in another request):
+			 * Briefly wait and then retry this request, basically waiting for
+			 * the above "was_processed" condition to become true.
+			 */
+			$this->logger->info( 'Token is currently being processed, waiting and retrying', array( 'token' => $log_token ) );
+			sleep( 1 );
+
+			// Get the current URL.
+			$current_url = add_query_arg( null, null );
```
Narek13, line 173: To be honest, I’m not a big fan of adding inline comments. I understand that they can make the code clearer, especially in this plugin where some logic is quite complex. However, in cases where the code is obvious like here, I would suggest not adding them. This isn't a strong opinion and open to discussion. Another point is that other devs aren’t adding them either, which leads to inconsistent styling.

Narek13 (approved):
Thanks for working on this. I’ve raised a point we can discuss in another context, but it’s absolutely not something that would prevent me from approving this PR, although haven't tested locally.

Dinamiko (approved):

PR #3398, opened 2025-05-20
QA Add PCP-4360 and various updates

sedatbsr:
* PCP-4360 added.
* Test data added to onboarding-ui.data for PCP-4360 `(onboardingSubscriptionComparisonByOPM)` (naming subject to change)
* PCP-4357 combined in the same scenario
* `snapshotContent()` function replaced with `snapshotLocator() `in PCP-4356
*  Fix lint issues

### DIFF
--- a/tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
+++ b/tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
@@ -8,6 +8,7 @@ import {
 	merchants,
 	Pcp,
 } from '../../resources';
+import { onboardingSubscriptionComparisonByOPM } from './_test-data';
 
 test.beforeAll( async ( { utils, pcpApi } ) => {
 	await utils.configureStore( storeConfigDefault );
@@ -34,8 +35,10 @@ test.describe( () => {
 			'* Business account is required for subscriptions.'
 		);
 		await pcpOnboarding.personalAccountRadio().click();
-		await pcpOnboarding.snapshotContent(
-			`${ testInfo.title } - Set up store type`
+		await pcpOnboarding.snapshotLocator(
+			pcpOnboarding.onboardingContentContainer(),
+			`${ testInfo.title } - Set up store type`,
+			{ timeout: 3000 }
 		);
 		await pcpOnboarding.continueButton().click();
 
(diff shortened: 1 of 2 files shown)


tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
```diff
@@ -34,87 +35,106 @@ test.describe( () => {
...
+						3000
+					);
+
+					await pcpPaymentMethods.visit();
+					await pcpOnboarding.snapshotContent(
+						`${ testInfo.title } - Payment methods - PayPal, Venmo enabled`,
+						3000
+					);
+
+					const snapshotName = testInfo.title;
```
mishautkin: A really minor thing but maybe better to use either `testInfo.title` or `snapshotName` everywhere?

tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
```diff
@@ -34,87 +35,106 @@ test.describe( () => {
...
 
 		await expect( pcpOnboarding.subscriptionsCheckbox() ).toBeVisible();
 		await expect( pcpOnboarding.subscriptionsCheckbox() ).toHaveAttribute(
 			'disabled'
 		);
-		await pcpOnboarding.snapshotContent(
-			`${ testInfo.title } - Select product types - Subscription type of product not allowed`
+		await pcpOnboarding.snapshotLocator(
+			pcpOnboarding.onboardingContentContainer(),
+			`${ testInfo.title } - Select product types - Subscription type of product not allowed`,
```
mishautkin, line 51: Just keep in mind that potentially you'll have to analyse this long strings among hundreds of screenshots:
image

Also the information in the SS name is duplicated.

This is also the reason why I insist so much on laconic test summaries (not only because I don't like reading 😆). When you come upon such a test summary I would suggest to discuss it with Ana and potentially agree on a more laconic version for both Xray and autotest.

tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
sedatbsr: snapshotName removed.

tests/qa/tests/02-onboarding/onboarding-subscription.spec.ts
sedatbsr: We will discuss this topic.

mishautkin (approved):

PR #3406, opened 2025-05-22
Add condition to not process already completed order (PCP-4631)

### LINKED ISSUE #3374: PayPalGateway::process_payment on completed order leads to order failure
### Describe the Bug

When `PayPalGateway::process_payment` is called again on a completed and paid order, the order fails (and no refund is happening). I do not think that this should happen and the code should be idempotent in that case.

### To Reproduce

I am not sure when exactly this can happen. For us, the code was invoked multiple times when another error occurred.

When manually invoking `PayPalGateway::process_payment` on a completed order, the line

```
$this->order_processor->process($wc_order);
```

can be invoked multiple times. The second time, it will throw (`ORDER_ALREADY_COMPLETED`) which leads to an order status of `failed` and an invocation of `PayPalGateway::handle_payment_failure` after the order was already completed.

I figure this might in some cases happen if duplicate requests are coming in and there may be a race condition involved. It very much also depends on 3rd-party plugins if that can happen.

### Expected Behavior

Probably best: do nothing, except adding a log entry, and return with success. I do not think most action handlers should be called again.

### Actual Behavior

A `PaypalApiException` is thrown and the order fails.

### Environment

 - Plugin Version 3.0.5

### Additional Details

There may be various fixes and I do not know which is best:

  - `process_payment` could check if there is a PayPal (not WooCommerce) order and that is completed already. If yes, just `return $this->handle_payment_success($wc_order);`. This may be required at various places in `process_payment` to handle all cases.
  - In `OrderProcessor::process`, after loading the PayPal order, before `add_paypal_meta`, check if the paypal order is already completed and if it is, just return (possibly logging that case). Invoking `woocommerce_paypal_payments_after_order_processor` again might or might not be a good idea.

The second solution is somewhat more generic and protects against failure if `process` is ever invoked again. On its own, many hooks in `process_payment` could still be invoked twice, though, and lead to side effects such as duplicated metadata. It may be useful to implement both.

In all cases, there still will be a chance for race conditions with concurrent invocations, which is hard to completely eliminate.

Dinamiko:
Fixes #3374

### DIFF
--- a/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
+++ b/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
@@ -231,7 +231,7 @@ class OrderProcessor {
 		// Do not continue if PayPal order status is completed.
 		$order = $this->order_endpoint->order( $order->id() );
 		if ( $order->status()->is( OrderStatus::COMPLETED ) ) {
-			$this->logger->warning('Could not process PayPal completed order #'. $order->id() );
+			$this->logger->warning( 'Could not process PayPal completed order #' . $order->id() . ', Status: ' . $order->status()->name() );
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
LGTM

mmaymo (approved):

PR #3407, opened 2025-05-23
Fix new settings UI in WC 9.9+

AlexP11223:
The CSS changes in WC 9.9 affected our UI.

So in this PR

- setting the white background in `#mainform`, not just `body`
- added `!important` for the welcome screen h2 title margin, to avoid `margin: 0` from `body.woocommerce_page_wc-settings #mainform h2`, not sure if there is a better way


Dinamiko (approved):

AlexP11223:
Also set the background for `#wpcontent`, looks like it can affect it on some screens.

PR #3426, opened 2025-05-28
Bring back PartnerReferralsDataTest.php and fix tests based on latest business logic

Dinamiko:
This PR brings back `PartnerReferralsDataTest.php` file which was removed for 3.0.6 release, it also fixes the tests based on the latest logic included in `PartnerReferralsData.php`.


InpsydeNiklas:
This is no longer needed.

PR #3431, opened 2025-05-30
Use experience_context

AlexP11223:
Now using the `experience_context` instead of the deprecated `application_context`.

At first I was trying to add a class for each payment source like in #1491 but it was not fitting well into the current code without much bigger refactoring, and the main goal for now is just the `experience_context`. (also in the old PR I was trying to use the same classes both for sending requests and parsing of responses, which is not correct because they are different)

So I ended up just creating a builder (to avoid putting everything into the constructor parameters) for creating an `ExperienceContext` object which helps reduce some code duplication. Compared to the old `ApplicationContextRepository` we can send just the properties that are needed for the payment source (e.g. just the return URLs) instead of all possible properties every time.

Usage example:

```
$experience_context_builder
    ->with_order_return_urls( $wc_order )
    ->with_current_payment_method_preference()
    ->build()
    ->with_locale( 'de-DE' )
    ->to_array()
```

The functions like `with_locale` that simply set a property to the specified value are not duplicated in the builder.

The `to_array` function uses reflection for retrieving the list of properties to make adding new properties simpler.

`with_default_paypal_config` sets all properties for the old smart buttons and similar use cases like the old `ApplicationContextRepository`.

The parameters for the `OrderEndpoint::create` are simplified, removed unused, also now there is no separate token parameter, it should be passed as payment source using the new `to_payment_source` function.

For Oxxo I am not sure which properties should be set in its `payment_source` and `experience_context` since they are not included in the API docs, but I assumed it's the same as in the second request (confirm payment source) + added the standard smart buttons properties for `experience_context` in case some of them were relevant.

---

Also I disabled the PHPCS warning about not adding types in PHPDocs of properties. Since we are now PHP 7.4+ it does not make sense to duplicate the type from the property definition. Ideally PHPCS should detect and skip such cases (iirc like in other projects using our rules) but the package used here seems to be abandoned several years ago. But Psalm raises an error if the type is completely missing, so it's ok.


Dinamiko (approved):

modules/ppcp-api-client/src/Entity/ExperienceContext.php
```diff
@@ -0,0 +1,244 @@
...
+
+	const SHIPPING_PREFERENCE_GET_FROM_FILE        = 'GET_FROM_FILE';
+	const SHIPPING_PREFERENCE_NO_SHIPPING          = 'NO_SHIPPING';
+	const SHIPPING_PREFERENCE_SET_PROVIDED_ADDRESS = 'SET_PROVIDED_ADDRESS';
+
+	const USER_ACTION_CONTINUE = 'CONTINUE';
+	const USER_ACTION_PAY_NOW  = 'PAY_NOW';
+
+	const PAYMENT_METHOD_UNRESTRICTED               = 'UNRESTRICTED';
+	const PAYMENT_METHOD_IMMEDIATE_PAYMENT_REQUIRED = 'IMMEDIATE_PAYMENT_REQUIRED';
```
Narek13: Do we need to set the visibilities for these?

AlexP11223: Looks like const properties are public by default, private const is allowed since PHP 7.1.

Narek13: Yes, even though it's not required, I think explicitly setting visibility for clarity and consistency is better (especially when private const or protected const are also already used). But not a strong opinion, so please feel free to ignore.

modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
```diff
@@ -259,11 +236,8 @@ static function ( PurchaseUnit $item ) use ( $shipping_preference ): array {
 
 		$response = $this->request( $url, $args );
 		if ( is_wp_error( $response ) ) {
-			$error = new RuntimeException(
-				__( 'Could not create order.', 'woocommerce-paypal-payments' )
-			);
-			$this->logger->log(
-				'warning',
+			$error = new RuntimeException( 'Could not create order.' );
```
hmouhtar, line 239: This isn’t being translated. I don’t think it’s strictly necessary, but other errors are, so it should be consistent

hmouhtar (approved):

modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php
AlexP11223: Yes, I think normally we should not translate errors/logs. So the ones that are translated just remained like that from the old times :D

PR #3435, opened 2025-06-02
Invoice prefix repeatable number causes an error on frontend (4793)

Dinamiko:
This issue relates to a problem with the invoice prefix functionality. When sending an existing invoice ID to PayPal, it's causing `DUPLICATE_INVOICE_ID` error on the frontend of the website. 

### How to reproduce
- Create a new business account on PayPal developer portal
- Onboard merchant
- Create and purchase a product using PayPal payment
   - Check WC order ID
- Log as merchant in PayPal sandbox
   - Go to last transaction and check Invoice ID value
- Reset WordPress installation including the database and onboard merchant again
   - This is important as WC order ID will start with the same value as before
- Create and purchase a product using PayPal payment
- `DUPLICATE_INVOICE_ID` Error is displayed

This PR fixes 2 issues:
1. In new settings UI invoice prefix was not sent to PayPal, only the WC order ID.
2. When reseting WordPress and onboard existing merchant with existing transactions DUPLICATE_INVOICE_ID is thrown when sending the same invoice ID to PayPal.

As `DUPLICATE_INVOICE_ID` issue affects mostly sandbox environments while developing, this PR only adds the random invoice prefix on sandbox and uses the same logic as before (prefix is generate based on site URL) for live environments.

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -2095,9 +2095,21 @@ return array(
 	},
 
 	/**
-	 * Returns random 6 characters length alphanumeric prefix, followed by a hyphen.
+	 * Returns a prefix for the site, ensuring the same site always gets the same prefix (unless the URL changes).
 	 */
 	'wcgateway.settings.invoice-prefix'                    => static function( ContainerInterface $container ): string {
-		return wp_generate_password(6, false) . '-';
+		$site_url = get_site_url( get_current_blog_id() );
+		$hash     = md5( $site_url );
+		$letters  = preg_replace( '~\d~', '', $hash ) ?? '';
+		$prefix   = substr( $letters, 0, 6 );
+
+		return $prefix ? $prefix . '-' : '';
+	},
+
+	/**
+	 * Returns random 6 characters length alphanumeric prefix, followed by a hyphen.
+	 */
+	'wcgateway.settings.invoice-prefix-random'             => static function( ContainerInterface $container ): string {
+		return wp_generate_password( 6, false ) . '-';
 	},
 );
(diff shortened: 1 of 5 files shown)


modules/ppcp-wc-gateway/services.php
```diff
@@ -2093,4 +2093,23 @@ static function( ContainerInterface $container ): DisplayManager {
...
+		$prefix   = substr( $letters, 0, 6 );
+
+		return $prefix ? $prefix . '-' : '';
+	},
+
+	/**
+	 * Returns random 6 characters length alphanumeric prefix, followed by a hyphen.
+	 */
+	'wcgateway.settings.invoice-prefix-random'             => static function( ContainerInterface $container ): string {
+		return wp_generate_password( 6, false ) . '-';
```
AlexP11223: I guess it should not cause issues for us currently since afaik we don't create multiple orders in a single request (or at least not re-connecting merchants during that), but implementing this as a service feels wrong, since it will be cached. Maybe it can be a Factory at least (they are not cached).

AlexP11223: ah, I did not notice that it's only for settings. But then there potentially could be the opposite problem, saving different values in different places. Ideally it seems like something that should be explicitly generated one time after connecting.

Dinamiko: yes it's only for getting the default value of the invoice prefix setting, once the setting is saved it's not used anymore. Also it's only used in sandbox environment.

AlexP11223 (approved):

Narek13 (approved):
Haven't tested but LGTM. Thanks

PR #3445, opened 2025-06-06
PayPal button should not be displayed when customer already has vaulted PayPal account (4791)

Dinamiko:
When PayPal is saved as customer’s vaulted payment method, the Pay Pal button on My Account / Payment methods / Add payment method shouldn’t be displayed.

### Steps To Reproduce
- Save PayPal as payment method for the customer
- Navigate to My Account / Payment methods / Add payment method

Actual result:
PayPal button and Gateway are displayed:

image-20250602-075747

### Expected behavior
Only one vaulted PayPal account should be available for the customer, thus it shouldn’t be displayed if vaulted already.


hmouhtar (approved):

modules/ppcp-save-payment-methods/src/SavePaymentMethodsModule.php
```diff
@@ -340,11 +341,19 @@ function () use ( $c ) {
 
 				add_action(
 					'woocommerce_add_payment_method_form_bottom',
-					function () {
+					function () use ( $c ) {
 						if ( ! is_user_logged_in() || ! is_add_payment_method_page() ) {
 							return;
 						}
 
+						// Do not display PayPal button if the user already has a PayPal payment token.
```
Narek13: I would suggest to remove the inline comment and add it for the whole action to describe why this action is added.

```php
/**
 * Disables the PayPal button if the user already has a PayPal payment token.
 */
add_action(
					'woocommerce_add_payment_method_form_bottom',
					...
```

modules/ppcp-save-payment-methods/src/Service/PaymentMethodTokensChecker.php
```diff
@@ -0,0 +1,56 @@
...
+	 *
+	 * @var PaymentTokensEndpoint
+	 */
+	private PaymentTokensEndpoint $payment_method_tokens_endpoint;
+
+	/**
+	 * PaymentMethodTokensChecker constructor.
+	 *
+	 * @param PaymentTokensEndpoint $payment_method_tokens_endpoint Payment method tokens endpoint.
+	 */
```
Narek13: After @AlexP11223 did [this](https://github.com/woocommerce/woocommerce-paypal-payments/pull/3474/files#diff-05ae9cddcaec1e845771a7db224961439f83ef5939ec67d3a48744cb34d7e58b) change we can avoid adding unnecessary constructor comments

PR #3446, opened 2025-06-06
Add WooCommerce Tracks integration and the Onboarding funnel (4649)

danieldudzic:
### Description
This PR introduces a tracking system for the onboarding flow with source-based field filtering, multi-funnel support, and extensible adapter architecture. The system prevents tracking loops and provides comprehensive analytics capabilities across the new Settings UI.

It monitors WordPress data stores rather than adding code to frontend components, ensuring comprehensive coverage of all state changes regardless of their source (user actions, API responses, system updates) while maintaining clean separation of concerns.

##### 1. **Registry System** (`registry.js`)
- Registers tracking funnels with configurations
- Maps stores to funnels to prevent duplicate subscriptions  
- Enables dynamic funnel activation based on conditions
```javascript
registerFunnel('ppcp_onboarding', onboardingConfig);
```

##### 2. **Subscription Manager** (`subscription-manager.js`)
- Prevents subscription conflicts when multiple funnels track the same store
- Routes store changes to relevant funnels
- Manages funnel lifecycle based on conditions
```javascript
// Single subscription handles multiple funnels
ensureStoreSubscription('wc/paypal/onboarding');
// Routes to: onboarding funnel, settings funnel, etc.
```

##### 3. **Dedicated Tracking Store** (`data/tracking/`)
- Separate Redux store handles all field source information
- Completely isolates tracking concerns from business logic
- Centralized field source storage with automatic cleanup
```javascript
// State: { storeName: { fieldName: { source, timestamp } } }
trackingStore.getFieldSource('wc/paypal/onboarding', 'step');
```

##### 4. **Source-Based Field Filtering** (`utils.js`, `subscription-manager.js`)
- Field-level source rules define which change sources trigger tracking
- Prevents tracking loops by filtering out subscription/internal updates
- Enables selective tracking (user-only vs user+system)
```javascript
fieldRules: {
    step: { allowedSources: ['user', 'system'] },
    isCasualSeller: { allowedSources: ['user'] }
}
// Only tracks user-initiated account type changes
```
*Source tracking needed because Redux store subscriptions and internal system updates create noise in analytics by triggering tracking events for non-user actions.*

##### 5. **Adapter Pattern** (`adapters/woocommerce-tracks.js`, `adapters/console-logger.js`)
- WooCommerce Tracks Adapter for analytics
- Console Logger Adapter for debugging
- Extensible for additional platforms
```javascript
trackingService.addAdapter(new WooCommerceTracksAdapter());
trackingService.addAdapter(new ConsoleLoggerAdapter());
// Events sent to both destinations
```

##### 6. **Funnel Configuration** (`funnels/onboarding.js`)
- Uses FunnelConfigBuilder pattern for flexible configuration
- Defines tracking setup for the onboarding flow
- Specifies stores, fields, events, and translation functions
```javascript
export const config = FunnelConfigBuilder.createBasicFunnel('ppcp_onboarding', {
    debug: false,
    adapters: ['woocommerce-tracks'],
    trackingCondition: {
        store: 'wc/paypal/common',
        selector: 'merchant', 
        field: 'isConnected',
        expectedValue: false
    }
})
.addStore('wc/paypal/onboarding', [
    createFieldTrackingConfig('step', 'persistent', {
        rules: { allowedSources: ['user', 'system'] }
    })
])
.build();
```

##### 7. **Core Tracking Service** (`services/funnel-tracking.js`)
- Orchestrates the entire tracking process for each funnel
- Manages adapters and processes state changes  
- Handles source filtering and field rule validation
- Coordinates funnel-specific translation functions

#### Universal Store Compatibility
Any Redux store becomes tracking-compatible via `createHooksForStore`:
```javascript
// Makes ANY store tracking-compatible
const { usePersistent, useTransient } = createHooksForStore('wc/paypal/any-store');

// In components
const [field, setField] = usePersistent('fieldName');
setField(newValue, 'user'); // Automatically tracked if configured
```

#### Store Registration
Stores register with funnels in their index files:
```javascript
// Required: stores must register with tracking funnels  
addStoreToFunnel(STORE_NAME, ONBOARDING_FUNNEL_ID);
```

#### Component Integration
Form components pass appropriate source values (`'user'`, `'system'`) when updating fields.

### Testing
1. **Enable WooCommerce Tracks Debug**:
   Add to the browser console:
   ```javascript
   localStorage.setItem('debug', 'wc-admin:*');
   ```
2. **Keep the browser console open** and trigger events by:
   - Starting onboarding flow
   - Changing account type/product selection
   - Completing onboarding steps
3. **Verify Events**: Look for `ppcp_onboarding_` prefixed events in console


Dinamiko (approved):

modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Steps/StepProducts.js
```diff
@@ -38,7 +38,7 @@ const StepProducts = () => {
 		};
 
 		initChoices();
-	}, [ canUseSubscriptions, optionState, products, setProducts ] );
+	}, [ canUseSubscriptions, isCasualSeller ] );
```
strangerkir, line 82: I assume these dependencies were removed intentionally, and this is okay?

modules/ppcp-settings/resources/js/data/common/reducer.js
```diff
@@ -73,21 +74,76 @@ const defaultPersistent = Object.freeze( {
...
+			timestamp: Date.now(),
+		},
+	};
+};
 
+const clearFieldSource = ( currentSources, fieldName ) => {
+	const newSources = { ...currentSources };
+	delete newSources[ fieldName ];
+	return newSources;
+};
```
stracker-phil: I'm uncertain if I like this direct integration into the reducer, as it implements tracking on this store only. This (a) does not scale well and (b) changes the responsibility of the store to handle two things at once

modules/ppcp-settings/resources/js/data/common/reducer.js
```diff
@@ -73,21 +74,76 @@ const defaultPersistent = Object.freeze( {
...
+		if ( action && action.source ) {
+			const fieldName = Object.keys( payload )[ 0 ];
+
+			newState.fieldSources = updateFieldSources(
+				newState.fieldSources,
+				fieldName,
+				action.source
+			);
+		}
+
```
stracker-phil: While working, this pattern is not ideal for several reasons:
1. it adds bloat to the code and makes reducers look more complex than they are.
2. I think tracking should not happen on reducer level, but on action level.
3. It violates the single responsibility

My solution would be to create a new store for tracking with two actions: `updateSources` and `clearSources`. Then extend [createHooksForStore()](https://github.com/woocommerce/woocommerce-paypal-payments/blob/803b2bfacc45eedfe662359c885b920334c029ba/modules/ppcp-settings/resources/js/data/utils.js#L123-L133) to dispatch the `updateSources` action right after calling the original dispatcher, like this:

```js
const setValue = useCallback(
    (newValue, source = null) => { // ← add optional tracking argument here
        actions[dispatcher](key, newValue);
        
        if (source) { // ← dispatch conditional tracking here
            const trackingActions = useDispatch('tracking-store');
            trackingActions.updateSources(key, source);
        }
    },
    [actions, key]
);
```

modules/ppcp-settings/resources/js/services/tracking/index.js
```diff
@@ -0,0 +1,30 @@
...
+	initializeTracking,
+	getRegisteredFunnels,
+	getTrackingInstances,
+	getTrackingInstance,
+	validateFunnelConfig,
+	getMultiFunnelStores,
+	getFunnelsForStore,
+	getTrackingStatus,
+} from './registry';
+
```
stracker-phil, line 24: With a dedicated Redux store for `tracking`, this registry would become much simpler and reliable, as most of its logic would be part of the Redux store (a pattern that's more aligned with the code base)

strangerkir (approved):
LGTM

Only one question to make sure there is no unintentional regression.

modules/ppcp-settings/resources/js/services/tracking/subscription-manager.js
```diff
@@ -0,0 +1,852 @@
...
+
+		registration.conditionCheckCount++;
+
+		try {
+			const conditionStore = select( trackingCondition.store );
+			if ( ! conditionStore ) {
+				return false;
+			}
+
+			// Check store readiness.
```
Narek13, line 338: I've noticed a growing tendency to add more inline comments in the codebase. Personally, I believe inline comments are valuable when they clarify something non-obvious or explain edge cases, but in general, I prefer using them only when truly necessary.

Since some developers use them frequently and others rarely, it might be helpful to align on a shared approach to inline comments as a team. Perhaps we can discuss this outside of this PR to establish a consistent convention?

Narek13 (approved):
This is a really great PR, thanks for the work on it @danieldudzic  🙌 I’ve also briefly tested it locally, and the code works fine for me.

One thing we might want to consider is what Philipp suggested: refactoring the tracking logic into a dedicated Redux store. But current PR looks good to me.

AlexP11223 (approved):
LGTM except the above comments.

mmaymo (approved):

hmouhtar:
LGTM :+1:

hmouhtar (approved):

PR #3449, opened 2025-06-10
Implement country-based feature availability check (4800)

stracker-phil:
_This PR extends #3447_

### Description

Extends the Contact Module feature flag with a second condition:

The feature is only available, if the merchant country is "US"

#### Technical Change

The PR introduces a new class named `MerchantDetails` which provides access to the country.

This class is a first implementation that should eventually in two ways:
1. Replace using the service in many places `api.shop.country`
2. Provide more information, especially eligibility checks (e.g. merging `FeaturesEligibilityService` into this)


danieldudzic (approved):
LGTM!

PR #3456, opened 2025-06-13
Modify Create Order payload to display email and phone in PayPal popup (4803)

stracker-phil:
_This PR extends on #3452_

### Description

Extend the "Pay with PayPal" popup with additional contact fields when
1. The merchant is eligible for the Contact Module feature (US merchant)
2. The "Custom Shipping Contact" toggle is active (or the merchant uses legacy UI)

### Screenshots

| Before | After |
|---|---|
| CleanShot 2025-06-12 at 18 54 56@2x | CleanShot 2025-06-12 at 18 55 26@2x |


modules/ppcp-wc-gateway/src/Helper/MerchantDetails.php
```diff
@@ -13,6 +13,34 @@
  * Main information source about merchant details.
  */
 class MerchantDetails {
```
danieldudzic, line 15: I'm 100% for having a centralized place for this data, but we need to make sure to update everything to use it and also stick to it as the standard/convention.

We need to use the same format and conventions for things like:
- Merchant eligibilities (e.g., local country/currency matrix, etc.)
- Merchant capabilities (e.g., PayPal SellerStatus "features" returned from the API)

Otherwise, we risk fragmentation or confusion in the future.

danieldudzic (approved):
LGTM!

modules/ppcp-wc-gateway/src/Helper/MerchantDetails.php
stracker-phil: I agree, and I will raise this topic in an internally discussion again.

As I see it, we currently have a very fragmented eligibility/capability detection in the code, and this is a first (incomplete) attempt to establish a standard.

Since this is a bit off-topic (not part of the contact-module-feature) I just want to provide a proof-of-concept integration for now. And we'll need to refactor _something_ in the coming weeks, after discussing the topic

PR #3459, opened 2025-06-16
Cache the reference transaction disabled value (4857)

Narek13:
The [reference_transaction_enabled](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8d8508618594bf9ec36c4eb283fe95a5744f651b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php#L122) method attempts to check if Reference Transactions are enabled for the PayPal account by making a call to [create_token()](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8d8508618594bf9ec36c4eb283fe95a5744f651b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php#L73)  (Which is called only [once](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8d8508618594bf9ec36c4eb283fe95a5744f651b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php#L131) in the codebase ) and caching the result with a transient:

1. On success, sets the transient to cache the result.
2. On failure, deletes the transient and returns false.

and If `create_token()` throws an exception, `set_transient()` never runs. This means the failure case is not cached, and the function maybe gets called again and again, causing API spam.

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php
@@ -135,12 +135,12 @@ class BillingAgreementsEndpoint {
 				);
 			} finally {
 				$this->is_request_logging_enabled = true;
-				set_transient( 'ppcp_reference_transaction_enabled', true, MONTH_IN_SECONDS );
 			}
 
+			set_transient( 'ppcp_reference_transaction_enabled', true, MONTH_IN_SECONDS );
 			return true;
 		} catch ( Exception $exception ) {
-			delete_transient( 'ppcp_reference_transaction_enabled' );
+			set_transient( 'ppcp_reference_transaction_enabled', false, HOUR_IN_SECONDS );
 			return false;
 		}
 	}


Dinamiko (approved):

mreishus:
@Narek13 Thanks for adding the failure caching! I noticed it might not be working as intended:

https://github.com/woocommerce/woocommerce-paypal-payments/blob/055f396c0be1f0d9e21501cc4336b372c6e68e11/modules/ppcp-api-client/src/Endpoint/BillingAgreementsEndpoint.php#L122-L146

The issue is that the early return logic only checks for true:
```php
if (wc_string_to_bool(get_transient('ppcp_reference_transaction_enabled')) === \true) {
    return \true;
}
```

However, when the transient is set to false (from the new caching line), it doesn't return early. This means `create_token()` gets called again even though we have a cached failure that should be valid for 1 hour.

I looked into it a bit further and found that `get_transient()` returns false on failure, meaning that we can't distinguish between a missing transient and a set false. [Doc page](https://developer.wordpress.org/reference/functions/get_transient/).

So, to stop the API requests, I think the code needs to set something other than `false` and handle all three cases (cached success, cached failure, no cache).

Narek13:
Thanks @mreishus for your comment. Currently we [refactor](https://github.com/woocommerce/woocommerce-paypal-payments/pull/3495) this part to determine merchant eligibility for Reference Transactions using Seller Status API, instead of using the billing agreements, so the caching issue hopefully should be resolved.

PR #3463, opened 2025-06-17
Update Shipping entity to include contact fields (email, phone) (4808)

stracker-phil:
_This PR extends #3457_

### Description

Extend the `Shipping` class to contain the two optional properties `email_address` and `phone_number`

Those properties are populated from the API's response object `purchase_units[].shipping`, so they depend on the presence of the shipping address in the order data


Narek13 (approved):
LGTM. Thanks

danieldudzic (approved):
Straightforward!

PR #3466, opened 2025-06-18
Override email recipient with shipping contact (4810)

stracker-phil:
_This PR extends #3465_

### Description

Swaps out the order's billing details (email and phone) with contact details provided by the contact module.

At the same time, the original billing details are preserved as custom order-meta entries that are visible in the order details page to the shop owner.

A "side effect" of this PR is that all customer-facing emails are sent to the custom contact address which


danieldudzic (approved):
Looks good!

PR #3474, opened 2025-06-20
Add server-side shipping callback

AlexP11223:
Added the [server-side shipping callback](https://developer.paypal.com/docs/checkout/standard/customize/shipping-module/) to replace the currently used JS callback.

Currently it is disabled by default, can be enabled via `PCP_SERVER_SIDE_SHIPPING_CALLBACK_ENABLED=1` env var or filter:

    add_filter( 'woocommerce.feature-flags.woocommerce_paypal_payments.server_side_shipping_callback_enabled', '__return_true' );

It is used when the shipping address or method can be changed in the PayPal popup, that is the block express buttons, product/cart buttons. Also works for Pay Later, and should work for Venmo but currently cannot test it.

In the callback we need to determine and return the available shipping methods for the received address, also get the updated amount. The available shipping methods may depend on the cart content and the customer, so the only way we found is via the Store API: https://developer.woocommerce.com/docs/apis/store-api/resources-endpoints/cart 
That is we call Get Cart when creating the PayPal order and add the cart token into the URL, then in the callback execute the Update Customer request and use the returned cart data.

The address data we receive is not complete, only country, state, city, postal code, but it is used only for determining the shipping methods and amount, we get the full address via API when completing the order.

Some of the improvements that may follow in the next PRs:

- some way to check if the URL is accessible, at least partially (from the same server) to be able to fallback to JS and avoid breaking payments, until a better solution is not provided by PayPal.
- (less important) add some kind of nonce into the URL to make sure that the callback comes from PayPal. Currently it would be somewhat redundant because we already pass the Store API Cart Token this way, but may be a good practice to add a dedicated check in case we change something in the future.
 
---

Also I disabled some of the PHPCS rules about requiring PHPDoc. As discussed recently, it does not really lead to better documentation and forces us to add lots of useless comments like "`<class> Constructor`", "`@param FooFactory $foo_factory Foo factory`". Especially one of the weirdest thing was that both class and file PHPDoc were required, so we had to either copy-paste or make one of them useless. (and the purpose of `@package <namespace>` is completely unclear, I guess it may be used by some tools generating documentation pages, but we don't use them, and sometimes even forget to update the namespace)

### DIFF
--- a/modules/ppcp-wc-gateway/src/Endpoint/ShippingCallbackEndpoint.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/ShippingCallbackEndpoint.php
@@ -16,8 +16,8 @@ use WP_REST_Response;
  */
 class ShippingCallbackEndpoint {
 
-	const NAMESPACE = 'paypal/v1';
-	const ROUTE     = 'shipping-callback';
+	private const NAMESPACE = 'paypal/v1';
+	private const ROUTE     = 'shipping-callback';
 
 	private CartEndpoint $cart_endpoint;
 
--- a/modules/ppcp-wc-gateway/src/Endpoint/ShippingCallbackEndpoint.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/ShippingCallbackEndpoint.php
@@ -1,0 +1,156 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\WcGateway\Endpoint;
+
+use Psr\Log\LoggerInterface;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\ShippingOption;
+use WooCommerce\PayPalCommerce\ApiClient\Factory\AmountFactory;
+use WooCommerce\PayPalCommerce\WcGateway\StoreApi\Endpoint\CartEndpoint;
+use WooCommerce\PayPalCommerce\WcGateway\StoreApi\Factory\ShippingRate;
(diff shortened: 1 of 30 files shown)


modules/ppcp-wc-gateway/src/StoreApi/Entity/CartTotals.php
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	public function total_price(): Money {
+		return $this->total_price;
+	}
+
+	public function total_tax(): Money {
+		return $this->total_tax;
+	}
+
```
Narek13, line 94: How cool and clean is this whole class without that unnecessary comments. Thanks for disabling those rules. 💪

modules/ppcp-wc-gateway/src/Endpoint/ShippingCallbackEndpoint.php
```diff
@@ -0,0 +1,156 @@
...
+use WooCommerce\PayPalCommerce\WcGateway\StoreApi\Factory\ShippingRate;
+use WP_REST_Response;
+
+/**
+ * Handles the shipping callback.
+ */
+class ShippingCallbackEndpoint {
+
+	const NAMESPACE = 'paypal/v1';
+	const ROUTE     = 'shipping-callback';
```
Narek13: Can you please add visibilities for these?

Dinamiko (approved):

PR #3476, opened 2025-06-20
Ditch items logic does not work when using saved card payment (4884)

Dinamiko:
The Subtotal mismatch fallback logic to prevent `ITEM_TOTAL_MISMATCH` error when creating the PayPal order was not executed when capturing the order using a vaulted card payment. This PR fixes it and handles the error adding a log entry when it fails.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,3 @@
+{
+  "enableAllProjectMcpServers": false
+}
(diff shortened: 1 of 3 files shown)


Narek13 (approved):
LGTM. Thanks

hmouhtar (approved):

PR #3479, opened 2025-06-23
[dev] Pin Github actions to full-commit SHA

adimoldovan:
### Description

Closes [QAO-42](https://linear.app/a8c/issue/QAO-42/pin-github-actions-to-full-commit-shas-for-immutability)
Contributes to [QAO-22](https://linear.app/a8c/issue/QAO-22), [QAO-23](https://linear.app/a8c/issue/QAO-23)

Pins Github actions to full commit SHAs as immutable references.
Skipped all official GitHub actions.  
No version updates - used the latest commit in the tag that was already used.

### Steps to Test

CI should pass.

### Documentation

### Changelog Entry

CI updates

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,7 +12,7 @@ jobs:
 
     name: PHP ${{ matrix.php-versions }} WC ${{ matrix.wc-versions }}
     steps:
-    - uses: ddev/github-action-setup-ddev@v1
+    - uses: ddev/github-action-setup-ddev@1c7ef18595da42355373cb6d9417a6f44d758b93 # v1
       with:
         autostart: false
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,7 +12,7 @@ jobs:
 
     name: PHP ${{ matrix.php-versions }} WC ${{ matrix.wc-versions }}
     steps:
-    - uses: ddev/github-action-setup-ddev@1c7ef18595da42355373cb6d9417a6f44d758b93 # v1
+    - uses: ddev/github-action-setup-ddev@1c7ef18595da42355373cb6d9417a6f44d758b93 # v1.10.1
       with:
         autostart: false
 
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Setup PHP
-      uses: shivammathur/setup-php@v2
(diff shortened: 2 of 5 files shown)


.github/workflows/php.yml
```diff
@@ -14,15 +14,15 @@ jobs:
...
+      uses: shivammathur/setup-php@0f7f1d08e3e32076e51cae65eb0b0c871405b16e # v2.34.1
       with:
         php-version: ${{ matrix.php-versions }}
 
     - name: Validate composer.json and composer.lock
       run: composer validate
 
     - name: Install dependencies
-      uses: ramsey/composer-install@v1
+      uses: ramsey/composer-install@994bb194a4fefcf39449ccf0f7766a4318f1ac76 # v1
```
annemirasol: How did you decide to use this commit? I see the last v1 as [1.3.0 and a different hash](https://github.com/ramsey/composer-install/commit/a7320a0581dcd0432930c48a0e7ced67e6ec17e8).

adimoldovan: I took the last commit of `v1`: https://github.com/ramsey/composer-install/commits/v1/

diegocurbelo: It seems we can should the commit `a7320a0581dcd0432930c48a0e7ced67e6ec17e8` then, which is the `1.3.0` release: https://github.com/ramsey/composer-install/releases/tag/1.3.0

diegocurbelo:
```suggestion
      uses: ramsey/composer-install@a7320a0581dcd0432930c48a0e7ced67e6ec17e8 # v1.3.0
```

.github/workflows/integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     name: PHP ${{ matrix.php-versions }} WC ${{ matrix.wc-versions }}
     steps:
-    - uses: ddev/github-action-setup-ddev@v1
+    - uses: ddev/github-action-setup-ddev@1c7ef18595da42355373cb6d9417a6f44d758b93 # v1
```
diegocurbelo:
```suggestion
    - uses: ddev/github-action-setup-ddev@1c7ef18595da42355373cb6d9417a6f44d758b93 # v1.10.1
```

diegocurbelo (changes requested):

.github/workflows/php.yml
adimoldovan: I guess we could do that, I don't see any harm. I wanted to use the same commit that is resolved by the version that I found in the workflow, to minimize the risk of breaking anything, which was `v1`. Interesting that the `v1` commit is more recent than the `v1.3.0` release one.

diegocurbelo (approved):
Thanks @adimoldovan!

PR #3487, opened 2025-06-25
OXXO payment gateway requires ORDER_COMPLETE_ON_PAYMENT_APPROVAL processing_instruction (4967)

Dinamiko:
When OXXO is used for payment, the transaction is not completed.

### Steps To Reproduce
— Store country is Mexico and currency is MXN
- Add product to the cart and navigate to classic checkout
- Fill out the form and select OXXO payment and place the order.

This PR fixes the issue by removing `application_context` from `OrderEndpoint.php::confirm_payment_source` and adding `processing_instruction` and `locale` to `OXXO/OXXOGateway::process_payment`.


Narek13 (approved):
LGTM. Thanks.

PR #3490, opened 2025-06-26
Legacy UI - Mexico - Notice on frontend when wp debugging is enabled  (4972)

Dinamiko:
### Steps To Reproduce
- Onboard as Mexico merchant in legacy UI
- Enable WP debugging
- Navigate to shop 
- Observe notice: `Try to access array offset on value of type bool`

This PR fixes the issue by setting empty `string` if given setting is `null`.


AlexP11223 (approved):

Narek13 (approved):
Thanks

PR #3491, opened 2025-06-26
UX Legacy Migration: phase 1 opt in via banner (4854)

Narek13:
## Issue Description
Update 3.0.0 launched the new UI for fresh plugin installations. We now need to implement mechanisms for existing users to migrate to the new UI.

**Phase 1 Requirements:**
Merchants using the legacy UX will see a banner on all admin panel pages with:
- **Learn more link** - directing to new experience documentation  
- **"Switch to New Settings" button** - for merchant opt-in migration

**Implementation Goals:**
- Detect if merchant is using legacy UX
- Display migration banner on admin panel pages
- Handle migration button click to save preference and reload with new UI

## Solution Overview
This PR implements:

1. **Legacy UX Detection** - Check if user should see old vs new interface
2. **Admin Banner Display** - Show migration notice on relevant admin pages
3. **Confirmation Flow** - Warn users that migration cannot be undone
4. **Migration Handler** - Process user opt-in and switch to new UI

## Technical Implementation
- Added banner notice with translatable text and proper links
- Implemented JavaScript handler for migration button with confirmation
- Created migration endpoint to save settings with the new models

## Testing
**Scenario: Old settings UI active**

**GIVEN** the old settings UI is active
**WHEN** the PayPal Payments settings page is visited
**THEN** a dismissable banner is visible, which prompts the user to activate the new UI
**AND** a button is visible that enables the new UI

### DIFF
--- a/modules/ppcp-settings/resources/js/switchSettingsUi.js
+++ b/modules/ppcp-settings/resources/js/switchSettingsUi.js
@@ -3,12 +3,21 @@ document.addEventListener( 'DOMContentLoaded', () => {
 	const button = document.querySelector(
 		'.button.button-settings-switch-ui'
 	);
+	const link = document.querySelector( 'a.settings-switch-ui' );
 
-	if ( ! typeof config || ! button ) {
+	if ( ! typeof config || ( ! button && ! link ) ) {
 		return;
 	}
 
-	button.addEventListener( 'click', () => {
+	const handleClick = ( event ) => {
+		event.preventDefault();
+
+		const confirmed = confirm( config.confirmMessage );
+
+		if ( ! confirmed ) {
+			return;
+		}
+
 		fetch( config.endpoint, {
 			method: 'POST',
 			headers: {
@@ -30,5 +39,13 @@ document.addEventListener( 'DOMContentLoaded', () => {
 			.catch( ( error ) => {
 				console.error( 'Error:', error );
 			} );
-	} );
+	};
+
+	if ( button ) {
+		button.addEventListener( 'click', handleClick );
+	}
+
+	if ( link ) {
+		link.addEventListener( 'click', handleClick );
+	}
 } );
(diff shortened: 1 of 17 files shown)


modules/ppcp-settings/resources/js/switchSettingsUi.js
```diff
@@ -3,12 +3,21 @@ document.addEventListener( 'DOMContentLoaded', () => {
 	const button = document.querySelector(
 		'.button.button-settings-switch-ui'
 	);
+	const link = document.querySelector( 'a.settings-switch-ui' );
 
-	if ( ! typeof config || ! button ) {
+	if ( ! typeof config || ( ! button && ! link ) ) {
```
Dinamiko: I see `! typeof config` was already there but I suggest to change it to: `typeof config === 'undefined'` as `typeof config` returns a string.

modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
```diff
@@ -0,0 +1,87 @@
...
+		$this->payment_settings = $payment_settings;
+		$this->local_apms       = $local_apms;
+	}
+
+	public function migrate(): void {
+		$allow_local_apm_gateways = $this->settings->has( 'allow_local_apm_gateways' ) && $this->settings->get( 'allow_local_apm_gateways' );
+
+		if ( $this->settings->has( 'disable_funding' ) ) {
+			$disable_funding = $this->settings->get( 'disable_funding' );
+			if ( ! in_array( 'venmo', $disable_funding, true ) ) {
```
Dinamiko, line 51: Here assumes `disable_funding` is always an array, better ensure is array before using it.

Dinamiko (changes requested):

Dinamiko (approved):

PR #3493, opened 2025-06-27
Add Fastlane 3D Secure support and enable for the UK (4487)

danieldudzic:
## Description
This PR will introduce 3D Secure authentication support for Fastlane payments and enable Fastlane availability in the UK market, with robust error handling and improved order processing for 3DS flows.

### 3D Secure Support

This PR will:
- Refactor `OrderFactory` with dedicated helper methods for processing PayPal order responses, improving maintainability and 3DS support.
- Add a `links` property to the `Order` entity to store 3DS payer action URLs, enabling proper payer authentication handling.
- Enhance the Fastlane gateway and return URL endpoint to support 3DS flows, including handling `PAYER_ACTION_REQUIRED` status, order capture after 3DS, and user-friendly error messages for failed authentication or payment declines.
- Update the `SettingsModel` with an enum mapping for 3D Secure configuration, ensuring correct API usage for SCA settings.
- Improve test coverage for 3DS-related code paths.

### Enable Fastlane for UK Market

This PR will:
- Update Fastlane's country-currency and card-type matrices to support the UK (GB/GBP) when enabled.
- Introduce a `axo.uk.enabled` feature flag controlled by the `PCP_AXO_UK_ENABLED` environment variable and a new filter, allowing conditional enablement of UK Fastlane.
- Update matrix logic to dynamically include UK support based on the feature flag.

## Testing
This PR should be tested for:
- 3D Secure authentication flows for Fastlane payments (including payer action redirects and post-auth completion)
- Fastlane payments in the UK market when the feature flag is enabled
- Error handling for failed or unavailable 3DS authentication, payment declines, and expired sessions
- Successful payment completion after 3DS verification
- Backwards compatibility for non-3DS flows and non-UK markets

## 3D Secure Test Scenarios
You can use the following credit card details to test various 3D Secure scenarios:  
https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/test/


modules/ppcp-axo/src/Gateway/AxoGateway.php
```diff
@@ -277,6 +322,100 @@ public function process_payment( $order_id ) {
...
+			return __( 'Payment session expired. Please try your payment again.', 'woocommerce-paypal-payments' );
+		}
+
+		return __( 'There was an error processing your payment. Please try again or contact support.', 'woocommerce-paypal-payments' );
+	}
+
+	/**
+	 * Extract payer action URL from PayPal order.
+	 */
+	private function get_payer_action_url( Order $order ) {
```
Narek13:
```suggestion
	private function get_payer_action_url( Order $order ): string {
```

Forgive me if I am wrong, but shouldn't this method return string?

modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
```diff
@@ -75,22 +78,37 @@ public function __construct(
...
+				$order = $this->complete_3ds_verification( $order );
+			} catch ( Exception $e ) {
+				$this->logger->warning( "3DS completion failed for order $token: " . $e->getMessage() );
+				wc_add_notice( $this->get_3ds_error_message( $e ), 'error' );
+				wp_safe_redirect( wc_get_checkout_url() );
+				exit();
+			}
 		}
 
+		// Get WooCommerce order ID.
```
Narek13: As discussed, please avoid adding inline comments for obvious things.

modules/ppcp-axo/src/Gateway/AxoGateway.php
```diff
@@ -266,7 +306,12 @@ public function process_payment( $order_id ) {
 				$this->order_processor->process_captured_and_authorized( $wc_order, $order );
 			}
 		} catch ( Exception $exception ) {
-			return $this->handle_payment_failure( $wc_order, $exception );
+			// Error handling for payment failures.
```
Narek13: Same here

modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
```diff
@@ -130,7 +161,95 @@ function( $allowed_hosts ) : array {
...
 		exit();
 	}
+
+	/**
+	 * Check if order needs 3DS completion.
+	 *
+	 * @param mixed $order The PayPal order.
+	 * @return bool
+	 */
+	private function needs_3ds_completion( $order ): bool {
```
Narek13:
```suggestion
	 * @param Order $order The PayPal order.
	 * @return bool
	 */
	private function needs_3ds_completion( Order $order ): bool {
```

I think this should have a [type](https://github.com/woocommerce/woocommerce-paypal-payments/blob/9ef34dc7737e575afac27dcd8673f2fdd6080440/modules/ppcp-api-client/src/Entity/Order.php#L15), otherwise how do we use `$order->status()` ?

modules/ppcp-axo/src/AxoModule.php
```diff
@@ -367,6 +369,56 @@ static function ( $methods ) {
...
+				$settings_model = $c->get( 'settings.data.settings' );
+				assert( $settings_model instanceof SettingsModel );
+
+				$three_d_secure = $settings_model->get_three_d_secure_enum();
+
+				if (
+					$three_d_secure === 'SCA_ALWAYS'
+					|| $three_d_secure === 'SCA_WHEN_REQUIRED'
+				) {
+					$data['payment_source']['card']->attributes = array(
```
Narek13: Are we sure that the `$data['payment_source']` and `$data['payment_source']['card']` exist?

Narek13 (commented):
Thanks for the great work @danieldudzic .
I haven't tested this locally but LGTM.
Just left minor suggestions or doubts/questions

modules/ppcp-api-client/src/Factory/OrderFactory.php
```diff
@@ -81,70 +81,223 @@ public function from_wc_order( \WC_Order $wc_order, Order $order ): Order {
...
+	 * @param bool      $is_3ds_response Whether this is a 3DS response.
+	 *
+	 * @return array Array of PurchaseUnit objects.
+	 * @throws RuntimeException When purchase units are required but missing.
+	 */
+	private function create_purchase_units( \stdClass $order_data, bool $is_3ds_response ): array {
+		// 3DS responses don't contain purchase units.
+		if ( $is_3ds_response ) {
+			return array();
+		}
```
AlexP11223, line 133: hm, do we really need this `is_3ds` flag here and in other places? Maybe we can just remove the "not contain" exceptions and return empty/default values where makes sense? And if we really need to check if e.g. PU exists, it probably fits more at some higher levels.

modules/ppcp-api-client/src/Factory/OrderFactory.php
```diff
@@ -81,70 +81,185 @@ public function from_wc_order( \WC_Order $wc_order, Order $order ): Order {
...
+
+	/**
+	 * Returns an Order object based off a PayPal 3DS Response.
+	 *
+	 * @param \stdClass $order_data The JSON object.
+	 *
+	 * @return Order
+	 * @throws RuntimeException When JSON object is malformed.
+	 */
+	public function from_paypal_response_with_3ds( \stdClass $order_data ): Order {
```
AlexP11223: About removing the flag, I meant is there something that prevents us from using the same `from_paypal_response` for both cases? So that we do not need to remember that we may need to check the status and call one of these methods if we retrieve the order response somewhere else.

If the main difference is just that we add links, then can't we just add it in `from_paypal_response` and remove `from_paypal_response_with_3ds`?

Or if they need to be separate, then maybe can handle the `->status === 'PAYER_ACTION_REQUIRED'` check in `OrderFactory`.

Narek13 (approved):
Thanks for making all the improvements.

modules/ppcp-api-client/src/Factory/OrderFactory.php
danieldudzic: Good point.

I initially created separate methods to be explicit about 3DS vs regular responses, but you're right - it creates unnecessary duplication and complexity.

Your suggestion to consolidate into one `from_paypal_response` method that automatically handles both cases (with links when present) is much cleaner. I'll refactor it to use a single method.

modules/ppcp-axo/src/AxoModule.php
```diff
@@ -367,6 +369,60 @@ static function ( $methods ) {
...
+					$card->attributes = array(
+						'verification' => array(
+							'method' => $three_d_secure,
+						),
+					);
+
+					$experience_context_builder = $c->get( 'wcgateway.builder.experience-context' );
+					assert( $experience_context_builder instanceof ExperienceContextBuilder );
+
+					$data['experience_context'] = $experience_context_builder
```
AlexP11223: I think it's set incorrectly? `experience_context` is supposed to be in `payment_source.*.experience_context`.

Also maybe it can be added where the order creation call is triggered initially (`AxoGateway`?)? hm, actually why this `ppcp_create_order_request_body_data` override is needed at all if it's only for `AxoGateway`? Wouldn't it be simpler to just put it in the `AxoGateway`?

Also if it's always only `card` then the API docs say it supports only cancel/return URLs.

AlexP11223 (changes requested):

modules/ppcp-axo/src/AxoModule.php
danieldudzic: Thanks for the feedback, again @AlexP11223 :) Much cleaner approach - appreciate your input! 🙏

The Fastlane 3DS docs show `experience_context` at the root level, not nested under `payment_source.card`.

I have kept it since it provides the return/cancel URLs for authentication redirects.

AlexP11223 (approved):

PR #3502, opened 2025-07-02
Add a logic to deactivate the plugin when Woo is not active (4780)

Narek13:
## Issue Description

When WooCommerce plugin is disabled but PayPal Payments is a active, the site crashes with message sometimes.

```
Fatal error: Uncaught LogicException: No PPCP container, probably called too early when the plugin is not initialized yet.
```

###Steps To Reproduce

- set up Multisite
- install WooCommerce & PayPal Payments
- network activate WooCommerce
- navigate to the main site and activate PayPal Payments on the site
- network deactivate WooCommerce
- navigate to main site
- → observe fatal error

## PR Description

Added a logic to deactivate the plugin when Woo is not active

### DIFF
--- a/woocommerce-paypal-payments.php
+++ b/woocommerce-paypal-payments.php
@@ -95,9 +95,13 @@ define( 'PPCP_PAYPAL_BN_CODE', 'Woo_PPCP' );
 					),
 					wp_kses(
 						sprintf(
-						// translators: %s is a link to WooCommerce.com.
-							esc_html__( 'WooCommerce PayPal Payments requires WooCommerce to be installed and active. You can download %s here.', 'woocommerce-paypal-payments' ),
-							'<a href="https://woocommerce.com/" target="_blank">WooCommerce</a>'
+						// translators: %s is a link to install WooCommerce.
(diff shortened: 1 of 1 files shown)


woocommerce-paypal-payments.php
```diff
@@ -43,29 +43,84 @@
...
+					'<div class="notice notice-error"><span class="notice-title">%1$s</span><p>%2$s</p></div>',
+					esc_html__(
+						'The plugin WooCommerce PayPal Payments has been deactivated',
+						'woocommerce-paypal-payments'
+					),
+					wp_kses(
+						sprintf(
+						// translators: %s is a link to WooCommerce.com.
+							esc_html__( 'WooCommerce PayPal Payments requires WooCommerce to be installed and active. You can download %s here.', 'woocommerce-paypal-payments' ),
+							'<a href="https://woocommerce.com/" target="_blank">WooCommerce</a>'
```
InpsydeNiklas: I suggest `network_admin_url('plugin-install.php?tab=plugin-information&plugin=woocommerce')` as target instead of  `woocommerce.com`. 
Additionally, I suggest including the entire sentence `You can download WooCommerce here.` in the linked text as well.

Dinamiko (approved):

PR #3504, opened 2025-07-04
Add support for AppSwitch (4712)

hmouhtar:
This PR introduces support for [AppSwitch](https://developer.paypal.com/docs/checkout/standard/customize/app-switch/) on the PayPal buttons for classic and block contexts.

As per the docs, since AppSwitch does not support continue flows, it will only be enabled on Pay Now flows with server shipping callback.

A new `ReturnUrlFactory` was added to handle the limitation of AppSwitch using hash parameters to send the order data when returning from the PayPal App, meaning that the old approach of using an AJAX endpoint as a middleman would not work since the hash parameters never make it to the server, so `ReturnUrlFactory` generates the correct URL based on the provided context. This is only used on the ppcp-button module endpoints, which rely on the JS SDK, so none of the regular API gateway logic will be affected by this change.

For classic contexts, there was a problem where the resume flow would end in an infinite loop on errors (resume flow triggered > error > re-render button > resume flow triggered...). Fixing this infinite loop meant refactoring a significant part of the existing logic, so I opted for manually cleaning the AppSwitch hash parameters from the URL before triggering the reload buttons event.

The `onApprove` callback for the block components relied on the `actions.order.get` function, which is not available on AppSwitch flows, so a new GetOrderEndpoint was created to get the PayPal order data in this scenario.

### Features for future releases

Under some circumstances, AppSwitch can send the user back to a different browser from where the request was initiated. PayPal expects us to handle this scenario, which probably means we will need to find a way to authenticate users based on the order data, though this will be trickier for guest checkouts.

Other possible changes are around the UX. On classic components, during the resume flow the buttons are not rendered, so there is no visual feedback to the user that the payment is ongoing for a couple of seconds. It is yet to be decided if this will be handled by us or by the PayPal SDK.


modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
```diff
@@ -69,6 +70,13 @@ class CreateOrderEndpoint implements EndpointInterface {
 	 */
 	private $shipping_preference_factory;
 
+	/**
+	 * The return URL factory.
+	 *
+	 * @var ReturnUrlFactory
+	 */
+	private $return_url_factory;
```
Narek13:
```suggestion
	private ReturnUrlFactory $return_url_factory;
```

modules/ppcp-button/src/Endpoint/GetOrderEndpoint.php
```diff
@@ -0,0 +1,84 @@
...
+use WooCommerce\PayPalCommerce\ApiClient\Endpoint\OrderEndpoint;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
+
+/**
+ * Class GetOrderEndpoint
+ */
+class GetOrderEndpoint implements EndpointInterface {
+
+	const ENDPOINT = 'ppc-get-order';
```
Narek13: Please add visibility.

```suggestion
	const public ENDPOINT = 'ppc-get-order';
```

modules/ppcp-api-client/src/Factory/ReturnUrlFactory.php
```diff
@@ -0,0 +1,57 @@
...
+			case 'product':
+				if ( ! empty( $request_data['purchase_units'] ) && is_array( $request_data['purchase_units'] ) ) {
+					$first_unit = reset( $request_data['purchase_units'] );
+					if ( ! empty( $first_unit['items'] ) && is_array( $first_unit['items'] ) ) {
+						$first_item = reset( $first_unit['items'] );
+						if ( ! empty( $first_item['url'] ) ) {
+							return $first_item['url'];
+						}
+					}
+				}
```
Narek13, line 37: What will happen if code enters the top level `if` and then `$first_unit['items']` or `$first_item['url']` does not exist or is empty? the code will just not return anything while it should return `string`. I would use early return when the code enters the top level `if`

Narek13 (commented):
Thanks for working on this @hmouhtar. I haven't tested but LGTM.
Left couple of suggestions and doubts.

modules/ppcp-api-client/src/Factory/ReturnUrlFactory.php
hmouhtar: @Narek13 The only scenario where something would be returned is if the item URL exists in the array. Otherwise, the code continues and the exception is thrown. I can't use an early return because AppSwitch will fail if the wrong return URL is passed, so this should be treated as a fatal exception

Narek13: Thanks for the explanation.

Narek13 (approved):
Thanks. LGTM

Dinamiko (approved):

Issue #438, opened 2022-01-14
doubled plugin module URL path causing failure (499)


AlexP11223:
Only happens on a few isolated merchant sites. But it does not appear to be a plugin conflict.

Smart buttons (and other JS, CSS files) won't load because the request URL is incorrect:

![image](https://user-images.githubusercontent.com/5680466/149537595-c928b88b-f63a-4a53-95ac-f6424c93e92a.png)

### Possible cause

`__FILE__` for some reason returns weird paths like this `....../wp-content/plugins/woocommerce-paypal-payments/vendor/composer/../../modules/ppcp-onboarding/src/../services.php` instead of `....../wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-onboarding/services.php`

PR #3505, opened 2025-07-04
Create a centralized service for a path to the main plugin file (4970)

Narek13:
## Issue Description

The codebase has multiple service files that use `realpath(__FILE__)` without proper error handling, which can cause fatal `TypeError` exceptions when the file path cannot be resolved. This issue was identified in production where `realpath()` returned `false`, causing `dirname()` to receive a boolean instead of a string.

## PR Description

This PR adds a centralized service for a path to the main plugin file with a fallback to use the `__FILE__` when `realpath()` fails


src/services.php
```diff
@@ -30,4 +30,6 @@
 	'http.redirector'    => function( ContainerInterface $container ) : RedirectorInterface {
 		return new WpRedirector();
 	},
+	'ppcp.plugin-path'   => static fn(): string =>
+		dirname( realpath( __FILE__ ) ?: __FILE__, 2 ) . '/woocommerce-paypal-payments.php',
```
AlexP11223: Maybe it would be better to find a way to avoid `dirname`, somehow pass `__FILE__` from the plugin file (or not sure if there is some WP function to get it). This would avoid #438 even if `realpath` fails.

Or actually maybe a simpler way is to use `baseUri` from `Package::PROPERTIES` for creating the URLs. https://github.com/inpsyde/modularity/blob/master/docs/Properties.md
Either directly `baseUri . myPath` in all these services or wrapping into a service like this accepting module name and asset path https://github.com/inpsyde/worldline-for-woocommerce/blob/main/inc/services.php#L27-L39 though it's a bit more error-prone because Psalm cannot check the usage of such service (parameters, types).

Narek13: Thanks @AlexP11223 .
I have added a service configs, which are using the `Package::PROPERTIES`:

- path-to-plugin-folder
- path-to-plugin-main-file as the `$properties->pluginMainFile()` returns path to `bootstrap.php` instead of the `woocommerce-paypal-payments.php` for some reason

AlexP11223: oh, this seems to be something that should be fixed. We need to pass the correct path here https://github.com/woocommerce/woocommerce-paypal-payments/blob/f99f3759335fee41df4f071a325a2754ad7b2cbb/bootstrap.php#L35 so either by adding `woocommerce-paypal-payments.php` to `$root_dir` or adding the plugin file parameter to bootstrap and passing `__FILE__` here https://github.com/woocommerce/woocommerce-paypal-payments/blob/f99f3759335fee41df4f071a325a2754ad7b2cbb/woocommerce-paypal-payments.php#L78

AlexP11223 (approved):

src/services.php
Narek13: - added `"coenjacobs/mozart"` as dev dependency so we can update packages within the `lib` dir.
- Updated all the packages using `mozart` within the `lib` dir.
- Fixed the `PluginProperties` by providing correct path to the main file: `"$root_dir/woocommerce-paypal-payments.php"`

PR #3515, opened 2025-07-09
Broken PayPal Subscriptions payment in blocks and classic cart (4982, 4943, 4981)

Dinamiko:
### Steps To Reproduce
- Onboard as US business merchant with Subscriptions and without ACDC
- Disable “Save PayPal and Venmo” and “Save Credit and Debit Cards“
- Create a subscription product and connect it to PayPal
- Add product to cart and try to pay in the following pages: classic cart, blocks cart and checkout

This PR fixes JS and PHP errors related to undefined array access in PayPal subscription handling. The main issues were:
- context is not defined error in `CartActionHandler.js` causing subscription creation failures
- "Undefined offset: 0" PHP notice in `WooCommerceOrderCreator.php`
- "Cannot read properties of undefined (reading '0')" JavaScript error in `Address.js`

It also fixes a wrong condition that loaded `paypal-subscription.js` file not only in admin product screen but in multiple non related screens.

### DIFF
--- a/modules/ppcp-blocks/resources/js/Helper/Address.js
+++ b/modules/ppcp-blocks/resources/js/Helper/Address.js
@@ -81,14 +81,14 @@ export const paypalShippingToWc = ( shipping ) => {
 export const paypalPayerToWc = ( payer ) => {
 	const firstName = payer?.name?.given_name ?? '';
 	const lastName = payer?.name?.surname ?? '';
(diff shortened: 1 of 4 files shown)


AlexP11223 (approved):

Narek13 (approved):
LGTM. Thanks

PR #3518, opened 2025-07-10
Fix contacts race condition in block

AlexP11223:
There was a race condition in express block checkout when setting a custom contact email/phone in the popup, 
they were reverted in the next order save call in another WC request. Not sure why since it's tricky to debug, I guess WC does not have any sync here, so if the second request loads the order data before my save it will have the old data and it will be detected as new changes.

So now setting the contacts again in the `woocommerce_store_api_cart_update_order_from_request` hook which is fired after that second save. The contacts are initially saved into the order meta to be able to retrieve them later without an API request. Also there is now `woocommerce_paypal_payments_contacts_added` hook, it is used for setting the contacts in other cases instead of directly in `add_paypal_meta` to avoid code duplication.

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -92,6 +92,7 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 	public function run( ContainerInterface $c ): bool {
 		$this->register_payment_gateways( $c );
 		$this->register_order_functionality( $c );
+		$this->register_contact_handlers();
 		$this->register_columns( $c );
 		$this->register_checkout_paypal_address_preset( $c );
 		$this->register_wc_tasks( $c );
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -1011,6 +1012,43 @@ private function is_order_paid_by_paypal( $order ) : bool {
...
+	 *
+	 * The contact module can provide a custom email and phone number.
+	 * These two properties are intended to be treated as primary contact details.
+	 */
+	private function register_contact_handlers(): void {
+		$set_order_contacts = function ( WC_Order $wc_order ): void {
+			$email = $wc_order->get_meta( PayPalGateway::CONTACT_EMAIL_META_KEY );
+			$phone = $wc_order->get_meta( PayPalGateway::CONTACT_PHONE_META_KEY );
+
+			if ( $email && is_string( $email ) ) {
```
Dinamiko: Maybe better use `is_email` instead of `is_string`?

Dinamiko (approved):

modules/ppcp-wc-gateway/src/WCGatewayModule.php
AlexP11223: Yeah, I thought to avoid that in case this validation is not the same as in PayPal and may incorrectly reject some emails, but probably makes sense to check it since WC checks it anyway and throws an exception.

PR #3533, opened 2025-07-15
Block UI during onApprove callback (4712)

hmouhtar:
When returning from AppSwitch, the `onApprove` callback takes a few seconds to finish processing. During this time, the user could click the PayPal button again or trigger other WooCommerce actions like "Proceed to Checkout", which disrupted the AppSwitch flow and caused unexpected issues reported by QA.

This PR addresses the problem by implementing the same approach used at checkout: blocking the entire UI while the payment is being processed. 

**Classic contexts:**

To address this in classic contexts, I've reused the `Spinner` class, triggering a full-page blocking overlay when the onApprove callback is triggered, and removing it in the `.finally` callback.

**Block contexts:**

Block contexts are trickier. Unblocking the spinner in the `.finally` callback of the `onApprove` method was too soon, since WooCommerce has its own async order processing in block contexts. This left a significant time window in which the user could perform actions like clicking "Proceed to Checkout".

To handle block contexts, I’ve set up a useEffect that tracks the value of `isFullPageSpinnerActive` and toggles the spinner when it changes. This allows the spinner to be enabled in the onApprove callback, remain active, and be removed only in exceptional circumstances such as checkout errors which are handled by WooCommerce events.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ActionHandler/CheckoutActionHandler.js
@@ -158,7 +158,7 @@ class CheckoutActionHandler {
 		};
 		return {
 			createOrder,
-			onApprove: onApprove( this, this.errorHandler, this.spinner ),
+			onApprove: onApprove( this, this.errorHandler ),
 			onCancel: () => {
 				spinner.unblock();
 			},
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
LGTM. Thanks for working on this.

Narek13 (approved):

PR #3536, opened 2025-07-16
Add "Things to do next" item after settings migration (5009)

Narek13:
## Issue Description

**Add "Things to do next" item after settings migration to encourage users to check their configuration**

**Description:** 
After users migrate from the old UI to the new UI using the migration button, we need to add a prominent action item in the "Things to do next" section to guide them through configuration verification.

**Requirements:**
* Add a new action item to the "Things to do next" section that appears immediately after successful migration
* The action item should be the first/top item users see post-migration
* Action item text: "**You're now using the new PayPal Payments interface!** Complete the items below to ensure your payment configuration is optimized for your store."

**Context:** 
The "Things to do next" section is a dedicated area in the plugin that serves as the primary user guidance interface. Since this is the first thing users see after migration, it's the ideal place to prompt configuration review.

**Technical Notes:**
* Should trigger only after successful migration from old to new UI
* May need to track migration completion status to show this item appropriately → automatically complete the item when all other items are completed

**Acceptance Criteria:**
* Action item appears in "Things to do next" section after migration completion
* Item is positioned as the top priority task
* Text matches the specified wording exactly
* Item is only shown for users who have just completed migration (not for all users)

## PR Description

This PR Adds "Things to do next" item after settings migration as described above.
It saves the `'woocommerce_ppcp-settings-migration-is-done'` option in DB when settings migration is completed and then uses it to the new todo item

### DIFF
--- a/modules/ppcp-settings/src/Ajax/SwitchSettingsUiEndpoint.php
+++ b/modules/ppcp-settings/src/Ajax/SwitchSettingsUiEndpoint.php
@@ -24,6 +24,7 @@ class SwitchSettingsUiEndpoint {
 
 	public const ENDPOINT                      = 'ppcp-settings-switch-ui';
 	public const OPTION_NAME_SHOULD_USE_OLD_UI = 'woocommerce_ppcp-settings-should-use-old-ui';
+	public const OPTION_NAME_MIGRATION_IS_DONE = 'woocommerce_ppcp-settings-migration-is-done';
 
 	protected RequestData $request_data;
 	protected LoggerInterface $logger;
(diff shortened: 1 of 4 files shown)


modules/ppcp-settings/src/Ajax/SwitchSettingsUiEndpoint.php
```diff
@@ -70,6 +71,8 @@ public function handle_request(): void {
 			$this->onboarding_profile->save();
 
 			$this->settings_data_migration->migrate();
+
+			update_option( self::OPTION_NAME_MIGRATION_IS_DONE, 'yes' );
```
hmouhtar: Any reason for using `yes` instead of `true`? 🤔

hmouhtar (approved):
LGTM, thanks for the PR. Just left a small question

Dinamiko (approved):

PR #3540, opened 2025-07-18
Remove data-sdk-client-token field when not needed (5011)

Dinamiko:
`data-sdk-client-token` was added to all JS SDK namespaced scripts in all pages, this PR only adds it when Fastlane script is required.

This PR introduces the following changes:
- Removes `data-sdk-client-token` from global localized configuration
- Adds `data-sdk-client-token` only  when Fastlane script is required
- Removes `data-client-metadata-id` field as per [Fastlane documentation](https://developer.paypal.com/studio/checkout/fastlane/integrate#initialize-sdk) is not needed


modules/ppcp-axo/src/Endpoint/AxoScriptAttributes.php
```diff
@@ -0,0 +1,69 @@
...
+use WooCommerce\PayPalCommerce\Button\Endpoint\RequestData;
+use WooCommerce\PayPalCommerce\Button\Helper\ContextTrait;
+
+/**
+ * Handles the request for the PayPal Axo script attributes.
+ */
+class AxoScriptAttributes implements EndpointInterface {
+	use ContextTrait;
+
+	const ENDPOINT = 'ppc-axo-script-attributes';
```
Narek13, line 18: Please add the visibility

modules/ppcp-axo/src/Service/AxoApplies.php
```diff
@@ -1,21 +1,24 @@
...
 
 use WooCommerce\PayPalCommerce\ApiClient\Helper\CurrencyGetter;
+use WooCommerce\PayPalCommerce\WcGateway\Helper\CardPaymentsConfiguration;
+use WooCommerce\PayPalCommerce\WcGateway\Helper\CartCheckoutDetector;
+use WooCommerce\PayPalCommerce\WcSubscriptions\Helper\SubscriptionHelper;
 
 /**
- * Class ApmApplies
+ * Class AxoApplies
  */
```
Narek13: We can remove redundant comments like this when touching the part of the code

modules/ppcp-axo/src/Service/AxoApplies.php
```diff
@@ -79,4 +89,28 @@ public function for_merchant() : bool {
...
+			   && ! $this->subscription_helper->cart_contains_subscription();
+	}
+
+	/**
+	 * Condition to evaluate if the current endpoint is excluded.
+	 *
+	 * @return bool
+	 */
+	private function is_excluded_endpoint(): bool {
+		// Exclude the Order Pay and Order Received endpoints.
```
Narek13: I think no need for this inline comment, the code perfectly describes itself

Narek13 (approved):
Thanks for working on this. Left a couple of minor suggestion which are not preventing from approval.

AlexP11223 (approved):

Narek13 (approved):
Thanks for making all the changes

PR #3543, opened 2025-07-21
Remove redundant authorize call in subscription renewal handler (1575)

hmouhtar:
The vaulted subscription renewal handler was attempting to authorize orders when the intent was set to "AUTHORIZE". This resulted in `ORDER_ALREADY_AUTHORIZED` errors because PayPal orders created with tokens as a payment source are authorized automatically, with the authorization returned in the order creation response.

This PR removes the redundant authorize call while maintaining the `AuthorizedPaymentsProcessor::CAPTURED_META_KEY` meta update call, which is necessary for the manual renewal action to be displayed.

This change will only affect vaulted subscription renewals. Since these renewals always use tokens as the payment source, no explicit authorization call is required.


Narek13 (approved):
LGTM

PR #3549, opened 2025-07-23
Subscriptions migrated from PPEC to PPC fails on renewal

Dinamiko:
When subscriptions originally purchased using the legacy PayPal Checkout gateway (PPEC) are migrated to PayPal Payments (PPC) fails on renewal.

### Steps To Reproduce
- Disable the PayPal Payments plugin (PPC).
- Enable WooCommerce PayPal Checkout Gateway (PPEC).
- Purchase a subscription product using PPEC.
- Disable PPEC and re-enable PPC.
- Confirm the subscription now shows paypal legacy as the gateway (expected behavior).
- Note the subscription post ID.
- In Tools → Scheduled Actions, locate the renewal action for that subscription and run action.

### Possible Cause
[should_mock_ppec_gateway](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php#L79) is executed before renewal action is triggered and therefore woocommerce_scheduled_subscription_payment [doing action](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php#L144) is always false. 

Because should_mock_ppec_gateway is false mocked PPEC gateway is not added and renewal triggered by action scheduler fails.

### Suggested Solution
Remove should_mock_ppec_gateway condition so PPCE mock gateway is added no matter the context when the site has existing PPEC subscriptions and PPEC plugin is not active.  

By doing so  woocommerce_scheduled_subscription_payment action check works as expected as PPCE mock gateway is available.

### DIFF
--- a/modules/ppcp-compat/src/CompatModule.php
+++ b/modules/ppcp-compat/src/CompatModule.php
@@ -94,28 +94,6 @@ class CompatModule implements ServiceModule, ExtendingModule, ExecutableModule {
 
 		$this->legacy_ui_card_payment_mapping( $c );
 
-		// Short-circuit RenewalHandler::get_token_for_customer() to use a Billing Agreement ID for PPEC orders.
-		add_filter('woocommerce_paypal_payments_subscriptions_get_token_for_customer', function ($token, $customer, $order ) {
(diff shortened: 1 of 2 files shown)


AlexP11223 (approved):

Narek13 (approved):
LGTM

PR #3550, opened 2025-07-23
When Apple Pay is enabled empty space for Apple Pay button is showing on all pages  (3957)

Dinamiko:
### Steps to reproduce
- Enable Apple Pay
- Enable Express Checkout and Mini Cart in Styling button location
- Navigate to block Checkout with non-Apple device
→ observe empty Apple Pay button div:

f2353af9-94dd-49c6-8547-322f07eb5b23

This PR adds a condition into `registerExpressPaymentMethod` `canMakePayment`  to display button only if Apple session and can make payments are available.


Narek13 (approved):
Thanks for fixing this.

hmouhtar (approved):

PR #3551, opened 2025-07-23
Fix undefined properties error in checkout block (5027)

hmouhtar:
The PayPal address may be empty in some cases, such as with virtual products. There were no undefined property checks in the address helpers, which was causing errors in the checkout block.

This PR fixes that by adding optional chaining to the address properties.


Narek13 (approved):
Thanks

PR #3553, opened 2025-07-24
New UI - On Goole Pay popup no shipping methods visible when Pay Now is enabled (4976)

Dinamiko:
Shipping callback currently does not work for Google Pay in new settings UI.

### Steps to reproduce
- Define for example free shipping and flat rate methods
- Navigate to shop, click on product, click on Google Pay
  - Observe shipping methods are not showing in Google Pay modal

### Expected behavior
When the “Pay Now Experience” setting is enabled, then both PayPal and Google Pay should have the shipping callback functionality automatically enabled.

### Possible cause
It seems the problem of non displaying shipping methods in Google Pay modal is caused by `this.buttonConfig.shipping?.enabled` returning false in [requiresShipping](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/resources/js/GooglepayButton.js#L220).

`this.buttonConfig.shipping?.enabled` get its value from setting [googlepay_button_shipping_enabled](https://github.com/woocommerce/woocommerce-paypal-payments/blob/trunk/modules/ppcp-googlepay/extensions.php#L222) which defaults to `no`.

In new setting UI we do not have this setting in the UI and therefore it’s not possible to change its value.

### Suggested solution
`requiresShipping` should get the value from "Pay Now Experience" setting for new settings UI while still using the `googlepay_button_shipping_enabled` value for legacy UI.

### DIFF
--- a/modules/ppcp-blocks/resources/js/paypal-config.js
+++ b/modules/ppcp-blocks/resources/js/paypal-config.js
@@ -99,7 +99,7 @@ export const handleApprove = async (
 			order = json.data;
 		}
 
-		const shippingAddress = order.purchase_units?.[ 0 ]?.shipping?.address;
+		const shippingAddress = order?.purchase_units?.[ 0 ]?.shipping?.address;
 		if ( shippingAddress ) {
 			const addresses = paypalOrderToWcAddresses( order );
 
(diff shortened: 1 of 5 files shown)


Narek13 (approved):
LGTM

PR #3556, opened 2025-07-25
Remove the specified gateway when the button is disabled (4520)

Narek13:
## Issue Description
 
The separate Apple Pay gateway appears in some cases on the Classic Checkout when it’s not supposed to. The button is usually not displayed in these cases.

**Steps To Reproduce**

- Onboard with merchant eligible for acdc
- Navigate to shop
- Add product to cart

## PR Description
 
A Filter is added to remove the available payment gateways (e.g., Google Pay or Apple Pay) when the button is disabled for the current location (e.g., classic checkout) in the styling settings.
This is necessary because WooCommerce automatically includes the gateway when it is enabled, even if the button is hidden via settings.

### DIFF
--- a/modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
+++ b/modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
@@ -319,7 +319,7 @@ class StylingSettingsMapHelper {
 			 *
 			 * @psalm-suppress MissingClosureParamType
 			 */
-			static function ( $methods ) use ( $button_name, $current_context ) {
+			static function ( $methods ) use ( $button_name, $current_context ): array {
 				if ( $current_context !== 'classic_checkout' ) {
 					return $methods;
 				}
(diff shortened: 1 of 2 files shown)


modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
```diff
@@ -305,26 +305,28 @@ protected function mapped_button_enabled_value( array $styling_models, string $b
...
+			'woocommerce_available_payment_gateways',
 			/**
-			 * Outputs an inline CSS style that hides the Google Pay gateway (on Classic Checkout)
-			 * In case if the button is disabled from the styling settings but the gateway itself is enabled.
+			 * Param types removed to avoid third-party issues.
 			 *
-			 * @return void
+			 * @psalm-suppress MissingClosureParamType
 			 */
-			add_action(
```
hmouhtar, line 315: Hi @Narek13, just double checking, was this action removed because it is no longer necessary?

hmouhtar (approved):

modules/ppcp-compat/src/Settings/StylingSettingsMapHelper.php
Narek13: Yes. Instead of hiding it with CSS, now we are unseting the method

PR #3557, opened 2025-07-25
Contact module -does not work for virtual products in block pages and classic checkout

Dinamiko:
When buying a virtual product (which uses NO_SHIPPING), the contact module information is displayed in the PayPal popup window but not saved in the WooCommerce order.

### Steps To Reproduce
- Connect US merchant (contact module should be enabled)
- Create virtual product

Try payment from block pages or from classic checkout selecting different email and phone then default one
Observe on order received page email and phone in billing data are different then selected

### Expected behaviour
- Contact information from PayPal popup is saved in the order when it contains only virtual products.

### Possible cause
It fails because `ShippingFactory::from_paypal_response` has checks for shipping fields (e.g. `name`) and the shipping data from PayPal only contains `phone_number` and `email_address`.

There is also a related issue in Cart which prevents order to proceed because it's checking shipping address but it does not exist.

### Solution
This PR makes shipping address fields optional and allows PayPal contact information (email/phone) to be captured even when no shipping address is provided. It ensures that contact details from the PayPal popup are properly saved to WooCommerce orders for virtual products in both block and classic checkout flows.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/ShippingFactory.php
+++ b/modules/ppcp-api-client/src/Factory/ShippingFactory.php
@@ -96,30 +96,30 @@ class ShippingFactory {
 	 * @throws RuntimeException When JSON object is malformed.
 	 */
 	public function from_paypal_response( \stdClass $data ): Shipping {
-		if ( ! isset( $data->name->full_name ) ) {
-			throw new RuntimeException( 'No name was given for shipping.' );
+		$name = $data->name->full_name ?? null;
+
+		$address = null;
+		if ( isset( $data->address ) ) {
+			$address = $this->address_factory->from_paypal_response( $data->address );
 		}
-		if ( ! isset( $data->address ) ) {
-			throw new RuntimeException( 'No address was given for shipping.' );
-		}
-		$contact_phone = null;
+
 		$contact_email = null;
-		$address       = $this->address_factory->from_paypal_response( $data->address );
+		if ( isset( $data->email_address ) ) {
+			$contact_email = $data->email_address;
+		}
+
+		$contact_phone = null;
+		if ( isset( $data->phone_number->national_number ) ) {
+			$contact_phone = new Phone( $data->phone_number->national_number );
+		}
 
 		$options = array_map(
 			array( $this->shipping_option_factory, 'from_paypal_response' ),
 			$data->options ?? array()
 		);
 
-		if ( isset( $data->phone_number->national_number ) ) {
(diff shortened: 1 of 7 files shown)


modules/ppcp-api-client/src/Factory/ShippingFactory.php
```diff
@@ -96,30 +96,30 @@ public function from_wc_order( \WC_Order $order ): Shipping {
...
+		if ( isset( $data->address ) ) {
+			$address = $this->address_factory->from_paypal_response( $data->address );
 		}
-		if ( ! isset( $data->address ) ) {
-			throw new RuntimeException( 'No address was given for shipping.' );
+
+		$contact_email = null;
+		if ( isset( $data->email_address ) ) {
+			$contact_email = $data->email_address;
 		}
```
Narek13: This can be simplified

```suggestion
		$contact_email = $data->email_address ?? null;
```

modules/ppcp-api-client/src/Factory/ShippingFactory.php
```diff
@@ -96,30 +96,30 @@ public function from_wc_order( \WC_Order $order ): Shipping {
...
 	 */
 	public function from_paypal_response( \stdClass $data ): Shipping {
-		if ( ! isset( $data->name->full_name ) ) {
-			throw new RuntimeException( 'No name was given for shipping.' );
+		$name = $data->name->full_name ?? null;
+
+		$address = null;
+		if ( isset( $data->address ) ) {
+			$address = $this->address_factory->from_paypal_response( $data->address );
 		}
```
Narek13:
```suggestion
		$address = isset( $data->address )
    		? $this->address_factory->from_paypal_response( $data->address )
    		: null;
```

modules/ppcp-api-client/src/Factory/ShippingFactory.php
```diff
@@ -96,30 +96,30 @@ public function from_wc_order( \WC_Order $order ): Shipping {
...
+		if ( isset( $data->email_address ) ) {
+			$contact_email = $data->email_address;
 		}
+
 		$contact_phone = null;
-		$contact_email = null;
-		$address       = $this->address_factory->from_paypal_response( $data->address );
+		if ( isset( $data->phone_number->national_number ) ) {
+			$contact_phone = new Phone( $data->phone_number->national_number );
+		}
```
Narek13: same here

Narek13 (approved):
Thanks for working on this. Made couple of suggestions which are not preventing from approval

PR #3562, opened 2025-07-29
Clean AppSwitch hash params on Pay for Order page (5031)

hmouhtar:
Unlike other contexts, the Pay for Order page does not process the payment over AJAX, it uses a regular form submission.

When there is an error during payment processing on this page, there is no way to notify that to the frontend before or after the page is reloaded. Since the AppSwitch hash params are still present in the URL upon reload, the AppSwitch flow is triggered again, causing an infinite reload loop.

To fix this, this PR adds code to preemptively remove the hash params from the URL on the onApprove callback, before the payment is processed.


Narek13 (approved):
Haven't tested but code wise looks good to me. Thanks

AlexP11223 (approved):

PR #3569, opened 2025-07-31
Cross browser AppSwitch

AlexP11223:
Added handling of cross-browser AppSwitch (started in non-default browser, then redirected to the default one).

We save the current cart + PayPal order ID into transient before going to PayPal and add the transient key into the return URL, then after returning we retrieve the saved cart and PayPal order, create a WC order and redirect to Pay for order.

The WC order creation is done using the existing `WooCommerceOrderCreator`, refactored to use `CartData` instead of the current WC cart. 
I had to change the `woocommerce_paypal_payments_shipping_callback_woocommerce_order_created` hook parameters. I hope it was used only internally :thinking: And renamed it to `woocommerce_paypal_payments_woocommerce_order_created_from_cart`, so that at least nothing crashes, and because it's not just the shipping callback.

Remaining issues:
- Cross-browser AppSwitch does not work when starting from block checkout because WC redirects to cart if the cart is empty on checkout losing all parameters.
- AppSwitch resume does not work in block cart if PayPal is not the default gateway.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js
+++ b/modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js
@@ -40,6 +40,8 @@ const onApprove = ( context, errorHandler, spinner ) => {
 					throw new Error( data.data.message );
 				}
 
+				// in some cases a different method may get selected,
+				// such as when returning from AppSwitch in a different browser and PayPal is not default
 				if ( ! getCurrentPaymentMethod().startsWith( 'ppcp-' ) ) {
 					jQuery(
 						`input[name="payment_method"][value="${ PaymentMethods.PAYPAL }"]`
--- a/modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js
+++ b/modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js
@@ -1,3 +1,8 @@
+import {
+	getCurrentPaymentMethod,
+	PaymentMethods,
+} from '../Helper/CheckoutMethodState';
+
 const onApprove = ( context, errorHandler, spinner ) => {
 	return ( data, actions ) => {
 		spinner.block();
@@ -34,6 +39,13 @@ const onApprove = ( context, errorHandler, spinner ) => {
 					}
 					throw new Error( data.data.message );
 				}
+
+				if ( ! getCurrentPaymentMethod().startsWith( 'ppcp-' ) ) {
+					jQuery(
+						`input[name="payment_method"][value="${ PaymentMethods.PAYPAL }"]`
+					).prop( 'checked', true );
+				}
+
 				document.querySelector( '#place_order' ).click();
 			} );
 	};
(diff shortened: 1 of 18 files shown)


modules/ppcp-button/resources/js/modules/OnApproveHandler/onApproveForPayNow.js
```diff
@@ -34,6 +39,13 @@ const onApprove = ( context, errorHandler, spinner ) => {
 					}
 					throw new Error( data.data.message );
 				}
+
```
hmouhtar, line 42: A short comment explaining why this is necessary would be helpful

modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
```diff
@@ -475,33 +489,52 @@ private function create_paypal_order( ?WC_Order $wc_order = null, string $paymen
 			$payment_source_key
 		);
 
+		$custom_args = array();
+
+		$cart_data = null;
+		// Save the cart to be able to access it after cross-browser AppSwitch.
+		if ( 'pay-now' !== $data['context'] ) {
+			try {
+				$cart_data = $this->cart_data_factory->from_current_cart();
```
hmouhtar, line 498: I would suggest having the key generated automatically when the a a new CartData instance is created, or in any case, moving that logic to the factory. I would also suggest removing the `set_key` method from the public interface if there is no scenario where we would manually set a key.

modules/ppcp-button/src/Session/CartDataTransientStorage.php
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+
+	public function get( string $key ): ?CartData {
+		$data = get_transient( $key );
+		if ( ! is_array( $data ) ) {
+			return null;
+		}
+
+		$cart_data = new CartData(
```
hmouhtar: Following on my recommendation about automatically creating a key for CartData, I would recommend having a `CartData::fromArray` static method that just receives the entire transient array and returns the instance. This would remove the need for having the public `set_key` method and is easier to maintain.

hmouhtar (commented):
Hi @AlexP11223, thanks for working on this!

I added some comments, and I have a couple of questions:

1. What happens if the EarlyOrderHandler is enabled? Wouldn't we end up with 2 orders?
2. If the user is authenticated, will the order be associated with that user upon returning from AppSwitch on a different browser? `configure_customer` just sets the current user ID, which will be 0, so it seems that there won't be an association

AlexP11223:
I think EarlyOrderHandler is not creating anything? It only adds a hook handler to set some meta.

For authenticated users as I understand currently AppSwitch will not trigger at all. But should be possible to save the customer id too. However then WC will ask to log in or enter the email to access the order pay page.

modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
AlexP11223: The idea was to have the key only when it is actually saved somewhere or is about to be saved. So `null` indicates that it is not saved anywhere, which can happen in the original `WooCommerceOrderCreator` use case.

Narek13 (approved):
Thanks for working on this Alex. LGTM

hmouhtar:
> I think EarlyOrderHandler is not creating anything? It only adds a hook handler to set some meta.
> 
> For authenticated users as I understand currently AppSwitch will not trigger at all. But should be possible to save the customer id too. However then WC will ask to log in or enter the email to access the order pay page.

@AlexP11223 AppSwitch does trigger for authenticated users. AFAIK PayPal expects the order to be associated with the customer account in the cross browser flow

hmouhtar (approved):

PR #3577, opened 2025-08-01
Add polling mechanism for renderer wrapper (4990)

hmouhtar:
There was a race condition with the PayPal messages renderer, which attempted to render onto a wrapper container that sometimes wasn't available in the DOM.

This PR solves the issue by adding a new async helper function, `waitForElement`, which uses setTimeout polling to return the wrapper instance within a specified timeframe.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/Helper/WaitForElement.js
+++ b/modules/ppcp-button/resources/js/modules/Helper/WaitForElement.js
@@ -1,0 +1,32 @@
+/**
+ * Waits for a DOM element using setTimeout polling
+ * @param {string} selector - CSS selector for the element
+ * @param {number} timeout  - Maximum time to wait in milliseconds (default: 3000)
+ * @param {number} interval - Polling interval in milliseconds (default: 100)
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/resources/js/modules/Helper/WaitForElement.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Waits for a DOM element using setTimeout polling
+ * @param {string} selector - CSS selector for the element
+ * @param {number} timeout  - Maximum time to wait in milliseconds (default: 3000)
+ * @param {number} interval - Polling interval in milliseconds (default: 100)
+ * @return {Promise<Element>} - Resolves with the element or rejects if timeout
+ */
+export function waitForElement( selector, timeout = 3000, interval = 100 ) {
```
Narek13, line 8: Feels more like an util to me rather than a helper, but not a strong opinion here.

Narek13 (approved):
Thanks for working on this. Looks good to me.

PR #3583, opened 2025-08-05
Prevent early is_enabled check for Apple Pay button in new UI (4711)

hmouhtar:
This PR prevents early `is_enabled()` checks for the Apple Pay button in the new UI.

In the new UI, the Apple Pay button availability is determined by several factors like the current context, not a global setting. The `is_enabled()` check depends on context functions (like `is_cart()`), which return false too early during block registration.

This issue and solution are the same as described in this PR: https://github.com/woocommerce/woocommerce-paypal-payments/pull/3566


Narek13 (approved):

PR #3598, opened 2025-08-16
Add Fastlane session restoration post payment failure (5032)

danieldudzic:
### Description

This PR will add Fastlane session restoration after payment failures for both Block and Classic checkouts.

It ensures customers have their Fastlane data pre-loaded after returning to the checkout page in the case of a payment failure.

### Screenshots

ScreenCast2025-08-16at1 07 03PM-ezgif com-resize

### Steps to test
1. In PayPal Payments > Settings > Expert Settings > 3D Secure select `Always require 3D Secure`
2. In Checkout authenticate with any Fastlane profile, add a new card - using one of the Failure [test case card details](https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/test/)
3. Confirm that after the failed payment you return to the relevant checkout page with a failure notice on top and after a few seconds the Fastlane profile data automatically loads again.


Narek13 (approved):
Haven't tested but code wise LGTM

PR #3599, opened 2025-08-18
Phase 2: Settings Migration - Add Woo Inbox item (5166)

Narek13:
## Issue Description

To inform users about the upcoming settings migration, we already display a notice in our settings page, and additionally want to add a WC Home Inbox item

## PR Description

This PR:

- Adds a logic to dynamically and separately enable/disable the inbox notes
- If the condition is not met, then the previously added note will be deleted
- Adds new Woo Inbox note, which also allows to migrate settings.

Screenshot 2025-08-18 at 16 15 12

### DIFF
--- a/modules/ppcp-compat/src/CompatModule.php
+++ b/modules/ppcp-compat/src/CompatModule.php
@@ -13,6 +13,9 @@ use Exception;
 use WC_Cart;
 use WC_Order;
 use WC_Order_Item_Product;
+use WooCommerce\PayPalCommerce\Button\Helper\MessagesApply;
+use WooCommerce\PayPalCommerce\Settings\Data\SettingsModel;
+use WooCommerce\PayPalCommerce\Settings\SettingsModule;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExecutableModule;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExtendingModule;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ModuleClassNameIdTrait;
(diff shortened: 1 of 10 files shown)


modules/ppcp-settings/resources/js/switchSettingsUi.js
```diff
@@ -36,20 +30,22 @@ document.addEventListener( 'DOMContentLoaded', () => {
...
 	document.addEventListener( 'click', ( event ) => {
-		const linkElement = event.target.closest( 'a.settings-switch-ui' );
-		if ( linkElement ) {
+		if (
+			event.target.closest( '.button.button-settings-switch-ui' ) ||
+			event.target.closest( '.ppcp-notice-wrapper:not(.inline) a.settings-switch-ui' ) ||
+			event.target.closest( 'a[name="settings-switch-ui"]' ) ||
+			event.target
+				.closest( '.woocommerce-inbox-note__action-button' )
+				.textContent.includes( 'Switch to New Settings' )
```
Dinamiko: I suggest to check if element exist before calling textContent:
```js
event.target
    .closest('.woocommerce-inbox-note__action-button')
    ?.textContent.includes('Switch to New Settings')
```

modules/ppcp-compat/src/CompatModule.php
```diff
@@ -91,6 +94,64 @@ function() use ( $c ) {
...
+					: $settings_model->get_stay_updated();
+
+				if ( ! $messages_apply->for_country() || ! $stay_updated ) {
+					return;
+				}
+
+				$selected_locations = $settings->has( 'pay_later_messaging_locations' ) ? $settings->get( 'pay_later_messaging_locations' ) : array();
+
+				$settings->set( 'pay_later_messaging_enabled', true );
+				$settings->set( 'pay_later_messaging_locations', array_merge( $selected_locations, array( 'product', 'cart', 'checkout' ) ) );
```
Dinamiko: If `$selected_locations` already contains 'product', 'cart', or 'checkout', they'll be duplicated, I suggest to add `array_unique` or `array_diff`:
```php
array_merge(
    $selected_locations, 
    array_diff(array('product', 'cart', 'checkout'), $selected_locations)
)
```

Dinamiko (changes requested):

modules/ppcp-compat/src/CompatModule.php
Narek13: Thanks, nice catch :)

Dinamiko (approved):

PR #3612, opened 2025-08-25
Fix continuation mode in Elementor checkout (3389)

hmouhtar:
The compat module had logic that hardcoded the context when Elementor checkout was detected. This logic was stale and no longer relevant following updates to the way we manage continuation mode in the global settings object, and was causing issues on the Elementor checkout page because the context was not the expected one.

To fix this, this PR removes the `set_elementor_checkout_context` function, making the context the expected one once again ('mini-cart' instead of 'checkout').


Narek13 (approved):
Thanks. Haven't tested but LGTM

PR #3613, opened 2025-08-26
Keep order customer association after cross-browser AppSwitch for logged in users

AlexP11223:
Now preserving the customer id in the order if the user was logged in when starting the checkout. That is we started checkout in a non-default browser logged in as `user1`, and finished in the default browser as a guest, the order will belong to `user1`. 

The log-in prompt after the redirect to the default browser as a guest is disabled.

There is one a bit weird edge, when logged in as different users in both browsers. For now just using the user from the default browser to avoid confusion like going to My Account after the payment and not seeing the order. Maybe a better approach could be to force logout in the default browser. Anyway, hopefully this happens very rarely.

Also I added the `woocommerce_order_email_verification_required` filter handler to skip email prompt for such orders. It should not be needed in most cases because it is already skipped if the order was created less than 10 minutes ago, but probably a good idea to add that, e.g. if AppSwitch failed and finishing a bit later via another method.

### DIFF
--- a/modules/ppcp-button/src/ButtonModule.php
+++ b/modules/ppcp-button/src/ButtonModule.php
@@ -375,7 +375,7 @@ class ButtonModule implements ServiceModule, ExtendingModule, ExecutableModule {
 		 */
 		add_filter(
 			'woocommerce_order_received_verify_known_shoppers',
-			static function( $result ) {
+			static function ( $result ) {
 				if ( ! is_order_received_page() ) {
 					return $result;
 				}
@@ -417,8 +417,8 @@ class ButtonModule implements ServiceModule, ExtendingModule, ExecutableModule {
 		add_filter(
 			'woocommerce_order_email_verification_required',
 			static function (
-			$email_verification_required,
-			$order
+				$email_verification_required,
+				$order
 			) {
 				if ( ! $order instanceof WC_Order ) {
 					return $email_verification_required;
--- a/modules/ppcp-button/src/ButtonModule.php
+++ b/modules/ppcp-button/src/ButtonModule.php
@@ -237,6 +237,10 @@ class ButtonModule implements ServiceModule, ExtendingModule, ExecutableModule {
 		);
 	}
 
+	private static function is_cross_browser_order( WC_Order $wc_order ): bool {
+		return wc_string_to_bool( $wc_order->get_meta( PayPalGateway::CROSS_BROWSER_APPSWITCH_META_KEY ) );
+	}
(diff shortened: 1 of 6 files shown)


modules/ppcp-button/src/ButtonModule.php
```diff
@@ -299,10 +305,133 @@ static function () use ( $container ) {
...
+
+				if ( ! self::is_cross_browser_order( $order ) ) {
+					return $email_verification_required;
+				}
+
+				return false;
+			},
+			10,
+			2
+		);
```
Narek13, line 435: Does it make sense to add return types for all 3 hook callback functions?

Narek13 (approved):
Thanks for working on this. Left a small suggestion which doesn't prevent me from approving.

modules/ppcp-button/src/ButtonModule.php
AlexP11223: We return the first parameter which is not guaranteed to have the correct type if other handlers mess it up.

modules/ppcp-button/src/ButtonModule.php
```diff
@@ -299,10 +305,133 @@ static function () use ( $container ) {
...
+		 * Should not affect anything in most cases because it is skipped for just created orders (< 10 min).
+		 *
+		 * @param bool $email_verification_required
+		 * @param WC_Order $order
+		 *
+		 * @returns bool
+		 *
+		 * @psalm-suppress MissingClosureParamType
+		 */
+		add_filter(
```
hmouhtar, line 417: AppSwitch flows are short-lived. If the email prompt is already disabled for orders that were created within the last 10 minutes, I think that's enough, and I don't see much value in extending the support timeframe compared to the side effects that removing this prompt might have, since essentially all orders would be marked with `CROSS_BROWSER_APPSWITCH_META_KEY`. I'm not even sure AppSwitch will work 10 minutes after it was triggered, so I would recommend removing this. 

Not a blocker anyway, thanks for working on this @AlexP11223

hmouhtar (approved):

modules/ppcp-button/src/ButtonModule.php
AlexP11223: I encountered it some time ago, though not sure how. It can be useful not for AppSwitch itself but when it fails and the customer retries using different methods.

>  essentially all orders would be marked

all orders? :thinking: Only cross-browser are supposed to be marked, which should be quite rare.

PR #3616, opened 2025-08-27
Fix shipping preference

AlexP11223:
The Pay for order page is not supposed to allow changing the shipping address/method, so now sending `SHIPPING_PREFERENCE_SET_PROVIDED_ADDRESS`.

Also for some reason it was possible to end up with `SHIPPING_PREFERENCE_GET_FROM_FILE` while `needs_shipping` is `false`. I think this is not correct, unless I forgot something. So now sending `SHIPPING_PREFERENCE_NO_SHIPPING` in this case.


Narek13 (approved):
LGTM

Dinamiko (approved):

PR #3624, opened 2025-08-29
Add ddev command to reset wp install

Dinamiko:
This PR adds `ddev wp-reset` command to reset the WordPress installation, by default it does not delete all plugins but can do so by passing `--clean-plugins` argument to the command.

### DIFF
--- a/.ddev/commands/web/wp-reset
+++ b/.ddev/commands/web/wp-reset
@@ -25,7 +25,7 @@ reset-database() {
 
 	echo "Installing WordPress..."
 	wp core install \
-		--url=wp-test-site.ddev.site \
+		--url="${DDEV_PRIMARY_URL}" \
 		--title="${WP_TITLE}" \
 		--admin_user="${ADMIN_USER}" \
 		--admin_password="${ADMIN_PASS}" \
--- a/.ddev/commands/web/wp-reset
+++ b/.ddev/commands/web/wp-reset
@@ -10,7 +10,7 @@ show-help() {
 remove-plugins() {
 	echo "Deactivating all plugins..."
 	wp plugin deactivate --all --path=.ddev/wordpress
-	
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
LGTM. Thanks for working on this.

.ddev/commands/web/wp-reset
```diff
@@ -0,0 +1,62 @@
...
+
+	if [ "$1" = "clean-plugins" ]; then
+		remove-plugins
+	fi
+
+	echo "Installing WordPress..."
+	wp core install \
+		--url=wp-test-site.ddev.site \
+		--title="WooCommerce PayPal Payments" \
+		--admin_user=admin \
```
AlexP11223: maybe should use the same config as in https://github.com/woocommerce/woocommerce-paypal-payments/blob/11f33651b48e538c9ee21ae95b22e460a54bd076/.ddev/commands/web/orchestrate.d/10_wp_install.sh#L27

AlexP11223 (commented):
So it's like `orchestrate -f` but keeping things like plugins?

Dinamiko:
@AlexP11223 thanks for the feedback, I've updated it to use config constants instead of hardcoded values.

.ddev/commands/web/wp-reset
```diff
@@ -0,0 +1,62 @@
...
+	echo "Resetting database..."
+	wp db reset --yes --path=.ddev/wordpress
+
+	if [ "$1" = "clean-plugins" ]; then
+		remove-plugins
+	fi
+
+	echo "Installing WordPress..."
+	wp core install \
+		--url=wp-test-site.ddev.site \
```
AlexP11223: This probably should be `${DDEV_PRIMARY_URL}` too.

PR #3628, opened 2025-09-02
Fix sync checkout fields and headers in Fastlane flow

strangerkir:
In the Fastlane flow, we had shipping and billing headers displayed while shipping and billing checkout fields were hidden.
It happened because before showing or hiding headers we checked not just whether fields are displayed, but also whether they are active. But whether they are active or not is unrelated; we only care about fields' visibility when hiding/showing headers.

Before:

https://github.com/user-attachments/assets/6503c3ea-caba-47d0-be11-1c5a417c267e

After:

https://github.com/user-attachments/assets/eff471f6-88b2-4c7f-8618-d363282d8337


Dinamiko (approved):

Narek13 (approved):
LGTM. Thanks

PR #3637, opened 2025-09-04
Refactor Fastlane email submission state management in Classic Checkout (5095)

danieldudzic:
_This PR is dependent on https://github.com/woocommerce/woocommerce-paypal-payments/pull/3598_

### Description

This PR will update the email submission state management with a centralized `ButtonStateManager`.

### Steps to test

1. In Classic Checkout, enter a Fastlane profile email address and verify that the OTP dialog appears
2. Try to complete the order, and make sure the transaction fails
3. Return to the checkout page
4. Confirm that the OTP gets triggered for the session pre-filled email address
5. Test against any regressions for other Fastlane transactions in the Classic Checkout

### DIFF
--- a/modules/ppcp-axo-block/resources/js/hooks/useSessionRestoration.js
+++ b/modules/ppcp-axo-block/resources/js/hooks/useSessionRestoration.js
@@ -1,0 +1,87 @@
+import { useEffect, useRef } from '@wordpress/element';
+import { useDispatch } from '@wordpress/data';
+import { setIsEmailLookupCompleted, STORE_NAME } from '../stores/axoStore';
+import { log } from '../../../../ppcp-axo/resources/js/Helper/Debug';
+
+/**
+ * Hook to restore Fastlane session after payment failures using triggerAuthenticationFlow
+ * Only runs when ppcp_fastlane_error=1 URL parameter is present
+ * @param {Object} fastlaneSdk - The Fastlane SDK instance
+ */
+const useSessionRestoration = ( fastlaneSdk ) => {
+	const { setShippingAddress, setCardDetails, setIsGuest } =
+		useDispatch( STORE_NAME );
+	const hasProcessed = useRef( false );
+
+	useEffect( () => {
+		if ( ! fastlaneSdk || hasProcessed.current ) {
+			return;
+		}
+
+		const urlParams = new URLSearchParams( window.location.search );
+		const hasErrorParam = urlParams.get( 'ppcp_fastlane_error' ) === '1';
+
+		if ( ! hasErrorParam ) {
+			return;
+		}
+
+		// Remove the error parameter from URL
+		urlParams.delete( 'ppcp_fastlane_error' );
+		const newUrl = new URL( window.location );
+		newUrl.search = urlParams.toString();
+		window.history.replaceState( {}, '', newUrl );
+
+		hasProcessed.current = true;
+
+		const restoreSession = async () => {
+			try {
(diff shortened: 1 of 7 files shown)


modules/ppcp-axo-block/resources/js/hooks/useSessionRestoration.js
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useRef } from '@wordpress/element';
+import { useDispatch } from '@wordpress/data';
+import { setIsEmailLookupCompleted, STORE_NAME } from '../stores/axoStore';
+import { log } from '../../../../ppcp-axo/resources/js/Helper/Debug';
```
puntope, line 4: 💅  Not related directly to this PR, but WDYT about using Aliases for the imports?

So instead of 

`import { log } from '../../../../ppcp-axo/resources/js/Helper/Debug';`

It can be something like

`import { log } from '@modules/ppcp-axo/resources/js/Helper/Debug';`

modules/ppcp-axo-block/resources/js/hooks/useSessionRestoration.js
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		const urlParams = new URLSearchParams( window.location.search );
+		const hasErrorParam = urlParams.get( 'ppcp_fastlane_error' ) === '1';
+
+		if ( ! hasErrorParam ) {
+			return;
+		}
+
+		// Remove the error parameter from URL
```
puntope, line 28: 💅 Redundant comment

modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
```diff
@@ -170,10 +170,19 @@ function( $allowed_hosts ) : array {
...
 		exit();
 	}
 
+	/**
+	 * Get checkout URL with Fastlane error parameter.
+	 *
+	 * @return string
+	 */
+	private function get_checkout_url_with_error(): string {
+		return add_query_arg( 'ppcp_fastlane_error', '1', wc_get_checkout_url() );
```
Dinamiko: Is `ppcp_fastlane_error` URL argument added in all contexts or only in the context of Fastlane? if added to all contexts I suggest not add it if it is not needed.

Dinamiko: Or even better, to not couple `ReturnUrlEndpoint.php` to Fastlane maybe would be better to apply a filter here that allows adding error URL arguments from the outside, then in the Fastlane module we can use the filter to add this specific URL argument.

Dinamiko (changes requested):

modules/ppcp-axo-block/resources/js/hooks/useSessionRestoration.js
danieldudzic: @puntope Great suggestion! I think this should definitely be discussed, agreed upon, and then applied consistently across all plugin code files. I'll open a separate ticket for this.

modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
danieldudzic: Thank you for catching this! I have now implemented a filter, like you suggested :)

Dinamiko (approved):

PR #3655, opened 2025-09-12
Prevent fatal error when address() is null (PCP-5296)

puntope:
### Description

- Prevent a potential Fatal Error when shipping->address() is null
- Add test coverage to prevent this error in the future


Narek13 (approved):
Thanks for fixing this.

Dinamiko (approved):

PR #3656, opened 2025-09-12
Fix Continuation with PayPal subscriptions (5216)

strangerkir:
## Issue Description
Continuation mode doesn't work with a PayPal subscription when started from the classic cart. Instead, customer sees a regular classic checkout page.

## The problem
When deciding whether to go into Continuation mode, we check [PayPal order status](https://github.com/woocommerce/woocommerce-paypal-payments/blob/988ed95ef47f4588f27ec93c7c7557d27426892a/modules/ppcp-button/src/Helper/ContextTrait.php#L262-L263). But with subscriptions, PayPal Order stays in `CREATED` status, while subscription becomes `ACTIVE` after approval.

## The solution
When we have a subscription ID in the WC session, we know this is a subscription flow and check for a subscription status === `ACTIVE`.

## Additional changes
I did some refactoring while trying to understand what was going wrong. Mainly, I turned ContextTrait into a class and injected it instead of including it with `use`. My reasoning for this are:
* ContextTrait was relying on the `sessionHandler`, which it didn't have. It created an implicit dependency, making a using class responsible for having it without a clear requirement.
* It looks like it was created for a few use cases, but now been widely used in very different contexts, so the concept of trait has outlived itself.
* Class better fits the overall project's DI pattern.

I believe this refactoring is not crucial, and the fix should work without it. So if you don't like the refactoring part, we can revert it and leave only the fix.

## Video before and after
Before:

https://github.com/user-attachments/assets/6f76554a-2824-419a-b8f8-62765b11fd8f

After:

https://github.com/user-attachments/assets/5f191935-f074-4954-ae17-eacea9191f67


modules/ppcp-button/src/Helper/Context.php
```diff
@@ -259,8 +260,24 @@ private function is_paypal_continuation(): bool {
...
+		}
+
+		try {
+			$subscription_status = $this->subscription_status->get_status( $subscription_id );
+		} catch ( RuntimeException | PayPalApiException $exception ) {
+			$subscription_status = '';
+		}
+
+		if ( $subscription_id &&
+			'ACTIVE' !== $subscription_status
```
strangerkir, line 280: This is the most important part, the actual fix.

Dinamiko (approved):
@strangerkir I like the refactor of ContextTrait, it makes dependencies more explicit now, good job!

strangerkir:
> @strangerkir I like the refactor of ContextTrait, it makes dependencies more explicit now, good job!

Thanks :)

PR #3657, opened 2025-09-12
Add fallback for getting subscription order address in classic contexts (5215)

hmouhtar:
The `WooCommerceOrderCreator` class relied on the PayPal order object having the `purchase_units` property to obtain the shipping address. For subscriptions, `purchase_units` is not set in the PayPal order, so subscription order creation was failing in classic non-checkout contexts because the address was missing.

This PR solves that by adding a fallback to build the `Shipping` object from the current `WC_Customer` object through the `ShippingFactory` when the address is missing.


Narek13 (approved):
Haven't tested but code wise looks good to me. Thanks

PR #3658, opened 2025-09-13
Fix for "Call to a member function country_code() on null"

waskosky:
Fixes a weird error I had: PHP Fatal error:  Uncaught Error: Call to a member function country_code() on null in /wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-wc-gateway/src/Checkout/CheckoutPayPalAddressPreset.php:80


puntope:
Duplicated with https://github.com/woocommerce/woocommerce-paypal-payments/pull/3655

PR #3663, opened 2025-09-16
Classic checkout - Guest user - Trial vaulting subscription validation does not work (5056)

Dinamiko:
Validation on classic checkout page triggered after finishing steps on PayPal popup.

If guest user does not populate mandatory billing info filed and click on PayPal button, validation should be triggered at this moment. But now it is showing after finishing process on PayPal popup.

### Steps To Reproduce
- Navigate to shop as guest
- Add trial vaulting subscription to cart
- Do not populate mandatory billing info fields
- Click on PayPal button
   - Observe validation is not triggered at this point
- Finish steps on PayPal popup
   - Observe validation is triggered now

### Changes in this PR
Free trial subscription for guest user has a different flow (creates setup token) than the regular subscriptions flow which validates checkout on server side. In this case we have to validate on client side before server side is reached.

It adds the free trial context to `doBasicCheckoutValidation` conditional to force checkout validation. 
I noticed that `actions.reject()` was not working when executed inside `onSmartButtonClick`, the reason was that the return value from `onSmartButtonClick` was not being handled in `Renderer.js` onClick. The PR modifies the `onClick` handler in `Renderer.js` to return the result from `onSmartButtonClick`.

### DIFF
--- a/modules/ppcp-button/resources/js/button.js
+++ b/modules/ppcp-button/resources/js/button.js
@@ -90,7 +90,7 @@ const bootstrap = () => {
 	};
 
 	const doBasicCheckoutValidation = () => {
-		if ( PayPalCommerceGateway.basic_checkout_validation_enabled ) {
+		if ( PayPalCommerceGateway.basic_checkout_validation_enabled || PayPalCommerceGateway.is_free_trial_cart ) {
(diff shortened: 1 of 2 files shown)


strangerkir (approved):
Looks good to me, thanks!

PR #3669, opened 2025-09-17
Add var-dumper debug helper

AlexP11223:
Added [symfony/var-dumper](https://symfony.com/doc/current/components/var_dumper.html) as a Composer dev dependency, which adds `dump` and `dd` (`dump` + `die`) functions. Sometimes it can help a lot with debugging because the output is much nicer than the standard `var_dump`/`print_r`.

image


Narek13 (approved):
Thanks.

PR #3677, opened 2025-09-19
Prepare the agentic commerce module [5265]

stracker-phil:
### Description

Provides a new module and feature flag for the agentic commerce integration.

### Documentation

While the agentic commerce feature is in development, it's protected by a feature flag, which can be set:

- Environment variable `PCP_AGENTIC_COMMERCE_ENABLED` must be set to `1` to load this module
- WP filter `woocommerce.feature-flags.woocommerce_paypal_payments.agentic_commerce_enabled` must return `true`


modules/ppcp-agentic-commerce/docs/README.md
```diff
@@ -0,0 +1,3 @@
+# Agentic Commerce Overview
+
+Enable PayPal's AI agents to browse and interact with the store. It handles three main tasks: Registering the store as "agentic," providing a product catalog for the AI agents, and exposing REST endpoints for cart and order processing.
```
Narek13:
```suggestion
Enable PayPal's AI agents to browse and interact with the store. It handles three main tasks: Registering the store as "agentic", providing a product catalog for the AI agents, and exposing REST endpoints for cart and order processing. 
```

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace WooCommerce\PayPalCommerce\AgenticCommerce;
+
+use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExecutableModule;
+use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ServiceModule;
+use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ModuleClassNameIdTrait;
+use WooCommerce\PayPalCommerce\Vendor\Psr\Container\ContainerInterface;
+
+/**
+ * Class AgenticCommerceModule
```
Narek13: Why do we need to add this? Is CS still complaining about this?

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class AgenticCommerceModule implements ServiceModule, ExecutableModule {
+	use ModuleClassNameIdTrait;
+
+	/**
+	 * {@inheritDoc}
```
Narek13: BTW, we don’t need to add these `@inheritDoc` comments unless CS complains, since modern IDEs automatically understand them.

Narek13 (commented):
Thanks for taking care of this @stracker-phil . I just left couple of questions and a small wording suggestion.

Dinamiko (approved):

modules/ppcp-agentic-commerce/docs/README.md
stracker-phil: CleanShot 2025-09-22 at 12 10 00@2x

I'll drop the quotes completely, as they are not needed

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
stracker-phil: This is pure convention, but not needed. I've improved the comment

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
stracker-phil: Old habits die hard - removed those comments. Thanks for the reminder!

PR #3686, opened 2025-09-23
Check for messages visibility setting before rendering (5317)

hmouhtar:
The `waitForElement` utility returns a promise that can be rejected with an error whenever the passed element does not exist.

This PR moves the `shouldShow` condition outside the `waitForElement` callback to reduce the number of console errors whenever the messages wrapper does not exist.


Narek13 (approved):
LGTM. Thanks

PR #3703, opened 2025-09-26
Restore BCDC button for ACDC merchants in legacy UI [5355]

stracker-phil:
### Description

Decouple an exclusive ACDC vs BCDC logic of the new UI from merchants on the legacy UI


Narek13 (approved):
Thanks

PR #3704, opened 2025-09-26
Pre-migration notice for ACDC merchants using BCDC buttons [5356]

stracker-phil:
### Description

Modifies the migration notice in the legacy UI in following case:

1. An ACDC merchant is detected
2. The shop uses a BCDC button

In this case, an additional paragraph is added to the migration notice, informing the merchant about the change.

### Screenshots

Regular message:
CleanShot 2025-09-26 at 17 50 56@2x

Extended message:
CleanShot 2025-09-26 at 17 51 36@2x

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -129,6 +129,29 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 						return $notices;
 					}
 
+					/**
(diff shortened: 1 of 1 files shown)


Narek13 (approved):
LGTM. Thanks

stracker-phil:
Update: This change was rejected in favor of temporarily removing the UI upgrade path for all merchants, until we have a stable migration strategy

PR #3705, opened 2025-09-26
Remove migration banner/buttons [5357]

stracker-phil:
### Description

Removes all migration paths that allowed merchants to upgrade from the legacy UI to the new settings UI.

This is a temporary change while prepare/fix some BCDC related upgrade issues for legacy merchants.

Removed upgrade options:
- The "Switch to New Settings" button in the settings page header
- The migration announcement/banner on the settings page
- The Inbox item on the WooCommerce → Home page

### Screenshots of removed items

CleanShot 2025-09-26 at 21 24 33@2x
CleanShot 2025-09-26 at 21 25 40@2x
CleanShot 2025-09-26 at 21 26 22@2x

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -129,6 +129,9 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 						return $notices;
 					}
 
+					// phpcs:ignore -- yes, this code is intentionally and temporarily commented out.
+					/*
+					# PCP-5357 - This notification is temporarily disabled
 					$message = sprintf(
 					// translators: %1$s is the URL for the startup guide.
 						__(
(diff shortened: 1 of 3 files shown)


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -115,6 +118,7 @@ public function run( ContainerInterface $container ): bool {
 					esc_html__( 'This action will permanently switch to the new settings interface and cannot be undone', 'woocommerce-paypal-payments' )
 				)
 			);
+			*/
```
Narek13, line 121: But why disabling for everyone? Shouldn't we disable it only for ACDC merchants who have BCDC enabled?

danieldudzic (approved):
Reviewed and tested using Playground 🛝

**Banner**

|Before|After|
|-|-|
|before_pr_obfuscated|after_pr_obfuscated|


**Inbox**

|Before|After|
|-|-|
|before_pr_inbox|after_PR_inbox|

Narek13 (approved):

PR #3755, opened 2025-10-08
Migrate AppSwitch cross-browser handling logic to frontend (5428)

hmouhtar:
The previous approach for handling cross-browser AppSwitch flows relied on adding query parameters to the return URL. This approach broke the same-tab AppSwitch flow because with the addition of the query parameters, the browser detected the return URL as a different URL and opened a new tab, causing all orders to go through the resume flow.

To solve this, this PR migrates the cross-browser detection and handling logic to the frontend.

PR #3761, opened 2025-10-09
Disable cross-browser AppSwitch query parameters (5441)

hmouhtar:
The current approach for handling cross-browser AppSwitch flows relies on adding query parameters to the return URL. This approach broke the same-tab AppSwitch flow because with the addition of the query parameters, the browser detects the return URL as a different URL and opens a new tab, causing all orders to go through the resume flow.

This will be solved on a later release by https://github.com/woocommerce/woocommerce-paypal-payments/pull/3755, in the meantime, PayPal has requested to disable support for cross-browser AppSwitch, which is done in this PR.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/ReturnUrlFactory.php
+++ b/modules/ppcp-api-client/src/Factory/ReturnUrlFactory.php
@@ -25,15 +25,6 @@ class ReturnUrlFactory {
 		array $request_data = array(),
 		array $custom_query_args = array()
 	): string {
-		// TODO: Temporarily disabled cross-browser appswitch query params.
-		// This logic will be migrated to the frontend using hash params instead of query params
-		// return add_query_arg(
-		// array_merge(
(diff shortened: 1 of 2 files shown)


modules/ppcp-api-client/src/Factory/ReturnUrlFactory.php
```diff
@@ -25,13 +25,16 @@ public function from_context(
...
 		array $custom_query_args = array()
 	): string {
-		return add_query_arg(
-			array_merge(
-				array( self::PCP_QUERY_ARG => 'button' ),
-				$custom_query_args
-			),
-			$this->wc_url_from_context( $context, $request_data )
-		);
+		// TODO: Temporarily disabled cross-browser appswitch query params.
```
Dinamiko: No need to comment code below, better remove it completely, we can always see what was here if we need to add it again in the future.

Dinamiko (changes requested):

Dinamiko (approved):

PR #3762, opened 2025-10-10
Disable SSSC if no webhooks

AlexP11223:
Currently there is a confusing generic error in PayPal when the server-side callback is not working/not accessible. So we should  only use SSSC after checking basic prerequisites such as that webhooks are working (not directly related, but if webhooks are not arriving then SSSC probably is not accessible too), unless enforced with a filter.

Added a check for webhooks via the last webhook storage, e.g. like on the Status page. And now automatically requesting a test webhook after (re-)registering webhooks, otherwise the status would update only after performing the payment or pressing the Simulate button.


Narek13 (approved):
Thanks for working on this Alex. Looks good to me

Dinamiko (approved):

PR #3769, opened 2025-10-14
Fix Uncaught TypeError on simlate cart calls.

### LINKED ISSUE #3684: Fatal error on Simluate cart when no variations sent (5711)
### Describe the Bug

```
PHP Fatal error: Uncaught TypeError: WooCommerce\PayPalCommerce\Button\Helper\CartProductsHelper::add_variable_product(): Argument #3 ($post_variations) must be of type array, string given, called in wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Helper/CartProductsHelper.php on line 121
```

### To Reproduce

Payload can be sent to `/?wc-ajax=ppc-simulate-cart`
Like this:
```{"nonce":"f98ec5b666","products":[{"id":"57585","quantity":"1","variations":null,"extra":{}}]}```
With `variations` being null

I have traced it down to this line https://github.com/woocommerce/woocommerce-paypal-payments/blob/3.1.0/modules/ppcp-button/resources/js/modules/ActionHandler/SingleProductActionHandler.js#L222

This can return `null` if it can't detect any variations on the page. However, the PHP code for simulate cart does not properly handle the `null` case and expects an array to come through.

`null` is also the fallback in https://github.com/woocommerce/woocommerce-paypal-payments/blob/3.1.0/modules/ppcp-button/src/Helper/CartProductsHelper.php#L108

This passes down to https://github.com/woocommerce/woocommerce-paypal-payments/blob/3.1.0/modules/ppcp-button/src/Helper/CartProductsHelper.php#L197
where it is expected to be an array and not null or empty string.

### Expected Behavior

It does not throw a fatal error.

### Actual Behavior

Fatal error.

### Environment

 - WordPress Version: 6.8.2
 - WooCommerce Version: 10.2.1
 - Plugin Version: 3.1.0
 - Browser [e.g. Chrome, Safari] and Version: Any
 - Any other plugins installed: not relevant.

matt-h:
Fix PHP error: ```Uncaught TypeError: WooCommerce\PayPalCommerce\Button\Helper\CartProductsHelper::add_variable_product(): Argument #3 ($post_variations) must be of type array, string given, called in wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Helper/CartProductsHelper.php on line 117```

**Issue**: #3684

---

### Description

Make sure that variations is always an array before calling `add_variable_product`

### Steps to Test

Payload can be sent to `/?wc-ajax=ppc-simulate-cart`
Like this:
`{"nonce":"f98ec5b666","products":[{"id":"57585","quantity":"1","variations":null,"extra":{}}]}`
With variations being `null`

### Documentation

### Changelog Entry

Closes #3684 Fix Uncaught TypeError on simlate cart calls.


AlexP11223:
@matt-h The fix seems fine, but I'm not sure if I understand, is there a way to reproduce it for a normal user without manually sending such request? If there is a variable product without variations, then I think the plugin does not send such requests. Or what was the issue caused by this?

Dinamiko (approved):

matt-h:
@AlexP11223 The root of this was because it was products not using the default Add to Cart form elements.

In this plugin it finds variations by using `this.formElement.querySelectorAll( "[name^='attribute_']" )` but on the product page those elements didn't exist so it couldn't find the variations to send, so null was sent which would cause a PHP fatal error.

PR #3772, opened 2025-10-15
Fix undefined $invoice_id warning

Dinamiko:
This PR fixes bugs found in 3.2.0-rc2 package.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/CaptureFactory.php
+++ b/modules/ppcp-api-client/src/Factory/CaptureFactory.php
@@ -89,9 +89,9 @@ class CaptureFactory {
 			),
 			$amount,
 			(bool) $data->final_capture,
-			(string) $data->seller_protection->status,
-			(string) $data->invoice_id,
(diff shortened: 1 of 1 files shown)


modules/ppcp-api-client/src/Factory/CaptureFactory.php
```diff
@@ -89,9 +89,9 @@ public function from_paypal_response( \stdClass $data ): Capture {
 			),
 			$amount,
 			(bool) $data->final_capture,
-			(string) $data->seller_protection->status,
-			(string) $data->invoice_id,
-			(string) $data->custom_id,
+			(string) ( $data->seller_protection->status ?? '' ),
+			(string) ( $data->invoice_id ?? '' ),
+			(string) ( $data->custom_id ?? '' ),
```
AlexP11223, line 94: Can they be `''`? Maybe they also should be removed in `to_array` when empty?

AlexP11223 (approved):

PR #3775, opened 2025-10-15
fix: Overview tab infinite loading for BCDC-only countries (Russia, Brazil, Japan) (5348)

Narek13:
## Issue Description
The Overview tab displays an infinite loading spinner for BCDC-only countries (Brazil, Japan, and Russia). The tab never finishes loading because `featuresIsReady` never becomes `true` for these countries, as they don't have APM (Alternative Payment Methods) feature support.

**Affected Countries:**
- Brazil
- Japan  
- Russia

**Symptoms:**
- Overview tab shows perpetual loading animation
- Tab becomes functional immediately when store country is changed to US, France, Germany, or other countries with APM support
- Issue occurs in both white-label and branded modes

## PR Description
Fixes infinite loading on Overview tab for BCDC-only countries by implementing a more flexible features readiness check.

**Root Cause:**
For Brazil, Japan, and Russia, the `featuresIsReady` flag never becomes `true` because these countries don't have APM features available. The Overview tab was waiting for this flag before rendering, causing an infinite loading state.

**Solution:**
Fixed the root cause in `useFeatures()` hook by removing the `featuresData.length > 0` condition that prevented `isReady` from being set to `true` when the features array is empty. The hook now properly sets `isReady` to `true` after a successful API fetch regardless of whether features are available, as an empty array is a valid response for countries without APM support (Brazil, Japan, Russia).

This ensures that any component using `FeaturesHooks.useFeatures()` will work correctly for all countries without requiring individual workarounds.

**Impact:**
- ✅ **BCDC-only countries**: Overview tab now loads properly for Brazil, Japan, and Russia
- ✅ **Countries with APM**: No change in behavior, continues working as before
- ✅ **User experience**: Merchants can access Overview tab regardless of store country
- ✅ **No breaking changes**: Gracefully handles both scenarios (with and without features)

<table>
<tr>
<td width="50%">

**Before**

</td>
<td width="50%">

**After**

</td>
</tr>

<tr>
<td width="50%">

Screenshot 2025-10-15 at 18 08 14

</td>
<td width="50%">

Screenshot 2025-10-15 at 18 08 24

</td>
</tr>
</table>

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Tabs/TabOverview.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Tabs/TabOverview.js
@@ -8,16 +8,12 @@ import usePaymentGatewaySync from '../../../../hooks/usePaymentGatewaySync';
 const TabOverview = () => {
 	const { isReady: areTodosReady } = TodosHooks.useTodos();
 	const { isReady: merchantIsReady } = CommonHooks.useMerchantInfo();
-	const { isReady: featuresIsReady, features } = FeaturesHooks.useFeatures();
+	const { isReady: featuresIsReady } = FeaturesHooks.useFeatures();
 
 	// Enable payment gateways after onboarding based on relevant flags.
 	usePaymentGatewaySync();
 
-	// For countries without features (Brazil, Japan, Russia), consider features ready if array is empty
-	const effectiveFeaturesReady =
-		featuresIsReady || ( features && features.length === 0 );
-
-	if ( ! areTodosReady || ! merchantIsReady || ! effectiveFeaturesReady ) {
+	if ( ! areTodosReady || ! merchantIsReady || ! featuresIsReady ) {
 		return (
 			<SpinnerOverlay
 				asModal={ true }
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Tabs/TabOverview.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Tabs/TabOverview.js
@@ -8,12 +8,16 @@ import usePaymentGatewaySync from '../../../../hooks/usePaymentGatewaySync';
 const TabOverview = () => {
 	const { isReady: areTodosReady } = TodosHooks.useTodos();
 	const { isReady: merchantIsReady } = CommonHooks.useMerchantInfo();
-	const { isReady: featuresIsReady } = FeaturesHooks.useFeatures();
+	const { isReady: featuresIsReady, features } = FeaturesHooks.useFeatures();
 
 	// Enable payment gateways after onboarding based on relevant flags.
 	usePaymentGatewaySync();
 
-	if ( ! areTodosReady || ! merchantIsReady || ! featuresIsReady ) {
+	// For countries without features (Brazil, Japan, Russia), consider features ready if array is empty
+	const effectiveFeaturesReady =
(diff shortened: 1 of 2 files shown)


modules/ppcp-settings/resources/js/Components/Screens/Settings/Tabs/TabOverview.js
```diff
@@ -8,12 +8,16 @@ import usePaymentGatewaySync from '../../../../hooks/usePaymentGatewaySync';
...
-	const { isReady: featuresIsReady } = FeaturesHooks.useFeatures();
+	const { isReady: featuresIsReady, features } = FeaturesHooks.useFeatures();
 
 	// Enable payment gateways after onboarding based on relevant flags.
 	usePaymentGatewaySync();
 
-	if ( ! areTodosReady || ! merchantIsReady || ! featuresIsReady ) {
+	// For countries without features (Brazil, Japan, Russia), consider features ready if array is empty
+	const effectiveFeaturesReady =
+		featuresIsReady || ( features && features.length === 0 );
```
puntope: It would be more efficient to fix the `isReady` logic in `FeaturesHooks.useFeatures()` instead. So, in case we use  `FeaturesHooks.useFeatures()` somewhere else, now or in the future, we won't need to add the patch again.

puntope: Mostly the issue comes from https://github.com/woocommerce/woocommerce-paypal-payments/blob/develop/modules/ppcp-settings/resources/js/data/features/hooks.js#L35

However, all that useEffect duplicated logic could be totally removed if we fix the resolver

https://github.com/woocommerce/woocommerce-paypal-payments/blob/develop/modules/ppcp-settings/resources/js/data/features/hooks.js#L35

Basically, this should be` response?.data?.features`

Narek13: Yes, thanks @puntope for the suggestion. I have changed the logic here: 1886ea6c7d10979fe3d05121c256bdcf0f2bce85

modules/ppcp-settings/resources/js/data/features/hooks.js
```diff
@@ -53,12 +53,11 @@ export const useFeatures = () => {
 				const response = await apiFetch( { path: REST_PATH } );
 				const featuresData = response.data?.features || [];
 
-				if ( featuresData.length > 0 ) {
-					await setFeatures( featuresData );
-					await setIsReady( true );
-					return { success: true, features: featuresData };
-				}
-				return { success: false, features: [] };
+				// Set features and ready regardless of array length
```
puntope: I would omit redundant comments.

modules/ppcp-settings/resources/js/data/features/hooks.js
```diff
@@ -32,10 +32,10 @@ export const useFeatures = () => {
...
 					const featuresData = response.data.features;
 
-					if ( featuresData.length > 0 ) {
-						await setFeatures( featuresData );
-						await setIsReady( true );
-					}
+					// Set features regardless of length
+					await setFeatures( featuresData );
+					// Always set ready after successful fetch
+					await setIsReady( true );
```
puntope: I would omit redundant comments

puntope (approved):

Dinamiko (approved):

PR #3798, opened 2025-10-21
[Fix] Google Pay does not update the WC address correctly (PCP-5247)

puntope:
### Description

Fix bug in Google Pay for  Cart and Checkout Blocks when no address is set in the store

[The Update Data from PayPal ](https://developers.google.com/pay/api/web/reference/client#onPaymentDataChanged) only gets intermediate shipping data; it's better to handle the shipping data when payment is processed and load it inside the WC Order in the backend. The removed Handler expected some shipping in the order data that will never be available, making the order to fail. 

### Steps to Test

1. As a guest with a clean browser session, add a physical product to the cart.
2. Go to the block cart.
3. Pay via Google Pay.
4. Order can be completed and fills the address correctly

1. As a guest with a clean browser session, add a physical product to the cart.
2. Go to the block checkout.
3. Pay via Google Pay.
4. Order can be completed and fills the address correctly

### Demo

https://github.com/user-attachments/assets/ae057809-eb70-4c2b-8eb2-8f3227534401

### DIFF
--- a/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
+++ b/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
@@ -5,7 +5,7 @@
  * @package WooCommerce\PayPalCommerce\Button\Helper
  */
 
-declare(strict_types=1);
+declare( strict_types=1 );
 
 namespace WooCommerce\PayPalCommerce\Button\Helper;
 
@@ -24,6 +24,7 @@ use WC_Tax;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Payer;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Shipping;
+use WooCommerce\PayPalCommerce\ApiClient\Factory\PayerFactory;
 use WooCommerce\PayPalCommerce\ApiClient\Factory\ShippingFactory;
 use WooCommerce\PayPalCommerce\Button\Session\CartData;
 use WooCommerce\PayPalCommerce\Button\Session\CartDataFactory;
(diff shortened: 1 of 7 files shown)


Dinamiko (approved):

modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
```diff
@@ -388,4 +399,49 @@ protected function create_subscription( WC_Order $wc_order, int $product_id ): W
...
+	 *
+	 * @param Order      $order The PayPal Order.
+	 * @param array|null $paypal_data The PayPal Response data.
+	 *
+	 * @return Payer|null The Payer object or null if no payer information is available.
+	 */
+	private function get_payer( Order $order, ?array $paypal_data = null ): ?Payer {
+		$payer = $order->payer();
+		if ( is_null( $payer ) && isset( $paypal_data['payer'] ) ) {
+			$payer_data = json_decode( wp_json_encode( $paypal_data['payer'] ) ?: '' );
```
AlexP11223, line 415: Why do we need encode and decode json here? :thinking: 
Maybe can just cast to `object`?

puntope: (object) does not convert nested arrays. But just the first layer of the array. So it will keep the nested arrays as arrays

puntope: We can study doing some function like 
```
function array_to_object_deep( $data ) {
    if ( is_array( $data ) ) {
        return (object) array_map( __FUNCTION__, $data );
    }
    return $data;
}
```

```
array_to_object_deep( $paypal_data['payer'] ?? array() );
```
But `json_decode( wp_json_encode(` does the job and it's kinda a status quo in PHP to doing it like this. Also, its used in several parts of the codebase. So I assumed it was the preferred way of doing this

Screenshot 2025-10-23 at 15 38 34

AlexP11223 (approved):

PR #3803, opened 2025-10-23
Changing Payment Methods for Subscriptions (5465)

Dinamiko:
This PR fixes the cases where a customer wants to change a subscription payment to PayPal payment, either by using an existing PayPal payment or by creating a new one during the process.

It also adds comments and extract vault v2 related code so it would be easy to locate and remove them when vault v3 becomes the only used by all merchants.

Finally it refactors and add test coverage to client scripts.

### Acceptance
A customer can change subscription payment method from card to existing PayPal payment
- Purchase a subscription with card payment
- Save a PayPal payment in My Account / Payment methods / Add payment method if there is none
- Edit subscription in My account / Subscriptions, select PayPal gateway and click "Change payment method"
- Payment method should be changed to PayPal correctly

A customer can change subscription payment method by saving new PayPal payment

- Purchase a subscription with card payment
- Edit subscription in My account / Subscriptions
- PayPal button should be visible below payment gateways
- By clicking PayPal button customer would save a new payment
- Payment method should be changed to PayPal correctly

My Account / Payment methods / Add payment method

A customer can change the PayPal account from PayPal button:

switch-paypal-account


danieldudzic (approved):
LGTM

PR #3810, opened 2025-10-27
Agentic: Add cart persistence (5272)

danieldudzic:
### Description
This PR implements the **cart persistence** and **retrieval**. It introduces a custom WooCommerce session handler and adds the necessary PayPal REST endpoints (`GET` and `PUT` for `/merchant-cart/{cart_id}`).

This change enables the Agentic Commerce flow to manage a persistent cart session that can be created, retrieved, and updated by a partner system using a unique cart identifier.

### Key changes include:
- Implement `AgenticSessionHandler` and `AgenticWcSession` to manage and persist Agentic Commerce carts using a custom WooCommerce session handler.
- Add the **GET /merchant-cart/{cart\_id}** endpoint (`GetCartEndpoint`) to allow retrieval of an existing, persistent cart session.
- Add the **PUT /merchant-cart/{cart\_id}** endpoint (`UpdateCartEndpoint`), including support for partial updates (merging new data with existing cart data).
- Update the **POST /merchant-cart** endpoint (`CreateCartEndpoint`) to save the new cart to the session handler and return a generated `cart_id` and `ec_token`.
- Update the cart response to include the **totals** object (item total, shipping, tax) in the response payload.
- Introduce cart **status** (`CREATED` and `ACTIVE`) in the response objects to comply with the PayPal specification.


modules/ppcp-agentic-commerce/src/Response/CartResponse.php
Biont, line 1: This is the second class with the name `CartResponse` in our codebase - the existing one being `\WooCommerce\PayPalCommerce\WcGateway\StoreApi\Entity\CartResponse`

I realize this is an outome of a reasonable convention and can be easily overcome with bespoke imports and/or aliases. 
But I wanted to raise awareness nonetheless - perhaps this is something we want to run from?

danieldudzic: Right, but do you believe it's an issue? Given we have namespaces? Happy to rename it if necessary, just want to make sure we're aligned on whether this is a standard we want to enforce :)

PR #3814, opened 2025-10-27
Fatal error in Visual Composer front-end editor due to strict $hook_suffix type in SettingsModule.php (5469)

Dinamiko:
This PR fixes type error `Argument #1 ($hook_suffix) must be of type string, null given` by removing param type and return early if it's not a string.


Dinamiko:
Closing, it should be recreated from release/3.3.0 branch.

PR #3816, opened 2025-10-27
Vault V2/V3 separation (5494) 

Dinamiko:
Several modules currently support both versions with conditional logic. This refactoring will separate V2-specific code from hybrid modules into dedicated VaultV2 namespaces, making it easier to eventually remove all V2 support.

It also removes `ppcp-saved-payment-checker` module completely, it was there disabled for long.


modules/ppcp-wc-subscriptions/src/Helper/FreeTrialSubscriptionHelper.php
```diff
@@ -0,0 +1,47 @@
...
+namespace WooCommerce\PayPalCommerce\WcSubscriptions\Helper;
+
+use WC_Subscriptions;
+use WC_Subscriptions_Product;
+
+class FreeTrialSubscriptionHelper {
+	/**
+	 * Checks if the cart contains only free trial.
+	 *
+	 * @return bool
```
AlexP11223: Can remove duplicated types like that, but not super important of course.

AlexP11223 (approved):

PR #3823, opened 2025-10-30
Remove use of PayPal JS SDK actions with AppSwitch (5502)

hmouhtar:
Usage of PayPal JS SDK actions has stopped being supported when AppSwitch is enabled, causing errors during the onApprove callback. 

This PR fixes it by switching the order retrieval to a server-side endpoint rather than SDK actions whenever AppSwitch is enabled.


Dinamiko (approved):

Narek13 (approved):
Thanks, LGTM, although haven't tested.

PR #3826, opened 2025-10-30
reCaptcha

AlexP11223:
Moving the reCaptcha implemantion from the separate plugin by @hmouhtar . Disabled by default, when enabled it uses the invisible v3 reCaptcha for checkout and our Create Order ajax requests, and fallback to the standard v2 reCaptcha when the v3 score is lower than the threshold.

image

For the WC checkout requests the reCaptcha is used only for our card and PayPal gateways to avoid accidentally breaking/affecting other third-party payment methods.

The settings for now are in the WC integration settings.

image


modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
```diff
@@ -0,0 +1,561 @@
...
+		$this->asset_version   = $asset_version;
+		$this->logger          = $logger;
+	}
+
+	protected function should_use_recaptcha(): bool {
+		if ( ! wc_string_to_bool( $this->integration->enabled ) ) {
+			return false;
+		}
+
+		if ( wc_string_to_bool( 'guest_only' ) && is_user_logged_in() ) {
```
Dinamiko: I think this should be `$this->integration->get_option( 'guest_only' )` instead of the hardcoded string `guest_only`, doesn't it?

modules/ppcp-fraud-protection/resources/js/recaptcha-handler.js
```diff
@@ -0,0 +1,305 @@
...
+			console.error( 'PPCP reCAPTCHA: grecaptcha v2 not loaded' );
+			return;
+		}
+
+		container.innerHTML =
+			'<div class="g-recaptcha" data-sitekey="' +
+			config.siteKeyV2 +
+			'" data-theme="' +
+			config.theme +
+			'"></div>';
```
Dinamiko: I suggest to use DOM methods here:
```js
container.innerHTML = '';

const recaptchaDiv = document.createElement( 'div' );
recaptchaDiv.className = 'g-recaptcha';
recaptchaDiv.setAttribute( 'data-sitekey', config.siteKeyV2 );
recaptchaDiv.setAttribute( 'data-theme', config.theme );

container.appendChild( recaptchaDiv );
```

Dinamiko (changes requested):

modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
AlexP11223: oops, yes :D

modules/ppcp-fraud-protection/services.php
```diff
@@ -4,42 +4,35 @@
...
-				'id'   => $prefix . 'enabled',
-				'name' => __( 'Enable', 'woocommerce-paypal-payments' ),
-				'desc' => __( 'Enable fraud protection', 'woocommerce-paypal-payments' ),
-			),
-
-			array(
-				'type' => 'sectionend',
-				'id'   => $prefix . 'end',
-			),
+			PayPalGateway::ID,
```
hmouhtar: Add a filter here so the payment methods can be customized

hmouhtar (approved):
@AlexP11223 Thanks for working on this, looks good. Just added a minor suggestion.

1aTa:
I would highly suggest adding the specific location of this vital feature to the changelog and plugin frontpage.

There was a line in the changelog "Enhancement – Customizable reCAPTCHA implementation for PayPal endpoints #3829" but I had to search the forum and github to find out that it's been placed under Integration and not under Payments.

PR #3830, opened 2025-10-31
Agentic: Add full cart replacement endpoint (5269)

danieldudzic:
### Description
This PR implements the **full cart replacement** functionality. It introduces a new REST endpoint that allows complete replacement of an existing cart's contents while preserving the payment token.

This change enables replacing an entire cart in a single operation, which is more efficient than partial updates when the cart contents need to be completely changed.

### Key changes include:
- Add the **PUT /merchant-cart/{cart_id}** endpoint (`ReplaceCartEndpoint`) to allow full replacement of an existing cart session.
- Verify cart existence before performing replacement operation.
- Preserve the `ec_token` (payment token) during cart replacement to maintain payment context.
- Return **200 OK** status (not 201 Created) to indicate successful replacement of existing resource.
- Add test coverage for successful replacement, cart not found scenarios, complete data replacement validation, and failure handling.

### DIFF
--- a/modules/ppcp-agentic-commerce/src/Endpoint/ReplaceCartEndpoint.php
+++ b/modules/ppcp-agentic-commerce/src/Endpoint/ReplaceCartEndpoint.php
@@ -1,0 +1,145 @@
+<?php
+/**
+ * Replace Cart Endpoint for Agentic Commerce.
+ *
+ * PUT /api/paypal/v1/merchant-cart/{cart_id}
+ *
+ * @package WooCommerce\PayPalCommerce\AgenticCommerce\Endpoint
+ */
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\AgenticCommerce\Endpoint;
+
+use WP_REST_Request;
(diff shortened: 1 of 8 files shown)


modules/ppcp-agentic-commerce/src/Endpoint/ReplaceCartEndpoint.php
```diff
@@ -0,0 +1,145 @@
...
+	 * @param ResponseFactory       $response_factory The response factory.
+	 */
+	public function __construct(
+		JwtAuthService $auth_service,
+		AgenticSessionHandler $session_handler,
+		ResponseFactory $response_factory
+	) {
+		parent::__construct( $auth_service, $response_factory );
+		$this->session_handler = $session_handler;
+	}
```
stracker-phil: Suggestion: I see all endpoints have the same signature (all use the session handler), I suggest to move the session handler to the base class, and remove the custom constructor from all Endpoint classes

CleanShot 2025-11-03 at 18 28 22

stracker-phil (approved):

PR #3831, opened 2025-10-31
Fix translation notices

AlexP11223:


puntope (changes requested):
Unfortunatelly that solves only 2 translations but then new trabnslations appear like


_x( $text = 'Express Checkout', $context = 'Name of Buttons Location', $domain = 'woocommerce-paypal-payments' ) | .../extensions.php:20

puntope (approved):

PR #3835, opened 2025-11-03
Log reCaptcha rejections

AlexP11223:
Added an option for enabling logging of requests rejected by reCaptcha v3, and also a counter.

The logs are saved to the WC logs using `woocommerce-paypal-payments-recaptcha` source, the "view logs" link opens the WC logs page filtered by this source.
The log entry includes the endpoint name, reCaptcha score, IP, user-agent, request body (omitting some data like reCaptcha keys to keep it more readable) and the current cart if available.

The rendering of the counter and the link was supposed to be inside the `RecaptchaIntegration` class but we cannot add any constructor parameters here because WC needs to create a class instance, so moved into `Recaptcha`.

image

image

image

### DIFF
--- a/modules/ppcp-fraud-protection/src/PersistentCounter.php
+++ b/modules/ppcp-fraud-protection/src/PersistentCounter.php
@@ -1,0 +1,24 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\FraudProtection;
+
+/**
+ * A counter saving the current state into WP option.
+ */
+class PersistentCounter {
+	protected string $option_id;
+
+	public function __construct( string $option_id ) {
+		$this->option_id = $option_id;
+	}
+
+	public function increment(): void {
+		update_option( $this->option_id, $this->current() + 1 );
+	}
+
+	public function current(): int {
+		return (int) get_option( $this->option_id, 0 );
+	}
+}
--- a/modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
+++ b/modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
@@ -7,6 +7,8 @@ namespace WooCommerce\PayPalCommerce\FraudProtection\Recaptcha;
 use Automattic\WooCommerce\Utilities\OrderUtil;
 use Psr\Log\LoggerInterface;
 use WC_Order;
+use WC_Product;
+use WooCommerce\PayPalCommerce\FraudProtection\PersistentCounter;
 use WP_Error;
 
 class Recaptcha {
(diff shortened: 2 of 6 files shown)


modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
```diff
@@ -564,7 +614,73 @@ private function customer_ip(): string {
...
+		unset( $request_data['g-recaptcha-response'] );
+		$request_data_str = wp_json_encode( $request_data );
+
+		$cart = wp_json_encode( $this->cart_contents() );
+
+		$this->logger->debug(
+			"Rejected by v3 reCAPTCHA at {$endpoint_name} with score {$this->last_v3_score}, IP: {$ip}, User Agent: {$user_agent}, request content: {$request_data_str}, cart: {$cart}",
+			array(
+				'source' => $this->rejection_logger_source,
+			)
```
hmouhtar, line 681: Source can be just a constant, no need to have a service for that. Same applies for the counter option ID, if all the service does is return a string with no manipulation, then it should be a constant IMO

modules/ppcp-fraud-protection/src/PersistentCounter.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\FraudProtection;
+
+/**
+ * A counter saving the current state into WP option.
+ */
+class PersistentCounter {
```
hmouhtar, line 10: What's the purpose of logging the amount of v3 rejections? That data is available in the reCAPTCHA panel (percentage of suspicious traffic, score distribution, total requests)

modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
hmouhtar: $cart and $request_data should be added as context of the debug method: https://developer.woocommerce.com/docs/best-practices/data-management/logging/#context.

The logger will handle converting it to JSON and making it legible in the log viewer

hmouhtar (changes requested):

modules/ppcp-fraud-protection/src/PersistentCounter.php
AlexP11223: hm, probably not super useful, but still may be good to have some basic stats in the plugin, e.g. in case the merchant sees it here but does not check the Google panel, or there are multiple plugins/(sub)sites etc. Also the requests may be blocked by a custom `woocommerce_paypal_payments_recaptcha_verify_v3_result` filter.

hmouhtar (approved):

PR #3838, opened 2025-11-04
Apple Pay - When coupon is applied on total and we click on Apple Pay it is not visible on popup (5456)

Dinamiko:
When coupons were applied after page load in WooCommerce Blocks checkout, Apple Pay continued showing the original price. The payment sheet displayed incorrect totals because the transaction info was never refreshed after cart updates.

This PR add a `useEffect` hook that subscribes to the WooCommerce Blocks cart data store using `wp.data.subscribe`. When the cart total changes (e.g., coupon applied), the Apple Pay manager reinitializes with fresh transaction info, ensuring the payment sheet displays the correct discounted amount.

### How to test
- Add product to cart
- Navigate to Blocks checkout
- Apply a coupon code
- Click Apple Pay button
- Verify payment sheet shows discounted total


modules/ppcp-applepay/resources/js/boot-block.js
```diff
@@ -57,13 +58,50 @@ const ApplePayComponent = ( { isEditing, buttonAttributes } ) => {
...
+			}
+
+			const totals = store.getCartTotals();
+			if ( ! totals ) {
+				return;
+			}
+
+			const currentTotal =
+				parseInt( totals.total_price, 10 ) /
+				10 ** totals.currency_minor_unit;
```
AlexP11223: Probably can just use `total_price` without any conversions.

AlexP11223 (approved):

PR #3859, opened 2025-11-13
[QA] Fixes for v3.3.0

mishautkin:
### Description

- Fixes for running autotests for 3.3.0
- Enabled video recording, full page screenshots and tracing in playwright-report (for local execution only because of memory consumption)
- Added "Critical" labels to more tests


tests/qa/tests/05-transactions/_test-scenarios/cart.scenario.ts
```diff
@@ -42,11 +42,11 @@ export const transactionsOnCart = ( testsData: ShopOrder[] ) => {
...
-				await payPalApi.assertOrder( orderJson, testData );
-				await payPalApi.assertPayment(
-					orderJson.transaction_id,
-					testData
-				);
+				// await payPalApi.assertOrder( orderJson, testData );
+				// await payPalApi.assertPayment(
+				// 	orderJson.transaction_id,
+				// 	testData
+				// );
```
puntope: Lets completely remove old code if doesnt work

puntope (commented):
Screenshot 2025-11-13 at 16 49 30

Some tests are failing for me

tests/qa/tests/05-transactions/_test-scenarios/cart.scenario.ts
mishautkin: It's been decided to skip these assertions recently. But we don't have other tests which would check it. That's why I only commented it out.

mishautkin: Removed. We can add in future.

puntope (approved):

PR #3864, opened 2025-11-14
Prevent APM Methods to show in the UI when not eligible (5466)

puntope:
### Description

- Add a new static function in service.php to check eligibility for APM based on Merchant country
- Use that check to hide or show APM Gateway methods

### Steps to Test

1. Set USA merchant in PayPal 
2. Feature APM is available
3. APM Methods are available and visible in UI
4.  Set Brazil, Rusia or Japan merchant in PayPal 
5. Feature APM is NOT available
6. APM Methods are not available nor visible in UI

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -500,6 +500,8 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 					unset( $payment_methods[ PWCGateway::ID ] );
 				}
 
+				$this->unset_apm_payment_methods_when_not_eligible( $container, $payment_methods );
+
 				return $payment_methods;
 			}
 		);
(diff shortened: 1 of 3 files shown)


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -914,4 +916,21 @@ protected function is_gateway_enabled( string $gateway_name ): bool {
 
 		return $gateway_enabled === 'yes';
 	}
+
+	protected function unset_apm_payment_methods_when_not_eligible( ContainerInterface $container, array &$payment_methods ): void {
```
Dinamiko: I suggest to be explicit about the method dependencies, instead of passing the whole container, pass the concrete dependencies this method relies on. Also I suggest to not pass `$payment_methods` as reference but make the method return it, it would make it easier to reason about (it avoids the function mutating the caller state) and easier to unit test.

Dinamiko (changes requested):

modules/ppcp-settings/src/SettingsModule.php
puntope: Hey @Dinamiko thanks for the feedback. I refactored it

https://github.com/woocommerce/woocommerce-paypal-payments/pull/3864/commits/2c5094351de2f05052d60af7d4f4e605237a5213

This is ready for a new round

puntope:
Hey @Dinamiko I applied your feedback and this PR is ready for a new round. Thanks

Dinamiko (approved):

PR #3885, opened 2025-11-25
Duplicate PayPal captures and double stock reductions (5461)

Dinamiko:
Two concurrent POST requests attempted to capture the same PayPal order. Different request IDs confirm two separate PHP contexts processed the same order simultaneously, resulting in duplicate PayPal transactions and double stock reduction.

Notes:
- Happens intermittently, difficult to reproduce.
- Likely race condition between webhook (`CHECKOUT.ORDER.APPROVED`) and return URL capture flow.

### PR Changes

- Added atomic processing lock using `_ppcp_processing` meta flag with direct SQL to prevent concurrent order processing. The lock uses a single atomic `UPDATE ... WHERE meta_value != 'yes'` query, eliminating the race condition window that exists with separate check-then-set operations.
- Added HPOS compatibility for the atomic lock, supporting both legacy (`wp_postmeta`) and HPOS (`wp_wc_orders_meta`) storage modes.
- Added validation checks to skip processing if order already has a transaction ID or is not in pending/on-hold status
- Added early return in `ReturnUrlEndpoint` for orders already processed
- Wrapped order processing in try-finally block to ensure lock is always released
- Improved error handling and logging for concurrent processing attempts
- Added comprehensive integration tests for concurrency scenarios
- Updated test configuration to exclude node_modules from coverage

**Why Atomic SQL Instead of WooCommerce API**
The WooCommerce API (get_meta + update_meta_data + save) performs separate read and write operations, allowing two processes to both read "unlocked" and then both set "locked". The direct SQL approach combines the check and set into a single atomic database operation, guaranteeing only one process can acquire the lock.


modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -212,84 +212,96 @@ public function __construct(
...
-				if ( $is_paypal_return ) {
-					$this->logger->warning(
-						sprintf(
-							'No PayPal order ID found for WooCommerce order #%d.',
-							$wc_order->get_id()
+		if ( ! $this->verify_order_can_be_processed( $wc_order ) ) {
+			return;
+		}
+
+		$wc_order->update_meta_data( '_ppcp_processing', 'yes' );
```
danieldudzic: Suggestion: I'd consider adding a timestamp-based lock expiration (to prevent a situation when PHP process crashes `_ppcp_processing` flag might remain set forever, blocking all future attempts):
```
  // When setting lock:
  $wc_order->update_meta_data( '_ppcp_processing', time() );
```

```
  // When checking lock:
  $processing_lock = (int) $wc_order->get_meta( '_ppcp_processing', true );
  if ( $processing_lock && ( time() - $processing_lock ) < 60 ) {
      $this->logger->warning( /* lock active */ );
      return false;
  }
```

danieldudzic (approved):
Nice work @Dinamiko! This should reduce the race condition window. In case the issue persists, I'd recommend implementing the atomic lock acquisition.

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
luzat: The lock mechanism should probably be implemented in a more robust way:

- There is still a race condition if multiple `verify_order_can_be_processed` run at once.
- The window for failure is even larger, because the order meta data may have been read way earlier during the request (when `$wc_order` is initially loaded) and `update_meta_data` won't actually save the data until `$wc_order->save()` if I recall correctly.

I am using a cache-based lock in a workaround (`wp_cache_add`), but this relies on a cache being available and persistent for the locking period. The nice thing about that is, that `wp_cache_add` returns whether the key already exists. This is more complicated with portable SQL, but a solution with a separate locking table with `INSERT … ON DUPLICATE KEY UPDATE …` can be done for MySQL/MariaDB and could be 100% robust. A timeout would be good, still.

The current large race condition window could render the lock mostly useless and too unreliable.

Dinamiko:
@luzat thanks for the feedback, I've updated the PR implementing atomic SQL locking instead of using the WooCommerce API. 

I've implemented it using existing tables rather than creating a dedicated locking table to avoid adding complexity for now, if issue persist then we'll explore adding it. 

cc. @danieldudzic

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -355,6 +368,152 @@ public function create_order( WC_Order $wc_order ): Order {
...
+					$order_id
+				)
+			);
+			return false;
+		}
+
+		$wc_order->update_meta_data( '_ppcp_processing', 'yes' );
+		$wc_order->save();
+
+		return true;
```
danieldudzic: Seems like the fallback path has race condition. When meta doesn't exist, the fallback is non-atomic.

Perhaps we can use the following approach:

```suggestion
// Row doesn't exist yet - try atomic insert
// INSERT IGNORE returns 1 if inserted, 0 if duplicate key (already exists)
$result = $wpdb->query(
    $wpdb->prepare(
        "INSERT IGNORE INTO {$table} ({$id_column}, meta_key, meta_value) VALUES (%d, %s, %s)",
        $order_id,
        '_ppcp_processing',
        'yes'
    )
);

if ( $result > 0 ) {
    return true; // We inserted, we got the lock
}

// Insert failed - row was created by another process between UPDATE and INSERT
$this->logger->warning(
    sprintf(
        'Order #%d is already being processed (lock active), skipping payment processing.',
        $order_id
    )
);
return false;
```

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -355,6 +368,152 @@ public function create_order( WC_Order $wc_order ): Order {
...
+		}
+
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$rows_updated = $wpdb->query(
+			$wpdb->prepare(
+				"UPDATE {$table}
+				SET meta_value = 'yes'
+				WHERE {$id_column} = %d
+				AND meta_key = '_ppcp_processing'
+				AND meta_value != 'yes'",
```
danieldudzic: Minor consideration: If PHP crashes mid-process, the lock stays forever since it's never released. Stripe handles this by storing `time() + 5 * MINUTE_IN_SECONDS` instead of `'yes'`, allowing stale locks to auto-expire.

Not blocking for this PR, but might be worth a follow-up issue if we see stuck orders in production.

danieldudzic (approved):
Solid approach - matches Stripe's pattern. Just one suggestion and one minor consideration :)

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
Dinamiko: @danieldudzic good catch, I've replaced the non-atomic fallback (`SELECT + update_meta_data`) with an atomic `INSERT ... WHERE NOT EXISTS` query.

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
Dinamiko: Added, we now store `time() + 5 * MINUTE_IN_SECONDS` instead `'yes'`

PR #3890, opened 2025-11-26
Consolidate merchant details before and after onboarding (5311)

puntope:
### Description

Use `api.merchant.country` instead of `api.shop.country` when applicable to select PayPal country when merchant is onboarded or shop country otherwise.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -18,6 +18,7 @@ use WooCommerce\PayPalCommerce\ApiClient\Helper\PartnerAttribution;
 use WooCommerce\PayPalCommerce\Applepay\ApplePayGateway;
 use WooCommerce\PayPalCommerce\Applepay\Assets\AppleProductStatus;
 use WooCommerce\PayPalCommerce\Axo\Gateway\AxoGateway;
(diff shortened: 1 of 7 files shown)


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -439,12 +438,15 @@ function ( array $payment_methods ) use ( $container ): array {
 				$general_settings = $container->get( 'settings.data.general' );
 				assert( $general_settings instanceof GeneralSettings );
 
-				$merchant_data    = $general_settings->get_merchant_data();
-				$merchant_country = $merchant_data->merchant_country;
+				$messages_apply = $container->get( 'button.helper.messages-apply' );
+				assert( $messages_apply instanceof MessagesApply );
+				$pay_later_eligible = $messages_apply->for_country();
```
Dinamiko: `$messages_apply` and `$pay_later_eligible` seems never used, check it and remove if not needed.

Dinamiko (changes requested):

modules/ppcp-settings/src/SettingsModule.php
puntope: Hi @Dinamiko, I removed the unused variables and also `$dcc_applies` and `$all_payment_methods`, both not used as well.

Dinamiko (approved):

PR #3894, opened 2025-12-01
Fix APM icon size in Classic Checkout (5555)

danieldudzic:
### Description

Limit APM icon height to 24px in the classic checkout.

### Note

The original issue is reproducible when Credit Card processing is not present in the checkout.

### Screenshot

|Before|After|
|-|-|
|before_pwc|after_pwc|

### DIFF
--- a/modules/ppcp-local-alternative-payment-methods/src/LocalAlternativePaymentMethodsModule.php
+++ b/modules/ppcp-local-alternative-payment-methods/src/LocalAlternativePaymentMethodsModule.php
@@ -74,6 +74,21 @@ class LocalAlternativePaymentMethodsModule implements ServiceModule, ExtendingMo
 
 		$this->register_pwc_feature_flag_filters();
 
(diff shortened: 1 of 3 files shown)


modules/ppcp-local-alternative-payment-methods/src/LocalAlternativePaymentMethodsModule.php
```diff
@@ -74,6 +74,21 @@ private function run_with_translations( ContainerInterface $c ): void {
 
 		$this->register_pwc_feature_flag_filters();
 
+		add_action(
+			'wp_enqueue_scripts',
+			function () use ( $c ) {
+				$module_url    = $c->get( 'ppcp-local-apms.url' );
```
puntope, line 84: We want this to be loaded on every page or just in the checkout page?

Dinamiko (approved):

PR #3901, opened 2025-12-02
Business rule validation framework (5271)

stracker-phil:
### Description

This PR introduces a pluggable validation framework for cart operations. The `CartValidationProcessor` runs validators through a common `ValidatorInterface`, accumulating validation issues before cart operations proceed.

The framework includes two core validators:

**ProductValidator** checks that cart items reference valid, purchasable WooCommerce products.

**InventoryValidator** verifies stock availability and quantity constraints.

Validators register through a WordPress action hook (`woocommerce_paypal_payments_agentic_commerce_validators`), enabling third-party extensions to add custom business rules. Each validator can skip execution if the cart already contains related issues, preventing duplicate errors.

The validation processor is injected into REST endpoints and runs automatically on cart operations. Supporting changes include product caching in `ProductManager` to reduce database queries, and cleaner cart operation methods in the REST endpoint base class.


modules/ppcp-agentic-commerce/services.php
```diff
@@ -144,12 +145,15 @@
 	},
 
 	// Validation services.
-	'agentic.validation.product'          => static function ( ContainerInterface $c ): ProductValidator {
+	'agentic.validation.processor'        => static function (): CartValidationProcessor {
+		return new CartValidationProcessor();
+	},
+	'agentic.validator.product'           => static function ( ContainerInterface $c ): ProductValidator {
```
puntope, line 151: Makes more sense to keep `validation` following the namespace structure.

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
```diff
@@ -43,6 +42,16 @@ class AgenticCommerceModule implements ServiceModule, ExecutableModule {
 		'agentic.rest.checkout',
 	);
 
+	/**
+	 * A list of default cart validation services that verify business rules.
+	 *
+	 * Validators are processed in the order they are listed here.
+	 */
+	private const CART_VALIDATION_SERVICES = array(
+		'agentic.validator.product',
```
puntope, line 51: Same as the previous comment. I would suggest using `validation` and keeping consistency with the namespace

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
```diff
@@ -127,6 +137,17 @@ static function () use ( $container ): void {
...
+		add_action(
+			'woocommerce_paypal_payments_agentic_commerce_validators',
+			static function ( CartValidationProcessor $processor ) use ( $container ) {
+				foreach ( self::CART_VALIDATION_SERVICES as $service_id ) {
+					$validator = $container->get( $service_id );
+					assert( $validator instanceof ValidatorInterface );
+					$processor->register_validator( $validator );
+				}
+			}
+		);
```
puntope, line 149: I wonder if it would be more convenient to define and register the CART_VALIDATION_SERVICES in the CartValidatorProcessor as "default validators" so we can prevent an extra hook.

puntope: As a second advantage, it will reduce Validation responsibility load from the module itself.

modules/ppcp-agentic-commerce/src/CartValidation/InventoryValidator.php
```diff
@@ -9,56 +9,72 @@
...
+		$product = $this->product_manager->find_product( $item );
+
+		if ( ! $product ) {
+			return null;
 		}
 
-		return $issues;
+		if ( ! $this->product_manager->is_in_stock( $product ) ) {
+			return new ItemOutOfStock(
+				'Product is no longer available',
```
puntope, line 57: Are the issues visible for the user at some point? If so, are we translating them?

modules/ppcp-agentic-commerce/src/CartValidation/ProductValidator.php
```diff
@@ -9,53 +9,63 @@
...
-class ProductValidator {
+class ProductValidator implements ValidatorInterface {
 	private ProductManager $product_manager;
 
 	public function __construct( ProductManager $product_manager ) {
 		$this->product_manager = $product_manager;
 	}
 
-	public function validate_products_exist( PayPalCart $cart ): array {
+	public function validate( PayPalCart $cart ): ?array {
```
puntope, line 26: Nice move adding ValidatorInterface. However, this can be even better by abstracting `validate()` method. If you think about it, it's the same in both Validators:

- Check if it has Inventory issue 
- Iterate over  PayPalCart items
- For each of them, run a concrete validator (this is the only part that should be implemented in the concrete Validators)
- If it has issues, return them in an array

puntope: Also, the ProductManager is common in both, so an `AbstractValidator` or `AbstractCartItemValidator` will do the job.

modules/ppcp-agentic-commerce/src/Endpoint/AgenticRestEndpoint.php
```diff
@@ -100,7 +100,7 @@ public function check_permission( WP_REST_Request $request ) {
 	 * @param int          $status_code HTTP status code.
 	 * @return WP_REST_Response The successful response.
 	 */
-	protected function cart_details( CartResponse $cart, int $status_code = 200 ): WP_REST_Response {
+	protected function cart_details( CartResponse $cart, int $status_code ): WP_REST_Response {
 		$this->logger->info( "[REST] $status_code Response", $cart->to_array() );
```
puntope, line 104: Im aware this comes from previous code and is not refactored. But are we sure we want to log all the requests?

modules/ppcp-agentic-commerce/src/Endpoint/AgenticRestEndpoint.php
```diff
@@ -25,7 +25,7 @@
 use WooCommerce\PayPalCommerce\AgenticCommerce\Response\CartResponse;
 use WooCommerce\PayPalCommerce\AgenticCommerce\Response\ResponseFactory;
 use WooCommerce\PayPalCommerce\AgenticCommerce\Session\AgenticSessionHandler;
-use WooCommerce\PayPalCommerce\AgenticCommerce\CartValidation\ProductValidator;
+use WooCommerce\PayPalCommerce\AgenticCommerce\CartValidation\CartValidationProcessor;
```
puntope, line 28: use WooCommerce\PayPalCommerce\AgenticCommerce\Errors\Http\BadRequestError;

Is not necessary anymore

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
```diff
@@ -109,57 +107,36 @@ public function complete_checkout( WP_REST_Request $request ): WP_REST_Response
...
-			$this->session_handler->destroy_cart_session( $cart_id );
-
-			// Build the response with payment confirmation.
-			$response = $this->response_factory->from_order(
-				$order,
-				$cart
-			);
-
-			return $this->cart_details( $response );
+		// It's time to create the WooCommerce order.
```
puntope, line 128: What is the value of this comment? The statement clearly says` $this->create_wc_order`

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
```diff
@@ -109,57 +107,36 @@ public function complete_checkout( WP_REST_Request $request ): WP_REST_Response
...
+		$order = $this->create_wc_order( $cart, $payment_method, $session['ec_token'] );
```
puntope, line 129: What is  `ec` in `ec_token` ?

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
```diff
@@ -109,57 +107,36 @@ public function complete_checkout( WP_REST_Request $request ): WP_REST_Response
...
-		$validation_issues = $this->product_validator->validate_products_exist( $cart );
-		$validation_issues = array_merge( $validation_issues, $this->inventory_validator->verify_inventory( $cart ) );
+		// If the cart has _any_ validation issue, stop here.
+		if ( $cart->issues() ) {
+			$cart_response = $this->response_factory->from_cart( $cart );
 
-		if ( ! empty( $validation_issues ) ) {
-			$cart = $cart->with_validation_issues( ...$validation_issues );
-			return $this->cart_details( $this->response_factory->active_cart( $cart, $cart_id, $cart_session['ec_token'] ) );
+			return $this->cart_details( $cart_response );
```
puntope, line 125: Missing status, that is not anymore set by default

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
```diff
@@ -45,22 +45,19 @@ class CheckoutEndpoint extends AgenticRestEndpoint {
 
 	protected AgenticCheckoutProcessor $checkout_processor;
 
-	protected InventoryValidator $inventory_validator;
-
 	public function __construct(
 		AuthServiceProvider $auth_provider,
 		AgenticSessionHandler $session_handler,
 		ResponseFactory $response_factory,
+		CartValidationProcessor $validation_processor,
```
puntope, line 52: This comment applies to all the endpoints.

The architecture works, but we are violating the single responsibility principle by creating a single class per endpoint that handles both route definition and handling 

Instead, I suggest a single MerchantCart endpoint with all the routes. You can do that even with a single `register_route` call.

Then it receives different Handlers for processing the request.

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
```diff
@@ -120,7 +120,6 @@ private function create_order( PayPalOrder $paypal_order, $cart_data, PayPalCart
 			$paypal_data['payer'] = $payer_data;
 		}
```
puntope, line 122: use WooCommerce\PayPalCommerce\AgenticCommerce\Schema\Address;


Is not used anymore. BTW there are PHPCS rules to prevent this

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
```diff
@@ -145,7 +144,7 @@ private function create_order( PayPalOrder $paypal_order, $cart_data, PayPalCart
 	/**
 	 * Build payer data from PayPal cart.
 	 *
-	 * @param PayPalCart $cart The PayPal cart with customer data.
+	 * @param PayPalCart $cart The PayPal cart.
 	 * @return array Payer data array.
 	 */
 	private function build_payer_data( PayPalCart $cart ): array {
```
puntope, line 150: Some inconsistencies with the return types in this file. Some are defined ... like `:array ` but some have missing return types

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
```diff
@@ -155,74 +154,39 @@ private function build_payer_data( PayPalCart $cart ): array {
 
 		$payer_data = array();
 
-		// Add email address.
 		$customer = $cart->customer();
 		if ( $customer ) {
+			// Note: name is an array with items 'given_name' and 'surname', which can be string or null.
 			$payer_data['name'] = $customer->name();
```
puntope, line 160: I dont have enough context, but I'm assuming we dont need the surname, otherwise we can do something like get_complete_name or load also the surname in the array.

puntope: Something like we do below with `'full_name' => CartHelper::full_customer_name( $cart )`

modules/ppcp-agentic-commerce/src/Helper/CartHelper.php
```diff
@@ -53,4 +54,69 @@ static function ( float $cart_total, CartItem $item ): float {
 			0.0
 		);
 	}
+
+	public static function full_customer_name( PayPalCart $cart, string $default = '' ): string {
```
puntope, line 58: Static helpers have been demonstrated to be hard to test. I recommend refactoring this.

modules/ppcp-agentic-commerce/src/Helper/CartTransformer.php
```diff
@@ -75,12 +75,10 @@ private function build_cart_items( PayPalCart $paypal_cart ): array {
 		$cart_items = array();
 
 		foreach ( $paypal_cart->items() as $item ) {
-			$variant_id = $item->variant_id();
-			$item_id    = $item->item_id();
-			$quantity   = $item->quantity();
+			$quantity = $item->quantity();
```
puntope, line 78: At the beginning of the file there is an extra empty line 

```
use WooCommerce\PayPalCommerce\Button\Session\CartData;

use WooCommerce\PayPalCommerce\AgenticCommerce\Schema\PayPalCart;
```

puntope:
```
private function generate_cart_item_key( int $product_id, int $variation_id ): string {
		return md5( (string) $product_id . '-' . (string) $variation_id );
	}
```
Not necessarily the explicit casting

puntope: (I know this is not for this PR, just adding here what I found for follow up PRs)

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
```diff
@@ -32,24 +32,24 @@ class PayPalOrderBuilder {
...
 		PayPalCart $cart,
-		CartData $cart_data
+		CartData $woo_cart_data
 	): PurchaseUnit {
 
-		$cart_items = $cart_data->items();
+		$cart_items = $woo_cart_data->items();
 
-		// Calculate total from cart items.
+		// TODO: Why not using the PayPalCart to calculate the total?
```
puntope, line 46: These TODOs questions are usually never answered :) and they stay in the code forever

I suggest creating a follow-up issue and discussing it there.

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
```diff
@@ -78,74 +78,31 @@ public function build_purchase_unit_from_cart(
...
 		}
 
 		$address = new Address(
-			$shipping->country_code() ?? '',
-			$shipping->address_line_1() ?? '',
-			$shipping->address_line_2() ?? '',
-			$shipping->admin_area_2() ?? '', // city.
-			$shipping->admin_area_1() ?? '', // state.
-			$shipping->postal_code() ?? ''
+			$shipping['country_code'],
```
puntope, line 98: I liked the object notation, which is typo-safe.

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderManager.php
```diff
@@ -67,54 +69,72 @@ public function create_order( PayPalCart $cart ): string {
...
-			// Step 3: Create PayPal Order (application_context filter is registered in AgenticCommerceModule).
+			// Create PayPal Order (application_context filter is registered in AgenticCommerceModule).
 			$paypal_order = $this->order_endpoint->create(
 				array( $purchase_unit ),
 				ExperienceContext::SHIPPING_PREFERENCE_NO_SHIPPING,
 				null,               // payer.
-				'agentic-commerce', // payment_method identifier.
-				array(),            // request_data.
-				null                // payment_source.
+				'agentic-commerce'  // payment_method identifier.
```
puntope, line 91: These identifiers in the comments are provided by the IDEs because we already cover that with PHPDocs

Screenshot 2025-12-04 at 17 15 55

modules/ppcp-agentic-commerce/src/CartValidation/InventoryValidator.php
danieldudzic: It probably makes sense to localize, even though I think the agent would translate it too.

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
danieldudzic: I believe `ec` stands for Express Checkout in this context.

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
danieldudzic: For context: `PayPalCart` contains _untrusted_ data from the AI agent. The AI could send manipulated prices.

We should fetch actual product prices from the WooCommerce database and calculate the total.

  This prevents the AI from:
  - Setting fake prices ($0.01 for expensive items)
  - Requesting invalid products
  - Manipulating totals or taxes

At least that's my understanding.

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
```diff
@@ -32,24 +32,24 @@ class PayPalOrderBuilder {
...
 
-		// Calculate total from cart items.
+		// TODO: Why not using the PayPalCart to calculate the total?
 		$total = 0.0;
 		foreach ( $cart_items as $item ) {
 			$total += (float) $item['line_total'];
 		}
 
-		// Use the WooCommerce currency.
+		// TODO: Why do we use Woo currency instead of the PayPalCart currency?
```
danieldudzic, line 52: My understanding is that the `PayPalCart` currency is used for validation (ensuring AI sends correct currency in items), but the actual PayPal order should use the store's currency.

We can't create an order in a currency that the WooCommerce store doesn't support, no?

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderManager.php
```diff
@@ -67,54 +69,72 @@ public function create_order( PayPalCart $cart ): string {
...
-			// Step 1: Transform PayPalCart to CartData.
-			$cart_data = $this->cart_transformer->paypal_cart_to_wc_cart( $cart );
+		// At this stage, the order intent is always AUTHORIZE, not CAPTURE.
+		$set_order_intent = static fn(): string => 'AUTHORIZE';
+
+		/*
+		 * Build a minimal PurchaseUnit directly from the PayPalCart details.
+		 * We can't use from_wc_order() yet because there's no WC order.
+		 */
+		$woo_cart_data = $this->cart_transformer->paypal_cart_to_wc_cart( $cart );
```
danieldudzic, line 79: A complete nitpick, but I'd stick to `wc` prefix rather than `woo` in all naming.

modules/ppcp-agentic-commerce/services.php
stracker-phil: Good point 👍 I'll not change it now, for the reason, that service names do not follow a namespace pattern, but are more semantic.

There is a difference between the "validator.*" instances (execution) and the main "validation.processor" (which orchestrates validator.* instances)

All ".validator." instances implement the `ValidatorInterface` which led me to also name the service "validator"

modules/ppcp-agentic-commerce/src/AgenticCommerceModule.php
stracker-phil: We generally use the `*Module` class to orchestrate and set up the module, so adding those configuration items here is the correct place.

Also: The hook is intentional (and using it here), as Validation logic must be extendable by third party code. Using our own hook in this setup logic ensures that the separation is working well + it allows other code to potentially add other validators _before_ our core validation rules (using a lower hook priority)

modules/ppcp-agentic-commerce/src/CartValidation/InventoryValidator.php
stracker-phil: Issues are returned to the AI agent. The module is only available for US merchants selling to US customers in this phase, so we have no need for translation yet.

I assume those errors will always stay English and maybe provide a localized version as additional argument, as the AI agent must be able to parse and react to those errors

modules/ppcp-agentic-commerce/src/CartValidation/ProductValidator.php
stracker-phil: Good observation, but this pattern doesn't extend to all validators. The current two current validators check cart items, but future validators will check different items (shipping address, billing data, customer details, ...), and don't iterate over items.

Similarly, the `has_validation_issue()` check is only relevant for those two validators, not a core validation behavior. 

However: If we later discover shared behavior across all validators, I would rather opt to add it into the `CartValidationProcessor` where validation is orchestrated, instead of creating an abstract class.

Short: The current validators have no real duplication but coincidentally similar implementations.

modules/ppcp-agentic-commerce/src/Helper/CartHelper.php
stracker-phil: These are pure utility functions with no side effects, no external dependencies, and no state. Taking a cart as input and returning a value - trivial to test without any mocking or DI setup.

Note: This class is intentionally a fully static class, as it's a convenience tool that can should be easily available to any code that needs to process a PayPalCart object. Changing it to an instantiable class forces us to drop it into all consuming classes via DI, which adds unneeded complexity to business logic and existing unit tests.

modules/ppcp-agentic-commerce/src/Helper/CartTransformer.php
stracker-phil: The empty line is intentional:
It separates the external dependencies for module internal imports

image

The `(string)` conversion is needed to pass psalm checks, as static code analysis sees that we pass an `int`value to `md5()`, but that method expects a string. So that cast is a CI requirement

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
stracker-phil: Price validation is part of PCP-5630, already documented.

The item in the comment is possibly covered by PCP-5634, but some requirements are still a bit vague, that's why I deiced to keep TODO comments in the code as a reminder for later, as we will touch this class a few times before finishing the module.

The idea of tracking TODOs in Jira is good, I've created a new issue for this: PCP-5640

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
stracker-phil: Type safety is a good point, agree 👍

But keep in mind that the CartHelper also returns a type-safe and guaranteed fully populated address element. The benefit of removing 6 repetitive null coalescing operators and centralizing that logic in a tested helper outweighs the marginal difference in runtime type safety.

Without the helper, every dev needs to remember which address props can be null and must handle that case. With the helper, we can just use the resulting array

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderManager.php
stracker-phil: I also agree - but I did not add them (looking at @danieldudzic 😉) and they will disappear in the final release. I kept them as they might be helpful during development at this stage

modules/ppcp-agentic-commerce/src/Helper/PayPalOrderBuilder.php
stracker-phil: Right. We'll resolve this in a future PR (PCP-5640)

It's basically semantics, as the woo currency must match the cart currency in all cases

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
stracker-phil: The name for this token is aligned with the PayPal lingo, to avoid confusion.

TBH I'm not 100% clear about the meaning of "ec" here, but the name was adopted based on those details:
2025-12-04 - ec_token name origin@2x

(and intentionally not "token" as that's ambiguous, and could refer to authentication or another token)

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
stracker-phil: having one class per endpoint is no SRP violation: A single class owns the full HTTP endpoint from end to end (route, permission, handling, response)

A consolidated endpoint manager with different handlers is very thin and creates a tight coupling (after all, the handlers still need to define the endpoint they handle, just delegating the register-call to the manager), and all handlers still need dependencies (logger, session handler, etc) - the things that can be extracted into such a handler are already defined in the abstract base class

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
stracker-phil: Thanks, will add a new PR to clean this up 🙌

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
stracker-phil: Unfortunately, this is as good as we can get in PHP 7.4 😢 

The methods without annotation have a union-type, which is only supported in PH P8.0+, so we can only document the type in the doc-block

modules/ppcp-agentic-commerce/src/Helper/AgenticCheckoutProcessor.php
stracker-phil: I've added the comment, as I think the $payer_data might have an unexpected format, but did not want to touch the PayPal API integration in this PR.

It's just a minor documentation effort to help debugging in the future. The code below the comment was already present before

modules/ppcp-agentic-commerce/src/Endpoint/AgenticRestEndpoint.php
stracker-phil: Thanks, already removed in a follow-up PR

modules/ppcp-agentic-commerce/src/Endpoint/AgenticRestEndpoint.php
stracker-phil: Yes, this module is in active development and we have no other means to understand what happens.

I'd even argue to keep it in production (at least for a while) as agentic integration is a communication between two servers where we have no other means to debug or investigate behavior using a browser, like we have for most other features

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
stracker-phil: haha, good catch 🙂 I was more talking to myself here. Will remove it in a cleanup PR

modules/ppcp-agentic-commerce/src/Endpoint/CheckoutEndpoint.php
stracker-phil: Thanks for pointing this out, I missed that change 🙌

PR #3908, opened 2025-12-04
Disable simulate-event endpoint for webhook verification (5621)

Dinamiko:
This PR disables calling simulate webhook and resubscribe webhooks on plugin upgrade.

### DIFF
--- a/modules/ppcp-webhooks/src/WebhookModule.php
+++ b/modules/ppcp-webhooks/src/WebhookModule.php
@@ -155,7 +155,12 @@ class WebhookModule implements ServiceModule, FactoryModule, ExtendingModule, Ex
 
 		add_action(
 			'woocommerce_paypal_payments_gateway_migrate',
-			static function ( $installed_plugin_version ) use ( $container ) {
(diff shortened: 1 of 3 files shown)


modules/ppcp-webhooks/src/WebhookModule.php
```diff
@@ -156,6 +156,9 @@ function () use ( $container ) {
 		add_action(
 			'woocommerce_paypal_payments_gateway_migrate',
 			static function () use ( $container ) {
+				// Disabled for 3.3.1 release.
```
AlexP11223: I would check the previous version to disable it only for upgrades from e.g. 3.3.0, so that we do not break things for someone updating from an old version that actually needs a resubscribe to refresh the used event list.

One way to do that would be to pass these values to the hooks:
https://github.com/woocommerce/woocommerce-paypal-payments/blob/95ca53343ad61921a8747ed39d46a644729e3ff3/woocommerce-paypal-payments.php#L161-L162

AlexP11223 (approved):

PR #3919, opened 2025-12-05
Fix PayPalCartResponse issues (5634)

stracker-phil:
### Description

This PR addresses some problems with invalid details returned by the REST endpoints.

**Critical problems:**
- The status `ACTIVE` does not exist and should not be returned
- The payment token is only expected when creating a new cart, not on get cart or replace cart
- The required cart ID was not present in all responses

Plus, some method did miss the mandatory status code `200`


Narek13 (approved):
Thanks. LGTM

PR #3929, opened 2025-12-09
[Refactor] Use PaymentMethodsEligibilityService for filtering Payment Methods in Settings (5561)

puntope:
### Description

- Create `PaymentMethodsEligibilityService` in order to check the eligibilities for the PaymentMethods
- Refactor the SettingsModule to use this class for filtering out deactivated Payment Methods
- Tests

### Steps to Test

1. Test with Mexico (see that Oxo is the only APM).  
2. Test with US -- Verify that Venmo, Apple Pay and Google Pay appears
3. Test with DE -- Verify Pay Upon Invoice appears
4. Test PWC method
5. Test Fastlane (aka AXO) appears when ADCD is enabled and is AXO eligible.

### DIFF
--- a/modules/ppcp-settings/services.php
+++ b/modules/ppcp-settings/services.php
@@ -11,9 +11,11 @@ namespace WooCommerce\PayPalCommerce\Settings;
 
 use WooCommerce\PayPalCommerce\ApiClient\Helper\Cache;
 use WooCommerce\PayPalCommerce\Applepay\ApplePayGateway;
+use WooCommerce\PayPalCommerce\Applepay\Assets\AppleProductStatus;
 use WooCommerce\PayPalCommerce\Axo\Gateway\AxoGateway;
 use WooCommerce\PayPalCommerce\Button\Helper\MessagesApply;
 use WooCommerce\PayPalCommerce\Googlepay\GooglePayGateway;
+use WooCommerce\PayPalCommerce\Googlepay\Helper\ApmProductStatus;
 use WooCommerce\PayPalCommerce\LocalAlternativePaymentMethods\BancontactGateway;
 use WooCommerce\PayPalCommerce\LocalAlternativePaymentMethods\BlikGateway;
 use WooCommerce\PayPalCommerce\LocalAlternativePaymentMethods\EPSGateway;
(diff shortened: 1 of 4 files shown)


modules/ppcp-settings/src/Service/PaymentMethodsEligibilityService.php
```diff
@@ -0,0 +1,167 @@
...
+	 *
+	 * @param string           $merchant_country PayPal Country (or Woo Country if not onboarded).
+	 * @param bool             $is_apm_eligible If alternative payment methods are eligible.
+	 * @param array            $merchant_capabilities Array of Merchant capabilities (true: enabled / false: disabled).
+	 * @param DCCProductStatus $dcc_product_status ACDC Seller status.
+	 * @param callable         $axo_eligible Whether Axo is eligible.
+	 * @param callable         $card_fields_eligible Whether Card Fields is eligible.
+	 * @param bool             $apple_pay_available Whether Apple Pay is available.
+	 * @param bool             $google_pay_available Whether Google Pay is available.
+	 */
```
Narek13: We no longer need to add constructor and property comments if their purpose is already obvious.
And also I think you have a redundant empty line (50-51)

Narek13 (approved):
Thanks, for working on this.
Left a minor suggestion, which doesn’t block approval.

modules/ppcp-settings/services.php
```diff
@@ -669,7 +671,24 @@
...
+
+		return new PaymentMethodsEligibilityService(
+			$container->get( 'api.shop.country' ),
+			$container->get( 'ppcp-local-apms.eligibility.check' ),
+			$container->get( 'settings.service.merchant_capabilities' ),
+			$container->get( 'wcgateway.helper.dcc-product-status' ),
+			$container->get( 'axo.eligibility.check' ),
+			$container->get( 'card-fields.eligibility.check' ),
+			$applepay_product_status->is_active() && $container->get( 'applepay.eligible' ),
+			$applepay_product_status->is_active() && $container->get( 'googlepay.eligible' ),
```
Dinamiko: This is wrong, it should be `$googlepay_product_status` instead.

modules/ppcp-settings/services.php
```diff
@@ -669,7 +671,24 @@
...
 	},
+	'settings.service.payment_methods_eligibilities'      => static function ( ContainerInterface $container ): PaymentMethodsEligibilityService {
+		$applepay_product_status = $container->get( 'applepay.apple-product-status' );
+		assert( $applepay_product_status instanceof AppleProductStatus );
 
+		$googlepay_product_status = $container->get( 'googlepay.helpers.apm-product-status' );
+		assert( $googlepay_product_status instanceof ApmProductStatus );
+
+		return new PaymentMethodsEligibilityService(
+			$container->get( 'api.shop.country' ),
```
Dinamiko: Here is just to confirm if `api.shop.country` service is correct as in SettingsModule.php it was using `api.merchant.country`: 
```
$merchant_country = $container->get('api.merchant.country');
```

Dinamiko (changes requested):

modules/ppcp-settings/src/Service/PaymentMethodsEligibilityService.php
puntope: Thanks @Narek13 I adapted the code based on your advide

modules/ppcp-settings/services.php
puntope: Thanks for catching this. Adjusted

modules/ppcp-settings/services.php
puntope: Thanks for catching this. Adjusted

Dinamiko (approved):

PR #3938, opened 2025-12-12
[Refactor] Create and use aliases for imports for cleaner code (5337)

puntope:
### Description

Use import aliases instead of relative paths (sometimes 4 or 5 levels)

### Steps to Test

1. Checkout this Branch 
2. Compile using `npm run build:modules` with no errors

⚠️  Another test step should be to run `npm run test:unit-js`however, test are failing in develop as well. That is a separate issue solved in https://github.com/woocommerce/woocommerce-paypal-payments/pull/3902


AlexP11223 (approved):
it seems to work but PHPStorm does not understand the imports... Though maybe it will be resolved after moving into a single webpack config

Dinamiko (approved):

PR #3942, opened 2025-12-15
ppcp-save-payment-methods SettingsProvider migration (5612)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-save-payment-methods` module.

### PR Changes
- Sets `save_paypal_and_venmo` via `SettingsModel`
- Replaces old settings methods with new provider methods:
    - `vault_enabled` → `save_paypal_and_venmo()`
    - `vault_enabled_dcc` → `save_card_details()`
    - `3d_secure_contingency` → `three_d_secure_enum()`
- Implements new method `three_d_secure_enum()` in `SettingsProvider` to return the 3D Secure setting as an API enum value


stracker-phil (approved):
While I think the migration logic is not ideal here, the PR is a clean transition from the legacy `Settings` to the new `SettingsProvider` class

Narek13 (approved):
LGTM

PR #3943, opened 2025-12-15
Agentic: Add Coupon Validation (5633)

danieldudzic:
### Description

Implements coupon validation with enhanced context and resolution options for the agent.

###  Technical Details

- Filters: Allow plugins to customize **validation messages**, **resolution options**, and **alternative coupon suggestions** for the agent.
- Case-insensitive coupon matching via `wc_sanitize_coupon_code()`
- Uses `WC_Discounts` for accurate discount calculation
- Caps discounts to prevent $0 orders


modules/ppcp-agentic-commerce/services.php
```diff
@@ -145,27 +149,57 @@
...
+			$c->get( 'agentic.validator.coupon.discount-calculator' ),
+			$c->get( 'agentic.validator.coupon.resolution-builder' )
+		);
+	},
 
-	// REST endpoints.
-	'agentic.response.factory'            => static function (): ResponseFactory {
-		return new ResponseFactory();
+	// Response services.
+	'agentic.response.applied-coupons-builder'     => static function ( ContainerInterface $c ): \WooCommerce\PayPalCommerce\AgenticCommerce\CartValidation\CouponValidator\AppliedCouponsBuilder {
```
stracker-phil: Suggestion: Use the project standard of importing classes at the top, and not using FQCN in the module code

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/ContextBuilder.php
```diff
@@ -0,0 +1,326 @@
...
+	 * @var DiscountCalculator
+	 */
+	private DiscountCalculator $discount_calculator;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param ProductManager     $product_manager Product manager instance.
+	 * @param DiscountCalculator $discount_calculator Discount calculator instance.
+	 */
```
stracker-phil: nitpick: IMO those block-comments do not add any value, that's why most classes in the agentic commerce module do not use them. If adding them is just a convention, I'd reconsider to remove comment noise.

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/ContextBuilder.php
```diff
@@ -0,0 +1,326 @@
...
+	 *
+	 * @param string         $issue_type The issue type.
+	 * @param string         $code The coupon code.
+	 * @param PayPalCart     $cart The cart context.
+	 * @param WC_Coupon|null $wc_coupon The WC coupon object.
+	 * @param array          $builders Array of builder names to call.
+	 * @param array          $extra Extra context data.
+	 * @return array The built context.
+	 */
+	public function build(
```
stracker-phil: Nitpick: instead of generic verbs (build, get, update, process, ...) it's more helpful to understand the code behavior by using more expressive names, like "verb_what()" - i.e. `build_coupon_context()`

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/CouponValidator.php
```diff
@@ -0,0 +1,536 @@
...
+		'not applicable'                              => 'COUPON_NOT_APPLICABLE',
+
+		// Email restrictions.
+		'not yours'                                   => 'COUPON_EMAIL_RESTRICTED',
+		'does not belong'                             => 'COUPON_EMAIL_RESTRICTED',
+
+		// Already applied.
+		'already applied'                             => 'COUPON_ALREADY_APPLIED',
+		'already been applied'                        => 'COUPON_ALREADY_APPLIED',
+	);
```
stracker-phil: Blocker: This likely works in the initial version (US shops with US customers), but error messages are localized by WooCommerce, making this pattern matching only reliable for english shops

https://woocommerce.github.io/code-reference/files/woocommerce-includes-class-wc-coupon.html#source-view.1060 →  `esc_html__(...)`

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/CouponValidator.php
```diff
@@ -0,0 +1,536 @@
...
+		$normalized_code = wc_sanitize_coupon_code( $code );
+
+		$wc_coupon = new WC_Coupon( $normalized_code );
+
+		if ( ! $wc_coupon->get_id() ) {
+			return $this->create_issue( 'COUPON_NOT_EXIST', $code, $field, $cart, null );
+		}
+
+		// Run WC validation via WC_Discounts.
+		$result = $discounts->is_coupon_valid( $wc_coupon );
```
stracker-phil, line 305: Suggestion: Instead of using the localized response value, we could access the [numeric error code](https://woocommerce.github.io/code-reference/files/woocommerce-includes-class-wc-coupon.html#source-view.58) via the filter `woocommerce_coupon_error`

Idea
```php
$coupon_error = 0;
$fn_error_code = static fn($error_message, $error_code) => $coupon_error = $error_code;
add_filter('woocommerce_coupon_error', $fn_error_code, 10, 2);
$is_valid = $discounts->is_coupon_valid( $wc_coupon );
remove_filter('woocommerce_coupon_error', $fn_error_code);

if ( true !== $is_valid ) {
	// Inspect the $coupon_error value.
}
```

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/CouponValidator.php
```diff
@@ -0,0 +1,536 @@
...
+		return apply_filters(
+			'woocommerce_paypal_payments_agentic_commerce_coupon_validation_user_message',
+			$message,
+			$issue_type,
+			$code,
+			$wc_coupon,
+			$cart,
+			$context
+		);
+	}
```
stracker-phil: Suggestion: Since this method is private and only used a single time, I'd move that filter inline into the `create_issue()` method and only extract it once it's reused or becoming conditional, etc.

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/DiscountCalculator.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * This method calculates the theoretical discount amount without validation.
+	 * Used for showing discount comparisons in stacking conflicts, where coupons
+	 * may not be individually valid but we still want to show their discount amounts.
+	 *
+	 * @param WC_Coupon  $wc_coupon The WC coupon object.
+	 * @param PayPalCart $cart The cart context.
+	 * @return string The discount amount formatted to 2 decimals, or '0.00' if calculation fails.
+	 */
+	public function calculate( WC_Coupon $wc_coupon, PayPalCart $cart ): string {
```
stracker-phil: Nitpick: Same as the "build()" method, I'd consider a more descriptive name that tells the reader what is calculated - i.e. "calculate_discount_amount()"

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/ContextBuilder.php
```diff
@@ -0,0 +1,326 @@
...
+
+		$subtotal = CartHelper::cart_item_total( $cart );
+		$minimum  = (float) $wc_coupon->get_minimum_amount();
+		$shortage = max( 0, $minimum - $subtotal );
+		$currency = CartHelper::currency( $cart, get_woocommerce_currency() );
+
+		return array(
+			'minimum_required' => wc_format_decimal( $minimum, 2 ),
+			'current_subtotal' => wc_format_decimal( $subtotal, 2 ),
+			'shortage_amount'  => wc_format_decimal( $shortage, 2 ),
```
stracker-phil: Suggestion: Instead of using the `wc_format_decimal()` method, I would directly use `number_format( $discount_value, 2, '.', '' )`

Reason:
- We do not use the `wc_format_decimal()` anywhere else in the plugin
- The WC function is largely built to re-format string values, while we have guaranteed numeric type here
- `number_format()` generates the format that is expected by the agent, the WC function could introduce a filter that interferes with the agent

modules/ppcp-agentic-commerce/src/Response/NewCartResponse.php
stracker-phil, line 1: Attention: This change is reverting a recent change from the upstream epic branch!

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/CouponValidator.php
danieldudzic: Good catch! I've updated the implementation to use numeric error codes instead of pattern matching on localized strings.

 The `extract_numeric_error_code()` method now pulls the numeric `WC_Coupon` error constants (100-116) directly from the `WP_Error` data structure, and I've added a mapping array that converts these to PayPal issue types.

modules/ppcp-agentic-commerce/services.php
```diff
@@ -153,37 +158,66 @@
...
+		);
+	},
+	'agentic.validator.coupon.context-builder'     => static function ( ContainerInterface $c ): ContextBuilder {
+		return new ContextBuilder(
+			$c->get( 'agentic.helper.product-manager' ),
+			$c->get( 'agentic.validator.coupon.discount-calculator' )
+		);
+	},
+	'agentic.validator.coupon.resolution-builder'  => static function (): ResolutionBuilder {
+		return new ResolutionBuilder();
```
stracker-phil: **Minor Suggestion**:

Rename the class to `CouponResolutionBuilder` to avoid confusion in case we have more than one Resultion builder class in the future

modules/ppcp-agentic-commerce/services.php
```diff
@@ -153,37 +158,66 @@
...
+	'agentic.validator.currency'                   => static function (): CurrencyValidator {
 		return new CurrencyValidator();
 	},
+	'agentic.validator.coupon.discount-calculator' => static function ( ContainerInterface $c ): DiscountCalculator {
+		return new DiscountCalculator(
+			$c->get( 'agentic.helper.product-manager' )
+		);
+	},
+	'agentic.validator.coupon.context-builder'     => static function ( ContainerInterface $c ): ContextBuilder {
+		return new ContextBuilder(
```
stracker-phil: **Minor Suggestion**:

Rename the class to `CouponContextBuilder` to avoid confusion in case we have more than one such context builder class in the future

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/AppliedCouponsBuilder.php
```diff
@@ -0,0 +1,130 @@
...
+
+		// Only show applied coupons if validation passed.
+		if ( $validation_status !== 'VALID' ) {
+			return array();
+		}
+
+		// WooCommerce not available - can't build coupon data.
+		if ( ! class_exists( 'WC_Coupon' ) ) {
+			return array();
+		}
```
stracker-phil: **Suggestion**:

Rearrange the early return checks:
1. valid_status !== VALID (the cheapest check)
2. class_exists (also very cheap)
3. array-filter check (array processing as last step)

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/ContextBuilder.php
```diff
@@ -0,0 +1,325 @@
...
+		$context = array(
+			'specific_issue' => $issue_type,
+			'coupon_code'    => $code,
+		);
+
+		foreach ( $builders as $builder ) {
+			$method = "build_{$builder}";
+			if ( method_exists( $this, $method ) ) {
+				$context = array_merge( $context, $this->$method( $code, $cart, $wc_coupon, array_merge( $context, $extra ) ) );
+			}
```
stracker-phil: **Architectural Suggestion**:

While this is a smart and short solution, it makes the code a bit more difficult to maintain:
The `$builder` array is provided by a different class, but it must know the exact method names. If we change one of the internal method names, we must update two classes. Also, if we don't do this, the failure is silent - no error anywhere.

I would lean more towards explicit branching here using a switch, possibly via a `call_builder()` helper that directly returns the built context:
```php
public function build_coupon_context(...) {
    foreach ( $builders as $builder ) {
        $built_context = $this->call_builder( $builder, $cart, ... );
        $context = array_merge( $context, $built_context );
    }
}

private function call_builder( string $builder, ... ): array {
    switch ( $builder ) {
        case 'alternatives':
            // note: no $wc_coupon in the function call
            return $this->build_alternatives( $code, $cart, $extra );
        case 'expiration':
            // We can refactor the "build_expiration" method without impacting the public method's contract
            return $this->build_expiration( code, $cart, $wc_coupon, $extra );
        // ...
        default:
          // log error or do something, as this should never happen and indicates the calling function has wrong/outdated details.
    };
}

```

modules/ppcp-agentic-commerce/src/CartValidation/CouponValidator/CouponContextBuilder.php
```diff
@@ -0,0 +1,325 @@
...
+	 * Gets alternative coupon suggestions via filter.
+	 *
+	 * @param string     $failed_code The coupon code that failed.
+	 * @param string     $reason The failure reason.
+	 * @param PayPalCart $cart The cart context.
+	 * @return array Array of alternative coupon codes.
+	 */
+	private function get_alternative_coupons( string $failed_code, string $reason, PayPalCart $cart ): array {
+		return apply_filters(
+			'woocommerce_paypal_payments_agentic_commerce_suggested_alternative_coupons',
```
stracker-phil, line 318: Nice 👍

modules/ppcp-agentic-commerce/src/Response/ResponseFactory.php
```diff
@@ -11,36 +11,98 @@
...
+	 * @param PayPalCart $cart The cart object.
+	 * @param string     $cart_id The cart ID.
+	 * @param string     $token The payment token.
+	 * @return NewCartResponse The response object.
+	 */
 	public function new_cart( PayPalCart $cart, string $cart_id, string $token ): NewCartResponse {
-		$wc_cart = $this->build_wc_cart_or_null( $cart );
+		$wc_cart         = $this->build_wc_cart_or_null( $cart );
+		$applied_coupons = $this->build_applied_coupons( $cart );
+		return new NewCartResponse( $cart, $cart_id, $token, 'CREATED', $applied_coupons, $wc_cart );
```
stracker-phil: **Important**:

The 4 `Response/*` classes still revert some changes of the epic branch
- there should be no `$status` in the constructor anymore
- The first three arguments of all classes should be `$cart, $cart_id, $wc_order` (as those 3 items are identical in all classes)
- We do not have an `active_cart()` factory/state -- see `CartResponse::ALLOWED_STATUS` values

stracker-phil (approved):

PR #3950, opened 2025-12-16
ppcp-vaulting SettingsProvider migration (5616)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-vaulting` module.

### PR Changes
- Removes `woocommerce_paypal_payments_gateway_migrate_on_update` hook because it triggers vault v2 payment tokens migration on each plugin upgrade, this migration was needed before vault v3 was implemented. Migration logic is still there and can be run via: `do_action('pcp_migrate_payment_tokens')`.
  - Note: the callback also updates `vault_enabled_dcc`, this was also needed before vault v3 but it's not needed anymore, that logic should belong to the new settings UI module.
- Replaces legacy `intent` setting with the new `authorize_only` from SettingsProvider.

### DIFF
--- a/modules/ppcp-vaulting/services.php
+++ b/modules/ppcp-vaulting/services.php
@@ -36,7 +36,7 @@ return array(
 			$container->get( 'api.endpoint.order' ),
 			$container->get( 'settings.environment' ),
 			$container->get( 'wcgateway.processor.authorized-payments' ),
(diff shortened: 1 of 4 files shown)


puntope (approved):
LGTM

stracker-phil (approved):

PR #3962, opened 2025-12-19
Fix merge of package.json commands and readme

AlexP11223:
One of the merge commits somehow reverted some changes from #3932


puntope (approved):

puntope:
@AlexP11223 Why we are removing ddev from the unit tests? The idea is to run the PHP set in DDEV and not in your enviromment

PR #3963, opened 2025-12-19
Use Modularity Properties instead of Dhii

AlexP11223:
Replacing the Dhii versions and plugin libraries with Properties from the Syde Modularity to reduce complexity and remove an invalid semver check.


modules/ppcp-wc-gateway/src/Settings/WcInboxNotes/InboxNoteRegistrar.php
```diff
@@ -20,11 +19,11 @@ class InboxNoteRegistrar {
 	 * @var InboxNoteInterface[]
 	 */
 	protected array $inbox_notes;
-	protected PluginInterface $plugin;
+	protected string $plugin_base_name;
 
-	public function __construct( array $inbox_notes, PluginInterface $plugin ) {
-		$this->inbox_notes = $inbox_notes;
-		$this->plugin      = $plugin;
+	public function __construct( array $inbox_notes, $plugin_base_name ) {
```
Dinamiko: Why not adding type `string` to `$plugin_base_name` here?

woocommerce-paypal-payments.php
```diff
@@ -158,7 +158,7 @@ function () {
 			add_action(
 				'init',
 				function () {
-					$current_plugin_version   = (string) PPCP::container()->get( 'ppcp.plugin' )->getVersion();
+					$current_plugin_version   = (string) PPCP::container()->get( 'ppcp.plugin-version' );
```
Dinamiko: Casting to string may be redundant here as `ppcp.plugin-version` service already returns a string.

Dinamiko (changes requested):

modules/ppcp-wc-gateway/src/Settings/WcInboxNotes/InboxNoteRegistrar.php
AlexP11223: ah, yeah

woocommerce-paypal-payments.php
AlexP11223: removed

stracker-phil (approved):
LGTM

Dinamiko (approved):

PR #3974, opened 2026-01-02
ppcp-session SettingsProvider migration (5613)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-session` module.

### PR Changes
- New method `PaymentSettings::get_method_title()` to retrieve the title of a WooCommerce payment gateway
- New method `SettingsProvider::paypal_gateway_title()` to return the PayPal gateway title
- Updates service from `wcgateway.settings` to use new `settings.settings-provider` in `CancelView.php`

### DIFF
--- a/modules/ppcp-settings/src/Data/PaymentSettings.php
+++ b/modules/ppcp-settings/src/Data/PaymentSettings.php
@@ -152,16 +152,17 @@ class PaymentSettings extends AbstractDataModel {
 	 * Gets the payment method title.
 	 *
 	 * @param string $method_id ID of the payment method.
+	 * @param string $default_title Default title to return if method not found.
 	 * @return string The method title, or an empty string if not found.
 	 */
-	public function get_method_title( string $method_id ): string {
+	public function get_method_title( string $method_id, string $default_title = '' ): string {
 		$gateway = $this->get_gateway( $method_id );
 
 		if ( $gateway ) {
 			return $gateway->title;
 		}
 
-		return '';
+		return $default_title;
 	}
 
 	/**
--- a/modules/ppcp-settings/src/Data/PaymentSettings.php
+++ b/modules/ppcp-settings/src/Data/PaymentSettings.php
@@ -149,6 +149,22 @@ class PaymentSettings extends AbstractDataModel {
 	}
 
 	/**
+	 * Gets the payment method title.
+	 *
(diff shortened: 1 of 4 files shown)


modules/ppcp-settings/src/Data/SettingsProvider.php
```diff
@@ -267,6 +268,21 @@ public function paylater_enabled(): bool {
...
+	/**
+	 * Gets the PayPal gateway title as configured in WooCommerce settings.
+	 *
+	 * @return string The gateway title, defaults to 'PayPal' if not set.
+	 */
+	public function paypal_gateway_title(): string {
+		$title = $this->payment_settings->get_method_title( PayPalGateway::ID );
+
+		if ( empty( $title ) ) {
+			return __( 'PayPal', 'woocommerce-paypal-payments' );
```
stracker-phil: Why do we need to translate the brand name "PayPal"? I know it was translatable before, but it makes no sense IMO; I'd switch to simply return:

```php
return 'PayPal';
```

modules/ppcp-settings/src/Data/PaymentSettings.php
```diff
@@ -148,6 +148,22 @@ public function set_method_description( string $method_id, string $description )
...
+	 * @return string The method title, or an empty string if not found.
+	 */
+	public function get_method_title( string $method_id ): string {
+		$gateway = $this->get_gateway( $method_id );
+
+		if ( $gateway ) {
+			return $gateway->title;
+		}
+
+		return '';
```
stracker-phil: suggestion: instead of returning an empty string, this could return a `$default_title` argument:

```php
public function get_method_title( string $method_id, string $default_title = '' ): string {
	$gateway = $this->get_gateway( $method_id );

	if ( $gateway ) {
		return $gateway->title;
	}

	return $default_title;
}
```

Then the SettingsProvider implementation becomes a 1-liner:

```php
public function paypal_gateway_title(): string {
	return $this->payment_settings->get_method_title( PayPalGateway::ID, 'PayPal' );
}
```

stracker-phil (approved):
The changes look correct and I approve.

I've made some comments that are more optimization/refactoring related, that are easy to implement now (but not strictly needed)

PR #3975, opened 2026-01-02
ppcp-applepay SettingsProvider migration (5599)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-applepay` module.

### PR Changes

> Unit tests that depends on base class `ProductStatus::clear_state` would fail until all modules are migrated, it is OK the merge PR with these tests failing.

ppcp-settings module:
- Added 8 Apple Pay settings to `PaymentSettings` with getters/setters: e.g., applepay_button_enabled
- Added 8 delegation methods to `SettingsProvider`: e.g., applepay_button_enabled                                                                                                                                                                       

ppcp-applepay module:
- Updated service injections from `wcgateway.settings` to use `settings.provider` and `settings.data.payment`
- Updated all classes to use `SettingsProvider` for reads and `PaymentSettings` for writes
- Removed legacy `Settings` import in `extensions.php`, replaced constant with literal string

ppcp-api-client module:
- Updated `ProductStatus` base class to use `PaymentSettings` instead of legacy `Settings`


modules/ppcp-settings/src/Data/PaymentSettings.php
```diff
@@ -275,4 +283,156 @@ private function get_gateway( string $method_id ): ?WC_Payment_Gateway {
...
+
+	/**
+	 * Set Apple Pay validated.
+	 *
+	 * @param bool $value The value.
+	 * @return void
+	 */
+	public function set_applepay_validated( bool $value ): void {
+		$this->data['applepay_validated'] = $value;
+	}
```
stracker-phil, line 339: I would prefer to have a dedicated class to hold the product status result, instead of putting it into the settings model again.

modules/ppcp-applepay/services.php
```diff
@@ -72,19 +70,19 @@
 	},
 
 	'applepay.has_validated'                   => static function ( ContainerInterface $container ): bool {
-		$settings = $container->get( 'wcgateway.settings' );
-		return $settings->has( 'applepay_validated' );
+		$settings = $container->get( 'settings.settings-provider' );
+		return $settings->applepay_validated() !== false;
```
stracker-phil: This condition is wrong.

- Original code checked if the "setting" (cache) is defined, i.e. non-null
- New condition checks if the validation result is not false

More correct version would be:

```php
$applepay_status = $container->get();
assert( $applepay_status instanceof ProductStatus );
return $applepay_status->check_local_state(); // ← returns the cached value or null if not defined
```

Some `check_local_state()` implementations start with a WP filter to override the result, which needs to be handled, either:
1. Create a new method in the base class without a filter (better, IMO)
2. Add an argument to `bool $skip_filter = false` to the method

Sample for option 1:

```php
class ProductStatus {
    protected $result_cache; // renamed from $settings
    
    public function __construct( $result_cache ) {
        $this->result_cache = $result_cache;
    }
    
    public function get_cached_result(): ?bool {
        $cached_value = $this->result_cache->get( static::KEY ); // Note: $result_cache is a new class, no settings-model

        if ( $cached_value ) {
			return wc_string_to_bool( $cached_value );
		}

		return null;
    }
}

// In AppleProductStatus:

class AppleProductStatus extends ProductStatus {
    public const KEY = 'products_apple_enabled'; // previously "SETTINGS_KEY"
    
    // Remove the "private $settings" prop, as it was moved to the base class → protected $result_cache

    // Remove the constructor, the base class receives all props
    
    protected function check_local_state(): ?bool {
        $status_override = apply_filters( 'woocommerce_paypal_payments_google_pay_product_status', null );
        if ( null !== $status_override ) {
            return $status_override;
        }
        
        return $this->get_cached_result(); // Use the cache response from the base class
    }
}
```

modules/ppcp-applepay/services.php
```diff
@@ -72,19 +70,19 @@
...
-		return $settings->has( 'applepay_validated' ) ? $settings->get( 'applepay_validated' ) === true : false;
+		$settings = $container->get( 'settings.settings-provider' );
+		return $settings->applepay_validated();
 	},
 
 	'applepay.apple-product-status'            => SingletonDecorator::make(
 		static function ( ContainerInterface $container ): AppleProductStatus {
 			return new AppleProductStatus(
-				$container->get( 'wcgateway.settings' ),
+				$container->get( 'settings.data.payment' ),
```
stracker-phil: Replace this with the new `ProductStatusCacheResult` class

stracker-phil (changes requested):
Settings refactoring is solid 👍

However, the ProductStatus changes could be cleaned up to avoid future problems:
I recommend adding a new class to cache the product status result, instead of putting it into the settings model again.

**Background:** We used the Settings class in the past, because it was easy to add random items to it, and it was already available in most modules. But a validation result is not a setting value; we "only" cache it to avoid repeat API calls, and not to make the result available to other modules.

Risks of using `PaymentSettings` for this:
- The model mixes actual settings with API response caching
- Eventually someone will use the `PaymentSettings` class to check if a feature is active, skipping the `ProductStatus` class

My suggestion:
- Create a new `ProductStatusResultCache` class for this purpose
- In this class we still use string locators - `get( 'myKey' )` and not `get_myKey()` - to keep it flexible
- Rename the property from "$settings" to "$result_cache" to clarify that it's a cached value

Sample:

```php
class ProductStatusResultCache {
	private const CACHE_KEY = 'woocommerce-ppcp-cache-product-status';

	private bool $loaded = false;
	private array $cache = array();

	public function get( string $key ): string {
		$this->load();

		return $this->cache[ $key ] ?? '';
	}

	public function set( string $key, string $value ): void {
		$this->cache[ $key ] = $value;
		$this->persist();
	}

	public function clear( string $key ): void {
		$this->set( $key, '' );
	}

	private function load(): void {
		if ( $this->loaded ) {
			return;
		}
		$this->loaded = true;

		$data = get_transient( self::CACHE_KEY );

		if ( is_array( $data ) ) {
			$this->cache = array_map(
				static fn( $value ) =>(string) $value,
				$data
			);
		}
	}

	private function persist(): void {
		set_transient( self::CACHE_KEY, $this->cache );
	}
}
```

modules/ppcp-applepay/src/Assets/ApplePayButton.php
```diff
@@ -27,11 +27,18 @@ class ApplePayButton implements ButtonInterface {
...
 
 	/**
-	 * The settings.
+	 * The settings provider.
 	 *
-	 * @var Settings
+	 * @var SettingsProvider
 	 */
-	private $settings;
+	private $settings_provider;
```
Narek13:
```suggestion
	private SettingsProvider $settings_provider;
```

And then you can remove the unnecessary comment with `@var`

modules/ppcp-applepay/src/Assets/ApplePayButton.php
```diff
@@ -27,11 +27,18 @@ class ApplePayButton implements ButtonInterface {
...
+
+	/**
+	 * The payment settings.
+	 *
+	 * @var PaymentSettings
+	 */
+	private $payment_settings;
```
Narek13: Same here

modules/ppcp-applepay/src/Assets/AppleProductStatus.php
```diff
@@ -30,20 +30,20 @@ class AppleProductStatus extends ProductStatus {
 	/**
 	 * The settings.
 	 *
-	 * @var Settings
+	 * @var PaymentSettings
 	 */
-	private Settings $settings;
+	private PaymentSettings $settings;
```
Narek13: Can we please remove the unnecessary comments while refactoring?

modules/ppcp-applepay/src/Assets/DataToAppleButtonScripts.php
```diff
@@ -28,17 +27,17 @@ class DataToAppleButtonScripts {
 	/**
 	 * The settings.
 	 *
-	 * @var Settings
+	 * @var SettingsProvider
 	 */
 	private $settings;
```
Narek13: Same here

modules/ppcp-settings/src/Data/PaymentSettings.php
```diff
@@ -292,4 +300,156 @@ private function get_gateway( string $method_id ): ?WC_Payment_Gateway {
...
+	 */
+	public function set_applepay_button_type( string $value ): void {
+		$this->data['applepay_button_type'] = $value;
+	}
+
+	/**
+	 * Get Apple Pay button color.
+	 *
+	 * @return string
+	 */
```
Narek13, line 364: Sorry to jump in again, but the method already speaks for itself, and to me it doesn’t make much sense to repeat the same information in the comments.

modules/ppcp-applepay/extensions.php
```diff
@@ -86,7 +85,7 @@
 						State::STATE_ONBOARDED,
 					),
 					'requirements' => array(),
-					'gateway'      => Settings::CONNECTION_TAB_ID,
+					'gateway'      => 'ppcp-connection',
```
Narek13, line 90: Does it make sense to create a separate interface to store these IDs instead of hardcoding them?

Narek13 (commented):
Thanks for working on this. Just left a minor suggestions and doughts.

Dinamiko:
@stracker-phil good catch, it makes sense to not store the product status cache into PaymentSettings, I’ve replaced `PaymentSettings` with the existing `Cache` class for storing product status results. Also fixed logic bug in `applepay.has_validated` service and code style fixes. cc @Narek13

modules/ppcp-api-client/src/Helper/ProductStatus.php
```diff
@@ -193,13 +193,13 @@ public function has_request_failure(): bool {
 	/**
 	 * Clears the persisted result to force a recheck.
 	 *
-	 * Accepts a Settings object to don't override other sequential settings that are being updated
+	 * Accepts a PaymentSettings object to don't override other sequential settings that are being updated
```
Narek13:
```suggestion
	 * Accepts a PaymentSettings object to not override other sequential settings that are being updated
```

Narek13 (approved):
Thanks for making all the changes

modules/ppcp-api-client/src/Helper/ProductStatus.php
stracker-phil: IGNORE: I'd ignore this comment, as the `ProductStatus` class is refactored in #4000 and this param is removed

modules/ppcp-applepay/extensions.php
stracker-phil: IGNORE: This file/code will be dropped later in the refactoring process. For now, I'd accept the code as-is, since it correctly removes the `Settings` dependency

stracker-phil (approved):

PR #3978, opened 2026-01-04
fix: defer currency check to prevent early translation loading (5709)

mondalaci:
WooCommerce PayPal Payments was calling get_woocommerce_currency() during plugins_loaded hook, which triggered Aelia Currency Switcher's filter before WordPress translations were ready. This caused PHP notices upon every page load:

  "Translation loading for the wc-aelia-foundation-classes domain was
   triggered too early"

Root cause: SettingsModule::run() called apply_branded_only_limitations() directly, which fetches GeneralSettings requiring the current currency. When Aelia hooks into woocommerce_currency filter, it calls __() for translations before the init action has fired.

Fix: Defer apply_branded_only_limitations() to the init action (priority 1) so currency detection happens after translations are available.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -198,7 +198,7 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 		assert( $loading_screen_service instanceof LoadingScreenService );
 		$loading_screen_service->register();
 
-		$this->apply_branded_only_limitations( $container );
+		add_action( 'init', fn() => $this->apply_branded_only_limitations( $container ), 1 );
 
 		/**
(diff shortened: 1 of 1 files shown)


daigo75:
@mondalaci Thanks for the contribution. For the benefit of the WooCommerce team, the `woocommerce_currency` filter is used to determine the active currency at a given time, and it's linked to the initialisation of WooCommerce. Indeed, if function `get_woocommerce_currency()` is called before the `init` event, that would also trigger the filter. The logic used to select the filter does call the localisation function `__()` at some point, and that would raise the notification. 

Moving the call to `get_woocommerce_currency()` to the `init` event should be sufficient to prevent this glitch.

Dinamiko (approved):
@mondalaci it looks good to me, I'll merge it, thanks!

PR #3979, opened 2026-01-05
ACDC payment method disappears from Checkout after updating the plugin (5693)

Dinamiko:
After updating the plugin, some users experience ACDC payment method and settings tab disappearing. This may impact legacy UI and users who migrated from legacy UI to new UI. Fresh setups on new UI appear unaffected.

### Possible Cause
Incomplete cache invalidation during plugin updates.

### PR Changes
This PR ensures more capability-related caches are properly invalidated during updates, including:
- Clear `RefreshFeatureStatusEndpoint` throttle cache to allow immediate feature refresh
- Clear `FailureRegistry` to reset API failure tracking
- Set `products_dcc_enabled` and `products_pui_enabled` to empty string ('') instead of false to force fresh capability checks from PayPal API

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -22,6 +22,7 @@ use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
 use WooCommerce\PayPalCommerce\ApiClient\Helper\ReferenceTransactionStatus;
 use WooCommerce\PayPalCommerce\ApiClient\Helper\Cache;
 use WooCommerce\PayPalCommerce\ApiClient\Helper\DccApplies;
+use WooCommerce\PayPalCommerce\ApiClient\Helper\FailureRegistry;
 use WooCommerce\PayPalCommerce\LocalAlternativePaymentMethods\LocalApmProductStatus;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExecutableModule;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExtendingModule;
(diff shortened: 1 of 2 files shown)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -337,12 +345,11 @@ static function () use ( $c ) {
 				$pui_status_cache->delete( PayUponInvoiceProductStatus::PUI_STATUS_CACHE_KEY );
 
 				$settings = $c->get( 'wcgateway.settings' );
-				$settings->set( 'products_dcc_enabled', false );
-				$settings->set( 'products_pui_enabled', false );
+				$settings->set( 'products_dcc_enabled', '' );
```
danieldudzic, line 348: Looks good, perhaps you can add a comment regarding this.
```suggestion
				// Empty string triggers fresh capability check.
				$settings->set( 'products_dcc_enabled', '' );
```

danieldudzic (approved):
Nice work!

stracker-phil (approved):
Looking good, nice cleanup!

modules/ppcp-wc-gateway/src/WCGatewayModule.php
stracker-phil, line 348: An alternative would be to use `null` which seems to be the default to reset those states

- LoginSellerEndpoint, line 146
- OnboardingRESTController, line 241
- SettingsListener, line 488

(null triggers the `has()` method to return false)

```php
$settings->set( 'products_dcc_enabled', null );
$settings->set( 'products_pui_enabled', null );
```

stracker-phil (approved):

modules/ppcp-wc-gateway/src/WCGatewayModule.php
stracker-phil: Another idea to make the logic more transparent could be using the actual `clear()` method of the ProductStatus classes:

```php
$dcc_status = $c->get('wcgateway.helper.dcc-product-status');
assert( $dcc_status instanceof DCCProductStatus );
$dcc_status->clear();

$pui_status = $c->get('wcgateway.pay-upon-invoice-product-status');
assert( $pui_status instanceof PayUponInvoiceProductStatus );
$pui_status->clear();
```

Dinamiko:
Fixed in #3983

PR #3982, opened 2026-01-07
Centralize JS/CSS building

AlexP11223:
Now all assets are compiled in a single place, using the root `package.json`/`webpack.config.js`, instead of duplicating that in every module. 

The Webpack config has a list like this at the top (the paths are relative to the `resources` dirs of the modules):

```js
{
	'ppcp-some-module': [
		'js/index.js',
		'css/styles.scss',
	],
...
```
and the compiled files are created in the root `assets/` dir with names like `ppcp-some-module-js-index.js` (`modulename-dirs-filename`).

And for using the assets in the PHP code there is the `AssetGetter`. Each module should create its own instance like this:

```php
	'some-module.asset_getter'          => static function ( ContainerInterface $container ): AssetGetter {
		$factory = $container->get( 'assets.asset_getter_factory' );
		assert( $factory instanceof AssetGetterFactory );

		return $factory->for_module( 'ppcp-some-module' );
	},
```

Available methods:

- `$asset_getter->get_asset_url( 'index.js' )` - the most commonly used, for getting the asset URL that can be put on the page etc.
- `$asset_getter->get_asset_php_path( 'index.js' )` - returns the path to the asset's `.asset.php` file (``ppcp-some-module-js-index.asset.php``) which you can `require` to retrieve the metadata liie `dependencies`.
- `$asset_getter->get_static_asset_url( 'images/icon.svg' )` - for the URL of static assets like images, stored in the `assets` dirs of the modules.

---

The dir name (`js-` or `css-`) is included in the compiled asset name because some modules have JS and CSS files with the same names, and Webpack generates `name.js` for CSS files, and also the `.asset.php` (`@wordpress/scripts` puts all files in the same dir and iirc it's not easy to change). I am not sure why this .js file is created, as I understand the normal way to avoid that is to either remove the CSS entrypoints and import them from JS, or possibly create groups for CSS and JS in the webpack entries list, but it does not seem like an important thing to solve. So for now the simplest workaround avoiding any confusing conflicts is to keep the dir name.

In the future it should be easy to change the generated names/paths if we want, since everything is handled by a single config + `AssetGetter`.

One more thing to note is that `@wordpress/scripts` has some special [conventions](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#importing-styles-in-javascript) for some names like `style.scss`, for some reason it adds the `style-` prefix (`style-ppcp-some-module-css-style.css`). I do not fully understand the purpose of it, and it seems we were not really following it anyway, so I just renamed `ppcp-settings` `style.scss` to `styles.scss`.


modules/ppcp-applepay/resources/css/styles.scss
```diff
@@ -26,7 +26,7 @@
 
 	&.ppcp-preview-button.ppcp-button-dummy {
 		/* URL must specify the correct module-folder! */
-		--apm-button-dummy-background: url(../../../ppcp-applepay/assets/images/applepay.png);
+		--apm-button-dummy-background: url(../../../ppcp-applepay/assets/images/applepay.svg);
```
Dinamiko, line 29: Now that css is compiled to the root `assets/` directory,  I think this path should likely be `../modules/ppcp-applepay/assets/images/applepay.svg`

modules/ppcp-googlepay/resources/css/styles.scss
```diff
@@ -15,7 +15,7 @@
 
 	&.ppcp-preview-button.ppcp-button-dummy {
 		/* URL must specify the correct module-folder! */
-		--apm-button-dummy-background: url(../../../ppcp-googlepay/assets/images/googlepay.png);
+		--apm-button-dummy-background: url(../../../ppcp-googlepay/assets/images/googlepay.svg);
```
Dinamiko, line 18: Same as comment in applepay css.

Dinamiko (approved):
PR approved, I just added a comment for you to check if those css paths to svg needs to be updated in Google and Apple modules.

modules/ppcp-applepay/resources/css/styles.scss
AlexP11223: These images seem to be embeded as base64 in the .css files now, so the path only needs to be correct at compile time.

Actually before this PR the path was not correct (.png instead of .svg), and the new build was failing, so I am not sure if these images are actually used. It's for some preview, maybe only in legacy UI?

PR #3991, opened 2026-01-12
ppcp-blocks SettingsProvider migration (5602)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-blocks` module.

### PR Changes
- ppcp-settings module:
   - Added `is_method_enabled()` method to `SettingsProvider` to check if a payment gateway is enabled
- ppcp-blocks module:
   - Replaced all `wcgateway.settings` references with `settings.settings-provider`
   - Add `card_icons` to `SettingsModel` and `SettingsProvider`
   - Removed unused `Settings` import and assertion from `BlocksModule.php`

### DIFF
--- a/modules/ppcp-blocks/services.php
+++ b/modules/ppcp-blocks/services.php
@@ -12,6 +12,7 @@ namespace WooCommerce\PayPalCommerce\Blocks;
 use WooCommerce\PayPalCommerce\Assets\AssetGetter;
 use WooCommerce\PayPalCommerce\Assets\AssetGetterFactory;
 use WooCommerce\PayPalCommerce\Blocks\Endpoint\UpdateShippingEndpoint;
+use WooCommerce\PayPalCommerce\Settings\Data\SettingsProvider;
 use WooCommerce\PayPalCommerce\Vendor\Psr\Container\ContainerInterface;
 use WooCommerce\PayPalCommerce\Button\Assets\SmartButtonInterface;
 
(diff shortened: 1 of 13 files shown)


modules/ppcp-blocks/services.php
```diff
@@ -51,17 +52,23 @@ function () use ( $container ): SmartButtonInterface {
...
-			(bool) $settings->get( 'blocks_final_review_enabled' ) :
-			true;
+		return ! $settings_provider->enable_pay_now();
+	},
+
+	'blocks.card-icons'                    => static function ( ContainerInterface $container ): array {
+		$settings_provider = $container->get( 'settings.settings-provider' );
+		assert( $settings_provider instanceof SettingsProvider );
+
+		return $settings_provider->card_icons();
```
stracker-phil: Suggestion: I see no reason to add this service, as it's only used once (20 lines further up) and IMO has two downsides:

1. It makes the code appear more complex than it really is, by adding a new dependency to `AdvancedCardPaymentMethod`
2. The icon-list is instantly resolved during plugin load and cached for the entire request; possibly no problem here, but a pattern I would avoid

I suggest to remove this service again and fetch the icon list when it's actually consumed: in `AdvancedCardPaymentMethod.php::get_payment_method_data()`

modules/ppcp-settings/src/Data/SettingsModel.php
```diff
@@ -109,6 +109,7 @@ protected function get_defaults(): array {
 
 			// Array of string values.
 			'disabled_cards'         => array(),
+			'card_icons'             => array(),
```
stracker-phil, line 112: Question: Is this change complete? it looks like the migration or assignment of `card_icons` is missing - if this is a separate issue/PR then everything is fine, just want to make sure we do not miss something here

stracker-phil (approved):

PR #4007, opened 2026-01-15
Use pl-PL for Polish APMs

AlexP11223:
Looks like it needs to be pl-PL, otherwise fails in live.


Dinamiko (approved):

stracker-phil (approved):

InpsydeNiklas:
This is not needed, confirmed by PayPal.

PR #4016, opened 2026-01-16
ppcp-axo-block SettingsProvider migration (5601)

Dinamiko:
This PR replaces `wcgateway.settings` service with the new `SettingsProvider` in `ppcp-axo-block` module.


Narek13 (approved):
Thanks. LGTM

PR #4030, opened 2026-01-20
Fix missing shipping selection for Google Pay on product pages when the cart is empty

AlexP11223:
Now setting the cart on the Google Pay button click (like with the standard PayPal button) so that the shipping is determined correctly. Also fixed the product shipping check for the assets flag, it was always set to false if the cart is empty.

### DIFF
--- a/modules/ppcp-googlepay/src/Assets/Button.php
+++ b/modules/ppcp-googlepay/src/Assets/Button.php
@@ -449,17 +449,17 @@ class Button implements ButtonInterface {
 	public function script_data(): array {
 		$use_shipping_form = $this->should_use_shipping();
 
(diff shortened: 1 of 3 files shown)


Dinamiko (approved):

modules/ppcp-googlepay/src/Assets/Button.php
stracker-phil, line 1: Suggestion:

Would it make more sense to move all the decision logic into the `should_use_shipping()` method?

That method is private and used in a single location. To me, it makes sense to keep shipping-decision inside a function called `should_use_shipping()` and not in the function `script_data()`

stracker-phil (approved):

PR #4042, opened 2026-01-26
Product ingestion syncs same product three times (5624)

mmaymo:
### Description

The fact of same product 3 times in the batch it was because we needed to do unique_merge in the resulting array, cause the product could be true to several clauses.

### Steps to Test

1. Create a product with image and description
2. Enable agentic module (flag and settings)
3. See logs for the sync, it should show the product just once in the array

### Documentation

### Changelog Entry

Closes # .


modules/ppcp-agentic-commerce/src/Ingestion/IngestionBatchProvider.php
stracker-phil, line 1: Instead of `array_unique()` I suggest to exclude duplicates directly in the query, using the `'exclude' => $batch` argument.

Also, I see that two queries use the string "publish" (line 62/89) while the first query `ProductStatus::PUBLISH` (line 39); looking at the criteria, I suggest to extract the wc_get_products() call to a private helper method, so we have a single place to define the status/type/limit/...

Sample:

```php
$meta_fresh = array(
    'key'     => '_ppcp_agentic_last_sync',
    'compare' => 'NOT EXISTS',
);
$meta_dirty = array(
    'key'     => '_ppcp_agentic_needs_sync',
    'compare' => 'EXISTS',
);
$meta_stale = array(
    'key'     => '_ppcp_agentic_last_sync',
    'value'   => $stale_date,
    'compare' => '<',
    'type'    => 'DATETIME',
);

$batch = $this->get_products( $meta_fresh );
$dirty = $this->get_products( $meta_dirty, $batch );
$batch = array_merge( $batch, $dirty );
$stale = $this->get_products( $meta_stale, $batch );
return array_merge( $batch, $stale );
```

stracker-phil (changes requested):

stracker-phil (approved):

PR #4045, opened 2026-01-27
Vaulting subscription - PayPal - Free trial order renewal (4913)

Dinamiko:
The localized script data filter returned early without setting `vaulted_paypal_email`, leaving it as undefined in JavaScript. Since undefined `!== ''` evaluates to `true`, the JS incorrectly treated guest users as having a vaulted PayPal account, which hid the smart button and showed the standard "Proceed to PayPal" button instead.                                                                       
                                                                                                                                                                                                                       
This PR changes the JS checks from `!== ''` to `!!` so that `undefined`, `null`, and `''` are all treated as false.

### DIFF
--- a/modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php
+++ b/modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php
@@ -396,8 +396,10 @@ class WcSubscriptionsModule implements ServiceModule, ExtendingModule, Executabl
 		add_filter(
 			'woocommerce_paypal_payments_localized_script_data',
 			function ( array $localized_script_data ) use ( $c ) {
-				// Return early if save payment methods (Vault v3) is enabled.
+				// When Vault v3 is enabled, set empty vaulted_paypal_email to prevent
+				// undefined JS property from affecting button visibility logic.
 				if ( $c->has( 'save-payment-methods.eligible' ) && $c->get( 'save-payment-methods.eligible' ) ) {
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php
```diff
@@ -396,8 +396,10 @@ function () use ( $c ) {
 		add_filter(
 			'woocommerce_paypal_payments_localized_script_data',
 			function ( array $localized_script_data ) use ( $c ) {
-				// Return early if save payment methods (Vault v3) is enabled.
+				// When Vault v3 is enabled, set empty vaulted_paypal_email to prevent
+				// undefined JS property from affecting button visibility logic.
```
AlexP11223: Should we maybe also improve the JS check? To check for any empty, undefined values instead of only `''`. Maybe something like simply `if (vaulted_paypal_email)`.

stracker-phil (approved):
Looking good

modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php
```diff
@@ -396,8 +396,10 @@ function () use ( $c ) {
...
 				if ( $c->has( 'save-payment-methods.eligible' ) && $c->get( 'save-payment-methods.eligible' ) ) {
+					$localized_script_data['vaulted_paypal_email'] = '';
```
AlexP11223: Can probably remove this now to keep things simpler.

PR #4053, opened 2026-01-30
Cache order requests

AlexP11223:
Added `OrderEndpointCached` that caches the `order` calls in an array, so that we do not perform the same API requests at least during the same page request. Such as when visiting a WC order page, there were 3 order requests, now there is only 1.

Also, the `order` method now can accept a WC order (to retrieve the PayPal ID from the meta), because we seem to need to do it in many places and it does not make much sense to duplicate it every time.

Currently using the cached version only for the 3 WC order page requests, but it probably should be fine to use it in most (all?) other places.

### DIFF
--- a/modules/ppcp-api-client/services.php
+++ b/modules/ppcp-api-client/services.php
@@ -267,7 +267,7 @@ return array(
 			$bn_code
 		);
 	},
-	'api.endpoint.order.cached'                      => static function ( ContainerInterface $container ): OrderEndpoint {
+	'api.endpoint.order.cached'                      => static function ( ContainerInterface $container ): OrderEndpointCached {
 		$session_handler = $container->get( 'session.handler' );
 		assert( $session_handler instanceof SessionHandler );
 		$bn_code         = $session_handler->bn_code();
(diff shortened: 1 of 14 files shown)


Narek13 (approved):
LGTM. Thanks

modules/ppcp-api-client/services.php
```diff
@@ -244,27 +245,46 @@
...
-			$logger,
-			$subscription_helper,
+			$container->get( 'woocommerce.logger.woocommerce' ),
+			$container->get( 'wc-subscriptions.helper' ),
+			$container->get( 'wcgateway.is-fraudnet-enabled' ),
+			$container->get( 'wcgateway.fraudnet' ),
+			$bn_code
+		);
+	},
+	'api.endpoint.order.cached'                      => static function ( ContainerInterface $container ): OrderEndpoint {
```
Dinamiko: Is `OrderEndpoint` return type (instead of `OrderEndpointCached`) intentional here?

Dinamiko (approved):
PR approved, I just left a small clarification about `OrderEndpoint` return type.

modules/ppcp-api-client/services.php
AlexP11223: Ah, yeah, here it probably should be `OrderEndpointCached` though of course it does not make much difference.

PR #4056, opened 2026-02-02
Use merchant country instead of shop country (5842)

Narek13:
## Issue Description

Working Capital promotional items (WOO task and inbox note) were using `api.shop.country` to determine eligibility, which checks the WooCommerce shop's base country. This is incorrect because Working Capital availability should be based on the merchant's PayPal account country, not their shop location.

This meant merchants with US PayPal accounts but non-US shops wouldn't see Working Capital promotions, and conversely, merchants with non-US PayPal accounts but US shops would incorrectly see them.

## PR Description

Updated Working Capital eligibility checks to use `api.merchant.country` instead of `api.shop.country` in both:
- Woo task configuration (`wcgateway.settings.wc-tasks.working-capital-config`)
- Inbox note display logic

The `api.merchant.country` service correctly returns:
- The connected merchant's PayPal account country when available
- Falls back to `api.shop.country` when not connected

This ensures Working Capital promotions are only shown to merchants with US-based PayPal accounts, regardless of their WooCommerce shop location, maintaining consistency across all promotional touchpoints.

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -2242,7 +2242,7 @@ return array(
 				Note::E_WC_ADMIN_NOTE_INFORMATIONAL,
 				'ppcp-working-capital-inbox-note',
 				Note::E_WC_ADMIN_NOTE_UNACTIONED,
-				$is_working_capital_feature_flag_enabled && $container->get( 'api.shop.country' ) === 'US' && $stay_updated,
+				$is_working_capital_feature_flag_enabled && $container->get( 'api.merchant.country' ) === 'US' && $stay_updated,
 				new InboxNoteAction(
 					'learn_more',
 					__( 'Learn More', 'woocommerce-paypal-payments' ),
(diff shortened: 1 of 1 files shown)


stracker-phil (approved):
The PR itself is looking good.

But possibly missing: There is no integration with our central `FeaturesEligibilityService` yet, which should be the authoritative source for all eligibility checks.

If time allows, I suggest to move the check to a new service + integrate it with that service. If no time, please add an internal task for this house-keeping

Narek13:
Thanks @stracker-phil for the comment. Regarding:

> But possibly missing: There is no integration with our central FeaturesEligibilityService yet, which should be the authoritative source for all eligibility checks.

I thought the `FeaturesEligibilityService` was tied to the **Features** section displayed in the Overview tab.
At the same time, I’ve [integrated](https://github.com/woocommerce/woocommerce-paypal-payments/blob/0c816f8e540f65e3d4f1f6ac98ef65061ea65ad8/modules/ppcp-settings/src/Service/TodosEligibilityService.php#L251) it with `TodosEligibilityService`, but this PR only touches Woo tasks and Woo Inbox notes.
Because of that, I don’t think we need to add this to `FeaturesEligibilityService`, as it wouldn’t be used there anyway and it also doesn't provide an easy way to check for individual eligibility, something like `$features_eligibility_service->is_working_capital_eligible()`

PR #4061, opened 2026-02-03
Refactor GermanizedShipmentIntegration to Shiptastic.

### LINKED ISSUE #4060: Refactor GermanizedShipmentIntegration to support Shiptastic instead
The `GermanizedShipmentIntegration` does currently support legacy shipments only although the Germanized Core refactored the shipments integration to Shiptastic (which is a separate plugin now). The shipments integration won't be part of the Germanized core in future versions. The current integration script will break in the future.

I'll add a pull request which contains the necessary changes and is backwards compatible with older versions of Germanized.

Best,
Dennis

dennisnissle:
**Issue**: #4060

**Ticket**:

**Slack Thread**:

---

### Description

This pull request refactors the Germanized Shipments integration to the new Shiptastic integration.

### Changelog Entry

> Support current Shiptastic integration instead of Germanized legacy shipments integration

### DIFF
--- a/.psalm/shiptastic.php
+++ b/.psalm/shiptastic.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Vendidero\Germanized\Shipments {
+namespace Vendidero\Shiptastic {
 
 	use WC_Data;
 	use WC_Order;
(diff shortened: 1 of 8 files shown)


Narek13 (approved):
Thank you for reporting and refactoring. I wonder what will happen with merchants who are still using the old version of Germanized though, but code wise, looking good to me

@dennisnissle

dennisnissle:
@Narek13 This works with Germanized >= 3.19.0 which was released one year ago.

PR #4073, opened 2026-02-09
Add pattern validation and improve error messages for client credentials (5888, 5889)

Dinamiko:
This PR fixes `client_secret` error message and add `client_id` pattern validation. It also adds exception chaining and generic error messages for direct API authentication.


Narek13 (approved):
Thanks. LGTM

PR #4100, opened 2026-02-18
Deposits remainder payments not completed for pending-deposit orders (5919)

Dinamiko:
Orders created by WooCommerce Deposits for remainder payments use status pending-deposit. In 3.4.0 these orders cannot be completed via PayPal.

Customer can go to payment page, choose payment method, click pay, gets redirected to order received page, but payment is not completed. No transaction ID is saved and no funds are captured. The order still shows Pay button and user can repeat the loop.

This PR extracts the status check into OrderStatusHelper::is_awaiting_payment() with a `woocommerce_paypal_payments_awaiting_payment_statuses` filter, replacing the hardcoded check in 5 locations. 

Third-party plugins can hook in:
```php
  add_filter( 'woocommerce_paypal_payments_awaiting_payment_statuses', function( $statuses ) {
      $statuses[] = 'pending-deposit';
      return $statuses;
  } );
```


Narek13 (approved):
LGTM, thanks for working on this

PR #4108, opened 2026-02-20
New UI - As ACDC is not enable by default and after enabling, saving and reloading page setting is not preserved (5907)

Dinamiko:
`PaymentSettings::toggle_method_state()` (and `set_method_title/set_method_description`) rely on `WC()->payment_gateways()` to find the gateway object. But gateways like CreditCardGateway are only registered             conditionally when `wcgateway.is-plugin-settings-page` is true. During REST API saves (`POST wc/v3/wc_paypal/payment`), that flag is false, so `get_gateway()` returns `null` and the write silently does nothing.

### Steps To Reproduce
- Onboard with merchant from US
- Navigate to Payment tab and enable ACDC, save settings and reload
  - Observe setting are not saved

This PR adds a fallback when the gateway object isn't available and writes directly to the `woocommerce_{id}_settings option`.

### DIFF
--- a/modules/ppcp-settings/src/Data/PaymentSettings.php
+++ b/modules/ppcp-settings/src/Data/PaymentSettings.php
@@ -87,6 +87,11 @@ class PaymentSettings extends AbstractDataModel {
 					$gateway->enabled = wc_bool_to_string( $is_enabled );
 
 					$this->modified_gateway( $gateway );
+				} else {
+					$option_key          = 'woocommerce_' . $method_id . '_settings';
+					$settings            = get_option( $option_key, array() );
+					$settings['enabled'] = wc_bool_to_string( $is_enabled );
+					update_option( $option_key, $settings );
 				}
 		}
 	}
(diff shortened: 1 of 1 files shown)


modules/ppcp-settings/src/Data/PaymentSettings.php
```diff
@@ -140,6 +150,11 @@ public function set_method_description( string $method_id, string $description )
 			$gateway->description = $description;
 
 			$this->modified_gateway( $gateway );
+		} else {
```
Narek13, line 153: These three blocks share the same pattern to fetch option, set a key, save. Consider extracting a helper:

```php
private function update_gateway_setting( string $method_id, string $key, $value ): void {
    $option_key            = 'woocommerce_' . $method_id . '_settings';
    $settings              = get_option( $option_key, array() );
    $settings[ $key ]      = $value;
    update_option( $option_key, $settings );
}
```

Narek13 (approved):
Thanks, I left a suggestion, which you can ignore if you prefer.
It would also be good to create an issue on our board so we don’t forget to refactor these methods to rely on DB values, as we previously discussed.

Dinamiko:
Thanks @Narek13 I take note, let's handle this once epic branch is merged into develop.

PR #4089, opened 2026-02-13
Agentic: Add ResolutionOption schema and resolution options to all validators (5856)

danieldudzic:
### Description                                                                                                                                                                 
                                                                                                                                                                                  
 Cart validation errors need actionable resolution suggestions to help the agent guide users toward fixing issues.                                                               
                                                                                                                                                                                  
 This PR creates a `ResolutionOption` schema with factory methods and a `Priority` enum to ensure all validators provide uniform, type-safe resolution suggestions. Added missing resolution options to 5 validators covering inventory, products, currency, shipping, and quantity validation scenarios.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                   
  ### Resolution Options                                                                                                                                                     
                                                                                                                                                                                  
  | Validator | Scenario | Resolution Options | Priority |                                                                                                                        
  |-----------|----------|-------------------|----------|                                                                                                                         
  | **InventoryValidator** | Out of stock | • Remove from cart<br>• Wait for restock | HIGH<br>MEDIUM |                                                                           
  | | Insufficient quantity | • Reduce quantity to [max]<br>• Remove from cart | HIGH<br>LOW |                                                                                    
  | **ProductValidator** | Invalid product | • Remove from cart | HIGH |                                                                                                          
  | | Unavailable product | • Remove from cart<br>• View similar items | HIGH<br>MEDIUM |                                                                                         
  | **CurrencyValidator** | Mixed currencies | • Set all items to [store currency]<br>• Remove from cart | HIGH<br>LOW |                                                          
  | | Wrong store currency | • Change to [store currency] | HIGH |                                                                                                                
  | **ShippingValidator** | Missing street address | • Provide street address<br>• Update shipping address | HIGH<br>LOW |                                                        
  | | Missing city | • Provide city<br>• Update shipping address | HIGH<br>LOW |                                                                                                  
  | | Missing postal code | • Provide postal code<br>• Update shipping address | HIGH<br>LOW |                                                                                    
  | | Invalid postal code format | • Update shipping address | MEDIUM |                                                                                                           
  | | PO Box restriction | • Update to physical address<br>• Contact support | HIGH<br>LOW |                                                                                      
  | | Country not supported | • Change shipping country<br>• Contact support | HIGH<br>MEDIUM |                                                                                   
  | **CartItem** | Quantity out of range (1-999) | • Set a valid quantity (1–999) | HIGH |                                                                                        
[...]

PR #4109, opened 2026-02-20
Missing shipping address validation not triggering requires additional information status (5867)

mmaymo:
### Steps to Test

1. Create a new  simple cart via postman, with no address
2. See that the response has a Status invalid, and the validation issues shows the defect

This PR takes https://github.com/woocommerce/woocommerce-paypal-payments/tree/dev/PCP-5707-rename-to-store-sync as reference

### DIFF
--- a/modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
+++ b/modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
@@ -4,10 +4,10 @@
  *
  * Validates shipping addresses and restrictions according to WooCommerce settings.
  * Covers four main scenarios:
- * 0. Missing Shipping Address (physical products, no address provided)
- * 1. Invalid Shipping Address (completeness, format)
- * 2. PO Box Restriction (signature-required items)
- * 3. Region Restricted (country not allowed)
+ * 1. Missing Shipping Address (physical products, no address provided)
+ * 2. Invalid Shipping Address (completeness, format)
+ * 3. PO Box Restriction (signature-required items)
+ * 4. Region Restricted (country not allowed)
  *
  * @package WooCommerce\PayPalCommerce\StoreSync\CartValidation
  */
@@ -49,11 +49,7 @@ class ShippingValidator implements ValidatorInterface {
 						'',
 						array( 'specific_issue' => ShippingIssue::MISSING_SHIPPING_ADDRESS ),
 						array(
-							array(
-								'action'   => 'PROVIDE_SHIPPING_ADDRESS',
-								'label'    => 'Add shipping address',
-								'metadata' => array( 'priority' => 'HIGH' ),
-							),
+							ResolutionOption::provide_shipping_address(),
 						)
 					),
 				);
--- a/modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
+++ b/modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
@@ -38,7 +38,7 @@ class ShippingValidator implements ValidatorInterface {
 
 	public function validate( PayPalCart $cart ) {
 		$shipping_address = $cart->shipping_address();
-
+		// Scenario 1: Missing Shipping Address.
 		if ( ! $shipping_address ) {
 			if ( $this->cart_needs_shipping( $cart ) ) {
 				return array(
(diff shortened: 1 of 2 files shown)


modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
```diff
@@ -37,6 +40,24 @@ public function validate( PayPalCart $cart ) {
...
+						'',
+						array( 'specific_issue' => ShippingIssue::MISSING_SHIPPING_ADDRESS ),
+						array(
+							array(
+								'action'   => 'PROVIDE_SHIPPING_ADDRESS',
+								'label'    => 'Add shipping address',
+								'metadata' => array( 'priority' => 'HIGH' ),
+							),
+						)
+					),
```
danieldudzic, line 54: Good job on the core logic here :) — the `cart_needs_shipping` check is exactly what we need.

One thing to update: the parent branch recently https://github.com/woocommerce/woocommerce-paypal-payments/pull/4089 that changed all validators to use `ResolutionOption` factory methods instead of raw arrays. This matters because `ValidationIssue::to_array()` now calls `->to_array()` on each resolution option object, so raw arrays would cause a runtime error.

Since `provide_shipping_address()` doesn't exist yet on `ResolutionOption`, we'd need to add it. It can build on the existing `provide_missing_field()` method — which already covers this case semantically (we're providing a missing field: the shipping address). So perhaps we don't need a new `PROVIDE_SHIPPING_ADDRESS` constant in `ResolutionAction` as I might've suggested in the original issue description (sorry!) - the existing `PROVIDE_MISSING_FIELD` action is the right fit, with the specifics (field name, label, priority) encoded via the factory method:

```
  // In ResolutionOption.php:
  public static function provide_shipping_address(): self {
      return self::provide_missing_field( 'shipping_address', 'Add shipping address' )->with(
          array( 'metadata' => array( 'priority' => Priority::HIGH ) )
      );
  }
```

Then the `MissingField` call becomes:

```suggestion
                      new MissingField(                                                                                                                                
                          'Shipping address is required',                                                                                                            
                          'Please provide a shipping address to continue.',                                                                                            
                          'shipping_address',                                                                                                                          
                          '',                                                                                                                                          
                          array( 'specific_issue' => ShippingIssue::MISSING_SHIPPING_ADDRESS ),
                          array(
                              ResolutionOption::provide_shipping_address(),
                          )
                      ),
```

You can see examples of this pattern in `PriceValidator`, `InventoryValidator`, `ProductValidator`, and `CouponResolutionBuilder` on the parent branch.

modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
```diff
@@ -3,27 +3,30 @@
  * Shipping Validator for Agentic Commerce.
  *
  * Validates shipping addresses and restrictions according to WooCommerce settings.
- * Covers three main scenarios:
+ * Covers four main scenarios:
+ * 0. Missing Shipping Address (physical products, no address provided)
  * 1. Invalid Shipping Address (completeness, format)
  * 2. PO Box Restriction (signature-required items)
  * 3. Region Restricted (country not allowed)
```
danieldudzic: 📝 Total nitpick: perhaps we should start the numered list of scenarios with 1, and then go from there?

danieldudzic (changes requested):
Thanks for working on this! 🙏

We should re-sync this with the parent, and I also left a note regarding the recent refactor of Resolution Options, which this should probably also follow.

modules/ppcp-store-sync/src/Schema/ResolutionOption.php
```diff
@@ -172,6 +172,12 @@ public static function provide_missing_field( string $field_name, string $label
 		);
 	}
 
+	public static function provide_shipping_address(): self {
+		return self::provide_missing_field( 'shipping_address', 'Add shipping address' )->with(
+			array( 'metadata' => array( 'priority' => Priority::HIGH ) )
+		);
+	}
+
```
danieldudzic:
```suggestion
    /**
     * Factory: Provide shipping address.
     *
     * @return self
     */
	public static function provide_shipping_address(): self {
		return self::provide_missing_field( 'shipping_address', 'Add shipping address' )->with(
			array( 'metadata' => array( 'priority' => Priority::HIGH ) )
		);
	}
```

🤏 Sorry, nitpicking again. We should also add the docblock to keep things consistent with other methods in `ResolutionOption`.

modules/ppcp-store-sync/src/CartValidation/ShippingValidator.php
```diff
@@ -3,10 +3,11 @@
...
  * Validates shipping addresses and restrictions according to WooCommerce settings.
- * Covers three main scenarios:
- * 1. Invalid Shipping Address (completeness, format)
- * 2. PO Box Restriction (signature-required items)
- * 3. Region Restricted (country not allowed)
+ * Covers four main scenarios:
+ * 1. Missing Shipping Address (physical products, no address provided)
+ * 2. Invalid Shipping Address (completeness, format)
+ * 3. PO Box Restriction (signature-required items)
+ * 4. Region Restricted (country not allowed)
```
danieldudzic, line 10: 🤏 One more nitpick thing:
The inline `// Scenario N`: comments inside the `validate()` method still say `// Scenario 1:, // Scenario 2:, // Scenario 3:` - those should now be `// Scenario 2:, // Scenario 3:, // Scenario 4`: to match the updated docblock.

danieldudzic (changes requested):
Nice! Got 2 super minor change requests 🙏

danieldudzic (approved):
This is good to go! cc: @stracker-phil

PR #4118, opened 2026-02-25
New UI - Standard card button gateway visible on Woocommerce Payment tab and on classic checkout without button (5906)

Dinamiko:
### Steps to reproduce
- Onboard with merchant from US 
  - Observe standard card button gateway in Payments page
  - Observe standard card button gateway

This PR prevents the Standard Card Button gateway from appearing when the merchant already has Advanced Card Payments (ACDC), since ACDC handles card payments. Mexico is exempted because ACDC works differently there.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -384,6 +384,16 @@ class SettingsModule implements ServiceModule, ExecutableModule {
 
 					unset( WC()->payment_gateways->payment_gateways[ $index ] );
 				}
+
+				$card_config   = $container->get( 'wcgateway.configuration.card-configuration' );
+				$store_country = $container->get( 'api.merchant.country' );
+				if ( $card_config->use_acdc() && $store_country !== 'MX' ) {
+					foreach ( WC()->payment_gateways->payment_gateways as $index => $gateway ) {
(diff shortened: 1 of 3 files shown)


Narek13 (approved):
Thanks. Code-wise looks good to me.
Also, keep in mind that we run [this method](https://github.com/woocommerce/woocommerce-paypal-payments/blob/2ec4339c0646bd389733eb4c3fbe2b4a22c64acd/modules/ppcp-settings/src/Service/SettingsDataManager.php#L231)
during onboarding. After that, we [toggle](https://github.com/woocommerce/woocommerce-paypal-payments/blob/2ec4339c0646bd389733eb4c3fbe2b4a22c64acd/modules/ppcp-settings/src/Service/SettingsDataManager.php#L257) credit card gateway, and then we [toggle](https://github.com/woocommerce/woocommerce-paypal-payments/blob/2ec4339c0646bd389733eb4c3fbe2b4a22c64acd/modules/ppcp-settings/src/Service/SettingsDataManager.php#L270) the the Standard Card Button gateway.

So it might make sense to refactor this method to enable the Standard Card Button only if ACDC is not enabled.

P.S. - Also, please check the CI

Dinamiko:
Thanks @Narek13 let's handle it together with fixing CI once all "New UI" tickets are merged into the epic branch.

PR #4124, opened 2026-02-26
New UI - Crypto visible on Feature tab on fresh site when flag not enabled (5953)

Dinamiko:
### Steps to reproduce
- Onboard with US merchant
- Visit Settings / Overview tab
  -  Notice Pay with Crypto is visible despite feature flag is disabled

This PR moves `register_pwc_feature_flag_filters()` before the `should_add_local_apm_gateways()` early return in `LocalAlternativePaymentMethodsModule`, so the PWC feature flag filters are always registered regardless of gateway status.


danieldudzic (approved):
Straightforward 👍

PR #4126, opened 2026-02-26
New UI - ACDC not visible on frontend for eligible countries (5954)

Dinamiko:
### Steps to reproduce
- Onboard as US merchant
- Visit Settings / Payment methods tab
  - Notice ACDC is disabled by default

This PR adds missing `toggle_payment_gateways()` call in `SettingsDataManager::apply_configuration()` so payment gateways are enabled server-side during onboarding.

### DIFF
--- a/modules/ppcp-settings/src/Service/SettingsDataManager.php
+++ b/modules/ppcp-settings/src/Service/SettingsDataManager.php
@@ -197,6 +197,9 @@ class SettingsDataManager {
 		// Apply defaults for the "Settings" tab.
 		$this->apply_payment_settings( $flags );
 
+		// Enable/disable payment gateways based on onboarding choices.
+		$this->toggle_payment_gateways( $flags );
+
 		// Assign defaults for the "Styling" tab.
 		$this->apply_location_styles( $flags );
 


danieldudzic:
@Dinamiko Thanks for looking into this! I was also able to replicate the issue, however, I believe this particular fix might reintroduce a timing issue that was intentionally resolved in e992f45c0.

The `toggle_payment_gateways()` call was removed from `apply_configuration()` because the ACDC product status dependency arrives too late during onboarding — the page has already loaded by the time PayPal confirms the merchant's eligibility. Moving the toggle to the async sync flow (`woocommerce_paypal_payments_sync_gateways` → `sync_gateway_settings()` → `toggle_payment_gateways()`) was the correct solution to that timing issue, and that flow is still working as expected.

The actual root cause is in `WCGatewayModule.php`: the `woocommerce_payment_gateways` filter checks `$settings->has('products_dcc_enabled') && $settings->get('products_dcc_enabled')` to decide whether to register the ACDC gateway on checkout pages. This reads from the legacy `woocommerce-ppcp-settings` option. After `DCCProductStatus` was refactored in 40d2a9a9f to use `ProductStatusResultCache` (transients), `products_dcc_enabled` is never written to that option anymore — so the check always fails and the gateway never gets registered on checkout.

Re-adding `toggle_payment_gateways()` here happens to mask the problem by enabling the gateway earlier, but the dead legacy check in `WCGatewayModule.php` would still remain as a latent issue.

I've opened #4128 with a targeted fix — replacing the dead legacy settings check with `$dcc_product_status->is_active()`, which reads from the transient cache.

Dinamiko:
Fixed in https://github.com/woocommerce/woocommerce-paypal-payments/pull/4128

PR #4131, opened 2026-03-02
Vault via Orders API for redirect gateway in Blocks Checkout (3796)

Dinamiko:
This PR extends PayPal vaulting support to the "Proceed to PayPal" button in blocks checkout. 

Previously, saving payment methods only worked through the Express PayPal button. The redirect flow triggered by "Proceed to PayPal" never injected vault attributes because `OrderProcessor::create_order()` passed empty values to the Orders API. This change forwards the correct payment method and funding source, enabling `SavePaymentMethodsModule` to activate during the redirect flow.

Additionally, customers can now pay with a previously saved PayPal token directly from the "Proceed to PayPal" button.


modules/ppcp-wc-gateway/src/Endpoint/CapturePayPalPayment.php
```diff
@@ -0,0 +1,139 @@
...
+
+	/**
+	 * @var SettingsProvider
+	 */
+	private $settings_provider;
+
+	/**
+	 * @var LoggerInterface
+	 */
+	private $logger;
```
Narek13: No need for these comments for the obvious types. Also types are missing, example:
`private LoggerInterface $logger;`

modules/ppcp-wc-gateway/src/Endpoint/CapturePayPalPayment.php
```diff
@@ -0,0 +1,139 @@
...
+		$this->host                  = $host;
+		$this->bearer                = $bearer;
+		$this->order_factory         = $order_factory;
+		$this->purchase_unit_factory = $purchase_unit_factory;
+		$this->settings_provider     = $settings_provider;
+		$this->logger                = $logger;
+	}
+
+	/**
+	 * Creates PayPal order from the given PayPal/Venmo vault id.
```
Narek13:
```suggestion
	 * Creates PayPal order from the given PayPal/Venmo vault ID.
```

modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
```diff
@@ -401,6 +424,50 @@ public function process_payment( $order_id ): array {
...
+					}
+
+					$order = $this->order_endpoint->order( $created_order->id() );
+					$this->add_paypal_meta( $wc_order, $created_order, $this->environment );
+					$wc_order->add_payment_token( $token );
+
+					if ( $order->intent() === 'AUTHORIZE' ) {
+						$order = $this->order_endpoint->authorize( $order );
+
+						$wc_order->update_meta_data( AuthorizedPaymentsProcessor::CAPTURED_META_KEY, 'false' );
```
Narek13, line 449: I haven't tested or maybe I am wrong, but You call `update_meta_data` twice (three times counting the subscription branch) but there's no `$wc_order->save()`. Meta won't persist to the database
`update_transaction_id` calls `$wc_order->save()` but what if it is not called at all?

Narek13 (changes requested):
Thanks a lot for working on this. Added some suggestions and doubts.

modules/ppcp-wc-gateway/src/Endpoint/CapturePayPalPayment.php
```diff
@@ -0,0 +1,115 @@
...
+				'PayPal-Request-Id' => uniqid( 'ppcp-', true ),
+			),
+			'body'    => wp_json_encode( $data ),
+		);
+
+		$response = $this->request( $url, $args );
+		if ( $response instanceof WP_Error ) {
+			throw new RuntimeException( $response->get_error_message() );
+		}
+
```
Narek13, line 111: Sorry, do we also want to check the status code here? like is done in [orderEndpoint](https://github.com/woocommerce/woocommerce-paypal-payments/blob/dev/develop/modules/ppcp-api-client/src/Endpoint/OrderEndpoint.php) for example?
```php
$status_code = (int) wp_remote_retrieve_response_code( $response );
if ( 200 !== $status_code ) {
	throw new PayPalApiException( $json, $status_code );
}
```

Narek13 (approved):
Thanks for making all the chnages. Left 1 more doubt. Everything else looks good to me.

modules/ppcp-wc-gateway/src/Endpoint/CapturePayPalPayment.php
Dinamiko: Done, thanks!

PR #4132, opened 2026-03-02
AVS code mismatch (5960)

Dinamiko:
PayPal response is passing an `S` but the plugin is showing an `N` which means not a match. Any code that is not exactly `M` (Match) gets mapped to `N` (No match), including `S` (Service not Supported), `U` (Unavailable), `Y` (Yes), `A` (Address only), etc. This is semantically incorrect.   

This PR removes the fraud processor response legacy  keys that were causing the problem.


Narek13 (approved):
LGTM

PR #4134, opened 2026-03-03
Vault Import - convert billing agreements to Vault tokens (2432)

Dinamiko:
PPEC (PayPal Express Checkout) subscriptions use legacy Billing Agreement IDs for renewal payments via reference transactions, these subscriptions need to be migrated Vault v3 token system.   

### PR Changes                                                                                                                                                                                             
  - Implements on-demand conversion of legacy PayPal Express Checkout (PPEC) billing agreements to Vault v3 tokens at subscription renewal time, with graceful fallback to billing agreement if conversion fails       
  - Adds fallback to use billing agreements directly for merchants not eligible for Vault v3
  - Removes obsolete PPEC compatibility code (`DeactivateNote`, `SettingsImporter`, billing agreement fallback logic) that is no longer needed
  - Adds `BillingAgreementTokenConverter` service


modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php
```diff
@@ -35,16 +35,29 @@ class SubscriptionsHandler {
 	 */
 	private $mock_gateway;
 
+	private BillingAgreementTokenConverter $token_converter;
+
+	private LoggerInterface $logger;
 
 	/**
-	 * Constructor.
+	 * Whether the merchant is eligible for Vault v3.
```
Narek13: Should this be on line 45?

modules/ppcp-compat/src/PPEC/BillingAgreementTokenConverter.php
```diff
@@ -0,0 +1,73 @@
...
+				)
+			);
+
+			$customer_id = $this->customer_repository->customer_id_for_user( $user_id );
+			$result      = $this->payment_method_tokens_endpoint->create_payment_token( $payment_source, $customer_id );
+
+			if ( isset( $result->customer->id ) ) {
+				update_user_meta( $user_id, '_ppcp_target_customer_id', $result->customer->id );
+			}
+
```
Narek13, line 56: `$result->customer->id` gets an `isset()` check, but `$result->id` does not, used in both the logger and the return. Are we sure it exists? Otherwise we can add a check

```php
if ( empty( $result->id ) ) {
    $this->logger->error(
        sprintf( 'Vault token creation for Billing Agreement %s returned no token ID.', $billing_agreement_id )
    );
    return null;
}
```

modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php
AlexP11223: Seems ok to me, it's for the property `$vault_v3_eligible`. Maybe the GitHub diff is confusing. https://github.com/woocommerce/woocommerce-paypal-payments/blob/2cbc6f499712e5434eb246b19472138ce78281fd/modules/ppcp-compat/src/PPEC/SubscriptionsHandler.php#L47

`@var bool` is redundant, however.

AlexP11223 (approved):

Narek13 (approved):
Thanks for working on this and making all the changes.

PR #4135, opened 2026-03-03
Playground: Fix Playground workflow (5926)

danieldudzic:
### Description

Replaces the inline plugin build with artifact reuse from the existing "Build and distribute" workflow.

The workflow now polls for the B&D run matching the PR's head SHA, downloads its artifact, and re-uploads it under the playground-expected name.

When B&D did not run for the commit, the workflow exits gracefully without error.

### DIFF
--- a/.github/workflows/pr-playground-demo.yml
+++ b/.github/workflows/pr-playground-demo.yml
@@ -48,15 +48,43 @@ jobs:
 
   build_plugin:
     needs: prepare_version
-    uses: inpsyde/reusable-workflows/.github/workflows/build-plugin-archive.yml@a9af34f34e95cbe18703198c7e972e97ebcd7473
-    with:
-      PHP_VERSION: 7.4
-      NODE_VERSION: 22
-      PLUGIN_MAIN_FILE: ./woocommerce-paypal-payments.php
-      PLUGIN_VERSION: ${{ needs.prepare_version.outputs.plugin_version }}
-      PLUGIN_FOLDER_NAME: woocommerce-paypal-payments
-      ARCHIVE_NAME: ${{ needs.prepare_version.outputs.artifact_name }}
-      COMPILE_ASSETS_ARGS: "-vv --env=root"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+
+      - name: Setup Node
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-playground-demo.yml
```diff
@@ -48,15 +48,40 @@ jobs:
...
+
+      - name: Set plugin version
+        run: |
+          sed -i "s/Version: .*/Version: ${{ needs.prepare_version.outputs.plugin_version }}/" woocommerce-paypal-payments.php
+
+      - name: Prepare plugin directory
+        run: |
+          mkdir -p /tmp/plugin-build
+          rsync -a --exclude-from=.distignore . /tmp/plugin-build/
+
```
AlexP11223, line 63: I think it is not good to duplicate the ways of building the plugin. This adds another place that we have to maintain when something changes in the build, and it can get out of sync, become different compared to the main build. For example, here I think it at least misses the PHP-Scoper step?

Can we maybe find some way to either use the artifact from [this workflow](https://github.com/woocommerce/woocommerce-paypal-payments/actions/workflows/build-and-distribute.yml) or the files from the branch that it creates (without the `dev/` prefix)?

Narek13 (approved):
Thanks a lot for fixing this

.github/workflows/pr-playground-demo.yml
```diff
@@ -9,27 +9,21 @@ on:
       - "modules/**"
       - "package.json"
       - "composer.json"
-  push:
-    paths:
```
AlexP11223, line 13: Maybe should keep this (the list is not complete though). Because otherwise it will fail when the build did not need to run. 
Or maybe instead of duplicating this list here, it can simply exit without error if the build workflow did not run for this commit.

.github/workflows/pr-playground-demo.yml
```diff
@@ -9,27 +9,21 @@ on:
...
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
 
 # Disable permissions for all available scopes by default.
 permissions: {}
 
 jobs:
-  prepare_version:
+  fetch_build:
```
AlexP11223, line 16: 👍 
I was thinking there could be a simpler built-in way of waiting for another workflow. But I guess this is fine too, and gives some flexibility.

AlexP11223 (approved):

PR #4136, opened 2026-03-03
Allow Cart-Session to Update the ec_token

mmaymo:
### Description
Allow PUT /merchant-cart/{id} to create PayPal orders when needed
Per PayPal spec, the PUT endpoint can create a new ec_token when the cart doesn't have one (e.g., POST validation failed) or the existing token expired.

update_cart_session() now accepts optional $ec_token parameter. When provided, replaces stored token; when null, preserves existing.
ReplaceCartEndpoint checks if cart needs a new token (empty or expired) and calls order_manager->create_order() if the cart is valid. New token is passed to session handler and included in response.
Added UpdatedCartWithTokenResponse for PUT responses that include a new token. payment_method.token is only present in response when a new token was actually created.
Files:

AgenticSessionHandler.php - optional $ec_token param
AgenticRestEndpoint.php - pass token through store_local_cart()
ReplaceCartEndpoint.php - token creation logic
ResponseFactory.php - new cart_with_token() method
UpdatedCartWithTokenResponse.php - new response class

Behavior:

POST with invalid items → no token created
PUT to fix items → token created, included in response
PUT on cart with valid token → token preserved, not in response
Checkout → only reads existing token, never creates


modules/ppcp-store-sync/src/Session/AgenticSessionHandler.php
danieldudzic: The new `$ec_token` parameter needs to be used here, otherwise the token will never actually get replaced:                                                                     
                                                                                                                                                                                                                                       
```suggestion
			'ec_token' => $ec_token ?? $existing['ec_token'],
```

modules/ppcp-store-sync/src/Response/UpdatedCartWithTokenResponse.php
```diff
@@ -0,0 +1,76 @@
...
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\StoreSync\Response;
+
+use WC_Cart;
+
+use WooCommerce\PayPalCommerce\StoreSync\Schema\PayPalCart;
+
+class UpdatedCartWithTokenResponse extends CartResponse {
```
danieldudzic, line 20: There seems to be a filename typo — `UpdatedCartWithTokenRespones.php` instead of `UpdatedCartWithTokenRespone.php`.

modules/ppcp-store-sync/src/Endpoint/ReplaceCartEndpoint.php
```diff
@@ -88,11 +111,48 @@ public function replace_cart( WP_REST_Request $request ): WP_REST_Response {
...
+	 *
+	 * A new token is needed when:
+	 * 1. No existing token (previous POST failed to create one)
+	 * 2. Existing token has expired (future: implement expiry check)
+	 *
+	 * @param string     $existing_token The current ec_token from session.
+	 * @param PayPalCart $cart The new cart data.
+	 * @return bool True if a new token should be created.
+	 */
+	private function needs_new_token( string $existing_token, $cart ): bool {
```
danieldudzic: Minor: `$cart` is missing its type hint 😊

```suggestion
	private function needs_new_token( string $existing_token, PayPalCart $cart ): bool {
```

danieldudzic (changes requested):
Thanks for working on this! Just a few points to address, and should be good to go 🙌

modules/ppcp-store-sync/src/Endpoint/ReplaceCartEndpoint.php
```diff
@@ -75,8 +80,26 @@ public function replace_cart( WP_REST_Request $request ): WP_REST_Response {
...
 
-		// Replace the cart session (preserving ec_token).
-		$update_result = $this->store_local_cart( $cart_id, $new_cart );
+		// Determine if we need to create a new PayPal order.
+		$existing_token = $session['ec_token'] ?? '';
+		$new_token      = null;
+
+		if ( $this->needs_new_token( $existing_token, $new_cart ) ) {
+			// Create new PayPal order - may return empty string if cart is invalid.
+			$new_token = $this->order_manager->create_order( $new_cart );
```
danieldudzic:
```suggestion
			$new_token = $this->order_manager->create_order( $new_cart ) ?: null;
```

We should consider a slight modification here - this way if `create_order()` returns `''`, `$new_token` stays `null`, so `store_local_cart()` receives `null` and the `??` in `update_cart_session()` correctly preserves the existing token.

danieldudzic (changes requested):
@mmaymo Thanks for working on this. We need just one more change, I think.

stracker-phil:
I had to rewrite this PR (actually Claude did) as this PR had a significant drift from the epic code already.

Thanks for the research and PR, it was very helpful to understand the problem and re-apply it to the current epic code!

PR #4148, opened 2026-03-06
refactor(settings): add `is_feature_eligible()` and rename `get()` to `eligible_features()` in `FeaturesDefinition` (5982)

Narek13:
### Issue Description

`FeaturesDefinition::get()` had a generic name that didn't communicate intent, and the class had no way to check eligibility for a single feature. As a result, any code needing to check one feature had to retrieve the full eligibility checks array and resolve the callable manually:

```php
$eligibility_service = $container->get( 'settings.service.features_eligibilities' );
assert( $eligibility_service instanceof FeaturesEligibilityService );
$eligibilities = $eligibility_service->get_eligibility_checks();
$is_pui_eligible = $eligibilities[ FeaturesDefinition::FEATURE_PAY_UPON_INVOICE ] ?? false;

// then use it like this
$is_pui_eligible ? $is_pui_eligible() : $is_pui_eligible;
```

This made `FeaturesDefinition` practically only usable in the REST endpoint context, pushing developers to fall back on individual eligibility services (e.g. `ppcp-local-apms.pui.eligibility.check`) instead of the intended centralized abstraction.

### PR Description

Refactors `FeaturesDefinition` in two steps and updates the `PartnerReferralsData` call site:

**1. Rename `get()` → `eligible_features()`**
Pure rename with no logic change. The new name reflects what the method actually returns — only the features whose eligibility check passes. Updates the existing call in `FeaturesRestEndpoint`.

**2. Add `is_feature_eligible(string $feature_name): bool`**
New method for checking individual feature eligibility directly on `FeaturesDefinition`. Resolves the eligibility callable internally so callers get a plain `bool` back. Returns `false` and logs a warning when an unregistered feature name is passed, since this always indicates a programming error rather than a runtime condition.

**3. Refactor `PartnerReferralsData` PUI eligibility check**
Replaces the verbose eligibility boilerplate with the new method:

```php
if ( $this->features_definition->is_feature_eligible( FeaturesDefinition::FEATURE_PAY_UPON_INVOICE ) ) {
```

### DIFF
--- a/modules/ppcp-api-client/services.php
+++ b/modules/ppcp-api-client/services.php
@@ -345,7 +345,7 @@ return array(
 	'api.repository.partner-referrals-data'          => static function ( ContainerInterface $container ): PartnerReferralsData {
 		return new PartnerReferralsData(
 			$container->get( 'api.helpers.dccapplies' ),
-			$container->get( 'ppcp-local-apms.pui.eligibility.check' )
+			$container->get( 'settings.data.definition.features' )
 		);
 	},
 	'api.repository.payee'                           => static function ( ContainerInterface $container ): PayeeRepository {
--- a/modules/ppcp-api-client/src/Repository/PartnerReferralsData.php
+++ b/modules/ppcp-api-client/src/Repository/PartnerReferralsData.php
@@ -10,6 +10,7 @@ declare( strict_types = 1 );
 namespace WooCommerce\PayPalCommerce\ApiClient\Repository;
 
 use WooCommerce\PayPalCommerce\ApiClient\Helper\DccApplies;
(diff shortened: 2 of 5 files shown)


Dinamiko (commented):
Not directly related to this PR, but worth consider adding an `isset()` guard in `eligible_features` to the foreach loop, to make it consistent with `is_feature_eligible()`:
```php
 // Line 116: change from
 if ( $eligibility_checks[ $feature_key ]() ) {
 // to
 if ( isset( $eligibility_checks[ $feature_key ] ) && $eligibility_checks[ $feature_key ]() ) {
```

Dinamiko (approved):

PR #4150, opened 2026-03-06
Merchant not onboarded when upgrading from previous version (5994)

Dinamiko:
Upgrading from older plugin versions (e.g. v3.0.3) leaves merchants stuck as "Not Connected" because:                                                                                                                
                                                                                                                                                                                                                       
- Migration fails silently: The `seller_status()` API call fails, the exception bubbles up, and the migration-done flag never gets set — so migration never retries.
- Seller type stays "unknown": The frontend hides Card Processing and APM groups since it checks `isBusinessSeller`.
- Cache poisoning: The API call during migration uses the wrong environment (production instead of sandbox), caching `false` for vaulting eligibility for 1 hour.

### PR Changes

- Resilient migration — Try-catch around the API call with fallback defaults; each sub-migration isolated so one failure doesn't block others; done-flag always set; admin_init retry for stuck merchants.
- Deferred seller type resolution — New `SellerTypeResolver` service + `admin_init` hook that re-queries the API for connected merchants with unknown seller type, also backfills empty merchant_country.
- Cache clearing — Fires `woocommerce_paypal_payments_clear_apm_product_status` after migration and after successful seller type resolution to flush stale entries.

### DIFF
--- a/modules/ppcp-settings/src/SettingsModule.php
+++ b/modules/ppcp-settings/src/SettingsModule.php
@@ -41,7 +41,9 @@ use WooCommerce\PayPalCommerce\WcGateway\Gateway\OXXO\OXXO;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
 use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayUponInvoice\PayUponInvoiceGateway;
 use WooCommerce\PayPalCommerce\Settings\Service\SettingsDataManager;
+use Exception;
 use WooCommerce\PayPalCommerce\Settings\DTO\ConfigurationFlagsDTO;
+use WooCommerce\PayPalCommerce\Settings\DTO\MerchantConnectionDTO;
 use WooCommerce\PayPalCommerce\Settings\Enum\ProductChoicesEnum;
 use WooCommerce\PayPalCommerce\Settings\Data\GeneralSettings;
 use WooCommerce\PayPalCommerce\Settings\Data\PaymentSettings;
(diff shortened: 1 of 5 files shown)


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -851,4 +888,75 @@ protected function is_gateway_enabled( string $gateway_name ): bool {
...
+	 * during migration), this retries the seller status call and persists the
+	 * resolved type. Also backfills empty merchant_country.
+	 *
+	 * @param FailureRegistry    $failure_registry    The failure registry.
+	 * @param GeneralSettings    $general_settings    The general settings.
+	 * @param PartnersEndpoint   $partners_endpoint   The partners endpoint.
+	 * @param SellerTypeResolver $seller_type_resolver The seller type resolver.
+	 * @param LoggerInterface    $logger              The logger.
+	 */
+	private function resolve_unknown_seller_type(
```
Narek13: I think this method can be a part of `SellerTypeResolver` class which you introduced, we dont need this in the module file

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -851,4 +888,75 @@ protected function is_gateway_enabled( string $gateway_name ): bool {
...
+	}
+
+	/**
+	 * Checks whether seller type resolution is needed.
+	 *
+	 * @param FailureRegistry $failure_registry The failure registry.
+	 * @param GeneralSettings $general_settings The general settings.
+	 * @return bool True if the merchant is connected but has an unknown seller type.
+	 */
+	private function needs_seller_type_resolution( FailureRegistry $failure_registry, GeneralSettings $general_settings ): bool {
```
Narek13: I think this method can be a part of `SellerTypeResolver` class which you introduced, we dont need this in the module file

Narek13 (commented):
Thanks for working on this Emili. I left some suggestions regarding consilidating all "seller type" methods. Please let me know if you have any questions.

Dinamiko:
@Narek13 good suggestions, done, thanks!

Narek13 (approved):
LGTM. Thanks

PR #4155, opened 2026-03-09
BCDC button under PayPal buttons not active after migration (5667)

Dinamiko:
When migrating from legacy settings with BCDC configured under PayPal buttons (not as a separate gateway), the BCDC gateway is not activated. Merchants must manually enable it on the Payment Methods page.         
                                                            
This PR adds `toggle_method_state( CardButtonGateway::ID, true )` in `PaymentSettingsMigration::migrate()` alongside the existing override flag.


Narek13 (approved):
Thanks for fixing this. LGTM

PR #4157, opened 2026-03-09
Migration - Capture virtual only orders when enabled not preserved when we migrate to new ui (5952)

Dinamiko:
When migrating from the legacy settings UI to the new one, the "Capture Virtual-Only Orders" setting was always lost. The `intent` case in the migration switch was deriving `capture_virtual_orders` from the intent value (`$value === 'capture'`), which is always `false` when intent is `'authorize'`, the only scenario where the setting matters. Additionally, the actual legacy key `capture_for_virtual_only` was never included in the migration map, so its value was never read.                                                                                                                                                                   
                                                                                                                                                                                                                       
This PR adds `capture_for_virtual_only` to the migration key map so it gets migrated directly via the default case, and removed the incorrect derivation from the intent case.


Narek13 (approved):
LGTM, thanks for fixing this

PR #4159, opened 2026-03-10
Migration - Case when BCDC is part of PayPal or as separated gateway Pay later messaging tab is not present (5929)

Dinamiko:
We were incorrectly checking for ACDC eligible on Pay Later Messaging, it should not rely on card eligibility but only on vault enabled and allowed countries.


Narek13 (approved):
LGTM, thanks

PR #4160, opened 2026-03-10
Migration - Enable BCDC when merchant have it active and is not ACDC eligible (6010)

Dinamiko:
After upgrading from v2.9.6, the "Credit and debit card payments" (CardButtonGateway) disappears for non-ACDC merchants. In the legacy version, the card button was part of PayPal Smart Buttons and shown by default. The migration only enabled CardButtonGateway for ACDC-eligible merchants using BCDC, leaving non-ACDC merchants without it.                                                                              
   
### PR Changes                                                                                                                                                                                          
  - Fixed the migration to enable CardButtonGateway for all merchants who had the card funding source active (not in `disable_funding`), regardless of ACDC eligibility
  - Added a retroactive fix for merchants who already migrated with the bug (versions ≤ 3.4.1)

### DIFF
--- a/modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
+++ b/modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
@@ -76,8 +76,14 @@ class PaymentSettingsMigration implements SettingsMigrationInterface {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
LGTM. I have briefly tested (only this case in isolation) and it works for me. Thank you

PR #4176, opened 2026-03-12
ACDC - For unsupported countries for save payment method check box toggle visible for ACDC (5629)

Dinamiko:
The "Save your card" checkbox appeared in ACDC checkout for countries ineligible for vaulting (e.g. Mexico). Both blocks and classic checkout paths only checked the DB setting, not country eligibility.            

### PR Changes                                                                                                                                                                                                      
  - Gate `is_vaulting_enabled` on `save-payment-methods.eligible` in `AdvancedCardPaymentMethod` (blocks).
  - Add `vault_v3_enabled` check in `SmartButton::render_dcc_wrapper()` (classic).

### DIFF
--- a/modules/ppcp-blocks/services.php
+++ b/modules/ppcp-blocks/services.php
@@ -53,7 +53,8 @@ return array(
 				return $container->get( 'button.smart-button' );
 			},
 			$container->get( 'settings.settings-provider' ),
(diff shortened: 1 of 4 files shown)


Narek13 (approved):
LGTM. Thanks

PR #4177, opened 2026-03-12
Fix BCDC override flag (6024)

stracker-phil:
### Description

In a recent preparation to drop the legacy Settings class, we introduced a regression concerning the BCDC override flag.

#### Bug

The BCDC override flag should mean "Do not use ACDC, but only display BCDC" - due to a this bug, the decision was inverted, resulting in the flag being completely ignored:

- Flag present: Incorrectly treated as "Use ACDC, not BCDC"
- Flag absent: Correctly treated as "Use ACDC"

#### Steps to Test

Add the BCDC override flag to the DB and verify that no "Online Card Payments" are displayed in the plugins "Payment Methods" settings tab

The official way:
1. Onboard in the legacy UI with an ACDC merchant
2. Add the standard card payment button (the black button) and ensure to keep the ACDC checkbox unchecked
3. Update to the new UI → during migration, the BCDC Override flag is written to the DB

The fast way:
`wp option add "woocommerce_paypal_payments_override_acdc_status_with_bcdc" "1"`

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/DCCProductStatus.php
+++ b/modules/ppcp-wc-gateway/src/Helper/DCCProductStatus.php
@@ -35,8 +35,6 @@ class DCCProductStatus extends ProductStatus {
 	}
 
 	public function check_local_state( bool $skip_filters = false ): ?bool {
-		$state = null;
-
 		if ( ! $skip_filters ) {
 			/**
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
Thanks a lot for fixing this and for adding Unit Tests

PR #4179, opened 2026-03-12
Fix early country access (5713)

stracker-phil:
### Description

`CardFieldsModule::run()` previously resolved `card-fields.eligible` immediately during `plugins_loaded`, which eagerly called `wc_get_base_location()` before themes and late-loading plugins had a chance to filter the base country.
                           
The eligibility check is now deferred to `init` using the `card-fields.eligibility.check` callable, ensuring WooCommerce and the theme are fully initialized before the country is read.


Narek13 (approved):
LGTM, thanks

PR #4180, opened 2026-03-13
BCDC button under PayPal buttons not active after migration (5667 v2)

Dinamiko:
Upgrading from 2.9.6 (before new UI code was added) with BCDC + Google Pay fails because the BCDC migration override is never set. The API-based DCC product status check (`dcc_status->is_active()`) fails on first run after upgrade due to an empty transient cache (2.9.6 used a different caching mechanism), causing `is_bcdc_enabled_for_acdc_merchant()` to return false. Additionally, when the override is set, the `toggle_payment_gateways` handler incorrectly disables Google Pay and Apple Pay.                                                                                                                                                                       

This PR adds a legacy-settings-based fallback that detects BCDC usage without relying on the API cache (checks `disable_funding` and `dcc_enabled` from `woocommerce-ppcp-settings`). Also protected Google Pay and Apple Pay from being disabled when the BCDC override is active.


Narek13 (approved):

modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -761,9 +781,17 @@ function ( PaymentSettings $payment_methods, ConfigurationFlagsDTO $flags ) use
...
+
 					$payment_methods->toggle_method_state( CreditCardGateway::ID, false );
-					$payment_methods->toggle_method_state( ApplePayGateway::ID, false );
-					$payment_methods->toggle_method_state( GooglePayGateway::ID, false );
+
+					// BCDC-override merchants may have Google Pay and Apple Pay configured
+					// alongside BCDC. Only disable these if NOT in BCDC-override mode.
+					if ( ! $check_override() ) {
+						$payment_methods->toggle_method_state( ApplePayGateway::ID, false );
+						$payment_methods->toggle_method_state( GooglePayGateway::ID, false );
```
stracker-phil: unclear: is this still relevant after merging #4186?

stracker-phil (approved):
LGTM, but the last condition seems unnecessary after PR #4186, but not entirely sure.

Apple Pay and Google Pay possibly still need ACDC _eligibility_ and disabling them here could be justified.

PR #4182, opened 2026-03-13
Migration - When APM-S are part of PayPal gateway - What to expect after switching to New UI (5694)

Dinamiko:
When merchants had `allow_local_apm_gateways` set to false in the legacy UI (APMs rendered inside the PayPal Smart Button), the migration to the new UI skipped enabling them as separate gateways. Since the new UI has no "inside PayPal" mode, those APMs ended up disabled everywhere,  invisible on checkout.

This PR removes the `allow_local_apm_gateways` guard from the migration so APMs are always enabled based on disable_funding. Added a retroactive fix hook for merchants who already migrated with the bug.


Narek13 (approved):

Issue #4170, opened 2026-03-11
PayPal payments not captured for WooCommerce Deposits follow-up orders (pending-deposit status)


namiokuzono:
## Summary

When customers pay the **remaining balance** of a [WooCommerce Deposits](https://woocommerce.com/products/woocommerce-deposits/) order via PayPal (order-pay page), the payment is approved in PayPal but **never captured**. The order stays in "Pending Deposit Payment" and no funds are taken. The failure seems to be caused by WooCommerce PayPal Payments **skipping payment processing** when the order status is `pending-deposit`, because it only allows literal `pending` or `on-hold` statuses.

## Environment

- **WooCommerce PayPal Payments:** 3.4.1 (and likely earlier)
- **WooCommerce Deposits:** 2.4.3
- **WooCommerce:** 10.6+
- **Reproduced on:** Fresh WordPress install with WooCommerce, WooCommerce Deposits, and WooCommerce PayPal Payments (sandbox).

## Steps to reproduce

1. Install and activate WooCommerce, WooCommerce Deposits, and WooCommerce PayPal Payments. Connect PayPal (sandbox or production).
2. Create a product with deposits enabled (e.g. pay 25% now, 75% later).
3. As a customer, place an order and complete the **initial deposit** with PayPal. Order # 1 goes to "Partially Paid" (or similar); a follow-up order # 2 is created with status **"Pending Deposit Payment"** (`pending-deposit`).
4. As the customer, open the "Pay" link for the remaining balance (from email or My Account → Orders) so you land on the order-pay page for order # 2.
5. Choose PayPal and complete approval in the PayPal popup/redirect.
6. Return to the store after approving. 

## Expected behavior

- The plugin captures the approved PayPal payment and marks order # 2 as paid (e.g. "Partially Paid" or "Completed" per Deposits flow).
- Funds are debited from the customer’s PayPal account.

## Actual behavior

- The customer sees a success-style experience, but **no capture is performed**.
- Order # 2 remains in "Pending Deposit Payment".
- No money is taken from the customer’s PayPal account.
- In logs:  
  `Order # 52 has status "pending-deposit", skipping payment processing.`

---
Any code analysis is done by AI. If not helpful, please disregard them.
---

## Root cause 
In **PayPal Payments** (`modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php`), `verify_order_can_be_processed()` decides whether to run the capture flow when the customer returns from PayPal. It used:

```php
[...]

PR #4185, opened 2026-03-13
Remove Order Status Requirement for PayPal Payment Processing (6042)

Dinamiko:
This PR removes the `OrderStatusHelper::is_awaiting_payment()` guard that restricted PayPal payment processing to orders with pending, on-hold, or failed statuses. Payments can now be processed regardless of WC order status.

Existing safeguards (transaction ID check, atomic processing lock, idempotent `payment_complete()`) prevent double-processing. The reversal webhook handler now uses `WC_Order::is_paid()` to protect already-paid  orders from being marked failed.    

Relates to #4170.


Narek13 (approved):
LGTM

PR #4194, opened 2026-03-17
Migration - PUI and OXXO do not persist after upgrade from 2.9.6 to 4.0.0 (6046)

Dinamiko:
This PR ensures PUI and OXXO are enabled after migrating from 2.9.6 to 4.0.0.

### DIFF
--- a/modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
+++ b/modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
@@ -65,11 +65,8 @@ class PaymentSettingsMigration implements SettingsMigrationInterface {
 		$this->local_apms        = $local_apms;
 		$this->dcc_configuration = $dcc_configuration;
 
-		$pui_option               = get_option( 'woocommerce_' . PayUponInvoiceGateway::ID . '_settings', array() );
-		$this->legacy_pui_enabled = is_array( $pui_option ) && ( $pui_option['enabled'] ?? 'no' ) === 'yes';
-
-		$oxxo_option               = get_option( 'woocommerce_' . OXXO::ID . '_settings', array() );
-		$this->legacy_oxxo_enabled = is_array( $oxxo_option ) && ( $oxxo_option['enabled'] ?? 'no' ) === 'yes';
+		$this->legacy_pui_enabled  = $this->payment_settings->is_method_enabled( PayUponInvoiceGateway::ID );
+		$this->legacy_oxxo_enabled = $this->payment_settings->is_method_enabled( OXXO::ID );
 	}
 
 	public function migrate(): void {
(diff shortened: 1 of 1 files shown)


modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
```diff
@@ -58,6 +64,12 @@ public function __construct(
 		$this->dcc_status        = $dcc_status;
 		$this->local_apms        = $local_apms;
 		$this->dcc_configuration = $dcc_configuration;
+
+		$pui_option               = get_option( 'woocommerce_' . PayUponInvoiceGateway::ID . '_settings', array() );
+		$this->legacy_pui_enabled = is_array( $pui_option ) && ( $pui_option['enabled'] ?? 'no' ) === 'yes';
```
Narek13: Why not using [this method](https://github.com/woocommerce/woocommerce-paypal-payments/blob/88a031c0d1d985bc7f3dce054140189f7644c41b/modules/ppcp-settings/src/Data/PaymentSettings.php#L109)?

modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
```diff
@@ -58,6 +64,12 @@ public function __construct(
...
+
+		$oxxo_option               = get_option( 'woocommerce_' . OXXO::ID . '_settings', array() );
+		$this->legacy_oxxo_enabled = is_array( $oxxo_option ) && ( $oxxo_option['enabled'] ?? 'no' ) === 'yes';
```
Narek13: Same here

Narek13 (changes requested):

modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
Dinamiko: @Narek13 good catch, done!

modules/ppcp-settings/src/Service/Migration/PaymentSettingsMigration.php
```diff
@@ -101,6 +110,14 @@ public function migrate(): void {
 			}
 		}
 
+		if ( $this->legacy_pui_enabled ) {
```
Narek13: Thanks for the changes. Another question after the latest changes:
Why do we need this local properties `protected bool $legacy_pui_enabled` and `protected bool $legacy_oxxo_enabled` ? we already have `payment_settings` as dependency so can directly use it


```suggestion
		if ( $this->payment_settings->is_method_enabled( PayUponInvoiceGateway::ID ) ) {
```

Dinamiko: @Narek13 done, much cleaner now, thanks!

Narek13 (approved):
Thanks a lot for making all the changes. LGTM

PR #4198, opened 2026-03-18
Migration banner — remove Fastlane US-only text (6056)

Narek13:
### Issue Description

Since v4.0.0, Fastlane is available for all ACDC countries, not just the US. The migration banner text still referenced "US only", which is no longer accurate and should be removed.

### PR Description

- Updated the migration banner description to remove the Fastlane US-only qualifier
- Minor CSS improvement to better align the description text

**New banner text:**

> Your card payments can do more. Unlock lower processing fees, advanced fraud protection, and a faster checkout with Apple Pay, Google Pay, and Fastlane one-click checkout.

Screenshot 2026-03-18 at 14 46 30

### DIFF
--- a/modules/ppcp-settings/resources/css/components/screens/settings/_migration-banner.scss
+++ b/modules/ppcp-settings/resources/css/components/screens/settings/_migration-banner.scss
@@ -36,7 +36,7 @@
 
 	.ppcp-r-settings-migration-banner__icon {
 		text-align: center;
-		flex: 0.6;
+		width: 35%;
 		height: 100%;
 		position: relative;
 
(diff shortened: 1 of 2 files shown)


InpsydeNiklas (approved):

modules/ppcp-settings/resources/css/components/screens/settings/_migration-banner.scss
```diff
@@ -36,7 +36,7 @@
 
 	.ppcp-r-settings-migration-banner__icon {
 		text-align: center;
-		flex: 0.7;
+		flex: 0.6;
```
stracker-phil: nitpick:

`flex: 0.6` makes the real width of the element difficult to anticipate. Using a width attribute would be more predictable.

Note: `flex: 0.6` means: This element grows by 60% compared to an element with `flex: 1`; ie. it's meaning is relative to how many other elements are in the row: With 2 elements, flex 0.6 roughly matches 33% of available width.


Suggestion:

```css
.ppcp-r-settings-migration-banner__icon {
  text-align: center;
  width: 35%;  // ← clear
  height: 100%;
  position: relative;

  ...
}
```

PR #4203, opened 2026-03-19
Migrate "capture for virtual only" (6026)

stracker-phil:
### Description

Adds unit tests to confirm correct migration of the `capture_virtual_orders` flag from the legacy settings to the new React settings.

Note: The issue itself was already resolved in commit 2e5547a24 - this PR only adds unit tests to prevent future regressions

#### Timing Error

This PR also fixes a one-time fatal error that happened when auto migrating legacy to React. The issue was caused by accessing some WooCommerce functions too early (during `init` at prio `-1`, while WooCommerce initializes at `init` prio `10`)

Testing this issue is difficult, but basically follows those steps:
1. Start with plugin version 2.9.6, ensure there are NO react UI settings in the DB; i.e. a "clean legacy DB"
2. Install the latest plugin version to trigger the auto-migration during update
3. Directly after the plugin update you should see a fatal error, which disappears on page reload

https://github.com/user-attachments/assets/c9ca08ed-d5e3-48a9-b102-86ed24de05fa

### DIFF
--- a/modules/ppcp-settings/src/Data/OnboardingProfile.php
+++ b/modules/ppcp-settings/src/Data/OnboardingProfile.php
@@ -235,9 +235,16 @@ class OnboardingProfile extends AbstractDataModel {
 	public function set_gateways_synced( bool $synced ): void {
 		$this->data['gateways_synced'] = $synced;
 
-		// If enabling the flag, trigger the action.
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
Thanks a lot for adding tests. LGTM

PR #4204, opened 2026-03-19
Migration - PUI and OXXO do not persist after upgrade from 2.9.6 to 4.0.0 (6046 v2)

Dinamiko:
This PR reverts #4194 to the first commit changes which fixed the issue,  the initial fix works because the `PaymentSettingsMigration` object is constructed before `MigrationManager::migrate()` is called. So `get_option()` in the constructor captures the real value before the sync overwrites it.


Narek13 (approved):
Thanks

PR #4207, opened 2026-03-19
Harden plugin init to prevent errors during update (6063)

stracker-phil:
### Description

Skip plugin bootstrap on `update.php` to prevent fatal errors during manual zip uploads ("Upload Plugin").

When WordPress replaces plugin files mid-request, old in-memory hooks can autoload new class files that changed during the update (different constructor arguments, changed base class, etc) with an incompatible, old parent classes still in memory. The `$pagenow` guard prevents this by not initializing the plugin on that page.

Also contains a small refactor/clean-up of `init()` to return bool indicating whether the container is available, replacing the separate `is_woocommerce_activated()` re-check in `plugins_loaded`.
                                    
### How to test

Not possible to test with this release, it catches _future_ errors

Only verify that the plugin can be updated and initializes as expected


Dinamiko (approved):

stracker-phil:
Replaced by #4211

PR #4208, opened 2026-03-20
On fresh site after payment with PUI orders stuck on status on hold as webhook are not registered (6062)

Dinamiko:
This PR ensures webhooks are registered on new plugin installs, also resubscribes on plugin upgrades for PUI and OXXO.

### DIFF
--- a/modules/ppcp-webhooks/src/WebhookModule.php
+++ b/modules/ppcp-webhooks/src/WebhookModule.php
@@ -74,6 +74,13 @@ class WebhookModule implements ServiceModule, FactoryModule, ExecutableModule {
 			}
 		);
 
+		/**
+		 * Auto-recover missing webhook registration on admin pages.
+		 *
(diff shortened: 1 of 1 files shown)


Narek13 (approved):
Thanks. I haven't tested but code wise looks good to me. Only a minor suggestion which you are welcome to ignore:
I would remove the inline comments and add an action level docblock comments to describe why these actions are added and what are they doing. 
Examples: 
- [Auto-trigger settings migration to new UI on plugin update](https://github.com/woocommerce/woocommerce-paypal-payments/blob/9ad13038e75d44b982935f69babf0261174ed694/modules/ppcp-settings/src/SettingsModule.php#L76-L91)
- [Ensure BCDC remains functional in branded-only mode.](https://github.com/woocommerce/woocommerce-paypal-payments/blob/9ad13038e75d44b982935f69babf0261174ed694/modules/ppcp-settings/src/SettingsModule.php#L889-L898)

PR #4209, opened 2026-03-20
Plugin upgrade - Venmo (disabled in legacy) is enabled after upgrade to new version (6066)

Dinamiko:
After upgrading , Venmo appeared on checkout even when the merchant had disabled it via "Disable Alternative Payment Methods" in legacy settings. The migration logic correctly set Venmo as disabled, but `SmartButton.php` unconditionally added `venmo` to the PayPal SDK's `enable-funding` parameter, which takes precedence over `disable-funding`.                     

This PR fixes it by conditionally adding Venmo to `enable-funding` only when `venmo_enabled()` returns `true`.


Narek13 (approved):
LGTM, thanks for fixing this

PR #4210, opened 2026-03-20
Migration - PayPal title and description not saved after migration (6068)

Dinamiko:
Custom PayPal and Credit Card gateway titles/descriptions set by merchants were lost when upgrading from legacy UI, because the settings migration didn't copy `title/description/dcc_gateway_title/dcc_gateway_description` from the legacy `woocommerce-ppcp-settings` option to the new WooCommerce gateway options.

This PR adds `title` and `description` migration for both gateways in `PaymentSettingsMigration::migrate()`, so custom values are now preserved during upgrade.


Narek13 (approved):
LGTM, thanks

PR #4213, opened 2026-03-20
refactor(plugin): remove `woocommerce-ppcp-is-new-merchant` flag option (6067)

Narek13:
### Issue Description

The `woocommerce-ppcp-is-new-merchant` flag is no longer needed but was still being set and read in at least two places in the plugin, as well as referenced in unit tests.

### PR Description

Remove all code that sets or reads the `woocommerce-ppcp-is-new-merchant` option, including the `gateway_migrate` hook that wrote it on first install and any conditional logic gated on it. The option itself is intentionally left in the database for backwards compatibility with any existing installations that may have it stored.

Unit tests related to this option are retained, as the migration manager still deletes it during cleanup and that behaviour continues to need test coverage for the foreseeable future.


stracker-phil (approved):
Good call on keeping the `MigrationManager` + relevant tests in place.

LGTM - codebase only _deletes_ the option during migration, but no condition reads or uses that flag in the 4.x code anymore

---

UPD: It turns out that this PR is complete. We need to remove _all_ mentions of the `woocommerce-ppcp-is-new-merchant` flag - this codebase should not read, write nor delete that flag.

Key scenario: It must not be deleted to preserve the new UI during downgrade from 4.0 to 3.4 for new merchants

stracker-phil (approved):

stracker-phil:
Replaced by #4235

PR #4215, opened 2026-03-23
Fix merchant config broken after upgrade when frontend page loads first

Dinamiko:
After upgrading from 2.9.6 to 4.0.0, loading a frontend page (e.g. blocks checkout) before visiting admin settings breaks merchant config: features disappear, wrong connection label, BCDC disabled. Admin visit first works fine because SellerTypeResolver only ran on admin_init.

### PR Changes
  - Pre-populate `GeneralSettings` with legacy credentials before migration so API calls have working credentials
  - Move `SellerTypeResolver` from `admin_init` to init so it also runs on frontend pages
  - Skip gateway sync during migration to prevent reset of payment method states


Narek13 (approved):
LGTM. Thanks

PR #4216, opened 2026-03-23
refactor(button): consolidate disabled funding sources card decision (5914)

Narek13:
### Issue Description

The logic for managing the `card` funding source was fragmented across multiple modules using an add-then-remove pattern:

- `DisabledFundingSources::apply_context_rules()` added `card` to the disabled list
- A filter in `CardFieldsModule` removed it when ACDC was enabled
- A filter in `SettingsModule` re-added it when BCDC was not eligible
- A filter in `SettingsModule::apply_branded_only_limitations()` removed it again for BCDC in branded-only mode

This made the final state of the `card` funding source difficult to trace and easy to break. The branded-only filter was needed specifically to undo decisions made earlier in the chain.

### PR Description

Consolidates the `card` funding-source decision into a single method —> `DisabledFundingSources::should_disable_card()` which evaluates all relevant conditions upfront and removes the need for any post-hoc corrections in other modules.

**`DisabledFundingSources`:**
- Replaces `apply_context_rules()` with `apply_card_rule()` + `should_disable_card()`
- `should_disable_card()` handles all cases: non-checkout pages, block vs classic checkout, ACDC vs BCDC state, Mexico country override
- Uses `is_acdc_enabled()` instead of `use_acdc()` to correctly handle transitions such as branded-only → normal mode, where ACDC capability exists but the gateway is disabled

**`CardFieldsModule`:**
- Removes the `woocommerce_paypal_payments_sdk_disabled_funding_hook` filter that re-enabled `card` for ACDC — now handled by `should_disable_card()`

**`SettingsModule`:**
- Removes the `woocommerce_paypal_payments_sdk_disabled_funding_hook` filter that disabled `card` when BCDC was not eligible which is now handled by `should_disable_card()`
- Removes the branded-only `woocommerce_paypal_payments_sdk_disabled_funding_hook` filter that re-enabled `card` for BCDC because `is_bcdc_enabled()` correctly reflects the `CardButtonGateway` toggle state, so no correction is needed


Dinamiko (approved):

stracker-phil (approved):
LGTM

Also, excellent documentation 👍

PR #4222, opened 2026-03-26
fix(settings): Restrict Venmo, Google Pay and Apple Pay to locations enabled in styling settings (6079)

Narek13:
## Description

Fixes per-location visibility of Venmo, Google Pay, and Apple Pay in the New UI. All three were visible on frontend pages even when disabled for specific locations in the styling tab.

## Root Cause

The per-location `button_styling()->methods` array was never consulted when deciding whether to render or register these payment methods. Each one only checked its global enabled toggle, causing them to appear on all pages regardless of styling tab configuration.

## Changes

**Venmo (SmartButton / DisabledFundingSources)**
- `DisabledFundingSources::get_sources_from_settings()` now adds venmo to disabled funding sources when it is not in the enabled methods for the current context
- `SmartButton::url_params()` gates venmo in `enable-funding` on the current context's methods array, not just the global toggle

**Google Pay**
- `GooglePayButton::is_enabled()` now checks `button_styling()` methods for the current context in addition to the global toggle
- `GooglePayButton::render()` refactored — removed per-location flags, replaced repeated hook registration blocks with a hook map and loop
- `GooglePayButton\BlocksPaymentMethod::is_active()` checks `button_styling()` for the current context; `enabled` in `get_payment_method_data()` now uses `is_active()` instead of `smartButtonsEnabled`
- `woocommerce_payment_gateways` filter skips Google Pay registration when disabled for the current location
- `woocommerce_available_payment_gateways` filter removes Google Pay from classic checkout when disabled for that location

**Apple Pay**
- Same fixes applied as Google Pay across `ApplePayButton::is_enabled()`, `render()`, `BlocksPaymentMethod`, and both gateway filters

## Testing

1. Onboard with a US merchant on New UI
2. In the styling tab, disable Google Pay / Apple Pay / Venmo for a specific page (e.g. product page)
3. Navigate to that page and confirm the button is not visible
4. Enable it again and confirm it reappears
5. Verify checkout and cart pages are unaffected when only product page is disabled
6. Repeat for block checkout and block cart

### DIFF
--- a/modules/ppcp-applepay/services.php
+++ b/modules/ppcp-applepay/services.php
@@ -157,7 +157,6 @@ return array(
 			$container->get( 'applepay.asset_getter' ),
 			$container->get( 'ppcp.asset-version' ),
 			$container->get( 'applepay.data_to_scripts' ),
-			$container->get( 'wcgateway.settings.status' ),
 			$container->get( 'button.helper.cart-products' ),
 			$container->get( 'button.helper.context' )
 		);
--- a/modules/ppcp-applepay/services.php
+++ b/modules/ppcp-applepay/services.php
@@ -158,7 +158,8 @@ return array(
 			$container->get( 'ppcp.asset-version' ),
 			$container->get( 'applepay.data_to_scripts' ),
 			$container->get( 'wcgateway.settings.status' ),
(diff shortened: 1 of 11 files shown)


Narek13:
wrong branch name

PR #4223, opened 2026-03-26
fix(settings): Restrict Venmo, Google Pay and Apple Pay to locations enabled in styling settings (6079)

Narek13:
## Description

Fixes per-location visibility of Venmo, Google Pay, and Apple Pay in the New UI. All three were visible on frontend pages even when disabled for specific locations in the styling tab.

## Root Cause

The per-location `button_styling()->methods` array was never consulted when deciding whether to render or register these payment methods. Each one only checked its global enabled toggle, causing them to appear on all pages regardless of styling tab configuration.

## Changes

**Venmo (SmartButton / DisabledFundingSources)**
- `DisabledFundingSources::get_sources_from_settings()` now adds venmo to disabled funding sources when it is not in the enabled methods for the current context
- `SmartButton::url_params()` gates venmo in `enable-funding` on the current context's methods array, not just the global toggle

**Google Pay**
- `GooglePayButton::is_enabled()` now checks `button_styling()` methods for the current context in addition to the global toggle
- `GooglePayButton::render()` refactored — removed per-location flags, replaced repeated hook registration blocks with a hook map and loop
- `GooglePayButton\BlocksPaymentMethod::is_active()` checks `button_styling()` for the current context; `enabled` in `get_payment_method_data()` now uses `is_active()` instead of `smartButtonsEnabled`
- `woocommerce_payment_gateways` filter skips Google Pay registration when disabled for the current location
- `woocommerce_available_payment_gateways` filter removes Google Pay from classic checkout when disabled for that location

**Apple Pay**
- Same fixes applied as Google Pay across `ApplePayButton::is_enabled()`, `render()`, `BlocksPaymentMethod`, and both gateway filters

## Testing

1. Onboard with a US merchant on New UI
2. In the styling tab, disable Google Pay / Apple Pay / Venmo for a specific page (e.g. product page)
3. Navigate to that page and confirm the button is not visible
4. Enable it again and confirm it reappears
5. Verify checkout and cart pages are unaffected when only product page is disabled
6. Repeat for block checkout and block cart

### DIFF
--- a/modules/ppcp-applepay/src/ApplepayModule.php
+++ b/modules/ppcp-applepay/src/ApplepayModule.php
@@ -18,6 +18,7 @@ use WooCommerce\PayPalCommerce\Applepay\Assets\PropertiesDictionary;
 use WooCommerce\PayPalCommerce\Button\Assets\ButtonInterface;
 use WooCommerce\PayPalCommerce\Button\Assets\SmartButtonInterface;
 use WooCommerce\PayPalCommerce\Applepay\Helper\AvailabilityNotice;
+use WooCommerce\PayPalCommerce\Button\Helper\Context;
 use WooCommerce\PayPalCommerce\Settings\Data\Definition\FeaturesDefinition;
 use WooCommerce\PayPalCommerce\WcGateway\Helper\Environment;
 use WooCommerce\PayPalCommerce\Vendor\Inpsyde\Modularity\Module\ExecutableModule;
--- a/modules/ppcp-applepay/src/ApplepayModule.php
+++ b/modules/ppcp-applepay/src/ApplepayModule.php
@@ -132,12 +132,23 @@ class ApplepayModule implements ServiceModule, ExecutableModule {
 				$settings = $c->get( 'settings.settings-provider' );
 				assert( $settings instanceof SettingsProvider );
 
-				if ( $settings->applepay_enabled() ) {
-					$applepay_gateway = $c->get( 'applepay.wc-gateway' );
-					assert( $applepay_gateway instanceof WC_Payment_Gateway );
-
-					$methods[] = $applepay_gateway;
+				if ( ! $settings->applepay_enabled() ) {
+					return $methods;
 				}
+
+				$context = $c->get( 'button.helper.context' );
+				assert( $context instanceof Context );
+
+				$page_methods = $settings->button_styling( $context->context() )->methods;
+
+				if ( ! in_array( ApplePayGateway::ID, $page_methods, true ) ) {
+					return $methods;
+				}
+
+				$applepay_gateway = $c->get( 'applepay.wc-gateway' );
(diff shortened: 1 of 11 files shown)


modules/ppcp-googlepay/src/GooglepayModule.php
```diff
@@ -187,11 +188,57 @@ static function ( $methods ) use ( $c ) {
...
 				assert( $settings instanceof SettingsProvider );
 
-				if ( $settings->googlepay_enabled() ) {
-					$googlepay_gateway = $c->get( 'googlepay.wc-gateway' );
-					assert( $googlepay_gateway instanceof WC_Payment_Gateway );
+				if ( ! $settings->googlepay_enabled() ) {
+					return $methods;
+				}
+
+				$context = $c->get( 'button.helper.context' );
```
danieldudzic, line 195: I have a concern that `woocommerce_payment_gateways` can fire before the main WP query is parsed (it's applied inside WC_Payment_Gateways::init(), which runs whenever the singleton is first instantiated potentially during init, REST API, or cron). At that point, `Context::context()` falls through to the default 'mini-cart'. If a merchant has Google Pay disabled for `mini-cart` but enabled for checkout, the gateway won't be registered at all. No?

modules/ppcp-applepay/src/ApplepayModule.php
```diff
@@ -132,11 +133,57 @@ static function ( $methods ) use ( $c ) {
...
+				if ( ! in_array( ApplePayGateway::ID, $page_methods, true ) ) {
+					return $methods;
+				}
+
+				$applepay_gateway = $c->get( 'applepay.wc-gateway' );
+				assert( $applepay_gateway instanceof WC_Payment_Gateway );
+
+				$methods[] = $applepay_gateway;
+
+				return $methods;
```
danieldudzic, line 154: Same concern as Google Pay.

danieldudzic (approved):
Clean approach - the consistent use of `button_styling()->methods` and the hook-map refactoring are nice improvements. One concern about the `woocommerce_payment_gateways` filter below.

Pre-approving as the rest looks solid!

modules/ppcp-googlepay/src/GooglepayModule.php
Narek13: Thanks @danieldudzic, that's a valid theoretical concern. In practice, `woocommerce_payment_gateways` is typically fired during the frontend request lifecycle after the main query is parsed and most commonly triggered by `WC()->payment_gateways()` calls in checkout, cart, or payment-related templates. I've tested across checkout, cart, product page, and pay-order page and haven't been able to reproduce the issue.
That said, if you can point to a specific scenario where the singleton gets instantiated early (e.g. a particular REST endpoint, cron job, or init hook) I'm happy to investigate further and add a guard if needed. Do you have a concrete case in mind?

PR #4220, opened 2026-03-25
PayPal Subscriptions renewals incorrectly processed (5580)

Dinamiko:
Fixes #4191.

`RenewalHandler::renew()` receives a renewal order but passes its ID to `wcs_get_subscription()`, which expects a subscription ID. This causes `wcs_get_subscription()` to return `false`, silently skipping the PayPal Subscriptions API check and failing renewals with "Subscription could not be loaded."                                                                                                                         
                                                                                                                                                                                                                       
This PR replaces `wcs_get_subscription($wc_order->get_id())` with `wcs_get_subscriptions_for_renewal_order($wc_order)`, which is the correct WooCommerce Subscriptions API for retrieving subscriptions from a renewal order.

Issue #4191, opened 2026-03-16
"Subscription could not be loaded." when doing "Retry Renewal Payment"


qvera:
### Describe the Bug
When trying to do "Retry Renewal Payment" on some subscription renewals, it will quietly fail with log message "Subscription could not be loaded."

The backtrace indicates, that the issue is within ``/plugins/woocommerce-paypal-payments/modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php:460``, when using function ``wcs_get_subscription()``.
I have tried running the function in local env and copied DB, which resulted in exception being thrown as $order is renewal and not subscription. 

I was able to get $subscription through this:

```php
$subscription_id = $order->get_meta('_subscription_renewal');
if ($subscription_id) {
    $subscription = wcs_get_subscripton($subscription_id);
}
```
and it resulted in the subscription being returned correctly.

I am not sure if something in DB got set wrongly, or what happened, but it happened on 19 of subscriptions, which until now, were completely fine. 

### Expected Behavior

Retry renewal payment goes through with loaded subscription and tries to renew subscription.

### Actual Behavior

Silent fail with "Subscription could not be loaded" in the logs

### Environment

- WP Version: 6.9.1
- WooCommerce Version: 10.5.3
- Plugin Version: 3.4.1
- Browser: Firefox 148.0.2

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 10.5.3
Legacy REST API Package Version: The Legacy REST API plugin is not installed on this site.
Action Scheduler Version: ✔ 3.9.3
Log Directory Writable: ✔
WP Version: ❌ 6.9.1 - There is a newer version of WordPress available (6.9.4)
WP Multisite: –
WP Memory Limit: 2 GB
WP Debug Mode: –
WP Cron: –
WP Environment Type: production
Language: en_US
External object cache: ✔

### Server Environment ###

Server Info: nginx/1.22.1
Server Architecture: Linux 6.8.8-2-pve x86_64
PHP Version: 8.3.15
PHP Post Max Size: 32 MB
PHP Time Limit: 90
PHP Max Input Vars: 5000
cURL Version: 7.88.1
OpenSSL/3.0.15

SUHOSIN Installed: –
MySQL Version: 10.11.6-MariaDB-0+deb12u1
Max Upload Size: 32 MB
[...]

PR #4227, opened 2026-03-27
Fix 'Subscription could not be loaded' in PayPalSubscriptionsModule admin meta box

cyphercodes:
Fixes the same root cause as #4220 (which fixed `RenewalHandler.php`), but in the second affected file.

## Problem

`PayPalSubscriptionsModule.php` calls `wcs_get_subscription( $order->get_id() )` inside the `add_meta_boxes` hook. When viewing a **renewal order** in WP admin, `$order->get_id()` returns the renewal order ID, not the subscription ID. `wcs_get_subscription()` expects a subscription ID, so it returns `false` and logs "Subscription could not be loaded."

This is the exact same bug pattern that PR #4220 fixed in `RenewalHandler::renew()`.

## Fix

Added a fallback: when `wcs_get_subscription()` fails (returns non-subscription), try `wcs_get_subscriptions_for_renewal_order( $order )` which is the correct WooCommerce Subscriptions API for retrieving subscriptions from a renewal order.

This approach is backward-compatible — it still works when viewing an actual subscription directly, and now also handles renewal orders correctly.

## Testing

1. Create a subscription with PayPal Checkout
2. Let it generate a renewal order
3. View the renewal order in WP admin
4. Confirm no "Subscription could not be loaded" debug log entries

Related: #4191


cyphercodes:
Closing — branch was based on wrong starting point causing 763 unrelated commits. Will resubmit with a clean branch.

PR #4235, opened 2026-03-30
Remove UI flag references (6067)

stracker-phil:
Replaces #4213

### Issue Description

Two UI toggle flags are no longer needed but are still set in some places in the plugin, as well as referenced in unit tests.

### PR Description

Remove all code that changes, reads or deletes the two options `woocommerce-ppcp-is-new-merchant` and `woocommerce_ppcp-settings-should-use-old-ui`

The options are intentionally left in the database for backwards compatibility with any existing installations that may have it stored.


Narek13 (commented):
Thanks a lot @stracker-phil for fixing and for documenting this

Narek13 (approved):

PR #4242, opened 2026-03-31
Update stable tag to 4.0.1

daledupreez:
This PR updates the stable tag to 4.0.1 in the `dev/release/4.0.1` branch, as it was not updated correctly during the automated release process.

This change was applied to the WordPress.org plugin directory earlier in https://plugins.trac.wordpress.org/changeset/3495642/, so this PR is mostly to ensure we have the same data across both sets of code.


InpsydeNiklas (approved):

InpsydeNiklas:
This is no longer needed.

PR #4244, opened 2026-04-01
API rate limiting fix (6101)

stracker-phil:
## Description

**Fix excessive API calls to merchant-integrations endpoint:**

Merchants with `seller_type = 'unknown'` triggered an API call on every page load.
The resolver retries on every `init` hook when the type is unresolved, but the throttle only kicks in on failed requests - successful calls that return `UNKNOWN` were never throttled.

### Fix

1. **Treat unknown as casual** - `GeneralSettings::is_casual_seller()` returns `true` for any non-business type, breaking the retry loop without changing the stored DB value.
2. **Cache at the endpoint** - `PartnersEndpoint::seller_status()` caches responses for 10 minutes. Cleared on plugin update and "Refresh features".

### Test plan

- Added 7 new unit tests covering both changes

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
@@ -15,6 +15,7 @@ use WooCommerce\PayPalCommerce\ApiClient\Entity\SellerStatus;
 use WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException;
 use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
 use WooCommerce\PayPalCommerce\ApiClient\Factory\SellerStatusFactory;
+use WooCommerce\PayPalCommerce\ApiClient\Helper\Cache;
 use WooCommerce\PayPalCommerce\ApiClient\Helper\FailureRegistry;
 
 /**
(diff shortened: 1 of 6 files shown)


modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
```diff
@@ -73,6 +74,23 @@ class PartnersEndpoint {
 	 */
 	private $failure_registry;
 
+	/**
+	 * The cache for seller status responses.
+	 *
+	 * @var Cache
+	 */
+	private $cache;
```
Narek13:
```suggestion
	private Cache $cache;
```

modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
```diff
@@ -73,6 +74,23 @@ class PartnersEndpoint {
...
+
+	/**
+	 * Cache lifetime for seller status responses, in seconds.
+	 */
+	const SELLER_STATUS_CACHE_TTL = 600; // 10 minutes.
+
+	/**
+	 * Cache key for the seller status response.
+	 */
+	const SELLER_STATUS_CACHE_KEY = 'seller_status';
```
Narek13: why not adding visibilities?

Narek13 (approved):
Thanks a lot, @stracker-phil  for working on this. I’ve added a couple of minor suggestions, but they’re not blocking approval.

modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
stracker-phil: great question and thanks for pointing it out - I was in a hurry 😅

PR #4254, opened 2026-04-03
Remove vault v2 code (4572)

Dinamiko:
Vault v3 (`ppcp-save-payment-methods` module) supports all countries that vault v2 covered, making v2 logic redundant. All v2 code paths were already guarded by `save-payment-methods.eligible` checks that returned early when v3 was active.


Narek13 (approved):
Thanks, haven't tested but code-wise looks good to me.

PR #4274, opened 2026-04-13
[QA] Google Pay E2E (block checkout) transaction - PCP-5987

syde-piotr-ostrowski:
**Description**

This change adds one end-to-end Google Pay purchase test for block checkout. Uses popup-focused handling for the Google Pay window. Applies browser/runtime tweaks so popup detection and cross-origin Google screens are stable.

- New transaction spec and shared **Google Pay popup / login** helpers
- Store is configured like other QA transaction tests: merchant connect, DB reset, orders cleared, then **ACDC + Google Pay** enabled (Google Pay depends on ACDC).
- Scenario is tied to test case **PCP-26555** (`googlepay-checkout.data.ts` title + Jira link).

**Credentials**

- The test signs into **Google** and completes the wallet flow. **CI** needs the shared test Google account in **secrets**; **local** needs the same values in **`tests/qa/.env`**. Without that, the run cannot finish.

**Possible follow-ups**

- More scenarios or locales; **`storageState` / `state.json`** to reuse login and reduce flakes.
- Team policy if Google shows **extra verification** often (retries, tagged/smoke job, etc.).


mishautkin (approved):
Looks good!

PR #4276, opened 2026-04-14
Duplicate WooCommerce orders created when buyer switches payment method at checkout (6189)

Dinamiko:
When a buyer starts checkout with one local APM payment method (e.g., iDEAL) and then goes back to complete with a different method (e.g., credit card), the plugin creates two separate WooCommerce orders for the same purchase attempt. The first order remains stuck in On-Hold status with stock locked, while the second order completes successfully.

### Steps To Reproduce
- Add product(s) to cart on a WooCommerce store with PPCP and iDEAL enabled.                                                                                                                                        
- Proceed to checkout and select iDEAL as the payment method.
- Initiate the iDEAL payment flow (PayPal order created with status PAYER_ACTION_REQUIRED).                                                                                                                         
- Before completing the iDEAL redirect, go back and switch to credit card.                                                                                                                                          
- Complete the purchase with credit card.    

This PR changes the order status from `on-hold` to `pending` before the APM redirect on all local APMs gateways.

The fix will also prevent premature emails, and allow payment retry in My Account page.

### DIFF
--- a/modules/ppcp-local-alternative-payment-methods/src/BancontactGateway.php
+++ b/modules/ppcp-local-alternative-payment-methods/src/BancontactGateway.php
@@ -145,7 +145,7 @@ class BancontactGateway extends WC_Payment_Gateway {
 				'redirect' => wc_get_checkout_url(),
 			);
 		}
-		$wc_order->update_status( 'on-hold', __( 'Awaiting Bancontact to confirm the payment.', 'woocommerce-paypal-payments' ) );
+		$wc_order->update_status( 'pending', __( 'Awaiting Bancontact to confirm the payment.', 'woocommerce-paypal-payments' ) );
 
 		$purchase_unit = $this->purchase_unit_factory->from_wc_order( $wc_order );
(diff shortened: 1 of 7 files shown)


Narek13 (approved):
LGTM

modules/ppcp-local-alternative-payment-methods/src/BancontactGateway.php
```diff
@@ -145,7 +145,7 @@ public function process_payment( $order_id ) {
 				'redirect' => wc_get_checkout_url(),
 			);
 		}
-		$wc_order->update_status( 'on-hold', __( 'Awaiting Bancontact to confirm the payment.', 'woocommerce-paypal-payments' ) );
+		$wc_order->update_status( 'pending', __( 'Awaiting Bancontact to confirm the payment.', 'woocommerce-paypal-payments' ) );
```
stracker-phil, line 148: Nitpick: "pending" is the default status, technically we should not need an `update_status()` here

PR #4277, opened 2026-04-14
Cart simulation causes checkout instability and cannot be disabled (6188)

Dinamiko:
The cart simulation endpoint (`SimulateCartEndpoint`) used for Pay Later messaging on product pages causes checkout instability by swapping `WC()->cart` globally and calling `remove_all_actions('shutdown')`, which nukes all plugins' shutdown hooks. Errors during simulation leave the cart in a corrupted state with no recovery path, and there is no server-side way to disable the feature.

### PR Changes
  - All products use `IsolatedCartSimulator` that creates a throwaway `WC_Cart` with session hooks disabled , it never touches `WC()->cart`, never removes shutdown hooks, and guarantees cleanup via try/finally.
  - Add server-side guard for the `woocommerce_paypal_payments_simulate_cart_enabled` filter.
  - Fix silent error swallowing in `AbstractCartEndpoint::add_products()`.


modules/ppcp-button/src/Helper/ProductPriceCalculator.php
```diff
@@ -0,0 +1,95 @@
...
+
+use WC_Data_Store;
+
+/**
+ * Computes product totals using WooCommerce tax functions (wc_get_price_including_tax)
+ * without any cart interaction. Used as the fast path for simple and variable products
+ * in the cart simulation endpoint, avoiding the overhead and side effects of creating
+ * a WC_Cart instance.
+ */
+class ProductPriceCalculator {
```
stracker-phil: I'm not sure if this class is handling all cases correctly, as it's implementing a quite different calculation flow compared to regular WC_Cart::calculate_totals()

Some differences:
- As logged in customer there's no way to enter "[vat exempt](https://github.com/woocommerce/woocommerce/blob/c7fad1764d6aa78acf87baf8401cb06e9c8edb69/plugins/woocommerce/includes/class-wc-cart-totals.php#L134)" mode
- The "[is product taxable](https://github.com/woocommerce/woocommerce/blob/c7fad1764d6aa78acf87baf8401cb06e9c8edb69/plugins/woocommerce/includes/class-wc-cart-totals.php#L230)" decision is not happening
- [Default shipping cost](https://github.com/woocommerce/woocommerce/blob/c7fad1764d6aa78acf87baf8401cb06e9c8edb69/plugins/woocommerce/includes/class-wc-cart-totals.php#L337) is not included (intentional?)
- None of the default filters allows plugins to customize the total amount, [like this here](https://github.com/woocommerce/woocommerce/blob/c7fad1764d6aa78acf87baf8401cb06e9c8edb69/plugins/woocommerce/includes/class-wc-cart-totals.php#L877-L880)

Maybe the simplified calculation is intentional and sufficient for the Pay Later Message; if the total should really reflect only `quantity * unit_price` then we should document the reason for this in a comment

Dinamiko: @stracker-phil good points, for Pay Later messaging `ProductPriceCalculator` probably would be enough, but cart simulator is also used by Google and Apple Pay to get transaction info so in that case we need more precision, I removed `ProductPriceCalculator` completelly and now it relies only on `IsolatedCartSimulator`.

stracker-phil (approved):
Looks good now, good work!

PR #4279, opened 2026-04-15
Fix AddressFactory::from_wc_order billing address

AlexP11223:
The `PayerFactory` needed the billing address, but `AddressFactory::from_wc_order()` actually did not support the `billing`/`shipping` `type` parameter (like other methods there) always returning the shipping address.

Now it is fixed and it is possible to retrieve billing and shipping addresses similarly to other `PayerFactory` methods.


Narek13 (approved):
Thanks. LGTM

Dinamiko (approved):

PR #4288, opened 2026-04-21
Guest Cart appears empty when product is added from product page (6252)

Dinamiko:
Guests lost their cart after adding a product from the product page. The `ppc-simulate-cart` was wiping the guest's WC session cart.   

`IsolatedCartSimulator::cleanup_cart()` was calling `$cart->empty_cart( false )` on the throwaway cart, which fires the global `woocommerce_cart_emptied` action.                                                                                                                         
                                                                                                                                                                                                                       
### PR Changes                                                                                                                                                                                                                                                                                                                                                                                                              
- Removed the `empty_cart()` call. `remove_cart_items()` still runs so third-party per-item reservations are released. 
- Added a regression test asserting `WC()->session->get( 'cart' )` survives a simulate call.


Narek13 (approved):
Thanks. Looks good to me

PR #4289, opened 2026-04-21
BCDC - Mexico - BCDC not visible on block checkout  (6251)

Dinamiko:
On block checkout, the BCDC button was missing from the express-checkout section for MX merchants with BCDC enabled and ACDC disabled. The PayPal SDK was loaded with card in disable-funding.                       
   
`DisabledFundingSources::should_disable_card()` block branch used `use_acdc()` (ACDC capability) instead of `is_acdc_enabled()` (capability + gateway on). MX merchants can have both `use_acdc() == true` and `is_bcdc_enabled() == true` at the same time, so the block rule incorrectly disabled card.
                                                                                                                                                                                                                       
This PR uses `is_acdc_enabled()` in the block branch, matching the classic branch. MX merchants with BCDC selected now keep card enabled on block checkout and the BCDC button renders.


Narek13 (approved):
LGTM, thanks

PR #4292, opened 2026-04-23
Drop shipping when city is missing to prevent CITY_REQUIRED (6265)

Dinamiko:
PayPal's Orders v2 API rejects shipping addresses without `admin_area_2` (city) with `CITY_REQUIRED`, which cascades to `PAYER_CANNOT_PAY` on order approval.                                                              
   
The shipping-address completeness check in `PurchaseUnitFactory` only gates on `postal_code` presence and the `country_without_postal_code()` allowlist. For countries in that allowlist where PayPal still requires city, most notably Ireland (IE), where Eircode is optional but city is mandatory, the check passes even when city is empty. The incomplete address is sent to v2/checkout/orders and rejected.
                                                                                                                                                                                                                       
Per the https://developer.paypal.com/api/rest/reference/orders/v2/country-address-requirements/, `admin_area_2` is required in virtually every country (only HK, JP, and SG mark it optional, and WooCommerce's locale rules require city for those anyway).                                                                                                     
   
### PR Changes                                                                                                                                                                                                           
                                                         
This PR adds an `admin_area_2` check to both shipping-address validation sites in PurchaseUnitFactory:                                                                                                                   
  - `from_wc_cart()` cart/customer flow                                                                                                                                                                                
  - `should_disable_shipping()` from_wc_order() flow     
                                                                                                                                                                                                                       
If city is empty, shipping is set to null locally before the PayPal API call, mirroring how `address_line_1` is already handled. The `country_without_postal_code` allowlist is untouched,  postal-code handling for IE, HK, VN, etc. is unchanged.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
+++ b/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
@@ -194,6 +194,7 @@ class PurchaseUnitFactory {
 			if (
 				! $shipping_address ||
 				2 !== strlen( $shipping_address->country_code() ) ||
+				! $shipping_address->admin_area_2() ||
 				( ! $shipping_address->postal_code() && ! $this->country_without_postal_code( $shipping_address->country_code() ) )
 			) {
 				$shipping = null;
@@ -391,6 +392,7 @@ class PurchaseUnitFactory {
 				! $shipping_address ||
 				empty( $shipping_address->country_code() ) ||
 				empty( $shipping_address->address_line_1() ) ||
+				empty( $shipping_address->admin_area_2() ) ||
 				( ! $shipping_address->postal_code() && ! $this->country_without_postal_code( $shipping_address->country_code() ) );
 	}
 }
--- a/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
+++ b/modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
@@ -191,12 +191,7 @@ class PurchaseUnitFactory {
 		if ( $this->shipping_needed( ...array_values( $items ) ) && $customer instanceof \WC_Customer ) {
 			$shipping         = $this->shipping_factory->from_wc_customer( \WC()->customer, $with_shipping_options );
 			$shipping_address = $shipping->address();
-			if (
-				! $shipping_address ||
-				2 !== strlen( $shipping_address->country_code() ) ||
-				! $shipping_address->admin_area_2() ||
-				( ! $shipping_address->postal_code() && ! $this->country_without_postal_code( $shipping_address->country_code() ) )
-			) {
+			if ( ! $shipping_address || ! $this->can_use_shipping_address( $shipping_address ) ) {
 				$shipping = null;
 			}
 		}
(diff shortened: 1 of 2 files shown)


modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
```diff
@@ -391,6 +392,7 @@ private function should_disable_shipping( array $items, ?Address $shipping_addre
 				! $shipping_address ||
 				empty( $shipping_address->country_code() ) ||
 				empty( $shipping_address->address_line_1() ) ||
+				empty( $shipping_address->admin_area_2() ) ||
```
stracker-phil: suggestion: convert those long or-chains into single conditions with a short comment explaining the reasoning.

Though it makes the code much longer, it's also making the decisions more explicit and understandable/maintainable.

Sample (note - the comments were written by Claude, they might be wrong):

```php
private function should_disable_shipping( array $items, ?Address $shipping_address ): bool {
    // No items require shipping at all.
    if ( ! $this->shipping_needed( ...array_values( $items ) ) ) {
        return true;
    }

    // Cannot calculate shipping without an address.
    if ( ! $shipping_address ) {
        return true;
    }

    // Incomplete address — missing required fields.
    if ( empty( $shipping_address->country_code() )
        || empty( $shipping_address->address_line_1() )
        || empty( $shipping_address->admin_area_2() )
    ) {
        return true;
    }

    // Postal code required but absent for this country.
    if ( ! $shipping_address->postal_code()
        && ! $this->country_without_postal_code( $shipping_address->country_code() )
    ) {
        return true;
    }

    // Default: do not disable shipping.
    return false;
}
```

modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
```diff
@@ -194,6 +194,7 @@ function ( Item $item ): bool {
 			if (
 				! $shipping_address ||
 				2 !== strlen( $shipping_address->country_code() ) ||
+				! $shipping_address->admin_area_2() ||
```
stracker-phil: suggestion: Move this to a private helper method

reason: I see there's a difference already between this condition and the "should_disable_shipping" method below.
- the `address_line_1()` prop is not checked here, but below it is
- `! $shipping_address->admin_area_2()` vs `empty( $shipping_address->admin_area_2() )`
...

those small changes could eventually lead to a different result/hard to track bug

Suggestion to refactor this condition into:

```php
if ( ! $shipping_address || ! $this->can_use_shipping_address( $shipping_address ) ) {
    $shipping = null;
}
```

```php
private function can_use_shipping_address( Address $shipping_address ): bool {
  // use this new helper in the should_disable_shipping() decision, or at least put them close together
}
```

stracker-phil (approved):
Approving as the logic looks sound; still suggest to refactor those conditions for more clarity

Good work on adding/updating the unit tests 👍

modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
Narek13: Agree with @stracker-phil . I think is even better to include method level docblock description and remove all inline comments

Narek13 (approved):
Thanks for working on this and thanks for adding the tests :muscle:

modules/ppcp-api-client/src/Factory/PurchaseUnitFactory.php
stracker-phil: @Narek13 I think the method-level comment will likely not be updated when the internal logic changes.

Esp when adding a change at the bottom of the method, even the code reviewer will not see the method comment in GitHub review. Keeping the comment close to the condition then it's easier to keep them in-sync

Dinamiko:
@stracker-phil @Narek13 thanks for the feedback, I extracted the logic to its own method and added comments to make it clear what each condition does.

PR #4296, opened 2026-04-23
Add unit tests for `GeneralSettings::own_brand_only()`

Narek13:
### Summary         

  - Adds unit tests for `GeneralSettings::own_brand_only()` to document and guard the branded-only detection logic.                                                                                    
  - Covers all four decision branches:
  	- branded installation paths (core-profiler, payment-settings) in a WCPay-eligible country -> `return true`
  	- direct path or a non-WCPay country -> `return false`.
  - This logic is what drives `settings.service.merchant_capabilities` to gate Apple Pay, Google Pay, and ACDC capabilities in branded-only mode — preventing "Things to do next" todos (e.g. "Register
   Apple Domain") from appearing when they shouldn't.


stracker-phil (approved):
Picture book PR 👍

PR #4298, opened 2026-04-23
[QA] Fix Google Pay e2e test for headless/CI runs

syde-piotr-ostrowski:
The Google Pay transaction test (PCP-26555) was failing in headless mode.
Three areas were fixed:

- `httpCredentials` not stripped from visitor context - `test.use({ httpCredentials: undefined })` had no effect because `visitorPage` reads credentials from `testInfo.project.use` project-level only). It sometimes might broke Playwright's popup tracking.

- Sign-in missed on first run - `waitForContent` returned as soon as the popup left `about:blank`, before Google's redirect to accounts.google.com had fired. The sign-in check  was skipped, and the payment sheet loaded without an authenticated account.

- Every test run started with a fresh browser context and no Google cookies, requiring a full sign-in each time. During reruns, Google's  detection closed the popup before finished auth.

### DIFF
--- a/tests/qa/.gitignore
+++ b/tests/qa/.gitignore
@@ -11,6 +11,7 @@ test-results
 .DS_Store
 /resources/files/woocommerce-paypal-payments*.zip
 /resources/files/woocommerce-subscriptions*.zip
+.google-session.json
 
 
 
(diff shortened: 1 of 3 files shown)


mishautkin (approved):
Looks like a lot of custom action is required to make GooglePay work.
Approving for now but let's keep an eye on it

PR #4299, opened 2026-04-24
Failing ACDC payment for Free Trial subscription on classic checkout (6262)

Dinamiko:
On classic checkout, adding a free-trial subscription product and selecting Debit & Credit Cards left users stuck: the PayPal CardFields never rendered over the stock WooCommerce card inputs, so clicking Place order submitted the form with no tokenized card and the order failed.

Root cause: `DisabledFundingSources::get_sources_for_free_trial()` disables all funding sources for free-trial carts and only re-enables card when BCDC is on. For classic + ACDC the card source stayed disabled in the PayPal SDK URL, so `paypal.CardFields().isEligible()` returned false and `renderFields()` short-circuited.

This PR delegates the card decision in the free-trial path to the existing `should_disable_card()` authority so the same classic/blocks + ACDC/BCDC decision table applies. Result: classic + ACDC + free-trial now keeps card enabled (bug fix); all other contexts (blocks, MX, BCDC-only, no-card-flow) retain their current behavior.

### DIFF
--- a/modules/ppcp-button/src/Helper/DisabledFundingSources.php
+++ b/modules/ppcp-button/src/Helper/DisabledFundingSources.php
@@ -52,7 +52,7 @@ class DisabledFundingSources {
 		// Free trials have a shorter, special funding-source rule.
 		if ( $flags['is_free_trial'] ) {
 			return $this->sanitize_and_filter_sources(
-				$this->get_sources_for_free_trial(),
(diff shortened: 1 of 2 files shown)


AlexP11223 (approved):

danieldudzic (approved):
Clean fix, good test coverage 👍

PR #4301, opened 2026-04-24
fix(amount-factory): derive total from breakdown sum to prevent PayPal PATCH rejection on shipping change (6279)

slash1andy:
## Problem

Buyers using one-time pay (Fastlane guest checkout) are blocked from changing their shipping address. The failure affects approximately **10,683 requests / 2,882 unique orders in a 7-day window**.

When a buyer changes their shipping address, `UpdateShippingEndpoint` fires `ppc-update-shipping`, which calls `AmountFactory::from_wc_cart()` to build the PATCH payload. PayPal requires that `amount.value` exactly equals `item_total + shipping + tax_total − discount` in the breakdown. The current code uses `WC_Cart::get_total()` for `amount.value` but computes breakdown components from separate getters. Due to per-item tax rounding in WooCommerce, these two values can differ by ±$0.01 — enough for PayPal to reject the PATCH.

The same structural issue exists in `from_store_api_cart()`, used by the shipping callback endpoint, which also passes `total_price()` directly as `amount.value` without including fees in `item_total`.

## Fix

**`from_wc_cart()`**: compute the total in integer cents from the same component floats used for the breakdown, so `amount.value` is always the exact formatted sum of its parts.

**`from_store_api_cart()`**: use integer minor-unit arithmetic (Store API values are already in cents, so the sum is exact), and include `total_fees` in `item_total` to match `from_wc_cart()` and prevent a mismatch when WooCommerce fees are present.

## How to reproduce

1. Configure WooCommerce with a tax-exclusive rate (8–12%)
2. Add a product whose price × rate has a sub-cent remainder (e.g. $13.33 × 8% = $1.0664)
3. Add a non-zero shipping method
4. Open Fastlane (Axo) guest checkout
5. Change the shipping address or method
6. Observe the PATCH order request fail in the browser network tab or the WC PayPal debug log

## Verification

- The PATCH succeeds for the reproduction scenario above
- Buyer can change shipping address without error
- A unit test asserting `Amount::value_str() === formatted sum of breakdown parts` passes across carts with varied tax rates and product prices

### DIFF
--- a/modules/ppcp-api-client/src/Factory/AmountFactory.php
+++ b/modules/ppcp-api-client/src/Factory/AmountFactory.php
@@ -149,7 +149,7 @@ class AmountFactory {
 				null,
 				null,
 				null,
-				$make( $discount_minor ),
+				$discount_minor > 0 ? $make( $discount_minor ) : null,
 			)
 		);
 	}
--- a/modules/ppcp-api-client/src/Factory/AmountFactory.php
+++ b/modules/ppcp-api-client/src/Factory/AmountFactory.php
@@ -111,11 +111,7 @@ class AmountFactory {
 			null, // shipping discounts?
 			$discount
 		);
-		$amount    = new Amount(
-			$total,
-			$breakdown
-		);
-		return $amount;
+		return new Amount( $total, $breakdown );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


modules/ppcp-api-client/src/Factory/AmountFactory.php
stracker-phil: suggestion: also consolidate this into a 1-liner, same as in line 215

```php
return new Amount( $total, $breakdown );
```

stracker-phil (approved):
Solid change, thanks for providing the unit tests with it

Besides the minor style inconsistency of refactoring one return value into a 1-liner but not the other, the PR appears to be safe and correct 👍

modules/ppcp-api-client/src/Factory/AmountFactory.php
slash1andy: Changed! https://github.com/woocommerce/woocommerce-paypal-payments/pull/4301/commits/a05da10b0fdab28bb8ca041283d4a7fffe148bd4

slash1andy:
@stracker-phil thanks for the review! I made the change noted :)

Dinamiko (approved):

PR #4304, opened 2026-04-27
[QA] Auto test fixes for 4.0.3

mishautkin:
### Description

- Remove deleteAllOrders from preconditions 
    Due to frequent reset of test env and on account that wp-env in CI is created clean - there's no need in additional requests. The deleteAllOrders kept orders clean but also removed orders from previous test set preventing analysis. In case of many orders the precondition might timeout because of long-taking request.
- Fix refund.scenario 
    - Move methods from utils to refund.scenario
    - Fix precondition orders to refund because they failed on expected order status not set within 5 sec
- Fix assertion for "Payment via" on order-edit page


tests/qa/resources/pcp-gateways.ts
```diff
@@ -323,6 +325,5 @@ export const gateways = {
 	multibanco,
 	oxxo,
 	payUponInvoice,
-	debitOrCreditCard,
-	standardCardButton,
```
syde-piotr-ostrowski, line 327: You have removed `standardCardButton`, but we are still referring to it here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/56ed19d838b497ea1018b01383bae5b2c18e069c/tests/qa/resources/pcp-payments.ts#L64

that looks to me like a problem to solve before merging

tests/qa/resources/pcp-gateways.ts
```diff
@@ -323,6 +325,5 @@ export const gateways = {
 	multibanco,
 	oxxo,
 	payUponInvoice,
-	debitOrCreditCard,
```
syde-piotr-ostrowski, line 326: You have removed `debitOrCreditCard`, but we are still referring to it here: https://github.com/woocommerce/woocommerce-paypal-payments/blob/56ed19d838b497ea1018b01383bae5b2c18e069c/tests/qa/resources/pcp-payments.ts#L59

that looks to me like a problem to solve before merging

mishautkin: Done, thanks for pointing out

tests/qa/resources/pcp-gateways.ts
mishautkin: Done

syde-piotr-ostrowski (approved):
Looks good to me. Ready to be merged

PR #4305, opened 2026-04-27
[QA] Add parallel E2E test execution in CI

syde-piotr-ostrowski:
Adds a separate CI option for running the full Playwright E2E suite in parallel across multiple wp-env setups.

The new `e2e:test:all:parallel` workflow input splits execution into dedicated shards for:

1. plugin foundation
2. onboarding/settings
3. transactions
4. refunds
5. vaulting
6. subscriptions

Each shard runs in its own CI job, allowing tests to execute independently instead of blocking each other.

For now this is added as a separate manual workflow option, so the existing default E2E runs remain unchanged. In the future, this parallel execution can be promoted to the default release test run.

# How to run

- Trigger the E2E Tests workflow manually.
- Select `e2e:test:all:parallel`.
- Verify that all shard jobs start independently and report results/artifacts correctly

# Test run
[Here you can view the how the workflow was executed](https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/25001249459)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -15,6 +15,7 @@ on:
           - e2e:test:critical
           - e2e:test:stress
           - e2e:test:all
+          - e2e:test:all:parallel
       XRAY_TEST_EXEC_KEY:
         description: 'Xray Test Execution Key'
         required: false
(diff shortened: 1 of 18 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -75,3 +77,51 @@ jobs:
       ENV_FILE_DATA: ${{ secrets.QA_ENV_FILE }}
       NPM_REGISTRY_TOKEN: ${{ secrets.DEPLOYBOT_GITHUB_TOKEN_PACKAGES }}
       COMPOSER_AUTH_JSON: ${{ secrets.COMPOSER_AUTH_JSON }}
+
+  e2e-playwright-parallel:
+    name: Run Playwright E2E Tests / ${{ matrix.shard }}
+    if: ${{ inputs.TEST_SUITE == 'e2e:test:all:parallel' }}
```
mishautkin: `e2e-playwright` and `e2e-playwright-parallel` are almost duplicate. It should be possible to resolve matrix in a separate job and avoid duplication of other code?

tests/qa/playwright.config.ts
```diff
@@ -139,5 +139,37 @@ export default defineConfig< BaseExtend >( {
 			dependencies: [ 'setup-woocommerce' ],
 			testMatch: /stress\.spec\.ts/,
 		},
+
+		// Parallel CI
+		{
+			name: 'shard:plugin-foundation',
+			dependencies: [ 'setup-woocommerce' ],
```
mishautkin, line 182: `plugin-foundation` was the only set of tests which required not installed PCP plugin.

Since `plugin-foundation` is already the separate project, I was about to extract the common preconditions from other test specs. But since your PR is quite close to it I'd suggest to do it here.

All of the projects depend on `setup-woocommerce` but almost all except `plugin-foundation` might also depend on `setup-pcp-usa-for-tests` (or similar, like `setup-pcp-vaulting`, `setup-pcp-subscription`). This would allow to shorten `beforeAll` hooks which in case if test fails are being reexecuted and cause significant time loss. So I was thinking about something like:

```ts
{
	name: 'setup-pcp-usa-for-tests',
	dependencies: [ 'setup-woocommerce' ],
	testMatch: /pcp\.setup\.ts/,
    grep: 'setup:pcp:usa;',
},
{
	name: 'shard:transactions',
	dependencies: [ 'setup-pcp-usa-for-tests' ],
	testMatch: /05-transactions\/.*\.spec\.ts/,
},
```
> **Note:** There's `setup-pcp` project already but it has many setups and also serves for `env:setup` npm scripts.

mishautkin (changes requested):

syde-piotr-ostrowski:
@mishautkin 
Thanks, I did following changes according to your suggestions:

- The E2E workflow now uses one Playwright job for both normal and parallel runs.
- The parallel shard list is resolved in a separate job.
- Added separate Playwright setup projects for transactions, refunds, vaulting, and subscriptions.
- Removed repeated PCP install/reset/connect setup from the affected tests.

Checked test execution locally. [Now running in GitHub here](https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/25106215926)

mishautkin (commented):
Hi, @syde-piotr-ostrowski, thank you for the updates. Have couple of general comments:
1. I see there are a lot of failed tests in CI runs (e.g. I cansee that ACDC gateway is not visible)
2. ~~All of the shards have the same name - `run-playwright-tests`. Is it possible to change the name per shard?~~
    **Update:** `run-playwright-tests` is a job name in reusable `test-playwright.yml`. Let's skip it in this PR

tests/qa/playwright.config.ts
syde-piotr-ostrowski: This is already done in this PR — exactly as you described.

Each shard project depends on a dedicated setup project that runs only its specific grep from `pcp.setup.ts`:

```
  { name: 'setup-pcp-usa-for-transactions',    grep: /setup:pcp:usa:transactions;/,    dependencies: ['setup-woocommerce'] },
  { name: 'setup-pcp-usa-for-refund',          grep: /setup:pcp:usa:refund;/,          dependencies: ['setup-woocommerce'] },
  { name: 'setup-pcp-vaulting',                grep: /setup:pcp:usa:vaulting;/,        dependencies: ['setup-woocommerce'] },
  { name: 'setup-pcp-subscription',            grep: /setup:pcp:usa:subscription;/,    dependencies: ['setup-woocommerce'] },
```
And each shard depends on its setup project:

```
  { name: 'shard:transactions',   dependencies: ['setup-pcp-usa-for-transactions'] },
  { name: 'shard:refund',         dependencies: ['setup-pcp-usa-for-refund'] },
  { name: 'shard:vaulting',       dependencies: ['setup-pcp-vaulting'] },
  { name: 'shard:subscription',   dependencies: ['setup-pcp-subscription'] },
```
The spec beforeAll hooks are also trimmed - they only call `configureStore()`.

syde-piotr-ostrowski:
thanks @mishautkin

1. ACDC Fix

ACDC not visible - fixed. `setup:pcp:usa:vaulting` and `setup:pcp:usa:subscription` were missing a `updatePcpPaymentMethods` call. Added now.

2. Remaining CI Failures

Remaining failures in the latest CI run are all pre-existing issues.

**Refund PCP-1409/1410**
PayPal webhook not delivered in sandbox - order never reaches processing.

**Subscription PayPal popup**
2+ min sandbox login timeout - pre-existing flakiness.

**Subscription ACDC free trial**
$0 first-payment vault flow - sandbox-specific behavior.

**Vaulting PCP-5380**
Passed on retry.

None are caused by cross-shard contamination - each shard runs in a fully isolated runner with its own `wp-env` instance.

3. Shard Job Names

Noted, skipping per your update.

mishautkin (approved):
Looks good! 🦾 Please merge

PR #4306, opened 2026-04-27
Add shipping options to Cart API responses (6271)

stracker-phil:
### Description

Extends and refactors the `CartResponse` class to address the following items:

1. Provide a new [`shipping_options`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/c8e3ca6ebc29ea2ae7ff3995abf9487a865c727c/modules/ppcp-store-sync/src/Response/CartResponse.php#L192) array in the cart response
    - This response is generated by inspecting the [`WC()->shipping()`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/c8e3ca6ebc29ea2ae7ff3995abf9487a865c727c/modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php#L30) data
    - Assigned to the cart response via a new setter [`shipping_options()`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/c8e3ca6ebc29ea2ae7ff3995abf9487a865c727c/modules/ppcp-store-sync/src/Response/CartResponse.php#L139-L149)
    - When the WC object has no shipping options, an empty array is returned

2. Consolidated the `CartResponse` classes, implementing the same factory-method-logic which we use in most other parts of the module
    - Three [static factory methods](https://github.com/woocommerce/woocommerce-paypal-payments/blob/c8e3ca6ebc29ea2ae7ff3995abf9487a865c727c/modules/ppcp-store-sync/src/Response/CartResponse.php#L82-L125) in the main class
    - Same pattern from [`ValidationIssue`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/40860fe544ad36ddb183a106d53d679f0b83bf33/modules/ppcp-store-sync/src/Validation/ValidationIssue.php#L58-L83) or in [`ResolutionOption`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/40860fe544ad36ddb183a106d53d679f0b83bf33/modules/ppcp-store-sync/src/Validation/Resolution/ResolutionOption.php#L32-L46)
    - Note how all classes have a private constructor; also applied to the `CartResponse`
    - Also dropped the now obsolete classes: `NewCartResponse` and `PaidCartResponse`

3. Refactored the `ResponseFactory`, which still is the only place that creates new `CartResponse` instances

4. Extended (and adjusted) the test suite to cover the response expectations more thoroughly


modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+
+		$packages = WC()->shipping()->get_packages();
+
+		if ( empty( $packages ) ) {
+			return array();
+		}
+
+		$chosen_methods = WC()->session->get( 'chosen_shipping_methods' );
+		$chosen_id      = ( is_array( $chosen_methods ) && ! empty( $chosen_methods ) )
+			? $chosen_methods[0]
```
Narek13: Should we be handling `chosen_shipping_methods` per package index instead of only using the first one?
I’m wondering if this could lead to incorrect `is_selected` values when multiple packages are present.

modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+				$rate_id = $rate->get_id();
+
+				if ( null === $first_rate_id ) {
+					$first_rate_id = $rate_id;
+				}
+
+				$options[] = array(
+					'id'          => $rate_id,
+					'label'       => $rate->get_label(),
+					'amount'      => number_format( (float) $rate->get_cost(), 2, '.', '' ),
```
Narek13: Any reason we’re using `number_format` instead of Woo helpers like `wc_format_decimal()`?

modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+
+			foreach ( $package['rates'] as $rate ) {
+				$rate_id = $rate->get_id();
+
+				if ( null === $first_rate_id ) {
+					$first_rate_id = $rate_id;
+				}
+
+				$options[] = array(
+					'id'          => $rate_id,
```
Narek13: If I am not wrong the `$rate->get_id()` is not guaranteed globally unique across packages.
What happens if 2 packages produce the same rate ID → your final array may contain duplicates with conflicting selection.
If yes then maybe use something like this `$rate_id = $package_index . ':' . $rate->get_id();`

Narek13 (commented):
Thanks a lot for working on this. LGTM, although haven't tested. Just some questions/doubts regarding `ShippingOptionsBuilder`

modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php
stracker-phil: Good observation!

According to PayPal's specs, we'll receive and store a single shipping option. I simply pick the first option, as that's always available, and silently ignore cases where more than one option might appear in the session.

Will add a comment to explain this is intentional

modules/ppcp-store-sync/src/Helper/ShippingOptionsBuilder.php
stracker-phil: The `number_format` usage is intentional in the store-sync module:

It formats the currency value for the REST response, not for processing by WooCommerce. PayPal expects the default formatting (dot as decimal separator, no thousand separators).

When using `wc_format_decimal` we have (a) risk of producing an unsupported number format, and (b) additional overhead that's simply not needed for the API response

Narek13 (approved):
Thanks for the clarifications. Looks good to me.

PR #4307, opened 2026-04-27
Add SellerStatus filtering layer with extension point for custom overrides (6276)

danieldudzic:
### Description

Adds two WordPress filter hooks to `PartnersEndpoint::seller_status()` so external plugins can override or replace the PayPal merchant-integrations API response.

**What this PR adds:**

- `woocommerce_paypal_payments_seller_status` — filters the `SellerStatus` object on every return (cache hit or API response)
- `woocommerce_paypal_payments_seller_status_fallback` — provides a fallback `SellerStatus` when the API fails, instead of throwing
- Extracted `fetch_seller_status_from_api()` private method for clarity
- Documentation at `docs/seller-status-filter.md`

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
@@ -135,9 +135,60 @@ class PartnersEndpoint {
 	public function seller_status(): SellerStatus {
 		$cached = $this->cache->get( self::SELLER_STATUS_CACHE_KEY );
 		if ( $cached instanceof SellerStatus ) {
-			return $cached;
+			/**
+			 * Filters the seller status object before it is returned.
+			 *
+			 * @param SellerStatus $status The seller status (from cache or API).
+			 */
+			return apply_filters( 'woocommerce_paypal_payments_seller_status', $cached );
 		}
 
+		try {
+			$status = $this->fetch_seller_status_from_api();
+		} catch ( RuntimeException $exception ) {
+			/**
+			 * Provides a fallback SellerStatus when the API call fails.
+			 *
+			 * Return a SellerStatus instance to use as fallback instead of
+			 * throwing. Return null to let the exception propagate.
+			 *
+			 * @param SellerStatus|null $fallback Default null (no fallback).
+			 */
+			$fallback = apply_filters( 'woocommerce_paypal_payments_seller_status_fallback', null );
+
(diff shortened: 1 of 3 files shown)


modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
```diff
@@ -135,9 +135,60 @@ public function __construct(
...
+				return $status;
+			}
+
+			throw $exception;
+		}
+
+		/** This filter is documented above. */
+		$status = apply_filters( 'woocommerce_paypal_payments_seller_status', $status );
+
+		$this->cache->set( self::SELLER_STATUS_CACHE_KEY, $status, self::SELLER_STATUS_CACHE_TTL );
```
Dinamiko: I suggest swapping the order here: cache first, then apply the filter on the return. 

That keeps the cache "raw" so the filter runs exactly once per call, otherwise it runs again on every cache hit until the TTL expires, stacking up the same change over and over.

modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
```diff
@@ -135,9 +135,60 @@ public function __construct(
...
+				$this->logger->log(
+					'info',
+					'Seller status API failed, using configured fallback.',
+					array( 'error' => $exception->getMessage() )
+				);
+
+				/** @var SellerStatus $status */
+				$status = apply_filters( 'woocommerce_paypal_payments_seller_status', $fallback );
+
+				$this->cache->set( self::SELLER_STATUS_CACHE_KEY, $status, self::SELLER_STATUS_CACHE_TTL );
```
Dinamiko: Same as commented above, I suggest to swap order here for the same reason mentioned above.

Dinamiko (changes requested):

modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
danieldudzic: Thanks, good suggestion :)

modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
danieldudzic: Valid. Same here.

Dinamiko (approved):

PR #4309, opened 2026-04-28
Remove redundant cache from ReferenceTransactionStatus (6280)

danieldudzic:
### Description
Remove the redundant cache from `ReferenceTransactionStatus` and rely on `PartnersEndpoint::seller_status()`, which already caches the `SellerStatus` object.

This cache is a leftover from the old `BillingAgreementsEndpoint` (removed in #3495), where it made sense because the check required an expensive `create_token()` API call.
                                                                                                                                                                                                                                 
Now that the class reads from `PartnersEndpoint::seller_status()`, which already caches the `SellerStatus` object in its own transient, the extra cache is redundant.         

This change:
- Removes the extra boolean cache and its service wiring
- Keeps `woocommerce_paypal_payments_seller_status` effective on repeated calls
- Removes the stale cache-clearing logic from `WCGatewayModule`
- Adds unit coverage for the updated helper behavior

### DIFF
--- a/modules/ppcp-api-client/services.php
+++ b/modules/ppcp-api-client/services.php
@@ -300,8 +300,7 @@ return array(
 		);
 	},
 	'api.reference-transaction-status'               => static fn ( ContainerInterface $container ): ReferenceTransactionStatus => new ReferenceTransactionStatus(
-		$container->get( 'api.endpoint.partners' ),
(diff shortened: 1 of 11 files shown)


Narek13 (approved):
LGTM 👍 Thanks a lot for adding the unit test!

PR #4310, opened 2026-04-28
Fix the API response keys of the "totals" property (6272)

stracker-phil:
### Description

The `CartResponse.totals` specs changed since we implemented the first version of the API, we need to rename a few keys:

| Old key | Correct new key |
| --- | --- |
| `.tax_total` | `.tax` |
| `.item_total` | `.subtotal` |
| `.amount` | `.total` |


danieldudzic (approved):
LGTM!

PR #4311, opened 2026-04-28
Always add currency to cart (6272)

stracker-phil:
This PR extends #4306

### Description

The `totals` calculation requires a valid currency code to work, as a money value without a currency code is invalid.

#### Current Behavior

- The Cart API waits until PayPal provides a currency.
- This does not happen in early requests, and as a result we do not include price or currency details in the response
- Postman collection: Expects to see minimal cart totals immediately after an item was added

#### New Behavior

- When there is no currency code defined in the incoming PayPal cart, the Cart API falls back to use Woo's currency
- This unlocks totals calculation as soon as an item is in the cart
- Postman collection: Confirms that our API is working as expected
- Consequence: The AI agent can infer the currency and product price from the response, instead of relying on the product catalog


modules/ppcp-store-sync/src/Config/StoreCurrencyValue.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
danieldudzic, line 1: Missing `declare(strict_types=1);` 😅

tests/PHPUnit/StoreSync/Config/StoreCurrencyValueTest.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * GIVEN a StoreCurrencyValue instance
+	 * WHEN the value() method is inspected
+	 * THEN it exists and is callable on the public API
+	 */
+	public function test_value_method_is_callable(): void {
+		$testee = new StoreCurrencyValue();
+
+		$this->assertTrue( is_callable( array( $testee, 'value' ) ) );
+	}
```
danieldudzic: The third test `test_value_returns_woocommerce_currency_code` instantiates the class and calls `value()` - if either of those failed, it would fail too. So these two are redundant and can be removed :)

modules/ppcp-store-sync/src/Response/ResponseFactory.php
```diff
@@ -15,28 +15,33 @@
...
 	private ShippingOptionsBuilder $shipping_options_builder;
+	private StoreCurrencyValue $store_currency;
 
 	/**
 	 * Constructor.
 	 *
 	 * @param AgenticCartBuilder     $cart_builder             Cart builder service.
 	 * @param AppliedCouponsBuilder  $applied_coupons_builder  Applied coupons builder service.
 	 * @param ShippingOptionsBuilder $shipping_options_builder Shipping options builder service.
+	 * @param StoreCurrencyValue     $store_currency
```
danieldudzic: Nit: `@param StoreCurrencyValue $store_currency` is missing a description - the other params all have one (e.g. "Cart builder service.") - perhaps we should add one for consistency? :)

danieldudzic (changes requested):
Thanks for working on this. Just a few minor issues.

modules/ppcp-store-sync/src/Config/StoreCurrencyValue.php
stracker-phil: This was also missing in a few other files, I created a new PR to address this: #4322

tests/PHPUnit/StoreSync/Config/StoreCurrencyValueTest.php
stracker-phil: Excellent feedback! I forgot to refactor and clean the test suite 😅

modules/ppcp-store-sync/src/Response/ResponseFactory.php
stracker-phil: Good suggestion, and I've gone a different route:

I removed the constructor doc-block completely, as it's 100% redundant noise. All params and types are visible in the method signature, and the description adds no value to any param.

Having only call-site (in `services.php`) means, we'll never read that comment. The only reason I personally see for a constructor comment, is when we accept an ambiguous value like `string` or `array`

PR #4320, opened 2026-04-30
CreateCartEndpoint always returns the generated cart ID (6285)

stracker-phil:
### Description

All cart endpoints returned a copy of the incoming payload with additional details - status, validation, totals, ...

**An edge case that was not considered:**

The incoming payload can include a cart ID, and that exact ID is included in the API response - this is expected for most endpoints, like "GetCartEndpoint."

However, when creating a new cart, this ID is generated by the endpoint, and must always be returned, regardless of whether the incoming payload included an ID or not.

We address this bug here


Narek13 (approved):
Thanks. LGTM

PR #4325, opened 2026-04-30
[QA] PCP-6283 Disable video recording in CI to reduce artifact size

mishautkin:
### Description

Disable video recording in CI tests to reduce artifact size, according to PCP-6283.
Tested with the following [job](https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/25312322205)


syde-piotr-ostrowski (approved):
Approved. Looks good to me

PR #4328, opened 2026-05-04
Add agentic beta banner for eligible US merchants (6057)

Narek13:
## Summary

Adds an invite banner for the PayPal Store Sync (agentic) beta program, visible on all PayPal Payments settings tabs to eligible US-based merchants.

<table>
  <tr>
    <th>Desktop</th>
    <th>Mobile</th>
  </tr>
  <tr>
    <td>Screenshot 2026-05-04 at 17 27 57</td>
    <td>Screenshot 2026-05-04 at 17 30 12</td>
  </tr>
</table>

**Banner behavior:**
- **✕ (close)** — permanently dismisses the banner (`ppcp_agentic_banner_dismissed`)
- **Remind me later** — snoozes for 7 days via a stored timestamp (`ppcp_agentic_banner_snoozed_until`); banner reappears automatically after the snooze expires
- **Apply for Early Access** — opens the survey URL in a new tab and records `ppcp_agentic_beta_status = applied`

**Eligibility (all conditions must be true):**
1. Merchant is connected to PayPal
2. A US shipping zone with active methods exists
3. At least 50 published products
4. At least 50 completed orders in the last 90 days
5. Banner is not currently snoozed
6. Banner has not been permanently dismissed

## Implementation

- New React component `AgenticBetaBanner` rendered conditionally in the settings screen
- New REST endpoint `AgenticBetaBannerEndpoint` with `/dismiss`, `/remind`, `/apply` routes
- New `AgenticBetaBannerEligibility` PHP service wired into `ScriptDataHandler` to expose the eligibility flag to JS via `ppcpSettings.isAgenticBetaBannerEligible`
- Generalized shared banner SCSS (previously migration-banner-specific)

## Test plan

- [ ] Ineligible store (non-US, <50 products, <50 orders) → banner not shown
- [ ] Eligible store → banner shown on all settings tabs
- [ ] Click ✕ → banner hidden permanently; `ppcp_agentic_banner_dismissed = 1` in DB
- [ ] Click "Remind me later" → banner hidden; `ppcp_agentic_banner_snoozed_until` set ~7 days from now; reappears after snooze expires
- [ ] Click "Apply for Early Access" → survey URL opens in new tab; `ppcp_agentic_beta_status = applied` in DB

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
@@ -19,7 +19,8 @@ import {
 const BANNER_ID        = 'ppcp-agentic-beta-banner';
 const BANNER_CLASS     = 'ppcp-r-settings-banner';
 const BANNER_TITLE_ID  = `${ BANNER_ID }-title`;
-const APPLY_URL        = 'https://example.com';
+// TODO: Replace with the real beta sign-up URL once it is confirmed by the PayPal team.
+const APPLY_URL = 'https://example.com';
 
 let isBannerDismissed = false;
 
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
@@ -85,7 +85,7 @@ const AgenticBetaBanner = ( {
 					</Action>
 				</Content>
 				<Content asCard={ false } className={ `${ className }__icon` }>
-					<PPIcon imageName="icon-button-payment-method-advanced-cards-large.svg" />
+					<PPIcon imageName="icon-paypal-business-loan.svg" />
 					<button
 						className={ `${ className }__icon-close` }
 						aria-label={ __(
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
@@ -19,7 +19,8 @@ import {
 const BANNER_ID = 'ppcp-agentic-beta-banner';
 const BANNER_CLASS = 'ppcp-r-settings-banner';
 const BANNER_TITLE_ID = `${ BANNER_ID }-title`;
-const APPLY_URL = 'https://example.com'; // TODO: Replace with the real beta sign-up URL once it is known.
+const APPLY_URL =
+	'https://docs.google.com/forms/d/e/1FAIpQLSeSMEJii54f4GfofEsPt1TbKK40NnOvq_Rs9ezXA6Ck5ZEDvg/viewform';
 
 let isBannerDismissed = false;
 
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
@@ -19,7 +19,7 @@ import {
 const BANNER_ID = 'ppcp-agentic-beta-banner';
 const BANNER_CLASS = 'ppcp-r-settings-banner';
 const BANNER_TITLE_ID = `${ BANNER_ID }-title`;
-const APPLY_URL = 'https://example.com'; // TODO: Replace with the real beta sign-up URL once it is confirmed by the PayPal team.
+const APPLY_URL = 'https://example.com'; // TODO: Replace with the real beta sign-up URL once it is known.
 
 let isBannerDismissed = false;
 
--- a/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
@@ -16,15 +16,14 @@ import {
 	applyForAgenticBeta,
 } from '@ppcp-settings/data/agentic-beta/actions';
 
+const BANNER_ID        = 'ppcp-agentic-beta-banner';
+const BANNER_CLASS     = 'ppcp-r-settings-banner';
+const BANNER_TITLE_ID  = `${ BANNER_ID }-title`;
+const APPLY_URL        = 'https://example.com';
+
 let isBannerDismissed = false;
 
-const AgenticBetaBanner = ( {
-	id,
-	className,
-	title,
-	description,
-	actionProps,
-} ) => {
+const AgenticBetaBanner = () => {
 	const [ isDismissed, setIsDismissed ] = useState( () => isBannerDismissed );
 
 	const dismiss = () => {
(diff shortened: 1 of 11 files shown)


modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Handles banner interactions for the agentic beta program:
+ * permanent dismissal, 7-day snooze (remind-me-later), and survey application.
+ */
+class AgenticBetaBannerEndpoint extends RestEndpoint {
+
+	public const OPTION_DISMISSED     = 'ppcp_agentic_banner_dismissed';
+	public const OPTION_STATUS        = 'ppcp_agentic_beta_status';
+	public const OPTION_SNOOZED_UNTIL = 'ppcp_agentic_banner_snoozed_until';
```
stracker-phil: suggestion: Almost all of the plugin options use the prefix `woocommerce-ppcp-` (and `-` instead of `_`)

```php
	public const OPTION_DISMISSED     = 'woocommerce-ppcp-agentic-banner-dismissed';
	public const OPTION_STATUS        = 'woocommerce-ppcp-agentic-beta-status';
	public const OPTION_SNOOZED_UNTIL = 'woocommerce-ppcp-agentic-banner-snoozed-until';
```

modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
```diff
@@ -0,0 +1,92 @@
...
+				'permission_callback' => array( $this, 'check_permission' ),
+			)
+		);
+	}
+
+	/**
+	 * Permanently dismisses the banner. Once dismissed it will never be shown again.
+	 */
+	public function handle_dismiss( WP_REST_Request $request ): WP_REST_Response {
+		update_option( self::OPTION_DISMISSED, true );
```
stracker-phil, line 67: suggestion: Why not move all `update_option()` and `get_option()` calls into the same class?

I would move the update_option() calls + the consts to the `AgenticBetaBannerEligibility` and inject that class as dependency into the Endpoint here.

Something like this:

```php
$eligibility = new AgenticBetaBannerEligibility();

$eligibility->dismiss();
$eligibility->snooze();
$eligibility->should_display(); // false if snoozed or dismissed
```

Does it make sense to have multiple options for this - could be a single option with the agentic beta test state, that recognizes several states:

```php
update_option( self::OPTION_DISMISSED, -1 ); // dismissed, never show again
update_option( self::OPTION_DISMISSED, time() + $snooze ) // snoozed, until the expiration time
delete_option( self::OPTION_DISMISSED ); // active banner
```

modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
```diff
@@ -0,0 +1,92 @@
...
+ * Handles banner interactions for the agentic beta program:
+ * permanent dismissal, 7-day snooze (remind-me-later), and survey application.
+ */
+class AgenticBetaBannerEndpoint extends RestEndpoint {
+
+	public const OPTION_DISMISSED     = 'ppcp_agentic_banner_dismissed';
+	public const OPTION_STATUS        = 'ppcp_agentic_beta_status';
+	public const OPTION_SNOOZED_UNTIL = 'ppcp_agentic_banner_snoozed_until';
+	public const STATUS_PENDING       = 'pending';
+	public const STATUS_APPLIED       = 'applied';
```
stracker-phil, line 26: Why do we need the status? it's not used anywhere. Could we simply dismiss the banner once the link was clicked (+ maybe add a log entry)?

modules/ppcp-settings/src/Service/AgenticBetaBannerEligibility.php
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Determines whether the agentic beta banner should be shown to the merchant.
+ *
+ * All seven conditions must be true for the banner to be eligible.
+ */
+class AgenticBetaBannerEligibility {
+
+	private const REQUIRED_PRODUCT_COUNT = 1;
+	private const REQUIRED_ORDER_COUNT   = 1;
```
stracker-phil: issue: those are test values. the specs say 50 products and 50 orders

modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
Narek13: Thanks. But why should the Eligibility do snooze or dismiss? I think that's the purpose of the Endpoint to handle these. I think the Endpoint is the right place to handle actions like dismiss/snooze since it owns the request/response cycle.

modules/ppcp-settings/src/Service/AgenticBetaBannerEligibility.php
```diff
@@ -0,0 +1,127 @@
...
+	 * @return bool
+	 */
+	public function is_eligible(): bool {
+		return $this->general_settings->is_merchant_connected()
+			&& $this->store_country === 'US'
+			&& $this->has_us_shipping_zone()
+			&& $this->has_enough_products()
+			&& $this->has_enough_recent_orders()
+			&& $this->is_not_snoozed()
+			&& ! get_option( AgenticBetaBannerEndpoint::OPTION_DISMISSED );
```
stracker-phil: suggestion: 
1. I would move this to the top. A simple option read is much cheaper than counting orders and products in the DB
2. Combine the banner-status into a single method, like `$this->should_display_banner()` (combining the is_not_snoozed + ! OPTION_DISMISSED)
3. **Store the result in a transient for 10 minutes**, otherwise it might slow down admin UI when recalculating order count, etc every time

modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
Narek13: Good point, it's not used yet. It was part of the original ticket spec, and the idea was to potentially drive future behaviour (e.g. showing a different banner state or skipping re-eligibility checks for merchants who already applied).

modules/ppcp-settings/resources/js/Components/Screens/Settings/index.js
```diff
@@ -14,6 +17,39 @@ const SettingsScreen = ( { activePanel, setActivePanel } ) => {
 				activePanel={ activePanel }
 				setActivePanel={ setActivePanel }
 			/>
+			{ isAgenticBetaBannerEligible && (
+				<AgenticBetaBanner
```
stracker-phil: I see no way that any of the props will ever change in the agentic banner

We could make this a bit cleaner by doing either of:
1. Rename/refactor `AgenticBetaBanner` to a generic banner component that really needs the props. Or:
2. Move all the props as hardcoded values into the component, so the call site is a simple

```js
{ isAgenticBetaBannerEligible && <AgenticBetaBanner /> }
```

modules/ppcp-settings/resources/js/Components/Screens/Settings/index.js
```diff
@@ -14,6 +17,39 @@ const SettingsScreen = ( { activePanel, setActivePanel } ) => {
...
+					) }
+					actionProps={ {
+						buttons: [
+							{
+								type: 'secondary',
+								text: __(
+									'Apply for Early Access',
+									'woocommerce-paypal-payments'
+								),
+								url: 'https://example.com',
```
stracker-phil: is this placeholder intentional?

if we do not know the exact URL yet, maybe add a `// todo` comment explaining why it's `example.com` ;)

modules/ppcp-settings/src/Service/AgenticBetaBannerEligibility.php
```diff
@@ -0,0 +1,127 @@
...
+	 * and the banner has not been permanently dismissed.
+	 *
+	 * @return bool
+	 */
+	public function is_eligible(): bool {
+		return $this->general_settings->is_merchant_connected()
+			&& $this->store_country === 'US'
+			&& $this->has_us_shipping_zone()
+			&& $this->has_enough_products()
+			&& $this->has_enough_recent_orders()
```
stracker-phil: note: we need one more check to verify the minimum PHP version (beta testers need PHP 8.1 or higher)

stracker-phil (changes requested):

modules/ppcp-settings/src/Endpoint/AgenticBetaBannerEndpoint.php
stracker-phil: My suggestion was more aiming to separate HTTP endpoint from state management.

Currently, the endpoint needs to accept and handle the REST calls **and** know state internals, like how snoozing works (ie. how the eligibility class interprets a snoozed value or which status values are recognized by the eligibility class)

My perspective: The eligibility class owns the decision "when do we show the banner?" and snoozing/dismissing is part of that scope.

In the future, if we have multiple banners, we'd more likely have one base class for eligibility that all inherit (eg only overriding an option name), which makes the state handling fast and consistent. If every future banner needs state management in a rest endpoint + in a read-only eligibility class, we're more likely running into issues, no?

And the single-option approach makes this even cleaner: the Eligibility class becomes the sole source of truth for both reading and writing that state, with no option-name leakage into the Endpoint at all.

Narek13: Thanks for the explaination.

Fair point on separation of concerns, but I'd flip the framing: if the Eligibility class owns writes, it's no longer just an eligibility class and should be renamed to something like `BannerStateManager`. The name isn't cosmetic here but it signals the contract. Right now `AgenticBetaBannerEligibility` answers one question: _should we show the banner?_ or _Is it eligible?_
So I think Mixing dismiss/snooze writes into it conflates two responsibilities.

On the single-option approach: I actually like that idea and happy to adopt it.
On the broader architecture: this banner is intentionally lightweight and temporary, and I'd rather keep it easy to delete. If we land a permanent banner system down the road, that's the right moment to design a proper base class.

stracker-phil (approved):
Thanks for the changes, looks very good now! 🙌

modules/ppcp-settings/src/Service/AgenticBetaBannerEligibility.php
```diff
@@ -0,0 +1,159 @@
...
+			return (bool) $cached;
+		}
+
+		$result = $this->general_settings->is_merchant_connected()
+			&& $this->store_country === 'US'
+			&& $this->has_us_shipping_zone()
+			&& $this->has_enough_products()
+			&& $this->has_enough_recent_orders();
+
+		set_transient( self::TRANSIENT_KEY, (int) $result, 10 * MINUTE_IN_SECONDS );
```
stracker-phil: idea: Personally I'd cache the result for 24+ hours, not only 10 minutes. The conditions are very unlikely to change within one day. But also happy to keep this as-is

modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/AgenticBetaBanner.js
```diff
@@ -0,0 +1,121 @@
...
+import {
+	dismissAgenticBetaBanner,
+	remindLaterAgenticBeta,
+	applyForAgenticBeta,
+} from '@ppcp-settings/data/agentic-beta/actions';
+
+const BANNER_ID = 'ppcp-agentic-beta-banner';
+const BANNER_CLASS = 'ppcp-r-settings-banner';
+const BANNER_TITLE_ID = `${ BANNER_ID }-title`;
+const APPLY_URL = 'https://example.com'; // TODO: Replace with the real beta sign-up URL once it is known.
```
stracker-phil: The URL is now known:

https://docs.google.com/forms/d/e/1FAIpQLSeSMEJii54f4GfofEsPt1TbKK40NnOvq_Rs9ezXA6Ck5ZEDvg/viewform

PR #4330, opened 2026-05-04
[QA] Disable playwright trace in CI to reduce artifact size

mishautkin:
### Description

Reduce Playwright report artifact size by disabling playwright trace in CI.


syde-piotr-ostrowski (approved):
all good

PR #4343, opened 2026-05-07
PayPal Subscriptions - Unsuccessfully payment when paypal subscription in cart from block checkout (6321)

Dinamiko:
This PR adds `default` to `reference_id` as fallback on PayPal subscription approval flow used by blocks checkout, it prevents the "No reference ID given." error on `ppc-approve-subscription` ajax endpoint.

### Steps To Reproduce
- Create subscription connected to plan
- Add this to cart and navigate to block checkout
- Click on PayPal button
- Observe payment can not be finished


Narek13 (approved):
LGTM, Thanks

PR #4349, opened 2026-05-11
Wrong payment method on order edit page, depending on sequence in WC Settings - Payments (6277)

Dinamiko:
When ACDC (or another PCP gateway) is sorted above PayPal in WC Settings → Payments, block-checkout express PayPal payments get routed to the wrong gateway. This causes the order edit page to show Payment via Debit & Credit Cards" instead of "Payment via PayPal".

This PR adds two hooks to fix the race condition:
- `woocommerce_rest_checkout_process_payment_with_context` (priority 100) re-routes the Store API checkout to PayPalGateway when a paypal_order_id is present.
- `woocommerce_before_order_object_save` prevents concurrent cart requests from overwriting _payment_method once the order has PayPal meta.
  
Additionally, `chosen_payment_method` is pinned to PayPal in the session during order approval to stop cart batch requests from resetting it before checkout.

### Steps To Reproduce
- Setup USA/USD store
- Enable block checkout pages
- Install PayPal and connect US merchant
- Enable PayPal and ACDC
- In WC Settings → Payments move ACDC to the very top
- On frontend add simple product to cart
- Navigate to block cehckout and pay with express PayPal (or Pay Later)
- In the dashboard navigate to newly created order edit page and check the “Payment via …” text

### Expected behaviour
- PayPal should be displayed in “Payment via …” instead of ACDC (see video and screenshots attached)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -993,6 +994,66 @@ function ( WC_Order $wc_order ) use ( $set_order_contacts ): void {
...
+
+				$order = $context->order ?? null;
+				if ( $order instanceof WC_Order ) {
+					$order->set_payment_method( PayPalGateway::ID );
+
+					$session_handler         = $c->get( 'session.handler' );
+					$funding_source_renderer = $c->get( 'wcgateway.funding-source.renderer' );
+
+					$funding_source = $payment_data['funding_source']
+						?: ( $session_handler->funding_source() ?: 'paypal' );
```
Narek13, line 1029: Using `?:` instead of `??` means an empty string `''` in `$payment_data['funding_source']` falls through to the session. Is that intentional?

Narek13 (approved):
LGTM. Thanks for fixing this

modules/ppcp-wc-gateway/src/WCGatewayModule.php
Dinamiko: @Narek13 yes `?:` is intentional, `''` is treated the same as missing/null on purpose, since we always want a valid funding source string.

PR #4351, opened 2026-05-11
PayPal subscription - Debit & Credit Cards is not available for this order error (6323)

Dinamiko:
Logged-in customers with saved credit card tokens see "Debit & Credit Cards is not available for this order" error when checking out with a PayPal subscription product and vaulting is disabled. This happens because the CC gateway no longer declares subscriptions support without vaulting, but the saved tokens cause block checkout to auto-select it anyway.

This PR unsets the CC gateway from available payment methods and strip saved CC tokens from the checkout page when the cart contains a PayPal subscription product that the CC gateway cannot support (i.e., vaulting is off).

### DIFF
--- a/modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
+++ b/modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
@@ -79,12 +79,13 @@ class DisableGateways {
 			unset( $methods[ CardButtonGateway::ID ] );
 		}
 
+		$payment_gateways = WC()->payment_gateways;
 		if (
 			isset( $methods[ CreditCardGateway::ID ] )
 			&& $this->subscription_helper->cart_contains_paypal_subscription_product()
-			&& ! is_null( WC()->payment_gateways )
+			&& ! is_null( $payment_gateways )
 		) {
-			$cc_gateway = WC()->payment_gateways->payment_gateways()[ CreditCardGateway::ID ] ?? null;
+			$cc_gateway = $payment_gateways->payment_gateways()[ CreditCardGateway::ID ] ?? null;
 			if ( $cc_gateway && ! in_array( 'subscriptions', $cc_gateway->supports, true ) ) {
 				unset( $methods[ CreditCardGateway::ID ] );
 			}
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-payment-tokens/src/WcPaymentTokensModule.php
```diff
@@ -106,6 +107,29 @@ function ( $tokens ) use ( $container ) {
...
+					&& ! $is_post
+					&& $container->has( 'wc-subscriptions.helper' )
+					&& ! is_null( WC()->payment_gateways )
+				) {
+					$subscription_helper = $container->get( 'wc-subscriptions.helper' );
+					if (
+						$subscription_helper instanceof SubscriptionHelper
+						&& $subscription_helper->cart_contains_paypal_subscription_product()
+					) {
+						$cc_gateway = WC()->payment_gateways->payment_gateways()[ CreditCardGateway::ID ] ?? null;
```
Narek13: Minor: `WC()->payment_gateways` is checked on line 115 and then `payment_gateways()` is called again. Maybe store it in a local variable to avoid repeating:

```php
$payment_gateways = WC()->payment_gateways->payment_gateways();
$cc_gateway       = $payment_gateways[ CreditCardGateway::ID ] ?? null;
```

modules/ppcp-wc-gateway/src/Checkout/DisableGateways.php
```diff
@@ -79,6 +79,17 @@ public function handler( array $methods ): array {
 			unset( $methods[ CardButtonGateway::ID ] );
 		}
 
+		if (
+			isset( $methods[ CreditCardGateway::ID ] )
+			&& $this->subscription_helper->cart_contains_paypal_subscription_product()
+			&& ! is_null( WC()->payment_gateways )
+		) {
+			$cc_gateway = WC()->payment_gateways->payment_gateways()[ CreditCardGateway::ID ] ?? null;
```
Narek13: Minor: same as below: `WC()->payment_gateways` is checked twice. Maybe store it in a local variable to avoid repeating:

Narek13 (approved):
Thanks. LGTM.
Just 1 minor suggestion which can be ignored.

PR #4362, opened 2026-05-13
Fix condition in AgenticCardBuilder (6398)

stracker-phil:
### Description

During code review, an incorrect condition was found in the `AgenticCartBuilder`.

Since the output schema always defines a `shipping_address`, we need to test for empty shipping address differently than for empty billing address:
- `$cart->shipping_address()->is_empty()` - never null, always an `Address` instance
- `$cart->billing_address() !== null` - can be null when not defined

This difference matches the API specs, that expect to always have shipping details, but omit billing data unless they are already known.


stracker-phil:
Already fixed in a different PR, closing this

PR #4370, opened 2026-05-13
Simplify checkout validation rules (6404)

stracker-phil:
### Description

The agentic commerce `CheckoutEndpoint` still contained outdated validation checks that are not needed, or should be refactored to match the current module architecture.

This PR addresses the legacy code and ensures the checkout uses the correct methods and validation structures.


Narek13 (approved):
LGTM.

PR #4372, opened 2026-05-14
Vaulted ACDC payments not captured + malformed order request custom_id/invoice_id misplaced (6293)

Dinamiko:
Vaulted ACDC (saved card) orders were created but never captured, two bugs found in the vault v3 code path:

  1. `custom_id` and `invoice_id` were placed at the root of the PayPal order creation payload instead of inside `purchase_units[0]`.
  2. For `CAPTURE` intent, `POST /v2/checkout/orders/{id}/capture` was never called after order creation, leaving the order uncharged.

### PR changes
  - Switch `CaptureCardPayment::create_order()` from `from_wc_cart()` to `from_wc_order()` so custom_id/invoice_id are correctly embedded inside `purchase_units[0]`.
  - Add the missing `order_endpoint->capture()` call for `CAPTURE` intent in `CreditCardGateway::process_payment()`.
  - Remove the now-unused `$prefix` constructor parameter from `CreditCardGateway`.

### DIFF
--- a/modules/ppcp-wc-gateway/services.php
+++ b/modules/ppcp-wc-gateway/services.php
@@ -140,7 +140,6 @@ return array(
 			$container->get( 'settings.environment' ),
 			$container->get( 'api.endpoint.order' ),
 			$container->get( 'wcgateway.endpoint.capture-card-payment' ),
-			$container->get( 'api.prefix' ),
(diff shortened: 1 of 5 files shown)


Narek13 (approved):
Thanks for working on this. Looks good to me

PR #4373, opened 2026-05-14
fix(settings): Add missing icon-paypal-business-loan.svg to ppcp-settings assets

Narek13:
## Problem

The file `modules/ppcp-settings/assets/images/icon-paypal-business-loan.svg` was inadvertently left uncommitted and is missing from the repository. It needs to be committed to the codebase for proper Agentic beta banner UI


stracker-phil (approved):
Thanks

PR #4376, opened 2026-05-15
Credit cards not saved when Save credit card is enabled (6410)

Dinamiko:
ACDC saved cards showed up under PayPal at `/my-account/payment-methods/` because the vault token was created with `gateway_id = ppcp-gateway` instead of `ppcp-credit-card-gateway`. 

Two causes: `SavePaymentMethodsModule` lacked a 'card' case in its payment-source switch, and WCGatewayModule's `before_order_object_save` hook rewrote any order with `_ppcp_paypal_order_id` meta to `PayPalGateway::ID`.

### PR changes
- Added case 'card': in `SavePaymentMethodsModule` calling `create_payment_token_card()`.
- Narrowed the `WCGatewayModule` hook to skip orders whose payment method already starts with `ppcp-`.


Narek13 (approved):
LGTM. Thanks

PR #4377, opened 2026-05-15
[QA] Autotest fixes for v4.0.4

mishautkin:
### Description

- Remove = in e2e: scripts in package.json
- Fix values in playwright.config
- Remove deleteAllOrders:
    Orders are deleted on env:reset. Removing orders restricts test analysis from checking orders for previous set of tests.
- Fix GooglePay test case key
- Add shard:transaction-mexico Playwright project
- Remove and gitignore snapshots/ dir
- Add .npmrc and E2E-TESTS.md to .distignore
- Rename setup files for correct sequence 
    When called in CLI "env:reset;|setup:wc;|setup:pcp:usa;" the setup files are sorted in alphabetic order. Env setup should run first, then woocommerce, then pcp
- Refactor Playwright projects and npm scripts:
    - Project "all" has been removed because it doesn't work since beforeAll hooks are moved to setups
    - Setup projects have been deduplicated and run sequentially: first - woocommerce, then - setup PCP, then - setup specific tested feature (e.g. transaction)
    - Refactor npm scripts to run with tags in parallel
- Replace @Dev test tag with @Smoke
- Update docs

See [CI run](https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/26151310477).


playwright.config.ts
```diff
@@ -20,15 +20,15 @@ const viewportSize: ViewportSize = { width: 1280, height: 850 };
...
+		timeout: 20_000,
 	},
 	timeout: 2 * 60_000,
 	/* Run tests in files in parallel */
 	fullyParallel: false,
 	/* Fail the build on CI if you accidentally left test.only in the source code. */
 	forbidOnly: !! process.env.CI,
 	/* Retry on CI only */
-	retries: process.env.CI ? 2 : 0,
+	retries: process.env.CI ? 1 : 0,
```
syde-piotr-ostrowski, line 31: @mishautkin 
Are you sure that we only need 1 retry on CI? from what I observe, some less stable tests needed 2 retries to pass sometimes.

mishautkin: In case of "real" fails which occured recently (for example multiple ACDC tests were failing and being retried) the test execution becomes hugely stretched in time. This was my major concern. What do you think?

syde-piotr-ostrowski: OK, let's try with `1` retry and observe the results during upcoming weeks.

syde-piotr-ostrowski (approved):
Looks good to me

PR #4381, opened 2026-05-19
OXXO - Extract + Block Checkout (3196)

Dinamiko:
OXXO lived in the `ppcp-wc-gateway` module and had no WooCommerce Block Checkout support, buyers using the block-based checkout page couldn't select OXXO as a payment method.

This PR moves OXXO into the dedicated `ppcp-local-alternative-payment-methods` module (where Multibanco, Blik, iDEAL, etc. already live) and register it as a block payment method via `AbstractPaymentMethodType`, using the standard shared APM component.


modules/ppcp-local-alternative-payment-methods/src/OXXOGateway.php
```diff
@@ -0,0 +1,208 @@
...
+use WooCommerce\PayPalCommerce\Assets\AssetGetter;
+use WooCommerce\PayPalCommerce\Button\Exception\RuntimeException;
+use WooCommerce\PayPalCommerce\Settings\Data\Definition\PaymentMethodsDefinition;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\TransactionUrlProvider;
+use WooCommerce\PayPalCommerce\WcGateway\Processor\RefundProcessor;
+
+class OXXOGateway extends WC_Payment_Gateway {
+
+	const ID = 'ppcp-oxxo-gateway';
```
Narek13:
```suggestion
	public const ID = 'ppcp-oxxo-gateway';
```

modules/ppcp-local-alternative-payment-methods/src/OXXOGateway.php
```diff
@@ -0,0 +1,208 @@
...
+
+			$response = $this->orders_endpoint->confirm_payment_source( $request_body, $body->id );
+			$body     = json_decode( $response['body'] );
+
+			$payer_action = '';
+			foreach ( $body->links as $link ) {
+				if ( $link->rel === 'payer-action' ) {
+					$payer_action = $link->href;
+				}
+			}
```
Narek13, line 150:
```suggestion
			foreach ( $body->links as $link ) {
				if ( $link->rel === 'payer-action' ) {
					$payer_action = $link->href;
					break;
				}
			}
```

adding `break` once the link is found avoids unnecessary iterations.

Narek13 (commented):
Thanks a lot for working on this. I left a few suggestions/questions.

I also noticed that you removed the unit tests for the old gateway, but tests for the new one were not added.

Dinamiko:
@Narek13 thanks for the suggestion, I've also added unit and integration tests.

Narek13 (approved):
Thanks a lot for adding suggestions. LGTM

PR #4389, opened 2026-05-20
fix(compat): disable legacy WC PayPal Standard gateway on PPCP merchant connect

slash1andy:
## Summary

- Hooks into `woocommerce_paypal_payments_authenticated_merchant` (fired after successful PPCP onboarding) to automatically disable the WC core PayPal Standard gateway.
- Sets both `enabled = 'no'` and `_should_load = 'no'` on `woocommerce_paypal_settings` — the only complete disable path (UI toggle alone is insufficient; see Problem section).
- Also disables `restore-paypal-standard-for-woocommerce` if installed (`woocommerce_restore_paypal_standard_settings['enabled'] = 'no'`) — a plugin (4,000+ installs) that sends no BN code and is not covered by the native WPS disable.
- WCS guard checks both `paypal` and `restore_paypal_standard` gateway IDs — if either has active subscriptions, stores a transient instead of disabling and renders a notice on the next admin page load.

---

## Proposed Change

Two additions to `modules/ppcp-compat/src/CompatModule.php`:

**1. Hook registration in `run()` method:**

```php
add_action(
    'woocommerce_paypal_payments_authenticated_merchant',
    array( $this, 'disable_legacy_paypal_standard_on_connect' )
);
add_action( 'admin_notices', array( $this, 'maybe_show_wps_subscriptions_notice' ) );
```

**2. New method `disable_legacy_paypal_standard_on_connect()`:**

```php
protected function disable_legacy_paypal_standard_on_connect(): void {
    if ( function_exists( 'wcs_get_subscriptions' ) ) {
        $total_count = 0;
        foreach ( array( 'paypal', 'restore_paypal_standard' ) as $gateway_id ) {
            $total_count += count(
                wcs_get_subscriptions(
                    array(
                        'payment_method'         => $gateway_id,
                        'subscription_status'    => array( 'active', 'pending-cancel' ),
                        'subscriptions_per_page' => -1,
                    )
                )
            );
        }

        if ( $total_count > 0 ) {
            set_transient( 'ppcp_wps_standard_subs_notice', $total_count, 30 * DAY_IN_SECONDS );
            return;
        }
    }

    $settings               = get_option( 'woocommerce_paypal_settings', array() );
    $settings['enabled']      = 'no';
    $settings['_should_load'] = 'no';
    update_option( 'woocommerce_paypal_settings', $settings );

    $rpsfw = get_option( 'woocommerce_restore_paypal_standard_settings', array() );
    if ( ! empty( $rpsfw ) ) {
        $rpsfw['enabled'] = 'no';
        update_option( 'woocommerce_restore_paypal_standard_settings', $rpsfw );
    }
}
```

**3. New method `maybe_show_wps_subscriptions_notice()`:**

```php
protected function maybe_show_wps_subscriptions_notice(): void {
    $count = get_transient( 'ppcp_wps_standard_subs_notice' );
    if ( false === $count ) {
        return;
    }
    delete_transient( 'ppcp_wps_standard_subs_notice' );
    $subscriptions_url = admin_url( 'edit.php?post_type=shop_subscription&payment_method=paypal' );
    printf(
        '<div class="notice notice-warning"><p>%s</p></div>',
        wp_kses_post(
            sprintf(
                _n( '...singular...', '...plural...', (int) $count, 'woocommerce-paypal-payments' ),
                (int) $count,
                esc_url( $subscriptions_url )
            )
        )
    );
}
```

The `_should_load = 'no'` write is the critical line — it clears the permanent latch that `enabled = 'no'` alone cannot clear.

The notice uses a transient (`ppcp_wps_standard_subs_notice`) to bridge the AJAX merchant-connect request and the next admin page load — a direct `add_action('admin_notices')` during the connect AJAX call would be in-memory only and never fire.

---

## Safety: WCS Guard

The WCS subscription check covers both `payment_method = 'paypal'` (native WPS) and `payment_method = 'restore_paypal_standard'` (restoration plugin). If either has active or pending-cancel subscriptions:

- Those subscriptions generate automatic renewal orders via PayPal IPN callbacks.
- The IPN endpoint is provided by `WC_Gateway_Paypal_IPN_Handler` (or the restoration plugin's equivalent).
- Disabling the gateway silently breaks those renewals — the merchant's subscribers would stop being billed with no error surfaced in WooCommerce.

The safe path is a persistent admin notice (via transient): tell the merchant how many subscriptions are still on PayPal Standard and link them to the subscription list. They can then migrate those subscriptions to PPCP or let them expire naturally before re-connecting.

---

## Test Plan

**Native WPS:**
- [ ] Fresh PPCP onboard on a site with no prior PayPal Standard orders → after connect: `woocommerce_paypal_settings['enabled'] = 'no'` and `woocommerce_paypal_settings['_should_load'] = 'no'`
- [ ] PPCP onboard on a site where WPS was previously enabled → same result: both keys set to `'no'`
- [ ] PPCP onboard on a site with active WCS PayPal Standard subscriptions (`payment_method = 'paypal'`) → transient `ppcp_wps_standard_subs_notice` set; on next admin page load, notice renders with correct subscription count and link; WPS NOT disabled
- [ ] PPCP onboard on a site with WCS installed but zero active PayPal Standard subscriptions → WPS disabled unconditionally
- [ ] PPCP onboard on a site without WCS installed → WPS disabled unconditionally (`function_exists` check passes silently)
- [ ] Confirm `WOOTHEMES_CART` BN code no longer appears in PayPal SDK calls after the change (check network requests during checkout on a post-connect site)
- [ ] Confirm IPN endpoint no longer responds after disable (verify `WC_Gateway_Paypal_IPN_Handler` is not instantiated)
- [ ] Re-connect / re-authenticate merchant → method runs again without side effects (idempotent: setting already-disabled keys to `'no'` is safe)

### DIFF
--- a/modules/ppcp-compat/src/CompatModule.php
+++ b/modules/ppcp-compat/src/CompatModule.php
@@ -86,6 +86,7 @@ class CompatModule implements ServiceModule, ExecutableModule {
 			'woocommerce_paypal_payments_authenticated_merchant',
 			array( $this, 'disable_legacy_paypal_standard_on_connect' )
 		);
+		add_action( 'admin_notices', array( $this, 'maybe_show_wps_subscriptions_notice' ) );
 
 		$this->legacy_ui_card_payment_mapping( $c );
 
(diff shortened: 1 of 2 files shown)


modules/ppcp-compat/src/CompatModule.php
```diff
@@ -87,10 +80,14 @@ function () use ( $c ) {
...
 
-		$this->initialize_blueprint_compat_layer( $c );
-
 		add_action( 'woocommerce_paypal_payments_gateway_migrate', static fn() => delete_transient( 'ppcp_has_ppec_subscriptions' ) );
 
+		add_action(
+			'woocommerce_paypal_payments_authenticated_merchant',
+			array( $this, 'disable_legacy_paypal_standard_on_connect' )
+		);
+		add_action( 'admin_notices', array( $this, 'maybe_show_wps_subscriptions_notice' ) );
```
AlexP11223, line 89: These methods are `protected`, so using the array syntax results in a fatal error. (the test would not catch this because it uses a subclass where it's allowed)

tests/PHPUnit/Compat/DisableLegacyPaypalStandardTest.php
```diff
@@ -0,0 +1,252 @@
...
+
+	protected function setUp(): void {
+		parent::setUp();
+		$this->sut = new TestableCompatModule();
+
+		// Reset option state between tests.
+		delete_option( 'woocommerce_paypal_settings' );
+		delete_option( 'woocommerce_restore_paypal_standard_settings' );
+		delete_transient( 'ppcp_wps_standard_subs_notice' );
+	}
```
AlexP11223, line 43: It's unclear how to run this test. It probably needs to be refactored either as our other unit tests with mocks/BrainMonkey, or as our integration tests in `tests/integration/PHPUnit`.

AlexP11223 (changes requested):

Issue #4388, opened 2026-05-20
WC Subscriptions cart silently captures with no renewal credential when vaulting path fails


superdav42:
### Describe the Bug

When the PPCP "vaulting" path is selected for a WooCommerce Subscriptions cart but the vault is **not actually established on PayPal's side** (vault eligibility flips false silently, merchant lacks vaulting on the live account, eligibility cache is stale, etc.), the initial CAPTURE still succeeds and the WC subscription is created with **no PayPal renewable artifact**. The renewal a month later has nothing to charge, WC Subscriptions cancels the subscription, and any downstream plugin tying user access to that subscription cascade-cancels too.

The customer sees a perfectly successful checkout (PayPal receipt, WC order complete, WC subscription active) and only learns months later that everything is gone, with no recoverable trace.

### To Reproduce

1. WC Subscriptions + PPCP active. PayPal account where vaulting is enabled in settings but vault is not actually granted/eligible for the live token (sandbox: a fresh personal buyer is enough).
2. Check out a recurring subscription product through the PayPal smart button.
3. CAPTURE succeeds. WC order `completed`, WC subscription `active`.
4. Inspect the WC subscription and parent order meta: no `_ppcp_billing_agreement_id`, no `_ppcp_subscription_id`, no `_ppcp_vault_id`, no `_paypal_subscription_id`, and `WC_Subscription::get_payment_tokens()` is empty.
5. Wait for renewal — the WC renewal attempt has no token to charge → fails → subscription cancels.

### Expected Behavior

When the vaulting path is chosen on a cart that contains a WC Subscription, PPCP should guarantee one of:
- a vault token is actually persisted on the subscription / parent order, OR
- the order is rejected before CAPTURE (loud failure the customer can act on), OR
- the path falls back to PayPal-side Subscriptions (`intent=SUBSCRIPTION`) automatically.

Silently capturing a recurring cart with no renewable artifact is the worst outcome.

### Actual Behavior

CAPTURE succeeds; vault never lands on PayPal's side; WC subscription is created with empty PPCP meta; renewal a month later cancels the subscription with no diagnostic that links the cancellation back to the original vault failure.

### Environment

- WooCommerce PayPal Payments: 4.0.4
- WooCommerce Subscriptions: 7.7.0
- WordPress: 7.0-RC2
- PHP: 8.3 / 8.4

### Code-Level Evidence (PPCP 4.0.4)

[...]

PR #4393, opened 2026-05-21
fix(settings): coerce stored values in typed-bool getters and vault_enabled migration

superdav42:
**Issue**: #4388 (related — same module, distinct symptom; this PR does not close it)

**Ticket**: n/a (community contribution from production triage)

**Slack Thread**: n/a

---

### Description

`SettingsModel` exposes ten typed-bool getters that return `$this->data[$key]` directly, trusting that whatever is stored under that key is a real `bool`. On PHP 8.x this throws

```
Return value must be of type bool, string returned
```

whenever the stored option holds a non-bool value. We hit this in production for `get_save_paypal_and_venmo()`, which fatals during `WcSubscriptionsModule` boot and prevents PPCP's REST routes (including the webhook endpoint at `/wp-json/paypal/v1/incoming`) from registering. The user-visible effect is that PayPal webhook events stop reaching the site silently — order/subscription state desyncs without any obvious error to the merchant.

Observed shapes of the stored value that produce the fatal:

- Legacy classic-WooCommerce `'yes'`/`'no'` strings (`vault_enabled` lineage).
- `'1'`/`'0'` from form-posted settings via paths that bypass the typed setter.
- Direct DB writes by import/export tools that re-serialise options without preserving scalar types.

The same defect pattern exists on all ten typed-bool getters; only `get_save_paypal_and_venmo()` has been observed firing in the wild, but the surface is identical and the fix is a uniform one-liner per getter.

#### Two-part defence

**1. Defensive coercion in every typed-bool getter.** Wrap each getter with the existing `DataSanitizer::sanitize_bool()` (which is `filter_var( ..., FILTER_VALIDATE_BOOLEAN )`). Pure pass-through for real bools; correctly handles `'yes'`/`'no'`/`'1'`/`'0'`/`'true'`/`'false'`/`'on'`/`'off'` strings. Each getter also gets a per-field default via `??` so a missing key falls back to the value declared in `get_defaults()` instead of producing a PHP warning.

Affected getters (all in `modules/ppcp-settings/src/Data/SettingsModel.php`):

- `get_authorize_only()`
- `get_capture_virtual_orders()`
- `get_save_paypal_and_venmo()`
- `get_instant_payments_only()`
- `get_enable_contact_module()`
- `get_save_card_details()`
- `get_enable_pay_now()`
- `get_enable_logging()`
- `get_stay_updated()`
- `get_payment_level_processing()` — note this one previously had a weak `(bool)` cast which incorrectly returns `true` for `'no'` (every non-empty, non-`'0'` string casts to true).

**2. Explicit `vault_enabled` arm in `SettingsTabMigration`.** Legacy classic-WooCommerce settings stored `vault_enabled` as `'yes'` / `'no'`. The migration currently routes unknown keys through a `default:` arm that copies the raw string into the new typed slot. From there, `from_array()` hands it to `set_save_paypal_and_venmo( bool $save )`, which either TypeErrors (strict mode) or silently persists `'no'` as `true` (weak mode, where every non-empty non-`'0'` string casts to true — including `'no'`). The new `case 'vault_enabled':` arm coerces explicitly via `filter_var( ..., FILTER_VALIDATE_BOOLEAN )` before assignment.

#### Out of scope

Other legacy bool fields routed through the `default:` arm may share the same latent defect, but I have no production evidence for them. Happy to expand the migration patch in a follow-up if maintainers prefer.

### Steps to Test

**Reproduce the fatal (without this patch):**

1. On a site with PPCP active, run:
   ```bash
   wp option patch update woocommerce-ppcp-data-settings save_paypal_and_venmo yes
   ```
   (or use any other typed-bool key from the list above).
2. Load any page that boots `WcSubscriptionsModule`, or hit `/wp-json/paypal/v1/incoming`.
3. Observe `TypeError: SettingsModel::get_save_paypal_and_venmo(): Return value must be of type bool, string returned` in the error log; the webhook route returns `rest_no_route`.

**Verify the fix:**

1. Apply this branch.
2. Repeat the steps above — the getter now returns `true` (and `'no'` returns `false`), no fatal, webhook route registers normally.
3. Run `vendor/bin/phpunit tests/PHPUnit/Settings tests/PHPUnit/PpcpSettings` — 240 tests pass.

**Verify the migration arm:**

1. Set up a site with the legacy WooCommerce PPCP settings containing `vault_enabled = "yes"`.
2. Trigger the settings-tab migration.
3. Confirm `save_paypal_and_venmo` is stored as bool `true` in the new option, not the string `'yes'`.

### Documentation

No user-facing behaviour change. Internal defensive coercion only.

### Changelog Entry

> Fix: defensively coerce stored values in `SettingsModel` typed-bool getters, and explicitly coerce `vault_enabled` during settings-tab migration, to prevent `TypeError: Return value must be of type bool, string returned` when the persisted option contains a non-bool value (legacy `'yes'`/`'no'` strings, direct DB writes, or third-party import/export tools).

Related: #4388.

### DIFF
--- a/modules/ppcp-settings/src/Data/SettingsModel.php
+++ b/modules/ppcp-settings/src/Data/SettingsModel.php
@@ -306,7 +306,7 @@ class SettingsModel extends AbstractDataModel {
 	 * @return bool True if authorize only is enabled, false otherwise.
 	 */
 	public function get_authorize_only(): bool {
-		return $this->data['authorize_only'];
+		return $this->sanitizer->sanitize_bool( $this->data['authorize_only'] ?? false );
 	}
 
 	/**
@@ -324,7 +324,7 @@ class SettingsModel extends AbstractDataModel {
 	 * @return bool True if capturing virtual orders is enabled, false otherwise.
 	 */
 	public function get_capture_virtual_orders(): bool {
-		return $this->data['capture_virtual_orders'];
+		return $this->sanitizer->sanitize_bool( $this->data['capture_virtual_orders'] ?? false );
 	}
 
 	/**
@@ -342,7 +342,9 @@ class SettingsModel extends AbstractDataModel {
 	 * @return bool True if saving PayPal and Venmo is enabled, false otherwise.
 	 */
 	public function get_save_paypal_and_venmo(): bool {
-		return $this->data[ FeaturesDefinition::FEATURE_SAVE_PAYPAL_AND_VENMO ];
+		return $this->sanitizer->sanitize_bool(
+			$this->data[ FeaturesDefinition::FEATURE_SAVE_PAYPAL_AND_VENMO ] ?? false
+		);
 	}
 
 	/**
@@ -360,7 +362,7 @@ class SettingsModel extends AbstractDataModel {
 	 * @return bool True if instant payments only setting is enabled, false otherwise.
 	 */
 	public function get_instant_payments_only(): bool {
-		return $this->data['instant_payments_only'] ?? false;
+		return $this->sanitizer->sanitize_bool( $this->data['instant_payments_only'] ?? false );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


Dinamiko (approved):
@superdav42 PR looks great, thanks!

PR #4405, opened 2026-05-28
PUI - Extract + Block Checkout (3197)

Dinamiko:
Pay upon Invoice (PUI) was implemented inside ppcp-wc-gateway, the monolithic gateway module. This made the module harder to maintain and was inconsistent with the pattern already used for all other local APMs (Bancontact, BLIK, OXXO, etc.). PUI also lacked Block Checkout support.

### PR Changes
Extracted PUI from `ppcp-wc-gateway` into `ppcp-local-alternative-payment-methods`, matching the existing APM pattern:
  - Moved all PUI classes to the new module
  - Added Block Checkout support via a custom React component that collects birth date + phone and passes them via onPaymentSetup
  - Removed cross-module class dependency in `PayUponInvoiceOrderEndpoint::create()`,  now accepts a plain array instead of the `PaymentSource` entity.


Narek13 (approved):

PR #4409, opened 2026-05-29
Negative fee on manually created order influence on wrong total amount on PayPal side (6463)

Dinamiko:
Fixes a regression from where a **negative fee** on a manually created order was counted twice (netted into `item_total` *and* added as a `discount`), undercharging the PayPal total (e.g. a $90 order sent as $80).

This PR adds the extracted item-discounts back into `item_total` in `AmountFactory::from_wc_order()` so the negative fee is represented only once. Orders without negative items are unaffected.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/AmountFactory.php
+++ b/modules/ppcp-api-client/src/Factory/AmountFactory.php
@@ -161,18 +161,25 @@ class AmountFactory {
 		$currency = $order->get_currency();
 		$items    = $this->item_factory->from_wc_order( $order );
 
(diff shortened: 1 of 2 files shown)


Narek13 (approved):
Thanks for working on this and thanks for adding the tests

hanneslsm:
Thanks for working on this. My client's shop is effected by this bug and lost some money.
I'm reverting back to 4.0.3 … or is there already a timeline on when this fix will be released?

Dinamiko:
@hanneslsm the next release is planed around mid/end June.

PR #4414, opened 2026-06-01
Avoid spamming OAuth requests when auth fails

AlexP11223:
Added rate-limiting for the failed auth token requests as well as some other small improvements related to that (checking empty credentials, adding expiration safety margin). 
The retry request on exception in `bearer()` was replaced with more explicit and limited retries (not 429 etc.).


Dinamiko (approved):

stracker-phil (approved):
Nice rate limiting solution :)

Narek13 (approved):
Thanks. LGTM

PR #4418, opened 2026-06-02
APM-S - Transaction link open paypal.com instead of sandbox.paypal.com (6487)

Dinamiko:
`TransactionUrlProvider` picked sandbox vs. live solely from the `_ppcp_paypal_payment_mode` order meta. That meta is set via OrderProcessor but the local APM gateways build orders directly and never set it, so the provider defaulted to the live URL.

`TransactionUrlProvider` now falls back to the current Environment (`is_sandbox()`) when the order meta is absent. When the meta is present it still takes precedence.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Gateway/TransactionUrlProvider.php
+++ b/modules/ppcp-wc-gateway/src/Gateway/TransactionUrlProvider.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 
 namespace WooCommerce\PayPalCommerce\WcGateway\Gateway;
 
+use WooCommerce\PayPalCommerce\WcGateway\Helper\Environment;
+
 /**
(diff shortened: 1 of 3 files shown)


modules/ppcp-wc-gateway/src/Gateway/TransactionUrlProvider.php
```diff
@@ -29,19 +31,29 @@ class TransactionUrlProvider {
 	 */
 	protected $transaction_url_base_live;
 
+	/**
+	 * The environment.
+	 *
+	 * @var Environment
+	 */
+	protected $environment;
```
Narek13, line 39:
```suggestion
	protected Environment $environment;
```

Narek13 (approved):
Thanks for working on this. I left a small suggestion, but it’s not blocking approval

PR #4421, opened 2026-06-03
Vault Management Component (5937)

Dinamiko:
This PR includes Vault Management Component  integration.


.gitignore
```diff
@@ -32,3 +32,4 @@ test-results
 /tests/qa/resources/files/woocommerce-paypal-payments*.zip
 /tests/qa/resources/files/woocommerce-subscriptions*.zip
 .google-session.json
+/*.png
```
stracker-phil: nit: Why do we need this rule?

It does not look related to any change here - it would be useful to have a short `# Comment` for new ignore rules to explain what case they handle.

modules/ppcp-button/resources/js/modules/Renderer/VaultRenderer.js
```diff
@@ -0,0 +1,141 @@
+import { loadScript } from '@paypal/paypal-js';
+
+const VAULT_NAMESPACE = 'ppcpVaultComponent';
+const CONTAINER_SELECTOR = '#ppcp-vault-component';
+
+// Module-level cache so repeated render calls don't trigger duplicate loads.
+let vaultSdkPromise = null;
+
+class VaultRenderer {
```
stracker-phil, line 11: suggestion: This renderer looks like it should belong into the new `ppcp-vault-component` module, not into `ppcp-button` ...

This class has zero dependencies on other `ppcp-button` components and could live anywhere.

If this sounds like a good idea, we could simply add a todo comment for the next iteration

modules/ppcp-button/resources/js/modules/Helper/ScriptLoading.js
```diff
@@ -69,10 +69,13 @@ export const loadPaypalScript = ( config, onLoaded, onError = null ) => {
 		scriptOptions = merge( scriptOptions, config.script_attributes );
 	}
 
-	// Adds data-user-id-token to script options.
-	const userIdToken = config?.save_payment_methods?.id_token;
-	if ( userIdToken && config?.user?.is_logged === true ) {
-		scriptOptions[ 'data-user-id-token' ] = userIdToken;
+	// data-sdk-client-token is reserved for the vault component's dedicated SDK
+	// namespace and must not appear on shared SDK script loads.
+	if ( config?.user?.is_logged === true && ! config?.vault_component?.is_eligible ) {
```
stracker-phil: nit: this condition is quite opaque. a better comment or structure could help;

`user.is_logged` checks if user is logged in, so this branch only modifies session for logged in users.

I assume that the `data-user-id-token` is only relevant for logged in customers, when the merchant is not eligible for the new Vaulting component, i.e. provide the current "fallback" logic, right?

modules/ppcp-blocks/resources/js/Components/vault-component.js
```diff
@@ -0,0 +1,139 @@
+import { useEffect, useRef, useState } from '@wordpress/element';
+import { loadScript } from '@paypal/paypal-js';
+
+const VAULT_NAMESPACE = 'ppcpVaultComponent';
+
+// Module-level cache so re-renders don't trigger duplicate loads.
+let vaultSdkPromise = null;
+
+export const VaultComponent = ( { config, onApproveOrder, onRenderError } ) => {
```
stracker-phil, line 11: suggestion: This component looks like it could live in the new `ppcp-vault-component` module.

Also, since it's a React component I would stick to the naming convention of using PascalCase for this file, ie `VaultComponent.js`

modules/ppcp-blocks/resources/js/Components/paypal-saved-token.js
```diff
@@ -0,0 +1,117 @@
...
+	}
+	const input = document.querySelector(
+		`input[name="radio-control-wc-payment-method-saved-tokens"][value="${ wcTokenId }"]`
+	);
+	return (
+		input?.closest( '.wc-block-components-radio-control__option' ) ?? null
+	);
+};
+
+export const PayPalSavedToken = ( {
```
stracker-phil, line 26: suggestion: This component looks like it could live in the new `ppcp-vault-component` module.

Also, since it's a React component I would stick to the naming convention of using PascalCase for this file, ie `PayPalSavedToken.js`

modules/ppcp-vault-component/services.php
```diff
@@ -0,0 +1,69 @@
...
+			$container->get( 'api.merchant.country' ),
+			$reference_transaction_status
+		);
+	},
+	'vault-component.supported-countries'             => static function ( ContainerInterface $container ): array {
+		return apply_filters(
+			'woocommerce_paypal_payments_vault_component_supported_countries',
+			array( 'US' )
+		);
+	},
```
stracker-phil: suggestion: IMO extracting this array into a service goes against the "high cohesion" rule.

Since it's tightly coupled to the logic inside the `VaultComponentApplies` class, I would suggest to move this filter/array into that class:

```php
class VaultComponentApplies {
    // ... current class content

    private function supported_countries(): array {
        return apply_filters(
            'woocommerce_paypal_payments_vault_component_supported_countries',
            array( 'US' )
            );
    }
}
```

If needed anywhere else, we can make the function `public` or even `public static`

modules/ppcp-vault-component/src/VaultComponentModule.php
```diff
@@ -0,0 +1,204 @@
...
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$token_db_id = wc_clean( wp_unslash( $_POST['wc-ppcp-gateway-payment-token'] ?? '' ) );
+		if ( ! $token_db_id || 'new' === $token_db_id ) {
+			return;
+		}
+
+		$token = WC_Payment_Tokens::get( (int) $token_db_id );
```
stracker-phil: suggestion: since the `$token_db_id` could be anything, we should check if it's actually a number > 0 before trying to find DB values.

Theoretically, `0` or lower values can return an empty `WC_Payment_Token` instance, which could create problems.

I'd add a simple check to directly reject unexpected IDs:

```php
if ( (int) $token_db_id < 1 ) {
    return;
}
```

modules/ppcp-vault-component/src/VaultComponentModule.php
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		return $localized_script_data;
+	}
+
+	/**
+	 * After Path A capture, update the WC payment token with the new FI details
+	 * from the PayPal order response.
+	 */
+	private function maybe_update_token_fi_details( Order $order ): void {
```
stracker-phil: suggestion: This function (and some of the above code in this file) are clearly business logic, which should possibly live in an isolated class. The module's responsibility is to orchestrate the services, not make business decisions or trigger side effects like writing meta data to the DB.

modules/ppcp-vault-component/src/Endpoint/CreateVaultOrderEndpoint.php
```diff
@@ -0,0 +1,154 @@
...
+					$paypal = $data['payment_source']['paypal'];
+					if ( is_array( $paypal ) && isset( $paypal['token'] ) ) {
+						unset( $data['payment_source']['paypal']['token'] );
+					} elseif ( is_object( $paypal ) && isset( $paypal->token ) ) {
+						unset( $data['payment_source']['paypal']->token );
+					}
+				}
+				return $data;
+			};
+			add_filter( 'ppcp_create_order_request_body_data', $strip_body, 99 );
```
stracker-phil: nit: This function looks a bit too complex for an anonymous inline declaration.

I'd rather make this a public class function that we register/unregister by its name + add a short unit test to verify all branches work - it has 4 if's and a single manual test will never verify all branches)

```php
add_filter( 'ppcp_create_order_request_body_data', array( $this, 'strip_request_body' ), 99 );

remove_filter( 'ppcp_create_order_request_body_data', array( $this, 'strip_request_body' ), 99 );
```

modules/ppcp-vault-component/src/Authentication/VaultClientToken.php
```diff
@@ -0,0 +1,131 @@
...
+			'body'    => $body,
+		);
+
+		$response = $this->request( $url, $args );
+		if ( $response instanceof WP_Error ) {
+			$this->logger->error( 'Vault client_token request failed: ' . $response->get_error_message() );
+			throw new RuntimeException( $response->get_error_message() );
+		}
+
+		$json        = json_decode( $response['body'] );
```
stracker-phil, line 93: issue: catch invalid JSON payloads using try-catch

When the response body cannot be decoded, `$json` will be null and the `$json->access_token` access will throw a fatal error.

Suggestion:
```php
try {
  $json = json_decode( $response['body'], false, 512, JSON_THROW_ON_ERROR );
} catch() {
  // handle invalid JSON
}

File: modules/ppcp-vault-component/src/Authentication/VaultClientToken.php [hunk 11]
```diff
@@ -0,0 +1,131 @@
...
+
+	use RequestTrait;
+
+	const CACHE_KEY_PREFIX = 'vault-client-token-key:';
+
+	/**
+	 * The host.
+	 *
+	 * @var string
+	 */
```
stracker-phil: suggestion: those doc-block comments are quite useless and can be removed; instead define the type of the properties directly:

```php
class VaultClientToken {

    private string $host;
    private ClientCredentials $client_credentials;
    // ...

```

stracker-phil (commented):
I've added several comments, none is a real blocker.

If some items are not addressed due to lack of time, I suggest to create a new issue or add some `// todo` comments for the next release.

The most "significant" item is this:
https://github.com/woocommerce/woocommerce-paypal-payments/pull/4421/changes#r3388500514 (potential fatal error when parsing invalid JSON strings)

Also, the `VaultComponentModule` seems to mix orchestration and business logic and should be refactored IMO
https://github.com/woocommerce/woocommerce-paypal-payments/pull/4421/changes#r3388122247

stracker-phil (approved):
Approving, to unblock the merge

modules/ppcp-wc-gateway/src/Gateway/PayPalGateway.php
```diff
@@ -334,6 +334,21 @@ public function get_description(): string {
 		return apply_filters( 'woocommerce_paypal_payments_gateway_description', wp_kses_post( $description ), $this );
 	}
 
+	/**
+	 * Renders payment fields including saved payment method radio buttons when tokenization is active.
+	 */
+	public function payment_fields(): void {
+		if ( $this->supports( 'tokenization' ) && is_checkout() ) {
```
AlexP11223: `supports( 'tokenization' )` - is it the right way to check if we need tokenization? I think it is always true, since it is just what we define in the constructor.

Dinamiko: @AlexP11223 This is the idiomatic WC guard for `saved_payment_methods()` and is filterable via
 `woocommerce_paypal_payments_paypal_gateway_supports`, so it is not strictly constant.

Dinamiko:
@stracker-phil @AlexP11223 thanks for all the feedback, I applied most of them (e.g., VaultClientToken JSON fatal, tokenization support...) and added todos for the rest, I'll merge it so we can start acceptance, thanks!

PR #4427, opened 2026-06-08
Show account or card last digits in payment method title

AlexP11223:
Addef saving of the card brand and last digits into the order meta, similar to other info that we save when finishing orders. 

And now the account or account info is displayed in the payment method title, for example, on the order received page and emails.

I'm not sure if it fully works for subscriptions, it seems to work for the first order, but maybe there are some issues with renewals in some cases.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
@@ -86,12 +86,7 @@ class PaymentMethodTitleEnricher {
 			return $title;
 		}
 
-		return sprintf(
-			/* translators: %1$s: payment method title, %2$s: contextual detail (payer email or card). */
-			__( '%1$s (%2$s)', 'woocommerce-paypal-payments' ),
-			$title,
-			$detail
-		);
+		return sprintf( '%1$s (%2$s)', $title, $detail );
 	}
 
 	/**
--- a/modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
+++ b/modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
@@ -1,0 +1,140 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\WcGateway\Helper;
+
+use WC_Order;
+use WooCommerce\PayPalCommerce\Applepay\ApplePayGateway;
+use WooCommerce\PayPalCommerce\Googlepay\GooglePayGateway;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\CreditCardGateway;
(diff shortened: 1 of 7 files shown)


modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
```diff
@@ -0,0 +1,140 @@
...
+	 * Gateways for which the title is enriched.
+	 *
+	 * @var string[]
+	 */
+	private const SUPPORTED_GATEWAYS = array(
+		PayPalGateway::ID,
+		CreditCardGateway::ID,
+		ApplePayGateway::ID,
+		GooglePayGateway::ID,
+	);
```
strangerkir, line 29: So it won't work, for example, for CardButtonGateway? I know that it makes no sense to enrich some of the gateways' titles, just wanted to make sure nothing is missing here.

strangerkir (approved):
Looks good to me.

One question, just in case, but I'm sure it is fine there.

modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
AlexP11223: I have not tried, but the task description said 

> Standard Card Button and local APMs are out of scope (no useful payment source data returned).

strangerkir: Ah, right, I see it now.

modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
```diff
@@ -0,0 +1,140 @@
...
+		if ( false !== strpos( $title, $detail ) ) {
+			return $title;
+		}
+
+		return sprintf(
+			/* translators: %1$s: payment method title, %2$s: contextual detail (payer email or card). */
+			__( '%1$s (%2$s)', 'woocommerce-paypal-payments' ),
+			$title,
+			$detail
+		);
```
stracker-phil: Nit: This return string contains nothing to translate, and the pattern `'%1$s (%2$s)'` could be used for anything else in the future.

My suggestion:
- Just use `sprintf( '%1$s (%2$s)', $title, $detail );` (my recommendation)
- Or, if we keep the translation, at least use `_x()` to provide context and avoid collisions in future

stracker-phil (approved):
Nice PR, well done! 👏 

Besides the minor nit (do we need a `__()`?) it's generally ready to merge IMO

modules/ppcp-wc-gateway/src/Helper/PaymentMethodTitleEnricher.php
AlexP11223: Removed.

PR #4430, opened 2026-06-09
PartnersEndpoint::seller_status() floods PayPal merchant-integrations endpoint with 403s (6520)

Dinamiko:
`SellerTypeResolver::resolve_unknown_seller_type()` was hooked to init, running on every frontend and admin page load. Whenever a merchant's `seller_type` was stored as unknown , each page load triggered a live merchant-integrations API call. These calls returned 403 Authorization Failed on PayPal's side. 

This PR stops resolving seller type on page loads. The endpoint is now only called when strictly necessary:
  - Removed the init hook that resolved seller type on every page load.
  - At onboarding: resolution now runs once in the `woocommerce_paypal_payments_authenticated_merchant` listener (clearing any stale failure first), before onboarding defaults are applied.
  - On the Overview "Refresh" button: `RefreshFeatureStatusEndpoint` now re-resolves seller type after clearing the seller-status cache/registry, giving merchants an explicit, throttled way to force a re-check after reconnecting.


stracker-phil (approved):
Solid and clean, great work! 👍

PR #4417, opened 2026-06-02
Vaulted ACDC capture lacks HTTP status validation, causing "Order does not contain an id" on transient PayPal errors (6500)

Dinamiko:
`CaptureCardPayment::create_order()` only guards against `WP_Error` and never checks the HTTP status code. On any non-2xx PayPal response the body has no id, so the factory throws "Order does not contain an id", masking the real error and breaking checkout

This PR validates the status code before parsing; on non-200/201 throw a PayPalApiException with the real PayPal error, mirroring `CapturePayPalPayment::create_order()`.

PR #4431, opened 2026-06-09
Vaulted ACDC saved-card checkout stuck in "payment pending" (6500 v2)

Dinamiko:
Saved-card (vaulted ACDC) checkouts ended in WooCommerce "payment pending". PR #4417 only handled non-2xx PayPal responses, but the failing orders returned HTTP 200, so its code never ran for them.

`CreditCardGateway::process_payment()` only captured the PayPal order when its status was `APPROVED`. Vaulted-card orders come back as `CREATED`, so capture was skipped, no order-status update happened, yet the gateway still returned success → the order silently stayed in "pending payment".

### PR Changes
In `CreditCardGateway::process_payment()`:
  1. Capture on `CREATED` as well as `APPROVED` (mirrors existing OrderProcessor logic) so vaulted-card orders actually get charged.
  2. Safety net, capture/status handling now runs inside the try (PayPal errors fail cleanly), and if the order doesn't reach a handled state (processing/completed/on-hold), it fails with a clear customer message instead of sticking in pending.


strangerkir (approved):
LGTM

Issue #4434, opened 2026-06-10
Advanced Card Processing in Checkout Blocks starts 3DS even when required checkout fields are invalid


marcofucito:
### Describe the Bug

When using WooCommerce PayPal Payments with **Advanced Card Processing / Card Fields** on the **WooCommerce Checkout Block**, the card payment flow starts even when required WooCommerce checkout fields are empty and already marked as invalid by WooCommerce Blocks.

The customer can leave all required checkout fields empty, enter only the card details, click the place order button, and the PayPal card/3DS flow still starts.

WooCommerce Blocks visually marks the required fields as invalid immediately, for example:

- Email address is required / invalid
- First name is required
- Last name is required
- Address is required

Despite these visible validation errors, the PayPal Advanced Card Processing 3DS modal opens.

This creates a confusing and risky checkout flow: the customer completes bank/3DS authentication before WooCommerce has accepted the required checkout data.

### To Reproduce

1. Set up WooCommerce with the Checkout Block.
2. Install and activate WooCommerce PayPal Payments.
3. Enable Advanced Card Processing / Card Fields.
4. Use the Storefront theme.
5. Disable all plugins except:
   - WooCommerce
   - WooCommerce PayPal Payments
6. Add a product to the cart.
7. Go to the Checkout Block page.
8. Do not enter email, name, address, or any other required checkout field.
9. Select the card payment method.
10. Enter only valid card details.
11. Click the place order button.

### Expected Behavior

WooCommerce/PayPal should not start the PayPal card submit / 3DS flow while required checkout fields are invalid.

The checkout should stop before any PayPal order/card/3DS flow begins, and the customer should be asked to complete the required WooCommerce checkout fields first.

### Actual Behavior

WooCommerce Blocks marks the required fields as invalid, but the PayPal Advanced Card Processing flow still starts and opens the 3DS/bank authentication modal.

### Reproduced With Conflict Test

This was reproduced on staging using a minimal conflict-test setup:

- Theme: Storefront 4.6.2
- Active plugins only:
  - WooCommerce 10.8.1
  - WooCommerce PayPal Payments 4.0.4
- Checkout type: WooCommerce Checkout Block
- Payment method: `ppcp-credit-card-gateway`

No other plugins were active during the conflict test.

The same behavior was also reproduced on the production-like staging setup before the conflict test.

### Technical Details Observed

[...]

PR #4435, opened 2026-06-10
Prevent Advanced Card Fields submission when Blocks checkout fields are invalid

marcofucito:
## Description

This PR prevents PayPal Advanced Card Fields from submitting when the WooCommerce Blocks checkout has validation errors for missing or invalid required checkout fields.

Currently, the Advanced Card Fields payment flow can call `cardFieldsForm.submit()` even when WooCommerce Blocks has already marked required checkout fields as invalid. In that scenario, the card authentication flow, including 3DS, may start before the shopper has completed required checkout information.

This change uses the native WooCommerce Blocks checkout validation flow for Advanced Card Fields:

- registers an `onCheckoutValidation` callback
- checks the Blocks validation store via `wc/store/validation`
- returns a Blocks error response when validation errors are present
- keeps a final `onPaymentSetup` guard before `cardFieldsForm.submit()`

When invalid checkout fields are present, the shopper is asked to complete the required checkout fields before continuing with payment, and the PayPal card fields submission is not started.

## Why

Starting card authentication before required checkout fields are valid can create a confusing shopper experience:

1. The shopper enters card details.
2. 3DS/authentication starts.
3. WooCommerce Blocks still has required checkout fields missing.
4. The payment flow cannot complete cleanly.

This PR keeps the Advanced Card Fields flow aligned with the native Blocks checkout validation state before starting the PayPal card submission.

## Notes

WooCommerce Blocks remains responsible for rendering validation errors, focusing fields, and positioning the shopper in the checkout form.

A separate, translated error message is returned if the PayPal card fields form is not ready yet, so initialization/runtime readiness problems are not misreported as invalid card field input.

## Testing

Tested with a WooCommerce Blocks checkout using Advanced Card Fields:

- With required checkout fields missing, clicking place order no longer starts the 3DS/card authentication flow.
- WooCommerce Blocks validation errors remain visible.
- The shopper sees an error asking them to complete required checkout fields.
- After completing the required checkout fields, the card payment flow can proceed.

Build:

```bash
npm run build
```

Build completed successfully. Existing upstream webpack warnings remain unrelated to this change.

Related issue: #4434

### DIFF
--- a/modules/ppcp-blocks/resources/js/Components/card-fields.js
+++ b/modules/ppcp-blocks/resources/js/Components/card-fields.js
@@ -19,8 +19,19 @@ import {
 import { cartHasSubscriptionProducts } from '../Helper/Subscription';
 import { __ } from '@wordpress/i18n';
 
+const CHECKOUT_FIELDS_NOT_VALID_MESSAGE = __(
+	'Please complete all required checkout fields before continuing with payment.',
+	'woocommerce-paypal-payments'
+);
+
+function hasCheckoutValidationErrors() {
+	const validationStore = wp?.data?.select?.( 'wc/store/validation' );
+
+	return validationStore?.hasValidationErrors?.() || false;
+}
+
 export function CardFields( { config, eventRegistration, emitResponse } ) {
-	const { onPaymentSetup } = eventRegistration;
+	const { onPaymentSetup, onCheckoutValidation } = eventRegistration;
 	const { responseTypes } = emitResponse;
 
 	const [ cardFieldsForm, setCardFieldsForm ] = useState();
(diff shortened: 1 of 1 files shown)


marcofucito:
The implementation has since been simplified to use the native WooCommerce Blocks checkout validation flow via `onCheckoutValidation` and `wc/store/validation`. The previous DOM/focus handling approach is no longer part of this PR.

modules/ppcp-blocks/resources/js/Components/card-fields.js
```diff
@@ -43,10 +54,53 @@ export function CardFields( { config, eventRegistration, emitResponse } ) {
...
+			if ( typeof onCheckoutValidation !== 'function' ) {
+				return undefined;
+			}
+
+			return onCheckoutValidation(
+				() => {
+					if ( hasCheckoutValidationErrors() ) {
+						return {
+							type: responseTypes.ERROR,
+							message: CHECKOUT_FIELDS_NOT_VALID_MESSAGE,
```
AlexP11223, line 68: It seems it should be `errorMessage` here, not `message`. Currently it is not displayed anywhere, so it should be either renamed to `errorMessage`, or removed completely (keep only `type`), since normally there already should be some messages from WooCommerce.

AlexP11223 (commented):
Thank you for the contribution.

Dinamiko (approved):

PR #4437, opened 2026-06-11
Vaulting subscription - Unsuccessfully payment with acdc from block chekcout (6529)

Dinamiko:
Logged-in **returning** customers couldn't pay for a vaulting subscription by card on blocks checkout, the order failed with `422 UNPROCESSABLE_ENTITY / VALIDATION_ERROR`. Guests and new logged-in customers were unaffected.

A PayPal order carrying `payment_source.card.attributes.customer.id` becomes immutable to any purchase-unit `PATCH` once its card source is confirmed. On block checkout the card is confirmed before the WooCommerce order exists, so `OrderProcessor`'s post-confirmation `patch_order()` (which only syncs `custom_id`/`invoice_id`) is rejected. Guests have no `customer.id`, so their order stays patchable. CAPTURE still succeeds on the un-patched order, the amount was already final from the cart.
  
### PR Changes
`OrderProcessor::patch_order()` now catches `PayPalApiException`: on a **422 with an unchanged purchase-unit amount**, it logs and skips the patch so capture proceeds. Any other failure, including an actual amount change is re-thrown. All other flows still patch normally, preserving the numeric `custom_id` webhooks rely on for order mapping.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
+++ b/modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
@@ -17,6 +17,8 @@ use WooCommerce\PayPalCommerce\ApiClient\Entity\ExperienceContext;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\OrderStatus;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentSource;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\PurchaseUnit;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\PayPalApiException;
 use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
 use WooCommerce\PayPalCommerce\ApiClient\Factory\ExperienceContextBuilder;
 use WooCommerce\PayPalCommerce\ApiClient\Factory\OrderFactory;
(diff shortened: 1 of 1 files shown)


modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -266,17 +268,78 @@ public function create_order( WC_Order $wc_order, string $funding_source = 'payp
...
+			 * confirmation happens before the WooCommerce order exists, so this patch
+			 * runs afterwards and PayPal rejects it with 422 VALIDATION_ERROR. The patch
+			 * here only syncs metadata (custom_id/invoice_id); the amount was already
+			 * final when the order was created from the cart. When the amount is
+			 * unchanged it is therefore safe to skip the patch and capture the order
+			 * as-is. Any other failure (including an actual amount change) is re-thrown.
+			 */
+			if ( $exception->status_code() !== 422 || ! $this->order_amount_unchanged( $order, $updated_order ) ) {
+				throw $exception;
+			}
```
stracker-phil, line 293: **issue**: this could pass-through legit business-rule violations

The skip is currently gated on `status_code === 422` + unchanged total amount. Because the patch replaces the whole purchase unit, that gate also swallows unrelated 422s where the total happens to match (like `ITEM_TOTAL_MISMATCH` / `TAX_TOTAL_MISMATCH`) or a malformed invoice_id. Those are real validation failures we'd be silently dropping the patch for.

Suggestion: add a `has_detail()` check on the specific issue code in addition to the current checks, so the skip is scoped to the documented card-immutability case. We already have the `PayPalApiException::has_detail( string $issue )` ability:

```php
// define the error code on class level
public const ORDER_LOCKED_ISSUE = '...'; // todo

// Then only allow this specific issue to pass
if (
    $exception->status_code() !== 422
    || ! $exception->has_detail( self::ORDER_LOCKED_ISSUE )
    || ! $this->order_amount_unchanged( $order, $updated_order )
) {
    throw $exception;
}
```

stracker-phil (approved):

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
stracker-phil: Update: After some more checks, I think this is a **nice-to-have** to tighten the logic, but not an issue

Reason: If this patch call fails for any reason, it means the latest patch either was malformed or the order was locked. But the initially created order exists and was valid in the first place. This valid order can be captured, and the total amount check is sufficiently protecting the merchant here.

PR #4439, opened 2026-06-11
Free trial vaulting subscription - Unsuccessfully payment with acdc and PayPal (6528)

Dinamiko:
Vaulting/subscription payments (free-trial ACDC + PayPal, and admin renewals) intermittently failed with `403 NOT_AUTHORIZED "Authorization failed due to insufficient permissions"` on Vault v3 calls. 

The PayPal access token is cached for ~9h and only refreshed on connect/disconnect. If a merchant connects before Advanced Vaulting is provisioned (or scopes change afterward), the cached token lacks the vault scope and every vault call fails until it naturally expires. A renewal also surfaced this as an admin-ajax 500.

### PR Changes
  - **Self-healing retry** (`RequestTrait` + `ApiModule`): on a `401` or `403 NOT_AUTHORIZED`, flush the cached bearer token and retry the request once with a freshly issued token (with current scopes).
  - **Proactive refresh** (`RefreshFeatureStatusEndpoint`): flush the API cache when a merchant clicks "Refresh features", so newly granted vaulting takes effect immediately instead of after ≤9h.
  - **Graceful degradation**: the renewal/gateway paths no longer surface a raw 500, a vault failure fails the order cleanly with an actionable message.

### DIFF
--- a/modules/ppcp-api-client/src/ApiModule.php
+++ b/modules/ppcp-api-client/src/ApiModule.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 namespace WooCommerce\PayPalCommerce\ApiClient;
 
 use WC_Order;
+use WooCommerce\PayPalCommerce\ApiClient\Authentication\PayPalBearer;
+use WooCommerce\PayPalCommerce\ApiClient\Exception\RuntimeException;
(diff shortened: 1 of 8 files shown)


danieldudzic (approved):
Thanks for working on this! Clean.

PR #4450, opened 2026-06-15
Dev/pcp 6500 fix vaulted card payments

stracker-phil:
# Description

## 🐛 The problem

Saved (vaulted) card payments failed consistently at checkout, while a fresh manually entered card always worked. Affected orders never left **"pending payment"**, or failed outright with a generic error.

## 💡 Why it happens

When a vaulted card's issuer mandates Strong Customer Authentication, PayPal does not charge the card immediately. It returns the order in `CREATED` / `PAYER_ACTION_REQUIRED` status together with a `payer-action` link the buyer must visit to complete a 3D Secure challenge.

The saved-card path never handled that step. It assumed the charge was always frictionless: it tried to capture right away and, when PayPal had instead handed back a challenge link, there was nothing to capture. The order was left stuck.

Two pieces were missing:
- The request never asked PayPal to run 3D Secure, so SCA-required cards had no path to authenticate.
- There was no return handling to resume the capture after the buyer completed the challenge.

## ✅ The fix

The vaulted-card charge now drives the full 3D Secure round trip.

- **Request the challenge.** `CaptureCardPayment::create_order()` adds the merchant's configured 3DS contingency (`SCA_WHEN_REQUIRED` / `SCA_ALWAYS`) and an `experience_context` (locale, brand name, return and cancel URLs) so PayPal can present a branded, localized challenge and return the buyer to the right place.
- **Redirect on step-up.** When PayPal returns a `payer-action` link, `CreditCardGateway` stores a one-time nonce on the order and redirects the buyer to the challenge instead of trying to capture.
- **Resume on return.** `ReturnUrlEndpoint::maybe_resume_card_3ds()` handles the return (which arrives without the PayPal order token), matches the nonce, and re-runs payment processing to capture the now-authenticated order.
- **Shared finalize.** Authorize/capture, transaction-id recording, status transition and the stuck-order safety net are extracted into `finalize_vaulted_card_order()`, used by both the initial charge and the resume so the two paths stay identical and idempotent.

A frictionless saved card (no `payer-action` link) charges directly, exactly as before, and now skips a redundant order fetch on that path.

## 🔍 What to review

- **The nonce is the only thing guarding the resume.** The WC order id travels in a public, guessable query argument on the return URL; the PayPal order token does not come back. Authorization rests entirely on the one-time `_ppcp_card_3ds_resume` nonce stored on the order, compared with `hash_equals` and cleared on first use. Please scrutinize that a crafted return URL cannot trigger a capture, and that the nonce is genuinely single-use (a new attempt overwrites it, so only the latest challenge is resumable).
- **Resume re-enters `process_payment`.** The return handler calls the gateway's `process_payment` again rather than capturing directly. Worth confirming this is safe to run twice and that `finalize_vaulted_card_order`'s status checks keep it idempotent if PayPal already advanced the order during the return.
- **3DS only fires when the merchant enabled it.** If the 3DS setting is empty, no verification is requested and behavior is unchanged. The new flow is only exercised for merchants with 3DS configured.

## 🧪 Tests

Unit tests cover the request-body construction (vault id, stored-credential, 3DS verification method present/absent, experience context). A new integration test, `VaultedCard3dsTransactionTest`, exercises the end-to-end flow: redirect to the challenge, resume-and-capture, resume-and-authorize, and the rejection branches (unknown gateway, nonce mismatch, failed re-processing). The CAPTURE and AUTHORIZE intents are both covered.

## 🧪 How to verify

1. In **PayPal Payments → Settings**:
   - **Common Settings:** enable "Save PayPal and Venmo" and "Save Credit and Debit Cards".
   - **Other Payment Method Settings:** set **3D Secure: ALWAYS**.
2. As a logged-in user, check out on classic checkout, enter card `4000 0000 0000 2503` (a PayPal 3DS test card), and tick **Save Card** to vault it.
3. Check out again using that saved card → you are redirected to the 3DS challenge.
4. Complete the challenge → you return to the shop and the order reaches **processing/completed** (or **on-hold** for AUTHORIZE intent).
5. Cancel the challenge instead → the order stays unpaid and you land back on checkout.

### Sample
https://github.com/user-attachments/assets/0d1b1fe0-5e94-4e70-966b-765c1b119324

### DIFF
--- a/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
@@ -118,6 +118,8 @@ class CaptureCardPayment {
 				'usage'             => 'SUBSEQUENT',
 			),
 			'experience_context' => $this->experience_context_builder
+				->with_current_locale()
+				->with_current_brand_name()
 				->with_custom_return_url( $card_3ds_return_url )
 				->with_custom_cancel_url( wc_get_checkout_url() )
 				->build()
--- a/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
@@ -52,12 +52,7 @@ class CaptureCardPayment {
 
 	private SettingsProvider $settings_provider;
 
-	/**
-	 * The experience context builder.
-	 *
-	 * @var ExperienceContextBuilder
-	 */
-	private $experience_context_builder;
+	private ExperienceContextBuilder $experience_context_builder;
 
 	/**
 	 * The logger.
--- a/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
@@ -97,10 +97,8 @@ class CaptureCardPayment {
 		 * PayPal's vaulted-card 3D Secure return hits the return URL WITHOUT the
 		 * order token (it appends liability_shift/code/state instead), so
 		 * ReturnUrlEndpoint cannot identify the order by token. Encode the WC order
-		 * id plus a per-order resume nonce in the return URL: the order id locates
-		 * the order, and the nonce (matched against the order meta on return)
-		 * prevents the resume from being triggered by a manually crafted URL. A
-		 * cancelled challenge returns to the checkout URL and the order stays unpaid.
+		 * id plus a per-order resume nonce in the return URL so the return handler
+		 * can locate and authenticate the order.
 		 */
 		$card_3ds_return_url = add_query_arg(
 			array(
--- a/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
@@ -89,7 +89,7 @@ class CaptureCardPayment {
 	 *
 	 * @throws RuntimeException When request fails.
 	 */
-	public function create_order( string $vault_id, WC_Order $wc_order ): Order {
+	public function create_order( string $vault_id, WC_Order $wc_order, string $resume_nonce = '' ): Order {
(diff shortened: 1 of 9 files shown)


modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
```diff
@@ -50,6 +52,13 @@ class CaptureCardPayment {
 
 	private SettingsProvider $settings_provider;
 
+	/**
+	 * The experience context builder.
+	 *
+	 * @var ExperienceContextBuilder
+	 */
+	private $experience_context_builder;
```
Narek13:
```suggestion
	private ExperienceContextBuilder $experience_context_builder;
```

modules/ppcp-wc-gateway/src/Gateway/CreditCardGateway.php
```diff
@@ -448,6 +459,33 @@ public function process_payment( $order_id ) {
...
+		$provided_resume_nonce = wp_unslash( $_GET['ppcp_resume_nonce'] ?? '' );
+		$provided_resume_nonce = is_string( $provided_resume_nonce ) ? sanitize_text_field( $provided_resume_nonce ) : '';
+		$stored_resume_nonce   = (string) $wc_order->get_meta( self::THREE_DS_RESUME_META );
+
+		if ( $stored_resume_nonce && $provided_resume_nonce && hash_equals( $stored_resume_nonce, $provided_resume_nonce ) ) {
+			$wc_order->delete_meta_data( self::THREE_DS_RESUME_META );
+			$wc_order->save();
+
+			try {
+				$order = $this->order_endpoint->order(
```
Narek13, line 478: JFYI, you are catching only `RuntimeException` but I see that `order_endpoint::order` also throws `PayPalApiException`.

Narek13 (approved):
Thanks @stracker-phil for working on this. Looks really good to me, although I added minor notes, which you can feel free to ignore.

modules/ppcp-wc-gateway/src/Gateway/CreditCardGateway.php
stracker-phil: I've checked the code, and it's safe:

```php
class PayPalApiException extends RuntimeException
```

Dinamiko (approved):

PR #4454, opened 2026-06-16
Add `item_group_id` to ingestion (6551)

stracker-phil:
# Description

## 🎯 The goal

PayPal's product ingestion API now requires `item_group_id` on **every** product entry. Variations already carried it (the parent product's id), but simple and other non-variant products were sent without it, so those entries are now rejected.

Per the new spec, a product with no variants sends its own `id` in both the `id` and `item_group_id` fields.

## ✅ The fix

Non-variant entries now set `item_group_id` to the product's own id; variations keep sending the parent id. This closes the gap that left simple products without the field.

Alongside the field addition, the inline array-building in `ProductsPayload` is replaced by a `ProductDTO` value object. The DTO owns the wire-format contract in one place:

- the camelCase `merchantStoreUrl` key (the lone outlier in an otherwise snake_case payload),
- the omit-when-empty behaviour for the six optional fields (`mpn`, `sale_price`, `product_type`, `color`, `size`, `gender`).

`ProductsPayload::get_array()` becomes `get_products()` and returns `ProductDTO[]`; `SyncJob` serialises each via `to_array()` when composing the request body. All extraction logic stays in `ProductsPayload`.

## 🧪 Tests

`ProductDTOTest` covers the wire-format contract directly: required keys always present, the `merchantStoreUrl` camelCase outlier, optionals omitted when null and present when set, and `item_group_id` carrying whatever was passed (so a simple product's own id and a variation's parent id are both exercised).

`ProductsPayloadTest` is updated to the new `get_products()` return type and adds a case asserting a simple product's `item_group_id` equals its `id`.

### DIFF
--- a/modules/ppcp-store-sync/src/Ingestion/ProductDTO.php
+++ b/modules/ppcp-store-sync/src/Ingestion/ProductDTO.php
@@ -1,0 +1,106 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WooCommerce\PayPalCommerce\StoreSync\Ingestion;
+
+/**
+ * Single product entry of the agentic-commerce ingestion payload.
+ *
+ * A passive value carrier: it defines the payload schema and serialises itself via
+ * {@see ProductDTO::to_array()}, which owns the wire-format contract — the
+ * camelCase `merchantStoreUrl` key and the omit-when-empty behaviour for optional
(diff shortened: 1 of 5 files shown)


Narek13 (approved):
Clean and straightforward implementation 🙂 Thanks for working on this!

PR #4482, opened 2026-06-28
Validate and retry vaulted card order when PayPal response is missing order id (6368)

danieldudzic:
### Description
PayPal's `v2/checkout/orders` endpoint intermittently returns HTTP 200/201 with a response body missing the required `id` field, causing vaulted card orders to fail with "Order does not contain an id". Customers can recover by refreshing and resubmitting, proving the issue is transient.

This PR validates the decoded response before passing it to the order factory, and retries once with a new idempotency key when the response is incomplete. Applied to both `CaptureCardPayment` and `CapturePayPalPayment`.

### Steps to Test

1. Enable Advanced Card Processing and Vault v3
2. Save a credit card to a customer account
3. Checkout using the saved card, order should complete successfully
4. Check `WooCommerce > Status > Logs` for normal flow (no warnings)
5. To verify retry behavior, place a [debug mu-plugin](https://gist.github.com/danieldudzic/6c0326d97fbc5b873a7a845eb0d4fe3a) that strips the `id` from the first API response and confirm the order still completes with `WARNING: PayPal order response missing id.` and `INFO: Retrying PayPal order creation after incomplete response.` in logs

### DIFF
--- a/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
+++ b/modules/ppcp-wc-gateway/src/Endpoint/CaptureCardPayment.php
@@ -164,19 +164,57 @@ class CaptureCardPayment {
 			'body'    => wp_json_encode( $data ),
 		);
 
-		$response = $this->request( $url, $args );
-		if ( $response instanceof WP_Error ) {
-			throw new RuntimeException( $response->get_error_message() );
-		}
-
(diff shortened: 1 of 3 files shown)


Narek13 (approved):

Dinamiko (approved):

PR #4487, opened 2026-07-01
Hide Apple Pay button for subscription carts

Dinamiko:
On classic checkout, the Apple Pay button showed under the PayPal gateway when a vaulting subscription was in the cart.

Apple Pay vault does not work with the current implementation based on customer id and payment tokens, so it can't renewal a subscription using a savedpayment token. Unlike Google Pay, Apple Pay lacked the guards to hide itself for subscription carts. 

### PR Changes                                                                                                                                                                       
  - Add a subscription guard to `ApplePayButton::render()`
  - Return `false` from `validateContext()` for subscription contexts; remove the unused `isVaultV3Mode()`
  - Drop the subscriptions feature push in block checkout.

### DIFF
--- a/modules/ppcp-applepay/resources/js/Context/BaseHandler.js
+++ b/modules/ppcp-applepay/resources/js/Context/BaseHandler.js
@@ -7,17 +7,9 @@ class BaseHandler {
 		this.ppcpConfig = ppcpConfig;
 	}
 
-	isVaultV3Mode() {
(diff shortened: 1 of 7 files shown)


Narek13 (approved):
LGTM. Thanks for working on this and adding tests :)

Dinamiko:
Fixed in #4546.

PR #4490, opened 2026-07-01
API spam on merchant-integrations endpoint for merchants missing MANAGE_SUBSCRIPTIONS scope (6625)

Dinamiko:
The plugin repeatedly called PayPal's merchant-integrations (seller status) endpoint on merchants hitting persistent errors (e.g. a 403 from under-scoped REST apps). Failed responses were never cached, and the failure backoff was only enforced in one caller (ProductStatus, 60s). `ReferenceTransactionStatus` which runs on frontend paths, called `seller_status()` directly with no backoff, so they hit the endpoint on every page load.                                                                                                                                                                                       
                                                                                                                                                                                                                         
This PR centralizes the failure backoff inside `PartnersEndpoint::seller_status()` so every caller backs off for 10 minutes (matching the success cache TTL) after a real failure, anchored to the last actual API error.  Fallback handling was extracted into a shared helper used by both the backoff and error paths. Also aligned ProductStatus's in-request check to the same window.

### DIFF
--- a/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
+++ b/modules/ppcp-api-client/src/Endpoint/PartnersEndpoint.php
@@ -143,38 +143,62 @@ class PartnersEndpoint {
 			return apply_filters( 'woocommerce_paypal_payments_seller_status', $cached );
 		}
 
+		/*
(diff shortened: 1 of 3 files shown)


Narek13 (approved):
LGTM, Thanks

PR #4508, opened 2026-07-09
Add filter to allow Pay Later together with Vaulting (6443)

AlexP11223:
Added a filter (`woocommerce_paypal_payments_pay_later_with_vaulting`) so that the merchants whitelisted by PayPal can enable Pay Later buttons and messaging together with vaulting ("Save PayPal and Venmo"), which is not normally possible right now.

After enabling this filter, the Pay Later method does not become automatically disabled and greyed out in the payment method tab in the settings, and the Pay Later messaging configuration page also becomes available. If the filter gets disabled while Pay Later and vaulting are enabled, then Pay Later is automatically disabled as before. 

Also renamed the `vaultingEnabled` property into `payLaterDisabledByVaulting` in the Pay Later blocks, since now it is not simply about the vaulting state. This property is only used for displaying a message that it is not possible to use Pay Later while vaulting is enabled.

### DIFF
--- a/modules/ppcp-settings/src/Data/SettingsProvider.php
+++ b/modules/ppcp-settings/src/Data/SettingsProvider.php
@@ -423,6 +423,41 @@ class SettingsProvider {
 	}
 
 	/**
+	 * Whether Pay Later may run alongside vaulting ("Save PayPal and Venmo").
+	 *
+	 * By default the plugin disables all Pay Later features when vaulting is
+	 * active. PayPal-whitelisted merchants can return true from the filter below
+	 * to bypass that restriction.
+	 *
+	 * @return bool True when Pay Later is allowed together with vaulting.
+	 */
+	public function pay_later_with_vaulting_enabled(): bool {
+		/**
+		 * Filters whether Pay Later features may run while "Save PayPal and Venmo"
+		 * (vaulting) is active.
+		 *
+		 * Intended for PayPal-whitelisted accounts that are allowed to offer Pay
+		 * Later and Vaulting at the same time.
+		 *
+		 * @param bool $enabled Whether Pay Later is allowed alongside vaulting. Default false.
+		 */
+		return (bool) apply_filters( 'woocommerce_paypal_payments_pay_later_with_vaulting', false );
+	}
+
+	/**
+	 * Whether Pay Later is currently disabled because of vaulting.
+	 *
+	 * Reflects the live, effective state: true only when vaulting ("Save PayPal and
+	 * Venmo") is currently enabled AND the merchant has not opted into the override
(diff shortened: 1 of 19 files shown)


modules/ppcp-wc-gateway/src/Helper/SettingsStatus.php
Narek13, line 50: A doubt: here we don't add the same vaulting gate as below which was added to the button check but not the messaging-location check. [SmartButton::should_load_messages()](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8895dcf900986a02ebb3ec26ed84251af4f0de43/modules/ppcp-button/src/Assets/SmartButton.php#L591) calls this method directly and bypasses the separate `should_render_pay_later_messaging` filter entirely, so when vaulting is active and the override is off, the button correctly reports disabled while messaging can still report enabled and cause the PayPal SDK's messages component to load.

modules/ppcp-settings/src/Data/SettingsProvider.php
```diff
@@ -422,6 +422,41 @@ public function save_paypal_and_venmo(): bool {
...
+	/**
+	 * Whether Pay Later is currently disabled because of vaulting.
+	 *
+	 * Reflects the live, effective state: true only when vaulting ("Save PayPal and
+	 * Venmo") is currently enabled AND the merchant has not opted into the override
+	 * via {@see self::pay_later_with_vaulting_enabled()}.
+	 *
+	 * @return bool True when Pay Later features must be suppressed because of vaulting.
+	 */
+	public function pay_later_disabled_by_vaulting(): bool {
```
Narek13, line 456: Note: [here](https://github.com/woocommerce/woocommerce-paypal-payments/blob/8895dcf900986a02ebb3ec26ed84251af4f0de43/modules/ppcp-settings/resources/js/Components/Screens/Settings/Components/Settings/Blocks/SavePaymentMethods.js#L43-L44) we stale toggle description. The "Save PayPal and Venmo" toggle still hardcodes "...This will disable the Pay Later payment method on your site," untouched by this PR. For a whitelisted merchant who applies the new filter, this is now `false`, so the backend correctly keeps Pay Later enabled while the UI keeps telling them it won't be.

AlexP11223: fixed

modules/ppcp-wc-gateway/src/Helper/SettingsStatus.php
AlexP11223: hm, I think I cannot reproduce it, but yes, it probably makes sense to add that check here, so added it.

Narek13 (approved):
Thanks for the changes. LGTM

PR #4526, opened 2026-07-14
Core SDK loading & PayPal buttons via v6 Web SDK (5780)

danieldudzic:
**_Note:  This PR is against the v6 SDK epic branch: `dev/PCP-5777-update-v-5-js-sdk-integration-to-v-6-web-sdk` and not `dev/develop`_**

### Description

Core v6 Web SDK frontend: `createInstance()` with a server-fetched client token, explicit eligibility via `findEligibleMethods()` (currency/country/amount), one-time payment sessions and Web Component buttons for PayPal/Venmo/Pay Later across product, cart, checkout, and classic mini-cart.

- **Server side is 100% shared with v5.** v6 calls the existing `ppc-*` endpoints through one JS adapter and reuses existing services/helpers (`SettingsStatus`, styling DTOs, `Context`, `SingleProductActionHandler`, `ErrorHandler`). Only change outside the module: one `do_action` in `CreateOrderEndpoint` (stores the session order for non-checkout contexts; v5 unchanged).
- **Completion flow matches classic v5 continuation:** approve, redirect to checkout, Place Order processes the session order.

**Migration methodology (v5 and v6 side by side):** during the epic, every payment surface is owned by exactly one stack. A single function (`SdkV6Manager::should_load_on_current_page()`) decides per page whether v6 loads. On those pages the v5 smart buttons are suppressed, since both SDKs claim `window.paypal`. Everywhere else (block cart/checkout, pay-now, wallets, messaging) v5 keeps running untouched. This keeps the site and the QA/e2e suites fully functional after every story lands: each subsequent ticket simply widens what v6 owns instead of adding per-surface toggles or shims. The handoff is migration-phase scaffolding only, marked for removal at release (see `extensions.php`). The end state is a whole-flow flag flip: v5 or v6, never a per-page mix.

### Screenshots

| Before | After |
| :--- | :--- |
|v5|v62|

*Missing on the v6 side by design: messaging, wallets - included in separate tickets*

### Steps to Test

**Setup:** flag on via `PCP_SDK_V6_ENABLED=1` (or filter `woocommerce.feature-flags.woocommerce_paypal_payments.sdk_v6_enabled` set to true), sandbox-connected store, `npm run build`.

1. **Product page:** gold PayPal/Venmo/Pay Later Web Components render in `#ppc-button-ppcp-gateway-v6`. Network shows `web-sdk/v6/core` plus one `ppc-sdk-v6-client-token` POST, and **no** `paypal.com/sdk/js`. No `[PPCP SDK v6]` console errors.
2. **Purchase:** click PayPal, approve as sandbox buyer, get redirected to checkout, click Place Order. Verify the order in wp-admin (Processing, PayPal txn ID, stock reduced).
3. **Shipping in popup:** from cart, change the address to another US state. `update-customer` (non-empty `state`) and `ppc-update-shipping` fire, popup totals update.
4. **Settings:** shape/color changes in Styling apply; disabling a the script there.
5. **Coexistence / flag off:** block cart/checkout still show v5 e off, no v6 anywhere (identical to `dev/develop`).
6. **Suites:** `npm run unit-tests`, `npm run test:unit-js`, `npm cp-blocks/checkout-handler` JS failures are unrelated; they fail on
`dev/develop` too).


Dinamiko (approved):

modules/ppcp-button/src/Endpoint/CreateOrderEndpoint.php
```diff
@@ -370,6 +370,17 @@ public function handle_request(): void {
 				throw $exception;
 			}
 
+			/**
+			 * Fires after the PayPal order has been created via the create-order endpoint.
+			 *
+			 * Unlike woocommerce_paypal_payments_paypal_order_created (fired for
+			 * every API order creation), this action also receives the request data.
+			 *
+			 * @param \WooCommerce\PayPalCommerce\ApiClient\Entity\Order $order The created PayPal order.
```
stracker-phil: nit: Replace the FQCN with a use statement

Reason: this is the general convention in the project; it also keeps the comment significantly shorter

```php
use \WooCommerce\PayPalCommerce\ApiClient\Entity\Order;

...

/**
 * ...
 * @param Order $order The created PayPal order.
 * @param array $data  The request data.
 */
```

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -39,21 +49,61 @@ class SdkV6Manager {
 	 */
 	private Environment $environment;
 
+	/**
+	 * The button style mapper.
+	 *
+	 * @var ButtonStyleMapper
+	 */
```
stracker-phil: suggestion: Remove those doc-block comments

Reason: They just repeat what's visible in the declaration (type + verbose version of the variable name) and add no value - neither for devs nor for AI context.

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -39,21 +49,61 @@ class SdkV6Manager {
...
-	 * @param string      $version The assets version.
-	 * @param Environment $environment The environment object.
+	 * @param AssetGetter       $asset_getter The asset getter.
+	 * @param string            $version The assets version.
+	 * @param Environment       $environment The environment object.
+	 * @param ButtonStyleMapper $style_mapper The button style mapper.
+	 * @param bool              $should_handle_shipping Whether to handle shipping in PayPal.
+	 * @param SettingsStatus    $settings_status The settings status helper.
+	 * @param Context           $context The page context helper.
 	 */
```
stracker-phil: suggestion: Remove this comment (noise + maintenance overhead)

Reason: Same as above, this is just noise. The constructor is called exactly once, in a DI service and this documentation does not provide any useful insight for devs that cannot be gained by just looking at the code below.

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -84,12 +139,103 @@ public function enqueue(): void {
...
+		// otherwise this only happens as a side effect of constructing the
+		// (discarded) v5 SmartButton.
+		$this->context->init_context();
+
+		if ( $this->settings_status->is_smart_button_enabled_for_location( 'product' ) ) {
+			add_action(
+				'woocommerce_single_product_summary',
+				function (): void {
+					$this->render_wrapper();
+				},
```
stracker-phil: nit: Could use arrow-functions (for most of the action hooks in this function)

Reason: IMO makes the code slightly more readable because it's more compact

```php
add_action(
	'woocommerce_single_product_summary',
	fn() => $this->render_wrapper(),
	31
);
```

modules/ppcp-sdk-v6/src/SdkV6Module.php
```diff
@@ -53,6 +63,39 @@ static function () use ( $c ) {
...
+			'wp',
+			static function () use ( $c ) {
+				if ( is_admin() ) {
+					return;
+				}
+
+				$manager = $c->get( 'sdk-v6.manager' );
+				assert( $manager instanceof SdkV6Manager );
+
+				$manager->register_render_hooks();
```
stracker-phil: suggestion: Move the `register_render_hooks()` method to this class

Reason: Almost all actions are wired in a `...Module` class in this project; IMO it makes the logic easier to understand - all add_action() calls are in the main Module

Idea/sample:

```php
private function register_render_hooks( SdkV6Manager $manager ): void {
    $places = $manager->determine_render_places(); // returns array ['product' => true, 'cart' => false, ...]

	if ( $places['product'] ) {
		add_action(
			'woocommerce_single_product_summary',
			static fn() => $manager->render_wrapper(),
			31
		);
	}

	// ...
}
```

Also, in the current implementation I see that all render-hooks are passed through a filter which is currently missing - need to confirm why we have those filters and if we want to carry this logic into the V6 implementation, eg:

```php
// SmartButton.php:1710
private function checkout_button_renderer_hook(): string {
	/**
	 * The filter returning the action name that PayPal button will use for rendering on the checkout page.
	 */
	return (string) apply_filters(
		'woocommerce_paypal_payments_checkout_button_renderer_hook',
		'woocommerce_review_order_after_payment'
	);
}
```

modules/ppcp-sdk-v6/resources/js/sessions/shippingHandler.js
```diff
@@ -0,0 +1,90 @@
...
+			'Content-Type': 'application/json',
+			Nonce: storeApi.nonce,
+		},
+		body: JSON.stringify( body ),
+	} );
+
+	if ( ! response.ok ) {
+		throw new Error( 'Store API request failed.' );
+	}
+}
```
stracker-phil: suggestion: move this function to the `utils/api.js` module

Reason: it's a generic API caller that sets headers with the nonce; it looks like a method that we might reuse in other modules and that is not really tied to the shipping handler.

modules/ppcp-sdk-v6/resources/js/boot.js
```diff
@@ -1,108 +1,209 @@
...
 
 	if ( document.readyState === 'loading' ) {
-		document.addEventListener( 'DOMContentLoaded', init );
+		document.addEventListener( 'DOMContentLoaded', renderAll );
 	} else {
-		init();
+		renderAll();
+	}
+
+	if ( typeof jQuery !== 'undefined' ) {
```
stracker-phil: question: Is the `jQuery` gate the behavior we want here?
    
The initial render runs unconditionally, but the live-update bindings (`updated_checkout`, `wc_fragments_refreshed`, `updated_cart_totals`, …) are jQuery-gated. On a store without jQuery, the button renders once and then silently stops updating - could lead to stale totals/eligibility/branding.

This could lead to unreplicable, store-specific support tickets if someone eventually adds load-bearing logic that silently no-ops on some sites.

I'd suggest to make this method cleanly run-all or skip-all, never halfway.

- For all classic-checkout/cart: jQuery is effectively a hard dependency; missing jQuery should fail loud and obviously.                                        
- For block pages: I think jQuery is not needed here, but this condition isn't enough, as blocks need their own React components.

A first step could be to at least show a clear console error to help with debugging of unexpected issues:

```js
// utility idea
export function hasJQuery() {
  if (typeof jQuery === 'undefined' ) {
    console.error( '[PPCP SDK v6] jQuery not present' );
    return false;
  }

  return true;
}

// here:
if ( hasJQuery() ) {
  jQuery(...)
}
```

stracker-phil (approved):
Ready to merge; all comments are nits/suggestions

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -10,14 +10,24 @@
...
+use WooCommerce\PayPalCommerce\Button\Endpoint\CreateOrderEndpoint;
+use WooCommerce\PayPalCommerce\Button\Helper\Context;
 use WooCommerce\PayPalCommerce\SdkV6\Endpoint\ClientTokenEndpoint;
+use WooCommerce\PayPalCommerce\SdkV6\Helper\ButtonStyleMapper;
 use WooCommerce\PayPalCommerce\WcGateway\Helper\Environment;
+use WooCommerce\PayPalCommerce\WcGateway\Helper\SettingsStatus;
 
 /**
  * Class SdkV6Manager
  */
```
Narek13: I think we should remove these comments, they don't convey anything

modules/ppcp-sdk-v6/src/Helper/ButtonStyleMapper.php
```diff
@@ -0,0 +1,75 @@
...
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\SdkV6\Helper;
+
+use WooCommerce\PayPalCommerce\Settings\Data\SettingsProvider;
+
+/**
+ * Class ButtonStyleMapper
+ */
```
Narek13:
```suggestion
```

Same here, I think this is redundant

modules/ppcp-sdk-v6/src/SdkV6Module.php
```diff
@@ -29,6 +32,13 @@ public function services(): array {
 		return require __DIR__ . '/../services.php';
 	}
 
+	/**
+	 * {@inheritDoc}
+	 */
```
Narek13:
```suggestion
```

These are also not needed cause modern IDEs will recognize them from interface

Narek13 (approved):
Really nice PR, thank you! It looks great to me. I’ve also added a couple of small suggestions on top of what @stracker-phil already mentioned, but I think we’re good to go. Thanks again! 🙂

modules/ppcp-sdk-v6/resources/js/boot.js
danieldudzic: Added the `hasJQuery()` first step you suggested - logs [PPCP SDK v6] jQuery not present and gates the live-update bindings in `boot.js`, so a jQuery-less store no longer fails silently.

Deferring the full run-all/skip-all split to a separate issue `5783` (Blocks): the block-side half needs the React components built there, so it can't be settled in this issue :)

PR #4531, opened 2026-07-15
[QA] Fix OXXO tests PCP-6638

mishautkin:
### Description

OXXO's hosted checkout page was being treated as a popup with the wrong wait sequencing, causing flaky/incorrect test completion. This PR introduces a proper ApmHostedCheckout page object and unifies the OXXO test flow with the rest of classic checkout.

> **Note:** Temporarily disabled the final order status assertion in CI - Ngrok is required for doing it but it can not be used for parallel executions in current state.

**Changes:**

- Add ApmHostedCheckout POM (apm-hosted-checkout.ts) wrapping PayPal's sandbox APM simulator page, registered as a Playwright fixture.
- Fix completeOxxoPayment (paypal-ui.ts / paypal-ui-classic.ts) to submit the order, wait for the hosted checkout page, click "Test Successful Payment", and wait for the redirect to Order Received — instead of the previous popup-close-without-clicking approach. Removed the duplicate classic override now that behavior matches the base class.
- Loosen getPayPalPaymentDetails (paypal-api.ts) to fetch real PayPal capture details for OXXO (previously hardcoded to undefined), since the webhook capture completes before we read it.
- Delete transactionsOnClassicCheckoutOxxo — it's now functionally identical to transactionsOnClassicCheckout once the above fixes landed. OXXO test data now runs through the shared scenario function directly.

CI test run: https://github.com/woocommerce/woocommerce-paypal-payments/actions/runs/29478489928/job/87557175339

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,6 +21,7 @@ on:
           - shard:settings
           - shard:transaction-usa
           - shard:transaction-mexico
+          - shard:transaction-germany
           - shard:refund
           - shard:vaulting
           - shard:subscription
@@ -71,6 +72,7 @@ jobs:
             settings
             transaction-usa
             transaction-mexico
+            transaction-germany
             refund
             vaulting
             subscription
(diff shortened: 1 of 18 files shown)


Narek13 (approved):

PR #4535, opened 2026-07-15
Enrich connection details with merchant country (6690)

stracker-phil:
# Description

## 🐛 The problem

With the Store Sync feature flag active, the **Store Sync toggle is missing** for connected merchants.

Root symptom: after onboarding, the merchant's real PayPal-account country (`merchant_country`) is left empty.

Affected surface: the empty value is hidden almost everywhere, because most code reads the country via `get_merchant_country()`, which falls back to the WooCommerce store country. Store Sync reads the raw value, so that is where the gap surfaces. It also blocks DE/MX APM auto-enable during onboarding.

## 💡 Why it happens

The post-connect country lookup runs in the same request that authenticates the merchant, where the shared API client is a **login-only bearer** that cannot consume `seller_status()`. The seller-status back-off compounds it, so the lookup returns nothing and `merchant_country` stays empty.

## ✅ The fix

A new `PartnersEndpointFactory` builds a `PartnersEndpoint` with a fresh authenticated bearer from the credentials just stored at connect, so the lookup no longer depends on the memoized login bearer. This is the key that makes `seller_status()` reachable in the connecting request.

`MerchantDataResolver` uses it to resolve the country in-process: inline on connect (settings page correct on first load) and as an upgrade backfill, with a bounded ActionScheduler retry (max 3) as the fallback so an unresolvable merchant is not retried forever.

This replaces the previous loopback approach: `InternalRestService` and `AuthenticationManager::enrich_merchant_details()` are removed.

## 🔍 What to review

- **The back-off is intentionally never cleared.** It is the primary flood protection, the retry cap is the backstop. A persistent 403 leaves the country empty and the store-country fallback in place, by design.
- **`needs_resolution()` relies on the DTO trim** (`'' ===`, no local trim). Removing the `MerchantConnectionDTO` trim would make a whitespace country read as resolved.
- **Triggers must stay one-shot** (connect, upgrade). Wiring the resolver to a per-request event like `admin_init` would re-run the lookup on every page load for an unresolvable merchant.

## 🧪 How to verify

1. With the Store Sync flag active, connect a fresh PayPal sandbox account.
2. Confirm the **Store Sync toggle appears** on first load of the settings page.
3. Confirm the stored `merchant_country` holds the PayPal account country, not just the store country.
4. On a site already connected with an empty `merchant_country`, trigger the update migration and confirm the country backfills.
5. Onboard a genuine DE (or MX) account and confirm Pay upon Invoice (or OXXO) is auto-enabled.

### Changelog Entry

Fixed the missing Store Sync toggle by reliably resolving the PayPal merchant country after onboarding and on plugin update.


modules/ppcp-settings/src/SettingsModule.php
```diff
@@ -371,6 +372,10 @@ static function () use ( $container ): void {
 					$logger
 				);
 
+				$country_resolver = $container->get( 'settings.service.merchant-data-resolver' );
+				assert( $country_resolver instanceof MerchantDataResolver );
+				$country_resolver->ensure_country_resolved();
```
Dinamiko, line 377: I suggest to move the country resolver above the seller-type resolver (line 369) so it runs while the shared failure flag (`FailureRegistry`) is still clear, otherwise could be blocked by the failure that the seller-type resolver's own failing call writes just before it.

Dinamiko (changes requested):

modules/ppcp-settings/src/SettingsModule.php
stracker-phil: Good catch; however, when looking at the code, I see no benefit of inverting the calls

- The failure cache is cleared in line 368, so next next `seller_status` request will always pass to the API
- The `resolve_unknown_seller_type()` is the original way to fill in the merchant country
- Our new `ensure_country_resolved()` method is a fallback that will schedule a cron task in case the country was not resolved

ie having the new function at the last position adds an fail-safe at the end of the onboarding process, that retries the check in case something in the current request failed

Btw: The actual fix for this problem is the new `PartnersEndpointFactory` that creates an authenticated version of the API handler and fixes the unknown-seller-type check

Dinamiko (approved):

PR #4542, opened 2026-07-20
Insonsistent PayPal button behavior for subscriptions on cart (6639)

Dinamiko:
The PayPal button behaved inconsistently for subscription carts. On the block cart it showed for one subscription product but appeared then disappeared when a second (free-trial) one was added; on the classic cart the opposite happened (hidden for one, shown for two). Separately, emptying the block cart left the mini-cart still showing products.

### PR Changes
  - Centralized the button-display rule into a single shared decision `SubscriptionHelper::paypal_subscription_button_allowed()` (PHP) and `paypalSubscriptionButtonAllowed()` (JS), exposed as one `subscription_button_allowed` server flag. It covers both modes: PayPal Subscriptions requires a single item with a PayPal plan; vaulting requires vaulting enabled; non-subscription carts are always allowed.                                                                                                                                                                  
  - All three cart paths (classic, block, mini-cart) now consume this one rule, so they behave identically.                                                                                  
  - The button is decided before rendering and the wrapper is withheld server-side for unsupported carts (no flicker), and truly hidden rather than greyed.                                  
  - The mini-cart now subscribes to wc/store/cart, clearing the button when the block cart empties and re-applying the same subscription rule.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ContextBootstrap/MiniCartBootstrap.js
+++ b/modules/ppcp-button/resources/js/modules/ContextBootstrap/MiniCartBootstrap.js
@@ -47,7 +47,8 @@ class MiniCartBootstrap {
 			wp.data.subscribe(
 				debounce( () => {
 					this._handleStoreCartChange();
-				}, 300 )
+				}, 300 ),
+				'wc/store/cart'
 			);
 		}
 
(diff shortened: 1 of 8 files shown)


modules/ppcp-button/resources/js/modules/ContextBootstrap/MiniCartBootstrap.js
```diff
@@ -34,6 +38,19 @@ class MiniCartBootstrap {
...
+		The block cart/mini-cart mutates the `wc/store/cart` data store instead of
+		firing the jQuery fragment events above, so subscribe to it to keep the
+		mini-cart button in sync (and clear it when the cart becomes empty).
+		 */
+		if ( typeof wp !== 'undefined' && wp.data?.subscribe ) {
+			wp.data.subscribe(
+				debounce( () => {
+					this._handleStoreCartChange();
+				}, 300 )
+			);
```
Narek13, line 52: This fires on every store change, not just `wc/store/cart`.
`subscribe()` accepts an optional `storeNameOrDescriptor` second argument to scope the listener to one store like `wp.data.subscribe(callback, 'wc/store/cart')`

Narek13 (approved):
Thanks for working on this. Left a small suggestion

PR #4544, opened 2026-07-21
Relocate shared order/AJAX plumbing into a neutral module (6677)

danieldudzic:
**_Note: This PR is stacked on the contract-tests branch `dev/PCP-6675-contract-tests-for-shared-order-ajax-endpoints-v-5-v-6-shared-plumbing`, not `dev/develop`. Merge order is bottom-up: #4526 → #4537 → this PR._**

### Description

Relocates the shared WC-AJAX order endpoints and their supporting helpers out of the v5 frontend modules (`ppcp-button`, `ppcp-blocks`) into a new neutral `ppcp-order-endpoints` module. These endpoints are SDK-version-agnostic server plumbing (they build purchase units from the cart and call the Orders v2 API) shared by both the v5 and v6 SDK frontends, so housing them in a v5 module is a coupling liability as the v6 stories (ACDC, Wallets, Blocks, Save Payment Methods, Subscriptions) all reuse them.

This is a **pure move, no logic changes.** v5-only branches inside the endpoints move along as-is and are pruned only after a future v5 sunset.

- **Moved from `ppcp-button`:** `CreateOrderEndpoint`, `ApproveOrderEndpoint`, `ChangeCartEndpoint` (+ `AbstractCartEndpoint`), `RequestData`, `CartProductsHelper`, `WooCommerceOrderCreator`, `EarlyOrderHandler`. `ApproveSubscriptionEndpoint` and `SimulateCartEndpoint` were evaluated and intentionally left in `ppcp-button`, re-pointed at the new namespace.
- **Moved from `ppcp-blocks`:** `UpdateShippingEndpoint`. Its `wc_ajax_ppc-update-shipping` hook is now registered unconditionally in the new module, so it no longer depends on WC Blocks booting.
- **Backward compatible:** every `wc_ajax_ppc-*` action name and nonce name is byte-identical, and the old container keys (e.g. `button.request-data`, `blocks.endpoint.update-shipping`) resolve via one-line aliases to the canonical `order-endpoints.*` keys, so existing wiring and third-party consumers keep working.
- **Registered unconditionally** in `modules.php` (not behind the `sdk_v6_enabled` flag): v5 depends on it.

**Scope note:** relocation only. Any behavior change discovered to be necessary is out of scope and gets its own ticket.

### Steps to Test

**Setup:** sandbox-connected store, `composer dump-autoload && npm run build`.

1. **v5 storefront (flag off):** complete a purchase from product, cart, classic checkout, and mini-cart. Behavior is identical to before the relocation.
2. **v6 (flag on via `PCP_SDK_V6_ENABLED=1`):** complete a purchase. Behavior is identical to before.
3. **Blocks-unavailable resilience:** with WC Blocks inactive, confirm `ppc-update-shipping` still responds (previously it was gated on `ppcp-blocks` booting).
4. **Backward-compat keys:** boot the container and confirm old keys resolve without fatals, e.g. `PPCP::container()->get('button.request-data')` and `->get('blocks.endpoint.update-shipping')` return the relocated classes.
5. **Suites:** `npm run test` (contract tests from PCP-6675 pass unchanged; full unit suite passes with no behavioral changes) and `npm run lint`.


Narek13 (approved):
Thanks. LGTM

PR #4546, opened 2026-07-21
Apple Pay visible under paypal gateway when vaulting subscription is in cart (6554)

Dinamiko:
Apple Pay subscription renewals failed because the Apple Pay payment method was never vaulted during the initial purchase. The `ppcp_create_order_request_body_data` filter that injects `store_in_vault` only handled the Card and PayPal/Venmo gateways, Apple Pay orders (which arrive as `PayPalGateway::ID` with `funding_source = apple_pay`) fell through and got no vault attributes. So PayPal never returned a token, none was saved, and renewals had nothing to charge, even though the token-save and renewal code already supported Apple Pay.                
                                                                                                                                                                                             
This PR adds an Apple Pay branch to that filter so, when a subscription is being purchased and wallet saving is enabled, the create-order request carries `payment_source.apple_pay.attributes.vault.store_in_vault = ON_SUCCESS` (plus the customer id when known). PayPal then vaults the token, it's saved as a PaymentTokenApplePay, and the existing renewal flow charges it. Also aligned the create-order funding source string to apple_pay. 

As result of the changes in this PR it should be possible to purchase a subscription using Apple Pay from cart and checkout pages including classic and blocks editor. Also it should be possible to renewal the purchased subscriptions.


Narek13 (approved):
LGTM. Thanks

PR #4548, opened 2026-07-21
Scope reCAPTCHA assets and markup to active PayPal payment locations (6703)

Narek13:
### Description                                                                                                                                                                                         
                                                                                                                                                                                                          
When reCAPTCHA protection is enabled, its scripts, configuration, badge, and v2 container currently load on checkout, cart, and single product pages regardless of whether PayPal payments are actually enabled at those locations. For example, with only Classic Checkout enabled, reCAPTCHA still loads on cart and product pages where no PayPal button/gateway can be initiated — causing unnecessary asset loading, an unexpected reCAPTCHA badge, and unnecessary processing of visitor data on unrelated pages.                                                                                                  
                                                                                                                                                                                                          
Root cause: `Recaptcha::enqueue_scripts()` and the `woocommerce_single_product_summary` hook rendering the v2 container only checked the WordPress page type (`is_checkout()`/`is_cart()`/`is_product()`), never the merchant's configured Smart Button Locations — unlike `SmartButton`, which already gates its own rendering on `SettingsStatus::is_smart_button_enabled_for_location()`.                                                                                                                                               
                                                                                                                                                                                                          
### PR Changes                                                                                                                                                                                          
                                                                                                                                                                                                          
- Injected `SettingsStatus` into `Recaptcha` and added a shared `should_enqueue_for_current_location()` check (mirroring `SmartButton`'s existing gating pattern), which detects the current page's location — `product`, `cart`, `checkout`, or `checkout-block-express` (classic vs. block/express checkout are distinguished, since a merchant may enable PayPal for one but not the other) — and confirms it's enabled before allowing reCAPTCHA to load.                                                                                                                                                
- Applied this same gate to both `enqueue_scripts()` (scripts, config, badge) and `render_v2_container()` (the v2 container markup), so all reCAPTCHA output is consistently scoped.                    
- Added a new `woocommerce_paypal_payments_recaptcha_should_enqueue` filter around the final decision, giving developers a supported way to override it.                                                
- Added the first PHPUnit test coverage for `Recaptcha` (none existed before), covering: skipping on unsupported/disabled locations, running on enabled locations, the classic vs. block checkout distinction, and the new filter.                                                                                                                                                                        
                                                                                                                                                                                                          
### Steps to Test                                                                                                                                                                                       
                                                                                                                                                                                                          
1. Connect PayPal and configure valid reCAPTCHA v3 and v2 keys.                                                                                                                                         
2. Go to **WooCommerce → Settings → Payments → PayPal → Styling**.                                                                                                                                      
3. Enable only **Classic Checkout**; disable Cart, Express Checkout, Mini Cart, and Single Product locations.                                                                                           
4. Visit a single product page as a logged-out visitor, then visit the cart page.                                                                                                                       
5. Inspect the page source and browser console.                                                                                                                                                         
                                                                                                                                                                                                          
**Expected**: no `ppcp-recaptcha`/`ppcp-recaptcha-handler` scripts, no reCAPTCHA badge, no `ppcpRecaptchaSettings`, and no `#ppcp-recaptcha-v2-container` on product/cart pages. Visiting Classic Checkout should still load reCAPTCHA normally.


Dinamiko (changes requested):
@Narek13 it seems that if the smart-button location that applies there is disabled, checkout for classic checkout, checkout-block-express for block checkout then `should_enqueue_for_current_location()` returns false, `enqueue_scripts()` bails out, and reCAPTCHA is not loaded for ACDC. 

reCAPTCHA should load wherever any of its protected payment methods can actually be initiated, not only where the smart button is enabled.

Recaptcha already holds `$this->payment_methods`, the exact set it protects (PayPalGateway, CreditCardGateway, CardButtonGateway, AxoGateway). Reuse that against WooCommerce's own availability, which is the same source of truth the validation hooks rely on:                                                                                                              
```php                                                                                                                                                                                             
  private function should_enqueue_for_current_location(): bool {                                                                                                                             
      $location = $this->current_smart_button_location();                                                                                                                                    
                                                                                                                                                                                             
      $enabled_for_location = null !== $location                                                                                                                                             
          && $this->settings_status->is_smart_button_enabled_for_location( $location );                                                                                                      
                                                                                                                                                                                             
      $should_enqueue = $enabled_for_location                                                                                                                                                
          || $this->has_protected_gateway_on_current_page();                                                                                                                                 
                                                                                                                                                                                             
      /** ... existing woocommerce_paypal_payments_recaptcha_should_enqueue filter ... */                                                                                                    
  }     

/**                                                                                                                                                                                        
   * True when a reCAPTCHA-protected gateway (ACDC card fields, AXO, card button,                                                                                                            
   * standard PayPal) is available on a checkout / order-pay / add-payment-method                                                                                                            
   * page — surfaces whose availability is independent of the smart-button location.                                                                                                         
   */                                                                                                                                                                                        
  private function has_protected_gateway_on_current_page(): bool {                                                                                                                           
      if ( ! is_checkout() && ! is_add_payment_method_page() ) {                                                                                                                             
          return false;                                                                                                                                                                      
      }                                                                                                                                                                                      
                                                                                                                                                                                             
      $available = WC()->payment_gateways->get_available_payment_gateways();                                                                                                                 
      foreach ( $this->payment_methods as $id ) {                                                                                                                                            
          if ( isset( $available[ $id ] ) ) {                                                                                                                                                
              return true;                                                                                                                                                                   
          }                                                                                                                                                                                  
      }                                                                                                                                                                                      
      return false;                                                                                                                                                                          
  }          
```   
`is_checkout()` already covers the order-pay page (`is_checkout_pay_page()`), and `is_add_payment_method_page()` covers the add-payment-method surface where can_render_dcc() also renders.

Dinamiko (approved):

PR #4552, opened 2026-07-22
First-party Woo plugin detection for product customizations

strangerkir:
### Description

Provides two services for:
* detecting which plugins of the listed below are active
* finding all products related to one of those plugins

### Steps to Test
1. Install and activate plugins from the list below
2. Import products from the attached file
3. Install and activate the attached plugin
4. Run `wp ppcp-compat-products scan`
5. Make sure that all 8 plugins are reported as active, and 8 products are reported as affected - one per plugin

**Plugins list**
* [Product Bundles](http://woocommerce.com/products/product-bundles/) 
* [Product Add-Ons](http://woocommerce.com/products/product-add-ons/) 
* [Min/Max Quantities](http://woocommerce.com/products/minmax-quantities/) 
* [Composite Products](http://woocommerce.com/products/composite-products/) 
* [Gift Cards](http://woocommerce.com/products/gift-cards/) 
* [Per Product Shipping](http://woocommerce.com/products/per-product-shipping/) 
* [WooCommerce Deposits](http://woocommerce.com/products/woocommerce-deposits/) 
* [WooCommerce Subscriptions](http://woocommerce.com/products/woocommerce-subscriptions/)

**Products to import**
[wc-product-export-22-7-2026-1784711315948.csv](https://github.com/user-attachments/files/30267255/wc-product-export-22-7-2026-1784711315948.csv)

**Plugin for testing**
[ppcp-compat-products.zip](https://github.com/user-attachments/files/30268303/ppcp-compat-products.zip)

### Documentation
So far, only internal documentation is needed. This PR doesn't produce any visible changes; it only registers services which aren't used yet.

### Notes
* Failing CI was a pre-existing issue, so I'm ignoring it here to avoid scope creep


tests/PHPUnit/Compat/PluginDetector/ProductCustomizationDetectorTest.php
```diff
@@ -0,0 +1,255 @@
...
+			'woocommerce-deposits'             => false,
+		);
+	}
+
+	/**
+	 * @test
+	 */
+	public function test_all_false_when_no_plugin_is_active(): void {
+		$this->plugin_detector->shouldReceive( 'scan' )->once()->andReturn( $this->all_inactive() );
+		$this->logger->shouldNotReceive( 'warning' );
```
stracker-phil: suggestion: Remove this

Reason: Log output should not be tested/expected; when we add or change log output, it should not impact the test

Usual mock for the logger:
```php
// Define the logger with shouldIgnoreMissing in the setUp() method, then forget about it:
$this->logger = Mockery::mock(LoggerInterface::class)->shouldIgnoreMissing();
```

modules/ppcp-compat/src/PluginDetector/PluginDetectorInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Compat\PluginDetector;
+
+/**
+ * Interface PluginDetectorInterface
+ */
+interface PluginDetectorInterface {
```
stracker-phil: suggestion: Delete this interface

Reason: I cannot see any hint that there will be more than one implementation of this interface. Instead of adding it right away, we generally try starting with a single class and add an interface once it's really needed

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
```diff
@@ -0,0 +1,104 @@
...
+namespace WooCommerce\PayPalCommerce\Compat\PluginDetector;
+
+/**
+ * Class PluginDetector
+ */
+class PluginDetector implements PluginDetectorInterface {
+
+	/**
+	 * @inheritDoc
+	 */
```
stracker-phil, line 28: nit: Remove the docblock

Reason: it's 100% noise, adding no value. PhpStorm expands the parent docblock automatically.

Of course, if the interface is removed (as I suggested), move the original docblock here 😉

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
```diff
@@ -0,0 +1,104 @@
...
+			'woocommerce-shipping-per-product' => $this->is_woocommerce_shipping_per_product_active(),
+			'woocommerce-deposits'             => $this->is_woocommerce_deposits_active(),
+		);
+	}
+
+	/**
+	 * Checks whether WooCommerce Subscriptions is active.
+	 *
+	 * @return bool
+	 */
```
stracker-phil: nit: Remove all the docblocks in this file.

Reason: the return value is documented in code. And the function names are excellent, they do not need any explanation

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -0,0 +1,223 @@
+<?php
+/**
+ * Detects whether a product has been customized by a first-party
+ * WooCommerce extension (product type change or plugin-specific meta).
+ *
+ * @package WooCommerce\PayPalCommerce\Compat\PluginDetector
+ */
+
+declare(strict_types=1);
```
stracker-phil: nit: apply code style rules to this file

Reason: The project style expects spaces around operators and parentheses

Sample:
```php
declare( strict_types = 1 );
```

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -0,0 +1,223 @@
...
+
+namespace WooCommerce\PayPalCommerce\Compat\PluginDetector;
+
+use Psr\Log\LoggerInterface;
+use WooCommerce\PayPalCommerce\Compat\Exception\RuntimeException;
+
+/**
+ * Class ProductCustomizationDetector
+ */
```
stracker-phil: nit: Remove this docblock

 😉

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -0,0 +1,223 @@
...
+class ProductCustomizationDetector implements ProductCustomizationDetectorInterface {
+
+	/**
+	 * @var PluginDetectorInterface
+	 */
```
stracker-phil: nit: Remove the docblock, and the others that just re-state what's visible in the code

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -0,0 +1,223 @@
...
+	 * @var LoggerInterface
+	 */
+	private LoggerInterface $logger;
+
+	/**
+	 * Cached result of $plugin_detector->scan(), since plugin activation
+	 * cannot change within a request but scan() may be called once per product.
+	 *
+	 * @var array<string, bool>|null
+	 */
```
stracker-phil, line 29: note: This `@var` annotation is good, it this docblock should stay

(just want to point out that this is different/useful in contrast to the comments above)

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
```diff
@@ -0,0 +1,104 @@
...
+	public function scan(): array {
+		return array(
+			'woocommerce-subscriptions'        => $this->is_woocommerce_subscriptions_active(),
+			'woocommerce-gift-cards'           => $this->is_woocommerce_gift_cards_active(),
+			'woocommerce-product-bundles'      => $this->is_woocommerce_product_bundles_active(),
+			'woocommerce-product-addons'       => $this->is_woocommerce_product_addons_active(),
+			'woocommerce-min-max-quantities'   => $this->is_woocommerce_min_max_quantities_active(),
+			'woocommerce-composite-products'   => $this->is_woocommerce_composite_products_active(),
+			'woocommerce-shipping-per-product' => $this->is_woocommerce_shipping_per_product_active(),
+			'woocommerce-deposits'             => $this->is_woocommerce_deposits_active(),
```
stracker-phil: idea: turn these strings into `public const`s

Reasons:
- Allows auto-completion in the IDE
- Avoids typos
- Gives us a single place with the full detection-catalog

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -0,0 +1,223 @@
...
+	 *
+	 * @param \WC_Product $product The product to check.
+	 * @return bool
+	 */
+	private function is_customized_by_product_addons( \WC_Product $product ): bool {
+		$this->assert_method_exists( \WC_Product_Addons_Helper::class, 'get_product_addons' );
+
+		$addons = \WC_Product_Addons_Helper::get_product_addons( $product->get_id(), false, false, false );
+
+		return array() !== $addons;
```
stracker-phil: nit: Rewrite as `count( $addons ) > 0` or `! empty( $addons )`

Reason: makes the decision a bit easier to understand

tests/PHPUnit/Compat/PluginDetector/ProductCustomizationDetectorTest.php
```diff
@@ -0,0 +1,255 @@
...
+		$this->plugin_detector->shouldReceive( 'scan' )->once()->andReturn( $active_plugins );
+		$this->logger->shouldReceive( 'warning' )
+			->once()
+			->with(
+				Mockery::on(
+					static function ( string $message ) use ( $plugin ): bool {
+						return false !== strpos( $message, $plugin ) && false !== strpos( $message, 'does not exist' );
+					}
+				)
+			);
```
stracker-phil: suggestion: Remove this logger expectation

Reason: Same as above, It's not helping with plugin stability and does not prevent a bug. If devs need additional or different logs, it should not require unit test changes.

The `->assertFalse()` check below is exactly what this test should verify

modules/ppcp-compat/src/Exception/RuntimeException.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Compat\Exception;
+
+/**
+ * Class RuntimeException
+ */
+class RuntimeException extends \RuntimeException {
```
stracker-phil: change: Find a unique name for this

Reason: This class shadows the native `RuntimeException` which can (and likely will) cause confusion. In the worst case, someone might think "we do not need this class" and delete it, breaking the `ProductCustomizationDetector` try-catch

(Claude suggests `PluginApiChangedException` or `DetectionMethodNotFoundException`)

strangerkir: Good point, renamed.

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
```diff
@@ -37,16 +37,18 @@ class ProductCustomizationDetector implements ProductCustomizationDetectorInterf
 	private ?array $active_plugins = null;
 
 	/**
-	 * @param PluginDetectorInterface $plugin_detector The plugin presence detector.
-	 * @param LoggerInterface         $logger The logger.
+	 * @param PluginDetector  $plugin_detector The plugin presence detector.
+	 * @param LoggerInterface $logger The logger.
```
Narek13: I would also suggest to remove these comments as they do not convey any info

tests/PHPUnit/Compat/PluginDetector/ProductCustomizationDetectorTest.php
strangerkir: I'd still keep it tested from a practical point of view. If a 3rd-party API changes and the function no longer exists, this warning is the only way to know that the check itself was a false negative. So I'd say this is an important part of the logic, not just regular logging.

But I agree that testing for a particular string in logs may be too much, so I'm leaving just a warning assertion, without any message checks.

UPD: removed here, this one is useless, as we never examine real 3rd-party functions here. My comment above related to the next warning assertion.

modules/ppcp-compat/src/PluginDetector/PluginDetectorInterface.php
strangerkir: Removed :+1:

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
strangerkir: Done. The interface was removed, and I only left the file header here, removing the class docblock. It looks like in this project we often prefer file headers, so I followed the same pattern.

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
strangerkir: Agree, this is nothing more than noise. Removed now.

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
strangerkir: Fixed, thanks

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
strangerkir: Done

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
strangerkir: Removed completely :)

modules/ppcp-compat/src/PluginDetector/PluginDetector.php
strangerkir: Good point, thanks! Done.

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
strangerkir: Done

tests/PHPUnit/Compat/PluginDetector/ProductCustomizationDetectorTest.php
strangerkir: I explained above why I'd rather keep this. But I'm open to a discussion if you really believe it shouldn't be here.

modules/ppcp-compat/src/PluginDetector/ProductCustomizationDetector.php
strangerkir: Agree, they are useless. Removed.

stracker-phil (approved):
Nice work :)

A small concern, but not sure if it's serious or hypothetical:
In `ProductCustomizationDetector` line 71 log output (`$this->logger->warning(...)`) might be called very often this scan happens per-product on the front-end; a failure thrown by `assert_method_exists()` could produce a lot of log noise

We should keep an eye on this and maybe add a transient/meta-key to cache the detection result if needed

PR #4554, opened 2026-07-24
Insonsistent PayPal button behavior for subscriptions on cart (6639 v2)

Dinamiko:
On block cart/checkout, a free-trial sub + a paying sub hid the PayPal button and payment method (classic worked). Cause: 2+ subscriptions make WooCommerce require the `multiple_subscriptions` feature, but the block only declared `['products','subscriptions']`, so Blocks filtered PayPal out.                                                                   
                                                                                                                                                                                             
This PR passes the gateway's real supported features from PHP (PayPalPaymentMethod.php → supportedFeatures) and have `checkout-block.js` declare those instead of a hardcoded list, so `multiple_subscriptions` is included.


stracker-phil (approved):
Addresses two issues:

- Fix the PayPal button appearance depending on subscription support
- Keep the mini cart in sync with block data via a new Redux subscriber

Both changes look solid  👍

PR #4555, opened 2026-07-24
Unsuccessfully sandbox onboarding via wizard (6714)

Dinamiko:
Enabling Sandbox Mode and clicking "Connect Account" opened PayPal onboarding in a new tab instead of minibrowser, and never connected. PayPal's `partner.js` uses a single global, takes its environment from its own script domain, and only wires the first `[data-paypal-button]`. The plugin always loaded the production SDK with both buttons present (live first), so only the live button got bound; the sandbox button stayed a plain link.                                                                                                                             
                                                                                                                                                                                             
### PR changes  
  - `useHandleConnections.js`: load the environment-specific partner.js (production/sandbox) only for the button matching the Sandbox toggle, resetting `window.PAYPAL` on switch.               
  - `ConnectionButton.js`: only the active-environment button is a real `[data-paypal-button]` (so PayPal binds it); added a click-guard that blocks clicks until the button is actually bound.

### DIFF
--- a/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/ConnectionButton.js
+++ b/modules/ppcp-settings/resources/js/Components/Screens/Onboarding/Components/ConnectionButton.js
@@ -75,6 +75,7 @@ const ConnectionButton = ( {
 	const {
 		onboardingUrl,
 		scriptLoaded,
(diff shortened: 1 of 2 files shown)


stracker-phil (approved):
LGTM, and local test also shows it solves the problem

PR #4556, opened 2026-07-24
Stop endless store-sync onboarding retries (6719)

stracker-phil:
*Changelog:* `Fix - Store sync no longer retries onboarding in a loop when PayPal rejects the request #4556`

# Description

## 🐛 The problem

When PayPal's onboarding API rejects a store-sync registration (for example after a backend schema change), the plugin retries the register/deregister call on every request, without ever enabling store-sync.

## 💡 Why it happens

Registration was driven off the `init` hook with no backoff. A failed attempt left the desired and actual state mismatched, so the next request tried again, and a failing deregister returned an error that kept the store in a half-registered state. Nothing broke the loop.

## 🔍 What to review

- **New public hook.** `woocommerce_paypal_payments_settings_saved` is exposed on `AbstractDataModel`. Listeners must type-check the passed model, since it fires for every settings model.
- **No automatic retry.** A failed registration disables the toggle; re-enabling it is the merchant's explicit retry. There is no backoff or scheduled retry by design.

### DIFF
--- a/tests/PHPUnit/PpcpSettings/Data/AbstractDataModelTest.php
+++ b/tests/PHPUnit/PpcpSettings/Data/AbstractDataModelTest.php
@@ -1,0 +1,69 @@
+<?php
+declare(strict_types=1);
+
+namespace PHPUnit\PpcpSettings\Data;
+
+use WooCommerce\PayPalCommerce\Settings\Data\AbstractDataModel;
+use WooCommerce\PayPalCommerce\TestCase;
+use function Brain\Monkey\Actions\expectDone;
+use function Brain\Monkey\Functions\when;
+
+/**
+ * @covers \WooCommerce\PayPalCommerce\Settings\Data\AbstractDataModel
+ */
+class AbstractDataModelTest extends TestCase
+{
+	public function setUp(): void
+	{
+		parent::setUp();
+		when('get_option')->justReturn([]);
+	}
+
+	private function create_testee(): TestableDataModel
+	{
+		return new TestableDataModel();
+	}
+
+	/**
+	 * GIVEN a listener on the "settings saved" action that calls save() again
+	 * WHEN the initial save() dispatches the action
+	 * THEN the nested save() still persists the data (a second update_option call)
+	 * AND the action is dispatched only once overall, preventing a notification loop
+	 */
+	public function testReentrantSaveStillPersistsButFiresActionOnlyOnce(): void
+	{
+		$update_calls = [];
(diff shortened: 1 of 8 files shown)


tests/PHPUnit/PpcpSettings/Data/AbstractDataModelTest.php
```diff
@@ -0,0 +1,69 @@
+<?php
+declare(strict_types=1);
+
+namespace PHPUnit\PpcpSettings\Data;
```
Narek13, line 4: Should this be `WooCommerce\PayPalCommerce\PpcpSettings\Data`?
```suggestion
namespace WooCommerce\PayPalCommerce\PpcpSettings\Data
```

tests/PHPUnit/StoreSync/Registration/ReconciliationServiceTest.php
```diff
@@ -0,0 +1,158 @@
...
+	public function test_reconcile_registers_store_when_desired_but_not_registered(): void {
+		$this->settings->shouldReceive( 'should_initialize_features' )->andReturn( true );
+		$this->registration->shouldReceive( 'is_registered' )->andReturn( false );
+
+		$this->registration->shouldReceive( 'register' )
+			->once()
+			->andReturn( new RegistrationResult( true, 'Registered', null ) );
+
+		$this->settings->shouldNotReceive( 'set_active' );
+		$this->settings->shouldNotReceive( 'set_last_registration_error' );
```
Narek13: maybe is this a leftover? I can't find this anywhere in the code

Narek13 (approved):
Thank you @stracker-phil for working on this. I left 2 doubts, but going to approve as these are not blockers

tests/PHPUnit/PpcpSettings/Data/AbstractDataModelTest.php
stracker-phil: Good find! I copied the namespace from other `PpcpSettings` tests, which use the same format, but other modules use your suggested `WooCommerce\PayPalCommerce\` prefix.

All 3 test files are working currently, so I'll leave the namespace as-is for now, leaving this for an update after the release (we should also rename the folder from `PpcpSettings` to `Settings` as other test suites do not add the Ppcp prefix either)

tests/PHPUnit/StoreSync/Registration/ReconciliationServiceTest.php
stracker-phil: You're right, I missed it. Removed that assertion!

PR #4560, opened 2026-07-27
Advanced Card Fields for classic checkout (5781)

Narek13:
### Description                                                                                                                                                                                         
                                                                                                                                                                                                          
Migrates the Advanced Card Fields (ACDC) checkout experience to the v6 Web SDK, for **classic (shortcode) checkout only** — block checkout is tracked separately, since it depends on the  WooCommerce Blocks v6 migration  landing first.                                                                                                                                               
                                                                                                                                                                                                          
The v6 card fields mount into the same WC card-form input slots v5's `paypal.CardFields()` used, and reuse the existing v5 order create/approve AJAX endpoints (new `createCardOrder()`/`approveCardOrder()` adapters, rather than new endpoints) with a card-specific payment method and funding source. 3D Secure is handled automatically by the v6 SDK inside `cardSession.submit()` — no separate contingency setup needed, unlike v5.                                                                                                                               
                                                                                                                                                                                                          
**Notable implementation details:**                                                                                                                                                                         
- `should_load_on_current_page()` now loads the v6 module on checkout whenever ACDC is enabled, independent of the smart-button location setting (the card gateway is a regular payment method, not tied to that setting).                                                                                                                                                                                       
- `createCardOrder()` deliberately never sends `save_order_in_session` — at create-order time no card is attached yet, so storing it in session that early would let an unconfirmed order slip through. `approveCardOrder()` is what stores it, once the disabled-card-brand and 3DS checks pass server-side.                                                                                                   
- The v6 SDK expects a different style schema than v5 (camelCase JS properties vs. v5's kebab-case CSS, and a different supported-property set), so this adds a dedicated `cardFieldStyles.js` rather than reusing v5's helper.                                                                                                                                                                               
- Each mounted field's box is sized explicitly from the original WC input's rendered dimensions — v6's `createCardFieldsComponent()` only returns a plain element for us to lay out ourselves, unlike v5's `.render()`, which owns its own container.                                                                                                                                                         
- The click interceptor and mounted fields re-attach on WooCommerce's `update_checkout` event, since that AJAX call replaces the entire `#payment` box (including `#place_order` and the card inputs) on every billing/shipping field change.                                                                                                                                                                    
                                                                                                                                                                                                          
**Out of scope here:** saving a new card, paying with an already-saved card, and the $0 free-trial variant.                                                                                                 
                                                                                                                                                                                                          
### Steps to Test                                                                                                                                                                                       
                                                                                                                                                                                                          
1. Ensure the `PCP_SDK_V6_ENABLED` feature flag is on and the ACDC ("Debit & Credit Cards") gateway is enabled.                                                                                         
2. Add a simple (non-subscription) product to the cart.                                                                                                                                                 
3. Go to the classic (shortcode) checkout page.                                                                                                                                                         
4. Select **Debit & Credit Cards** as the payment method and confirm the card number/expiry/CVC (and cardholder name, if enabled) render as PayPal-hosted fields in place of the native WC inputs.      
5. Enter an invalid card number and confirm the field shows a red/invalid border without submitting.                                                                                                    
6. Change a billing/shipping field (e.g. postcode) *after* selecting the card gateway to trigger WooCommerce's checkout refresh, then confirm the card fields still work.                               
7. Enter a valid PayPal sandbox test card (e.g. `4111111111111111`, any future expiry, any CVC) and click Place Order — confirm the order completes and is captured in PayPal.                          
8. Trigger a 3D Secure challenge with a sandbox card configured for it, and confirm both completing and canceling the challenge behave correctly.                                                       
9. Select a different payment method (e.g. the PayPal wallet button) and confirm no interference from the card-fields code.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/boot.js
+++ b/modules/ppcp-sdk-v6/resources/js/boot.js
@@ -17,6 +17,7 @@ import { checkEligibility } from './eligibility';
 import { createSession } from './sessions/createSession';
 import { renderButtons } from './components/buttonRenderer';
 import { createOrder, fetchCartTotal } from './endpointsAdapter';
+import { initCardFields } from './cardFields/renderer';
 import { hasJQuery } from './utils/api';
 import { setErrorLabels } from './utils/errorHandler';
 
@@ -28,6 +29,20 @@ import { setErrorLabels } from './utils/errorHandler';
 	}
 
 	setErrorLabels( config.labels );
+
+	/**
+	 * Advanced Card Fields (ACDC): independent of the button render loop
+	 * below — it mounts into the existing WC card-form inputs rather than
+	 * a button wrapper, and only actually does anything when the card
+	 * gateway is enabled on this (checkout) page. Deferred the same way
+	 * as renderAll(), since it also queries checkout-form DOM elements.
+	 */
+	function initCardFieldsSafely() {
+		initCardFields( config ).catch( ( error ) => {
+			// eslint-disable-next-line no-console
+			console.error( '[PPCP SDK v6]', error );
+		} );
+	}
 
 	// The page-context and mini-cart wrappers are independent render
 	// targets; PHP only prints wrappers for enabled locations, so target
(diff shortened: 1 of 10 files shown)


danieldudzic (approved):
Looks good. Thank you for working on this! 

One bug: `attach()` nulls `cardSessionPromise` before its early return, so a checkout refresh that doesn't replace the payment box (WC only swaps changed fragments) drops the session without remounting - the next Place Order then submits a fieldless session (`SdkInitError: PostMessenger` does not exist at renderer.js:184).

I'll push the fix plus a regression test and merge.

PR #4565, opened 2026-07-29
Apply the buyer's state in Google Pay and Apple Pay shipping callbacks (6725)

stracker-phil:
*Changelog:* `Fix - Google Pay and Apple Pay now return shipping options for addresses in state-based shipping zones #4565`

# Description

## 🐛 The problem

On stores whose shipping rates depend on the customer's state or region, selecting a shipping address inside the Google Pay or Apple Pay sheet returns no shipping options. Google Pay shows an empty list; Apple Pay rejects the address with **"No shipping methods are available for your address."**

## 💡 Why it happens

Both wallet shipping callbacks discard the state (`administrativeArea`) from the wallet payload before WooCommerce calculates shipping. Google Pay wrote an empty billing and shipping state; Apple Pay never parsed `administrativeArea` into its address and then forced an empty destination state. With no state, a state-scoped shipping zone cannot match, so the rate list comes back empty.

## ✅ The fix

Both callbacks now carry the buyer's state through to WooCommerce's shipping calculation.

- **Google Pay:** `UpdatePaymentDataEndpoint::update_addresses()` reads `administrativeArea` for the customer state.
- **Apple Pay:** `ApplePayDataObjectHttp::simplified_address()` parses the state, and `ApplePayButton::customer_address()` and `getShippingPackages()` apply it to the customer and the shipping package.

## 🔍 What to review

The state from the wallet (`administrativeArea`) is trusted as-is as a WooCommerce state code; a country that returns a region name rather than a state code could still fail to match a zone.

## 🧪 How to verify

1. Create a shipping zone scoped to a single state (e.g. US → New York) with a rate, and no catch-all zone that would also match.
2. Add a shippable product and open the Google Pay sheet (and the Apple Pay sheet in Safari).
3. Select an address in that state, and confirm the state's shipping options appear.

### DIFF
--- a/modules/ppcp-applepay/src/Assets/ApplePayButton.php
+++ b/modules/ppcp-applepay/src/Assets/ApplePayButton.php
@@ -466,6 +466,9 @@ class ApplePayButton implements ButtonInterface {
 		WC()->customer->set_shipping_country(
 			$address['country'] ?? $shop_country_code
 		);
+		WC()->customer->set_shipping_state(
+			$address['state'] ?? ''
+		);
 		WC()->customer->set_billing_country(
 			$address['country'] ?? $shop_country_code
 		);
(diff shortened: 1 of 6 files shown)


Narek13 (approved):
LGTM. Thanks

PR #4567, opened 2026-07-30
Fix plugin calling Vault API with empty customer id (6731)

Narek13:
### Description                                                                                                                                                                                         
                                                                                                                                                                                                          
`WooCommercePaymentTokens::customer_tokens()` looks up the PayPal customer id from usermeta (`_ppcp_target_customer_id`, falling back to the legacy `ppcp_customer_id`) before fetching that customer's saved PayPal payment tokens. It never checked whether either lookup actually found a value before calling `payment_tokens_for_customer()`, so a user missing both meta keys (e.g. the usermeta row was deleted) triggered:                                                                                                                                                                                     

```                                                                                                                                                                                                          
GET https://api-m.paypal.com/v3/vault/payment-tokens?customer_id=                                                                                                                                       
Response: 400 Bad Request                                                                                                                                                                               
```
                                                                                                                                                                                                     
This adds the same early-return guard already used at the endpoint's other two call sites (`PayPalGateway::process_payment()` and `PaymentMethodTokensChecker::has_paypal_payment_token()`) — if no customer id is found, return no tokens without making the request.                                                                                                                                      
                                                                                                                                                                                                          
This doesn't change any user-visible outcome: `customer_tokens()` already caught the resulting exception and returned an empty array, so callers (subscription renewal, the free-trial checkout flow) already saw "no saved tokens" in this situation. The fix only removes the wasted, always-failing API call and its error-log noise.                                                                      
                                                                                                                                                                                                          
### Steps to Test                                                                                                                                                                                       
                                                                                                                                                                                                          
1. Set up Vaulting and a merchant connection.                                                                                                                                                           
2. Buy a subscription (this saves a payment token, storing the customer id in usermeta).                                                                                                                
3. Delete the customer id from the usermeta table (both `_ppcp_target_customer_id` and `ppcp_customer_id`, if present) for that user.                                                                   
4. Attempt to renew the subscription (or trigger `customer_tokens()` via the free-trial checkout flow).                                                                                                 
5. Confirm no `v3/vault/payment-tokens?customer_id=` request appears in the logs, and the renewal fails the same way it did before (just without the wasted API call).

### DIFF
--- a/modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
+++ b/modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
@@ -257,6 +257,10 @@ class WooCommercePaymentTokens {
 			$customer_id = get_user_meta( $user_id, 'ppcp_customer_id', true );
 		}
 
+		if ( ! $customer_id ) {
+			return array();
+		}
+
 		try {
 			$customer_tokens = $this->payment_tokens_endpoint->payment_tokens_for_customer( $customer_id );
 		} catch ( RuntimeException $exception ) {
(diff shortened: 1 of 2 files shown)


stracker-phil (approved):
LGTM

PR #4573, opened 2026-08-03
Improve 3DS rejection error with clear, customer-friendly decline messages

Narek13:
### Description

When a shop requires 3D Secure and a customer pays with a card whose 3DS liability does not shift, the plugin rejects the card **locally, before capture**  but the checkout showed **"Could not capture the PayPal order."** That reads as a technical failure, even though nothing was ever sent to PayPal to capture.

This replaces that message with clear, customer-friendly declines, chosen by the actual rejection reason:

- **`CardCaptureValidator::rejection_reason()`** (new) classifies the rejection:
  - `malformed` — order carries no `payment_source`
  - `3ds_failed` — `liability_shift = NO` (authentication attempted, not passed)
  - `3ds_unclear` — `liability_shift = UNKNOWN` or absent
  - `none` — capturable
  `is_valid()` now delegates to it (`rejection_reason() === REASON_NONE`), so its existing accept/reject behaviour is unchanged.
- The capture gate maps each reason to a customer-facing message and logs the reason:
  - `3ds_failed` → "This card could not be authorized. Please try a different payment method."
  - `3ds_unclear` → "The 3D Secure validation could not be completed. Please try again or use another card."
  - `malformed` / default → "This payment could not be processed. Please try a different payment method."

The accept/reject **decision** is intentionally unchanged — only the message the shopper sees (and a more descriptive log line).

### Steps to Test

> Requires Advanced Card Processing (card fields) enabled and **3D Secure set to "Require 3D Secure"** (`SCA_ALWAYS`). The gate lives in the card-fields capture path, so PayPal buttons alone won't trigger it.

1. Pay with the hosted card fields using a valid sandbox card that reaches the capture step (one that passes card-fields validation; complete the 3DS challenge if prompted).
2. Depending on the returned `liability_shift`, confirm the shopper sees the matching friendly decline (above) instead of "Could not capture the PayPal order."
3. Confirm the store log records the reason, e.g. `Could not capture order <id>: 3ds_failed`.
4. A capturable card (`liability_shift = POSSIBLE`/`YES`) still completes normally — no regression.

> Note: PayPal's sandbox does not reliably return a specific `liability_shift` on demand. All three messages were verified end-to-end by temporarily forcing each reason at the capture gate; the reason→message mapping is also covered by unit tests (`CardCaptureValidatorTest`).

### Changelog

Fix - Replace the generic "Could not capture the PayPal order" error with a clear, customer-friendly decline when a card is rejected locally on a 3DS-required shop.

### DIFF
--- a/modules/ppcp-card-fields/src/CardFieldsModule.php
+++ b/modules/ppcp-card-fields/src/CardFieldsModule.php
@@ -181,11 +181,12 @@ class CardFieldsModule implements ServiceModule, ExecutableModule {
 				$validator = $c->get( 'card-fields.service.card-capture-validator' );
 				assert( $validator instanceof CardCaptureValidator );
 
-				if ( ! $validator->is_valid( $order ) ) {
+				$rejection_reason = $validator->rejection_reason( $order );
+				if ( $rejection_reason !== CardCaptureValidator::REASON_NONE ) {
 					$logger = $c->get( 'woocommerce.logger.woocommerce' );
 					assert( $logger instanceof LoggerInterface );
 
-					$logger->warning( "Could not capture order {$order->id()}" );
+					$logger->warning( "Could not capture order {$order->id()}: {$rejection_reason}" );
 
 					// Only set session flag if WC session exists (not in API/agentic context).
 					/**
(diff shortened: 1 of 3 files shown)


stracker-phil (approved):
Beautiful, thanks!

PR #4574, opened 2026-08-03
Pin the built plugin folder name in the distribution workflow (6749)

stracker-phil:
*Changelog:* (none - build tooling)

# Description

The Build and Distribute workflow derived the plugin folder name from the GitHub repository name. When the workflow ran from a repository whose name did not match `woocommerce-paypal-payments`, the built package used the wrong folder name.

This passes an explicit `PACKAGE_NAME` to the workflow so the distributed plugin always uses the correct, stable folder name regardless of which repository the build runs in.


Narek13 (approved):
Thanks

InpsydeNiklas (approved):

PR #4575, opened 2026-08-03
Fix PayPal button hidden for manual-renewal-only subscriptions

mmaymo:
## 🐛 What and why

When WooCommerce Subscriptions has **Accept Manual Renewals** enabled and vaulting (Save PayPal and Venmo) is disabled, PayPal vanished from checkout for subscription carts.

Three separate resolvers each decided the subscription routing, and all three fell back to `subscriptions_api` whenever `save_paypal_and_venmo()` was false. That imposed PayPal Subscriptions semantics — which require a linked PayPal `Plan` — on a cart that only ever needed a single one-time charge.

The visible result: Block Checkout hid the PayPal gateway entirely and Classic Checkout rendered the button disabled. A merchant accepting manual renewals could not take PayPal at all, with no setting that explained why.

## 📂 What changed

- `modules/ppcp-button/services.php` — the `button.subscriptions-mode` resolver now returns the disabled mode for manual-renewal-only carts. Also swaps three magic strings for the `SubscriptionHelper` constants.
- `modules/ppcp-wc-subscriptions/src/WcSubscriptionsModule.php` — same branch in `get_subscriptions_mode()`, the resolver behind the `woocommerce_paypal_payments_paypal_gateway_supports` filter.
- `modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php` — same branch in `get_subscriptions_mode()`; restructured to early-return instead of precomputing the mode value.
- `modules/ppcp-wc-subscriptions/src/Helper/SubscriptionHelper.php` — `paypal_subscription_button_allowed()` returns `true` when manual renewals are accepted, before falling back to `$can_save_vault_token`. This is the change that unhides the button on both checkouts.
- `tests/PHPUnit/WcSubscriptions/WcSubscriptionsModuleTest.php` — two resolver tests: one for the new branch, one guarding the auto-renewal path.

## ⚙️ How it works now

Each of the three resolvers checks `SubscriptionHelper::accept_manual_renewals()` before the vaulting fallback and returns `SubscriptionHelper::SUBSCRIPTION_MODE_VALUE_DISABLED`. Existing consumers — `SmartButton::has_subscriptions()`, the gateway-supports filter, and the `Plan` gating in `PayPalSubscriptionsModule` — then treat the cart as a plain **`Order`**.

The checkout gate is server-driven, so no JS changed. `SubscriptionHelper::paypal_subscription_button_allowed()` feeds `SmartButton::script_data()['subscription_button_allowed']`, and the Block Checkout gate `paypalSubscriptionButtonAllowed()` already prefers that authoritative flag over its client-side fallbacks.

### Covered by unit tests

- ✅ With manual renewals accepted and vaulting disabled, `WcSubscriptionsModule::get_subscriptions_mode()` resolves to `disable_paypal_subscriptions`.
- ✅ With auto-renewal required and vaulting disabled, the same resolver still returns `subscriptions_api`.

### Not covered by unit tests

- ❌ **Block Checkout renders the PayPal option for a manual-renewal-only cart and pays through the Orders API** — the spec assigned this to JS tests for a `requiresVaultingOrSubscriptionPlan()` helper that does not exist in the codebase. The real gate is `paypalSubscriptionButtonAllowed()`, whose tests predate this branch and cover no manual-renewal case.
- ❌ `SubscriptionHelper::paypal_subscription_button_allowed()` — the change that actually unhides the button has no test at any level.
- ❌ `button.subscriptions-mode` and `PayPalSubscriptionsModule::get_subscriptions_mode()` — the identical branch in the other two resolvers is untested; only the `WcSubscriptionsModule` copy is covered.
- ❌ Classic Checkout button enabled and payment completing as a standard order — requires the full checkout flow. `e2e:test:shard:subscription` covers this ground.
- ❌ Automatic fallback when the merchant is not vaulting-eligible — lives in `SmartButton::can_save_vault_token()`, which this change must not touch.

## 🚨 What must not regress

- `SmartButton::can_save_vault_token()` — its eligibility logic is untouched; the new branch reaches it only through the existing fallback.
- `WcSubscriptionsModule::get_subscriptions_mode()` for auto-renewal carts — a subscription requiring automatic renewal must still resolve to `subscriptions_api` or `vaulting_api`, keeping PayPal hidden when neither vaulting nor a linked `Plan` is configured.


tests/PHPUnit/WcSubscriptions/WcSubscriptionsModuleTest.php
```diff
@@ -0,0 +1,103 @@
...
+	 * proving the manual-renewal fix does not regress the auto-renewal-required path.
+	 */
+	public function test_returns_subscriptions_api_when_auto_renewal_required_and_vaulting_disabled(): void {
+		expectApplied( 'woocommerce_paypal_payments_subscription_mode_disabled' )
+			->once()
+			->with( false )
+			->andReturn( false );
+
+		$settings_provider   = $this->create_settings_provider( false );
+		$subscription_helper = $this->create_subscription_helper( false );
```
stracker-phil, line 97: suggestion: this is the only difference between both tests

I would consolidate to a single test and use `@dataProvider` with 2 test cases instead of repeating the full test

tests/PHPUnit/WcSubscriptions/WcSubscriptionsModuleTest.php
```diff
@@ -0,0 +1,103 @@
...
+	private WcSubscriptionsModule $testee;
+
+	public function setUp(): void {
+		parent::setUp();
+
+		$this->testee = new WcSubscriptionsModule();
+	}
+
+	/**
+	 * @param bool   $accept_manual_renewals Value returned by SubscriptionHelper::accept_manual_renewals().
```
stracker-phil, line 28: nit: this `@param` is not present anymore, should be removed :)

modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
```diff
@@ -826,11 +826,18 @@ private function get_subscriptions_mode( ContainerInterface $container ): string
...
+			return SubscriptionHelper::SUBSCRIPTION_MODE_VALUE_DISABLED;
+		}
+
+		if ( $subscription_helper->accept_manual_renewals() && ! $settings_provider->save_paypal_and_venmo() ) {
+			return SubscriptionHelper::SUBSCRIPTION_MODE_VALUE_DISABLED;
+		}
+
+		return $settings_provider->save_paypal_and_venmo()
+			? SubscriptionHelper::SUBSCRIPTION_MODE_VALUE_VAULTING
+			: SubscriptionHelper::SUBSCRIPTION_MODE_VALUE_SUBSCRIPTIONS;
```
stracker-phil, line 841: suggestion: This exact block repeats in 3 modules, it could be extracted to a central helper like `SubscriptionHelper::resolve_subscription_mode()` or similar

Reason: A single deciding function is easier to test and avoid drift when we need to change it in the future.
Also, current duplicated code is not 100% identical (white space diffs) which makes it a bit harder to identify by find-in-files

stracker-phil (approved):
Approved: Fixes the issue, found no blocker

Still, left some suggestions to clean up the PR before merging

PR #4576, opened 2026-08-03
Fix PPEC billing agreement conversion using a real PayPal customer ID

Narek13:
### Description

When a subscription created with the legacy PayPal Express Checkout (PPEC) plugin renews through WooCommerce PayPal Payments, the compat layer converts the legacy Billing Agreement into a Vault v3 payment token. `BillingAgreementTokenConverter::convert()` passed `CustomerRepository::customer_id_for_user()` to the Vault API but for a user who has never vaulted with PayPal Payments, that method returns a **fabricated local ID** (`prefix . user_id`, `CustomerRepository.php:60`). PayPal doesn't recognise that ID, so the Vault call fails and the billing-agreement → token conversion never succeeds for those users.

**Fix**
- New `CustomerRepository::paypal_customer_id_for_user()` returns the real `ppcp_customer_id` or an empty string, and **never** fabricates a fallback.
- `convert()` now uses it: with a real ID PayPal reuses the customer; with an empty ID `create_payment_token()` omits `customer.id`, so PayPal assigns one and its ID is persisted on the user (existing behaviour). The wasteful, always-failing call with the bogus ID is gone.

`customer_id_for_user()` is left untouched. Its fallback is intentional for order creation elsewhere.

**Note on scope / severity:** `SubscriptionsHandler::use_billing_agreement_as_token()` already falls back to the legacy `BILLING_AGREEMENT` token path when conversion returns null, so on accounts where legacy BA charges still work the renewal wasn't hard-broken but the Vault v3 migration silently never happened (and logged an error) for every non-vaulted user. This makes the migration actually succeed (the ticket's preferred outcome), with the legacy fallback still in place as a safety net.

### Steps to Test

> Full end-to-end repro needs legacy PPEC subscription data + a Billing Agreement + a triggered renewal, which is a heavy setup; the logic is covered by unit tests.

1. Create a subscription via the legacy PPEC plugin; ensure the customer has **no** `ppcp_customer_id` user meta (never vaulted with PayPal Payments).
2. Trigger a renewal order for that subscription.
3. **Before the fix:** the Vault v3 conversion fails (PayPal rejects the local `prefix.user_id` customer ID); the log shows the failed conversion.
4. **After the fix:** conversion succeeds — PayPal assigns a customer, the Billing Agreement is stored as a Vault v3 token (`_ppec_ba_converted_to_vault_v3` on the subscription), and the returned customer ID is saved to `_ppcp_target_customer_id`.
5. A user who **does** have a real `ppcp_customer_id` still converts against that customer — no regression.

### Changelog

Fix - PPEC subscription renewals: convert legacy Billing Agreements to Vault v3 tokens using a real PayPal customer ID instead of a locally generated one.

### DIFF
--- a/modules/ppcp-api-client/src/Repository/CustomerRepository.php
+++ b/modules/ppcp-api-client/src/Repository/CustomerRepository.php
@@ -67,10 +67,21 @@ class CustomerRepository {
 	 * Unlike customer_id_for_user(), this never fabricates a local prefix-based
 	 * ID: PayPal does not recognize such IDs, so callers that pass the result to
 	 * the PayPal API should treat an empty string as "let PayPal assign one".
+	 *
+	 * `_ppcp_target_customer_id` holds the most recent ID PayPal assigned and so
+	 * takes precedence over `ppcp_customer_id`, matching how the rest of the
+	 * plugin resolves it. Reading only the latter would miss an ID PayPal has
+	 * already issued and have it create a second customer for the same user,
+	 * orphaning the tokens stored against the first.
 	 */
 	public function paypal_customer_id_for_user( int $user_id ): string {
 		if ( 0 === $user_id ) {
 			return '';
+		}
+
+		$customer_id = get_user_meta( $user_id, '_ppcp_target_customer_id', true );
+		if ( is_string( $customer_id ) && $customer_id ) {
+			return $customer_id;
 		}
 
 		$customer_id = get_user_meta( $user_id, 'ppcp_customer_id', true );
--- a/modules/ppcp-api-client/src/Repository/CustomerRepository.php
+++ b/modules/ppcp-api-client/src/Repository/CustomerRepository.php
@@ -59,4 +59,22 @@ class CustomerRepository {
 
 		return get_user_meta( $user_id, 'ppcp_customer_id', true ) ?: $this->prefix . (string) $user_id;
 	}
+
+	/**
+	 * Returns the real, PayPal-generated customer ID for the given user, or an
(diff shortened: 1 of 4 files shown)


Dinamiko (changes requested):
@Narek13 it seems `paypal_customer_id_for_user()` ignores `_ppcp_target_customer_id`, the plugin's canonical customer id, risking duplicate PayPal customers and orphaned tokens.
 
Recommendation: Mirror the established pattern, prefer `_ppcp_target_customer_id`, then `ppcp_customer_id`:
```php
     public function paypal_customer_id_for_user( int $user_id ): string {                                                                                                                   
         if ( 0 === $user_id ) {                                                                                                                                                             
             return '';                                                                                                                                                                      
         }                                                                                                                                                                                   
                                                                                                                                                                                             
         $customer_id = get_user_meta( $user_id, '_ppcp_target_customer_id', true );                                                                                                         
         if ( is_string( $customer_id ) && $customer_id ) {                                                                                                                                  
             return $customer_id;                                                                                                                                                            
         }                                                                                                                                                                                   
                                                                                                                                                                                             
         $customer_id = get_user_meta( $user_id, 'ppcp_customer_id', true );                                                                                                                 
         return is_string( $customer_id ) ? $customer_id : '';                                                                                                                               
     }  
```
This avoids creating duplicate PayPal customers, avoids clobbering an existing target customer id, and keeps the "let PayPal assign one" behaviour only for genuinely new customers. The two new unit tests would need a matching get_user_meta(`'_ppcp_target_customer_id', …`) expectation.

Dinamiko (approved):

PR #4577, opened 2026-08-04
Google Pay - Unsuccessfully payment with Google Pay when Pay Now disabled (6748)

Dinamiko:
Google Pay express checkout failed with "Order approve failed: Failed to create WooCommerce order: No shipping method has been selected." when the "Pay Now" setting was disabled.

Express wallets always create the WC order server-side (ApproveOrderEndpoint), but with Pay Now disabled the plugin doesn't send shipping options to PayPal. WooCommerceOrderCreator read the shipping method only from the PayPal order, so it had nothing to use, even though the chosen rate was already in WC()->session['chosen_shipping_methods'].                            
                                                                                                                                                                                             
### Solution                                                                                                                                                                                   
`WooCommerceOrderCreator` now resolves the shipping method in order of preference: the PayPal option flagged selected, then the first PayPal option, then WooCommerce's own chosen method via the existing `ShippingOptionFactory::from_wc_cart()`. It only throws when all of those are empty.                                                                                            

Also fixes a second bug in the same line: the old code took `options()[0]` and ignored the selected flag.

### DIFF
--- a/modules/ppcp-button/services.php
+++ b/modules/ppcp-button/services.php
@@ -430,7 +430,8 @@ return array(
 			$container->get( 'wc-subscriptions.helper' ),
 			$container->get( 'button.session.factory.card-data' ),
 			$container->get( 'api.factory.shipping' ),
-			$container->get( 'api.factory.payer' )
+			$container->get( 'api.factory.payer' ),
+			$container->get( 'api.factory.shipping-option' )
 		);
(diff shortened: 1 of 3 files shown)


stracker-phil (approved):
LGTM

Dinamiko:
Not needed for now.

PR #4595, opened 2026-08-10
Advanced Card Fields block checkout (6726) 

Dinamiko:
On a v6-enabled block checkout, the "Debit & Credit Cards" (ACDC) gateway was still rendered by the old v5 hosted-fields integration. The v6 card-fields flow existed only for classic checkout, so block checkout had no v6 card path.                                                                                                                                   
                                                                                                                                                                                             
This PR adds a v6 Advanced Card Fields component to the block checkout that mounts the SDK card fields and drives submission through the Blocks onPaymentSetup event — create order → card session confirms (running 3D Secure) → approve server-side → the checkout submit captures via the existing card gateway. 

Registered it as the block's card payment method, suppressed the v5 card block when v6 renders in its place, and extended the PHP config/gating to enable card fields on block checkout.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/blocks/V6CardFieldsComponent.test.js
+++ b/modules/ppcp-sdk-v6/resources/js/blocks/V6CardFieldsComponent.test.js
@@ -1,28 +1,28 @@
 const mockLoadSdkV6 = jest.fn();
-jest.mock( '../../sdkLoader', () => ( {
+jest.mock( '../sdkLoader', () => ( {
 	loadSdkV6: ( ...args ) => mockLoadSdkV6( ...args ),
 } ) );
 
 const mockCreateCardOrder = jest.fn();
 const mockApproveCardOrder = jest.fn();
-jest.mock( '../../endpointsAdapter', () => ( {
+jest.mock( '../endpointsAdapter', () => ( {
 	createCardOrder: ( ...args ) => mockCreateCardOrder( ...args ),
 	approveCardOrder: ( ...args ) => mockApproveCardOrder( ...args ),
 } ) );
 
 const mockCardFieldStyles = jest.fn();
-jest.mock( '../../cardFields/cardFieldStyles', () => ( {
+jest.mock( '../cardFields/cardFieldStyles', () => ( {
 	cardFieldStyles: ( ...args ) => mockCardFieldStyles( ...args ),
 } ) );
 
 const mockCardFieldContainer = jest.fn( () => null );
-jest.mock( '../../blocks/V6CardFieldContainer', () => ( {
+jest.mock( './V6CardFieldContainer', () => ( {
 	V6CardFieldContainer: ( props ) => mockCardFieldContainer( props ),
 } ) );
 
 import { render, waitFor, act } from '@testing-library/react';
 import { createElement } from '@wordpress/element';
-import { V6CardFieldsComponent } from '../../blocks/V6CardFieldsComponent';
+import { V6CardFieldsComponent } from './V6CardFieldsComponent';
 
 function cardConfig( overrides = {} ) {
 	return {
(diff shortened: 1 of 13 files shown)


modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -178,6 +179,19 @@ public function should_load_on_current_page(): bool {
 			&& is_active_widget( false, false, 'woocommerce_widget_cart' );
 	}
 
+	/**
+	 * Whether the v6 Advanced Card Fields should render on the current page.
+	 *
+	 * Gates both the JS `card_fields.enabled` flag and the suppression of the
+	 * v5 card block, so a page never ends up with neither card option.
+	 *
+	 * @return bool
```
stracker-phil: nit: not needed :)

Reason: the function already has `: bool` return type

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -167,7 +167,8 @@ public function should_load_on_current_page(): bool {
 			return true;
 		}
 
-		if ( 'checkout' === $page_location && $this->card_payments_configuration->is_acdc_enabled() ) {
+		if ( in_array( $page_location, array( 'checkout', 'checkout-block' ), true )
+			&& $this->card_payments_configuration->is_acdc_enabled() ) {
```
stracker-phil: suggestion: use the new `->is_card_fields_enabled()` function with an optional argument

Reason: this looks like an exact duplicate of the condition that was extracted into the is_enabled check

Result:
```php
if ( $this->is_card_fields_enabled( $page_location ) ) ...

public function is_card_fields_enabled( ?string $location = null ): bool {
	$location = $location ?? $this->get_page_context();

	return in_array( $location, array( 'checkout', 'checkout-block' ), true )
		&& $this->card_payments_configuration->is_acdc_enabled();
}
```

modules/ppcp-sdk-v6/resources/js/blocks/V6CardFieldsComponent.test.js
stracker-phil, line 1: issue: this file is in the wrong location.

Reason: JS tests are co-located in the same folder as the file they test.

This file should be in the folder `resources/js/blocks/` - there should be no `js/test/*` folder

modules/ppcp-sdk-v6/resources/js/endpointsAdapter.test.js
stracker-phil, line 1: issue: same as above, file is in wrong folder

stracker-phil (changes requested):
We should move the `.test.js` files before merging this

modules/ppcp-sdk-v6/resources/js/checkout-block.js
```diff
@@ -142,3 +143,25 @@ if ( config && config.page_context && config.continuation ) {
...
+		label: createElement( 'div', null, config.card_fields.title ),
+		ariaLabel: config.card_fields.title,
+		content: createElement( V6CardFieldsComponent, { config } ),
+		// A static placeholder, not the live fields: the SDK does not boot in
+		// the block editor.
+		edit: createElement(
+			'div',
+			{ className: 'ppcp-sdk-v6-editor-preview' },
+			config.card_fields.title
+		),
```
stracker-phil, line 161: suggestion: reuse the `V6EditorPreview` component

Reason: Other payment methods above already implement this via a preview component instead of creating a fully self controlled DOM element; if title is dynamic you could extend the preview component (otherwise maybe move the title to `fundingSources.js`)

```js
edit: createElement( V6EditorPreview, { title: config.card_fields.title } ),
```

stracker-phil (approved):

PR #4603, opened 2026-08-12
Save Payment Methods PayPal + Cards (5786)

Dinamiko:
The v6 Web SDK migration added wallet buttons and advanced card fields (classic + block checkout) but no vaulting, this PR implements v6 vaulting across all surfaces. 

Save-during-purchase now forwards the save choice for cards (WooCommerce's native tokenization checkbox on classic, Blocks' native showSaveOption/shouldSavePayment in blocks) into the order payload; PayPal/Venmo continue vaulting server-side from the merchant setting. 

A new v6 boot fully owns the Add Payment Method page, rendering the PayPal save button (`createPayPalSavePaymentSession`) and card-save fields (`createCardFieldsSavePaymentSession`), exchanging setup tokens via the existing Vault endpoints, and retiring the v5 stack there.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
+++ b/modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
@@ -78,39 +78,50 @@ import { handleError, setErrorLabels } from './utils/errorHandler';
 	}
 
 	/**
-	 * Shows the PayPal button only when PayPal is the selected method; the WC
-	 * submit button drives the card save flow for the card method. Replaces
-	 * the visibility handling the (now-suppressed) v5 script used to do.
+	 * Shows the PayPal button only when PayPal is the selected method and a
+	 * button actually rendered; otherwise the WC submit button stays visible
+	 * (card save flow, or PayPal ineligible). Replaces the visibility handling
+	 * the (now-suppressed) v5 script used to do.
 	 */
 	function syncVisibility() {
 		const isPayPal =
 			getCurrentPaymentMethod() === PaymentMethods.PAYPAL;
-		setVisibleByClass( ORDER_BUTTON_SELECTOR, ! isPayPal, 'ppcp-hidden' );
-		setVisible( buttonSelector, isPayPal );
+		// Only hand the submit role to the PayPal button when one actually
+		// rendered; otherwise (PayPal ineligible / no wrapper) the buyer must
+		// keep the native submit control instead of being left with nothing.
+		const hasPayPalButton = Boolean(
+			document.querySelector( `${ buttonSelector } paypal-button` )
+		);
+		const usePayPalButton = isPayPal && hasPayPalButton;
+		setVisibleByClass(
+			ORDER_BUTTON_SELECTOR,
+			! usePayPalButton,
+			'ppcp-hidden'
+		);
+		setVisible( buttonSelector, usePayPalButton );
 	}
 
 	async function init() {
 		const wrapper = document.querySelector( buttonSelector );
-		if ( ! wrapper ) {
-			return;
-		}
 
 		const sdk = await loadSdkV6( config, 'checkout' );
 		const eligibility = await checkVaultEligibility( sdk, {
 			currencyCode: config.currency,
 		} );
 
-		if ( eligibility.paypal ) {
+		// Render the PayPal button only when both its wrapper is present and
+		// PayPal is eligible.
+		if ( wrapper && eligibility.paypal ) {
 			const session = createSavePayPalSession( sdk, config );
 			wrapper.innerHTML = '';
 			wrapper.appendChild( createPayPalButton( session ) );
 		}
 
+		// Card saving is independent of the PayPal wrapper; gate it only on
+		// card eligibility and the card fields being enabled.
 		if ( eligibility.card && config.card_fields?.enabled ) {
 			initCardSaveFields( config );
 		}
-
-		syncVisibility();
 	}
 
 	function setupListeners() {
@@ -132,7 +143,12 @@ import { handleError, setErrorLabels } from './utils/errorHandler';
 
 	function boot() {
 		setupListeners();
-		init().catch( ( error ) => handleError( error ) );
+		// Always sync visibility once init settles: even if init() rejects
+		// (SDK load, client token, eligibility) the page must not be left with
+		// unmounted fields and an inconsistent submit control.
+		init()
+			.catch( ( error ) => handleError( error ) )
+			.finally( syncVisibility );
 	}
 
 	if ( document.readyState === 'loading' ) {
--- a/modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
+++ b/modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
@@ -1,0 +1,143 @@
+/**
+ * PayPal SDK v6 Bootstrap for My Account › Add Payment Method.
+ *
+ * Renders the v6 "save for later" surfaces that replace the v5
+ * add-payment-method.js: the PayPal wallet button (createPayPalSavePaymentSession)
+ * and the advanced card fields (createCardFieldsSavePaymentSession). The WC
+ * "Add payment method" submit button drives the card save; it is hidden while
+ * the PayPal method is selected, where the PayPal button is the submit control.
+ *
+ * @package
+ */
+
+import {
+	getCurrentPaymentMethod,
+	ORDER_BUTTON_SELECTOR,
+	PaymentMethods,
+} from '@ppcp-button/Helper/CheckoutMethodState';
+import { setVisible, setVisibleByClass } from '@ppcp-button/Helper/Hiding';
+import { loadSdkV6 } from './sdkLoader';
+import { checkVaultEligibility } from './eligibility';
+import { createSavePayPalSession } from './sessions/createSaveSession';
+import { initCardSaveFields } from './cardFields/saveRenderer';
+import { postJson } from './utils/api';
+import { handleError, setErrorLabels } from './utils/errorHandler';
+
+( function ( config ) {
+	'use strict';
+
+	if ( ! config ) {
+		return;
+	}
+
+	setErrorLabels( config.labels );
+
+	const buttonSelector = config.button.wrapper;
+
+	/**
+	 * Requests a PayPal wallet setup token from the existing WC AJAX endpoint.
+	 *
+	 * The returned promise is passed to session.start() without being awaited
+	 * first — awaiting it before the call loses the transient user activation
+	 * and breaks the popup. It resolves to a { vaultSetupToken } object,
+	 * mirroring the { orderId } shape the one-time session expects.
+	 *
+	 * @return {Promise<{vaultSetupToken: string}>} The setup token.
+	 */
+	function createVaultSetupToken() {
+		return postJson( config.ajax.create_setup_token ).then( ( data ) => ( {
+			vaultSetupToken: data.id,
+		} ) );
+	}
+
+	/**
+	 * Builds the PayPal save button and binds the click handler.
+	 *
+	 * @param {Object} session - The v6 save payment session.
+	 * @return {HTMLElement} The configured button element.
+	 */
+	function createPayPalButton( session ) {
+		const button = document.createElement( 'paypal-button' );
+		button.setAttribute( 'type', 'pay' );
+		if ( config.button.color_class ) {
+			button.className = config.button.color_class;
+		}
+
+		button.addEventListener( 'click', async () => {
+			try {
+				await session.start(
+					{ presentationMode: 'auto' },
+					createVaultSetupToken()
+				);
+			} catch ( error ) {
+				handleError( error );
+			}
+		} );
+
+		return button;
+	}
+
+	/**
+	 * Shows the PayPal button only when PayPal is the selected method; the WC
+	 * submit button drives the card save flow for the card method. Replaces
+	 * the visibility handling the (now-suppressed) v5 script used to do.
+	 */
+	function syncVisibility() {
+		const isPayPal =
+			getCurrentPaymentMethod() === PaymentMethods.PAYPAL;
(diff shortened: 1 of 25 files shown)


modules/ppcp-sdk-v6/services.php
```diff
@@ -49,11 +50,32 @@
 			// so this module does not depend on the ppcp-blocks module it replaces.
 			! $container->get( 'settings.settings-provider' )->enable_pay_now(),
 			$container->get( 'settings.settings-provider' )->save_paypal_and_venmo(),
-			$container->get( 'wcgateway.configuration.card-configuration' )
+			$container->get( 'wcgateway.configuration.card-configuration' ),
+			// Card "save during purchase" eligibility, mirroring the v5 block
+			// card method (AdvancedCardPaymentMethod): reference-transaction
+			// eligible AND the "save card details" setting on.
+			$container->get( 'save-payment-methods.eligible' )
+				&& $container->get( 'settings.settings-provider' )->save_card_details(),
```
Narek13: **Fatal when `ppcp-save-payment-methods` is disabled but v6 is enabled.**

`save-payment-methods.eligible` only exists while that module is loaded, and it has its own feature flag  independent of the v6 flag. The codebase already treats it as optional — `modules/ppcp-settings/services.php:114` guards it with `has()`.

This is not an edge path: `SdkV6Module` resolves this manager inside its `wp_enqueue_scripts` callback and `extensions.php` resolves it for every `button.smart-button` decoration, so the combination "save payment methods off + v6 on" throws on every front-end page rather than degrading.

```suggestion
			$container->has( 'save-payment-methods.eligible' )
				&& $container->get( 'save-payment-methods.eligible' )
				&& $container->get( 'settings.settings-provider' )->save_card_details(),
```

(`wc-subscriptions.helper` on the next line is safe — that module loads unconditionally, `modules.php:27`.)

modules/ppcp-sdk-v6/services.php
```diff
@@ -49,11 +50,32 @@
...
+		$settings_provider = $container->get( 'settings.settings-provider' );
+
+		return new AddPaymentMethodManager(
+			$container->get( 'sdk-v6.asset-getter' ),
+			$container->get( 'ppcp.asset-version' ),
+			$container->get( 'settings.environment' ),
+			$container->get( 'button.helper.context' ),
+			$settings_provider->save_paypal_and_venmo(),
+			$container->get( 'save-payment-methods.eligible' )
+				&& $settings_provider->save_card_details(),
```
Narek13: Same unguarded resolve as above, in the new service.

```suggestion
			$container->has( 'save-payment-methods.eligible' )
				&& $container->get( 'save-payment-methods.eligible' )
				&& $settings_provider->save_card_details(),
```

modules/ppcp-sdk-v6/resources/js/cardFields/saveRenderer.js
```diff
@@ -0,0 +1,160 @@
...
+			await postJson( config.ajax.create_payment_token, {
+				vault_setup_token: setupTokenId,
+			} );
+
+			submitting = true;
+			navigation.assign( config.payment_methods_page );
+		} catch ( error ) {
+			handleError( error );
+		}
+	}
```
Narek13, line 158: **A double click can vault the card twice.**

`submitting` is set only after the token exchange succeeds, right before the redirect, so it acts as a "leaving the page" latch rather than an in-flight guard — and nothing disables the button meanwhile. A second click during the async chain re-enters `handleSubmit`: a second `create_setup_token`, a second `session.submit()` (second 3DS challenge), a second `create_payment_token` exchange — i.e. two saved cards from one buyer action.

Compare the one-time flow in `cardFields/renderer.js:181-221`, which latches around the programmatic re-click *and* calls `spinner?.block()`, clearing the flag in `finally`.

```suggestion
	async function handleSubmit( event ) {
		if ( submitting || ! isCardGatewaySelected() ) {
			return;
		}

		event.preventDefault();
		event.stopImmediatePropagation();

		// Latch before the first await: a second click must not create a
		// second setup token and a second vaulted card.
		submitting = true;

		try {
			const cardSession = await ensureCardSession();
			const setupTokenId = await createCardSetupToken();
			const result = await cardSession.submit( setupTokenId );

			// Buyer dismissed the 3DS challenge; let them retry.
			if ( result.state === 'canceled' ) {
				submitting = false;
				return;
			}
			if ( result.state !== 'succeeded' ) {
				throw new Error( 'Card could not be saved.' );
			}

			await postJson( config.ajax.create_payment_token, {
				vault_setup_token: setupTokenId,
			} );

			navigation.assign( config.payment_methods_page );
		} catch ( error ) {
			submitting = false;
			handleError( error );
		}
	}
```

Blocking the button visually (the classic renderer's spinner equivalent) would be a nice addition on top.

modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
```diff
@@ -0,0 +1,143 @@
...
+	 * Shows the PayPal button only when PayPal is the selected method; the WC
+	 * submit button drives the card save flow for the card method. Replaces
+	 * the visibility handling the (now-suppressed) v5 script used to do.
+	 */
+	function syncVisibility() {
+		const isPayPal =
+			getCurrentPaymentMethod() === PaymentMethods.PAYPAL;
+		setVisibleByClass( ORDER_BUTTON_SELECTOR, ! isPayPal, 'ppcp-hidden' );
+		setVisible( buttonSelector, isPayPal );
+	}
```
Narek13, line 102: **The submit button is hidden even when no PayPal button rendered.**

`#place_order` is hidden whenever PayPal is the selected method, regardless of whether a button was actually appended. When `eligibility.paypal` is false (line 103) the wrapper stays empty, so the buyer gets neither a PayPal button nor a submit control — a dead end with no error shown.

Deriving it from what actually rendered keeps the two in sync without extra state:

```suggestion
	function syncVisibility() {
		const isPayPal =
			getCurrentPaymentMethod() === PaymentMethods.PAYPAL;
		// Only hand the submit role to the PayPal button when one actually
		// rendered; otherwise the buyer is left without any control.
		const hasPayPalButton = Boolean(
			document.querySelector( `${ buttonSelector } paypal-button` )
		);
		setVisibleByClass(
			ORDER_BUTTON_SELECTOR,
			! ( isPayPal && hasPayPalButton ),
			'ppcp-hidden'
		);
		setVisible( buttonSelector, isPayPal && hasPayPalButton );
	}
```

Related: if `init()` rejects (SDK load, client token, eligibility), `boot()`'s `.catch( handleError )` swallows it and `syncVisibility()` on line 113 never runs at all, leaving the page in its server-rendered state with unmounted card fields. Calling it in a `finally` would at least keep the visibility consistent.

modules/ppcp-sdk-v6/resources/js/boot-add-payment-method.js
```diff
@@ -0,0 +1,143 @@
...
+
+	async function init() {
+		const wrapper = document.querySelector( buttonSelector );
+		if ( ! wrapper ) {
+			return;
+		}
```
Narek13: **A missing PayPal wrapper also disables card saving.**

This early return skips `initCardSaveFields( config )` on line 110, so the card save fields silently depend on the *v5* module still printing `<div id="ppc-button-ppcp-gateway-save-payment-method">` (`SavePaymentMethodsModule.php:401-410`, on `woocommerce_add_payment_method_form_bottom`).

That markup is unaffected by this PR's script suppression, so it works today — but nothing expresses the dependency, and a merchant with only card vaulting enabled has no reason to expect the PayPal wrapper to matter. WE can scope the check to the branch that needs it:

```js
const wrapper = document.querySelector( buttonSelector );
const sdk = await loadSdkV6( config, 'checkout' );
const eligibility = await checkVaultEligibility( sdk, { currencyCode: config.currency } );

if ( wrapper && eligibility.paypal ) {
    // …render the PayPal button
}

if ( eligibility.card && config.card_fields?.enabled ) {
    initCardSaveFields( config );
}
```

modules/ppcp-sdk-v6/resources/js/eligibility.js
```diff
@@ -45,3 +45,25 @@ export async function checkEligibility(
...
+ */
+export async function checkVaultEligibility( sdkInstance, { currencyCode } ) {
+	const methods = await sdkInstance.findEligibleMethods( {
+		currencyCode,
+		paymentFlow: 'VAULT_WITHOUT_PAYMENT',
+	} );
+
+	return {
+		paypal: methods.isEligible( 'paypal' ),
+		card: methods.isEligible( 'advanced_cards' ),
```
Narek13, line 67: Is `advanced_cards` the right eligibility key?

modules/ppcp-sdk-v6/resources/js/cardFields/renderer.js
```diff
@@ -79,6 +79,19 @@ function isSavedTokenSelected() {
...
+ * Whether the buyer opted to save the card during purchase, read from
+ * WooCommerce's native tokenization checkbox for the card gateway (the same
+ * checkbox v5's CheckoutActionHandler reads). Absent when vaulting is off.
+ *
+ * @return {boolean} True when the "save payment method" checkbox is checked.
+ */
+function shouldSavePaymentMethod() {
+	return !! document.querySelector(
+		'#wc-ppcp-credit-card-gateway-new-payment-method'
+	)?.checked;
```
Narek13: The gateway id is already available as `config.card_fields.payment_method`, and WooCommerce derives this checkbox id from it (`wc-<gateway_id>-new-payment-method`). Hardcoding it here means a changed or filtered gateway id degrades silently to "never save" rather than failing loudly.

v5 hardcodes the same string (`CheckoutActionHandler.js:79-81`), so this is inherited — but since the function is new, passing the id in costs little:

```js
function shouldSavePaymentMethod( paymentMethod ) {
    return !! document.querySelector(
        `#wc-${ paymentMethod }-new-payment-method`
    )?.checked;
}
```

with `shouldSavePaymentMethod( paymentMethod )` at the call site on line 199.

modules/ppcp-sdk-v6/resources/js/cardFields/renderer.js
```diff
@@ -180,7 +193,11 @@ export async function initCardFields( config ) {
 
 		try {
 			const cardSession = await ensureCardSession();
-			const { orderId } = await createCardOrder( config );
+			const { orderId } = await createCardOrder(
+				config,
+				'checkout',
+				shouldSavePaymentMethod()
```
Narek13: **Classic and block checkout now disagree about subscriptions.**

The block component force-saves when `config.card_fields.has_subscriptions` is true (`V6CardFieldsComponent.js`, `savePaymentRef`), but the classic path here only reads WooCommerce's tokenization checkbox. If WC Subscriptions suppresses or unchecks that checkbox for a subscription cart, the card is not vaulted and renewals have no token.

This matches v5 behaviour, so it may well be intentional — but the two v6 surfaces behave differently now, which is the kind of asymmetry that gets discovered in production. Worth either applying `has_subscriptions` here too, or confirming with an explicit test: subscription product → classic checkout → v6 card fields → token stored.

Also `has_subscriptions` is built from `cart_contains_subscription()` only, while `SavePaymentMethodsModule.php:147-151` additionally checks `current_product_is_subscription()` and `order_pay_contains_subscription()` — fine for the block cart, possibly relevant for order-pay.

modules/ppcp-sdk-v6/src/Assets/AddPaymentMethodManager.php
```diff
@@ -0,0 +1,176 @@
...
+			: 'https://www.paypal.com';
+
+		/**
+		 * Filters the 3DS/SCA contingency used when creating the card setup
+		 * token for the save-for-later flow.
+		 *
+		 * @param string $verification_method The default 3D Secure enum value.
+		 */
+		$verification_method = (string) apply_filters(
+			'woocommerce_paypal_payments_three_d_secure_contingency',
```
Narek13: This duplicates the 3DS contingency localization that `SavePaymentMethodsModule.php:320-338` already performs for v5 (same filter, same `three_d_secure_enum()` source). Two places to keep in sync while both stacks coexist; a small shared service or helper would prevent drift.

No trust issue introduced — the endpoint whitelists the client-supplied value (`CreateSetupToken.php:88-91`) — so this is purely about duplication.

Narek13 (changes requested):
Thanks for working on this @Dinamiko 
In addition to what I have found and mentioned below, please note:

- No test for `boot-add-payment-method.js` at all: visibility sync per selected method, eligibility gating, missing-wrapper path.
- No PHP test for `AddPaymentMethodManager` (`should_load_on_current_page()` matrix + `script_data()`), although `SdkV6Manager` has one.

Dinamiko:
@Narek13 I added the changes, regarding `advanced_cards` eligibility I got it from the PayPal AI-Toolkit skill in [best practices](https://github.com/paypal/AI-Toolkit/blob/main/skills/paypal-best-practices/references/js-sdk-v6.md#card-fields-advanced-card-processing):
  - `services.php` (fatal fix): Guarded `save-payment-methods.eligible` with has() in both `sdk-v6.manager` and `sdk-v6.add-payment-method-manager`. Prevents the fatal on every front-end page when `save-payment-methods` is off but v6 is on (mirrors the existing pattern in `ppcp-settings/services.php`).                                                                                     
  - `saveRenderer.js` (double-vault fix): submitting is now latched before the first await (reset on cancel/error), so a double click can no longer create two setup tokens / two saved cards. 
  - `boot-add-payment-method.js` (dead-end + coupling fix): `syncVisibility()` only hides `#place_order` when a PayPal button actually rendered (no more empty-wrapper dead end when PayPal is ineligible); card save fields are decoupled from the PayPal wrapper's presence; `syncVisibility` now runs in `.finally()` so a failed `init()` still leaves consistent visibility.                                                                                          
  - New tests: `boot-add-payment-method.test.js` (JS) and `AddPaymentMethodManagerTest.php` (PHP), covering the two files the reviewer flagged as untested.

Narek13 (approved):
@Dinamiko Thanks for making all the changes and thanks for the info. LGTM

PR #4609, opened 2026-08-13
Webhook registration deregisters all webhooks for the current PayPal account (6770)

### LINKED ISSUE #4604: Webhook registration deregisters all webhooks for the current PayPal account (6770)
### Describe the Bug

When registering webhooks for a server, _all_ existing webhooks for the current PayPal account are deregistered, even if they are pointing at other servers or other services. This means connecting or reconnecting a staging site or similar secondary site to the same PayPal account will remove all webhooks before adding webhooks for the current site.

### To Reproduce

1. Set up a primary site with PayPal Payments installed, and connect using a PayPal account
2. Verify that webhooks are set up to deliver to the site
3. Set up a second site with PayPal Payments installed, and connect to the same PayPal account
4. The webhooks are only pointed at the second site after they have been registered.

### Expected Behavior

The registration process should deregister webhooks _for the current site_, not webhooks for other sites or services that may be needed for other purposes.

### Actual Behavior

All webhooks for the current account are removed.

### Environment

 - WordPress 7.0
 - WooCommerce 10.7.0
 - PayPal Payments 4.1.0
 - WooCommerce Subscriptions 9.0.0
 - PHP 8.4.22

### Additional Details

  ```
  ```

</details>

Dinamiko:
Fixes #4604.

`WebhookRegistrar` deletes every webhook on the connected PayPal account before re-registering, and always re-registers at the URL of whichever install runs it. So any second site sharing the same PayPal credentials (a staging/dev clone) wipes the primary site's webhook and silently redirects all webhook delivery to itself. Nothing warns the merchant, and the SSR "Webhook status" row stays green because it only checks whether this install ever received a webhook , not where delivery actually points.           
                                                                                                                                                                                             
### PR Changes                                                                                                                   
  - `WebhookRegistrar: deregistration` is now scoped by host, it deletes only webhooks pointing at this site and leaves other sites' webhooks intact, logging a warning when it skips a foreign one.
  - System Status Report: a new "Webhook delivery URL" row compares the live PayPal-registered webhook host against home_url(), showing a warning ("Delivered to …") when events are going to a different host, making the hijack visible. 

   
<img width="1433" height="397" alt="webhook-delivery" src="https://github.com/user-attachments/assets/6a8463bc-c2fe-462e-8c6e-c971ca3309d9" />

### DIFF
--- a/modules/ppcp-status-report/src/StatusReportModule.php
+++ b/modules/ppcp-status-report/src/StatusReportModule.php
@@ -34,6 +34,11 @@ class StatusReportModule implements ServiceModule, ExecutableModule {
 	use ModuleClassNameIdTrait;
 
 	/**
+	 * Transient key caching the live PayPal-side registered webhooks for the status page.
+	 */
+	private const REGISTERED_WEBHOOKS_TRANSIENT = 'ppcp-status-registered-webhooks';
+
+	/**
 	 * {@inheritDoc}
 	 */
 	public function services(): array {
@@ -139,10 +144,10 @@ class StatusReportModule implements ServiceModule, ExecutableModule {
 						'value'          => $this->bool_to_html( ! $last_webhook_storage->is_empty() ),
 					),
 					array(
-						'label'          => esc_html__( 'Webhook delivery URL', 'woocommerce-paypal-payments' ),
-						'exported_label' => 'Webhook delivery URL',
+						'label'          => esc_html__( 'Webhook delivery host', 'woocommerce-paypal-payments' ),
+						'exported_label' => 'Webhook delivery host',
 						'description'    => esc_html__( 'Whether PayPal delivers webhooks to this site or to a different host.', 'woocommerce-paypal-payments' ),
-						'value'          => $this->webhook_delivery_host_status( $c, $is_connected ),
+						'value'          => $this->webhook_delivery_host_status( $this->registered_webhooks( $c, $is_connected ) ),
 					),
 					array(
 						'label'          => esc_html__( 'PayPal Vault enabled', 'woocommerce-paypal-payments' ),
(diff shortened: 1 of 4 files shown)


modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -282,4 +289,75 @@ private function bool_to_html( bool $value ): string {
...
+			if ( '' !== $webhook_host && $webhook_host === $site_host ) {
+				return $this->bool_to_html( true );
+			}
+
+			if ( '' !== $webhook_host ) {
+				$foreign_hosts[ $webhook_host ] = $webhook_host;
+			}
+		}
+
+		return sprintf(
```
Narek13, line 371: **Blocking: this can render an empty host list.**

The loop above only adds to `$foreign_hosts` when `'' !== $webhook_host`, and it `continue`s past anything that isn't a `Webhook` instance - but this `sprintf` runs unconditionally. If PayPal returns webhooks whose URLs don't parse, or the service hands back a non-empty array of non-`Webhook` items, the merchant sees:

> Delivered to  (this site: mysite.com)

a red warning with nothing between "to" and the parenthesis. The early guard (`! is_array( $webhooks ) || array() === $webhooks`) doesn't cover it.

```php
if ( array() === $foreign_hosts ) {
    return $this->bool_to_html( false );
}
```

Neither new branch is covered by a test - see my note on the data provider.

modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -282,4 +289,75 @@ private function bool_to_html( bool $value ): string {
...
+	 * @return string
+	 */
+	private function webhook_delivery_host_status( ContainerInterface $c, bool $is_connected ): string {
+		if ( ! $is_connected ) {
+			return $this->bool_to_html( false );
+		}
+
+		try {
+			$webhooks = $c->get( 'webhook.status.registered-webhooks' );
+		} catch ( \Exception $exception ) {
```
Narek13: `\Exception` won't catch a `\Throwable` - a `TypeError` or `Error` raised inside the service factory bypasses this and fatals the whole status page. For a diagnostics row whose job is to degrade gracefully, `catch ( \Throwable $exception )` is the safer choice.

modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -282,4 +289,75 @@ private function bool_to_html( bool $value ): string {
...
+	 * @param bool               $is_connected Whether onboarding is complete.
+	 * @return string
+	 */
+	private function webhook_delivery_host_status( ContainerInterface $c, bool $is_connected ): string {
+		if ( ! $is_connected ) {
+			return $this->bool_to_html( false );
+		}
+
+		try {
+			$webhooks = $c->get( 'webhook.status.registered-webhooks' );
```
Narek13, line 321: JFYI: This adds an uncached live PayPal API call to the status page.

`webhook.status.registered-webhooks` is registered in `modules/ppcp-webhooks/factories.php`, and Modularity factories are non-shared, so every `get()` re-runs `WebhookEndpoint::list()` - nothing is reused with `webhook.status.registered-webhooks-data`.

A short transient would be cheap insurance.

modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -282,4 +289,75 @@ private function bool_to_html( bool $value ): string {
...
+	 * Reads the live PayPal-side webhook list (not local state) and compares each
+	 * webhook's host against this site's host. When they differ, webhook events are
+	 * being delivered elsewhere - typically a staging or dev clone connected to the
+	 * same PayPal account - which the "Webhook status" row above cannot detect.
+	 *
+	 * @param ContainerInterface $c            The services container.
+	 * @param bool               $is_connected Whether onboarding is complete.
+	 * @return string
+	 */
+	private function webhook_delivery_host_status( ContainerInterface $c, bool $is_connected ): string {
```
Narek13: Minor: passing the container into a private method makes it a service locator. The enclosing closure already resolves every other service up front, so resolving the webhook list there and passing `Webhook[]` in would match the file's existing shape - and would let the test drop its `ReflectionMethod` call.

modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -137,6 +138,12 @@ function () use ( $c ) {
 						'description'    => esc_html__( 'Whether we received webhooks successfully.', 'woocommerce-paypal-payments' ),
 						'value'          => $this->bool_to_html( ! $last_webhook_storage->is_empty() ),
 					),
+					array(
+						'label'          => esc_html__( 'Webhook delivery URL', 'woocommerce-paypal-payments' ),
```
Narek13: Copy: `Webhook delivery URL` -> `Webhook delivery host`.

The value only ever renders a hostname, never a full URL, and the description right below already says "host". Same change for `exported_label` on the next line.

The description and the `Delivered to %1$s (this site: %2$s)` string are both fine as written.

modules/ppcp-webhooks/src/WebhookRegistrar.php
```diff
@@ -133,4 +146,26 @@ private function do_unregister(): void {
...
+	 *
+	 * The comparison is host-based against this install's own endpoint URL, so the
+	 * NGROK_HOST development override is honoured automatically. A webhook whose URL
+	 * cannot be parsed is treated as not belonging to this site (conservative: it is
+	 * left in place rather than deleted).
+	 *
+	 * @param Webhook $webhook The webhook to check.
+	 * @return bool
+	 */
+	private function is_own_webhook( Webhook $webhook ): bool {
```
Narek13: **Blocking: this site's own stale webhooks are now never deleted.**

`do_unregister()` is called by `do_register()` on every re-registration, and "delete everything" used to clean up our own leftovers after a URL change. Now anything registered under a former host is foreign by definition and is skipped forever:

- a domain migration, or `http` -> `https` where the host also changed
- every `NGROK_HOST` rotation - which is precisely the dev workflow the docblock says host matching supports, so each restart leaves one more orphan

PayPal limits how many webhooks an app may have, so these accumulate until `create()` starts failing with no obvious cause.

Suggestion: alongside the host check, also delete the webhook recorded in `get_option( WebhookRegistrar::KEY )` by its stored id. That one is provably ours regardless of what the host is today.

modules/ppcp-webhooks/src/WebhookRegistrar.php
```diff
@@ -133,4 +146,26 @@ private function do_unregister(): void {
...
+		$own_host     = wp_parse_url( $this->incoming_webhook_endpoint->url(), PHP_URL_HOST );
+		$webhook_host = wp_parse_url( $webhook->url(), PHP_URL_HOST );
+
+		if ( ! is_string( $own_host ) || ! is_string( $webhook_host ) || '' === $own_host || '' === $webhook_host ) {
+			return false;
+		}
+
+		return strtolower( $own_host ) === strtolower( $webhook_host );
```
Narek13: Host-only comparison still lets same-host installs wipe each other:

- `example.com/shop-a` vs `example.com/shop-b`
- subdirectory multisite
- the very common `example.com/staging`

All of those share a host, so each still deletes the other's webhook - the exact failure in #4604, just on one domain. Comparing host **and** path (or the full incoming endpoint URL) would close it.

tests/PHPUnit/StatusReport/StatusReportModuleTest.php
```diff
@@ -0,0 +1,141 @@
...
+		$result = $this->webhook_delivery_host_status($container, true);
+
+		$this->assertSame('<mark class="no">&ndash;</mark>', $result);
+	}
+
+	public function empty_or_invalid_webhooks_provider(): array
+	{
+		return [
+			'no webhooks registered at all' => [[]],
+			'container returns a non-array value' => ['not-an-array'],
```
Narek13: Two branches of the new code aren't reachable from this provider:

- a non-empty array whose items aren't `Webhook` instances (the `continue` path)
- webhooks whose URLs have no parseable host (`'' === $webhook_host`)

Both land on the unconditional `sprintf` and produce the malformed `Delivered to  (this site: ...)` output I flagged in `StatusReportModule`. Adding them here would have caught it.

Narek13 (changes requested):
Thanks for working on this @Dinamiko 
I've found a few things that I'd like you to take a look at. Although I haven't tested it locally yet, I have some questions and concerns that I think are worth reviewing.

Dinamiko:
@Narek13 I addressed all your comments: webhook ownership now matches on stored id and host+path, and the SSR row is hardened, empty-host guard, `\Throwable` catch, a short transient around the live API call, a pure formatter instead of passing the container, the "Webhook delivery host" rename, and added coverage for the previously-untested branches.

Narek13 (approved):
Thanks for implementing the suggestions. LGTM

modules/ppcp-status-report/src/StatusReportModule.php
```diff
@@ -282,4 +294,102 @@ private function bool_to_html( bool $value ): string {
...
+	private function webhook_delivery_host_status( array $registered_webhooks ): string {
+		$site_host = $this->normalized_host( home_url() );
+
+		$foreign_hosts = array();
+		foreach ( $registered_webhooks as $webhook ) {
+			if ( ! $webhook instanceof Webhook ) {
+				continue;
+			}
+
+			$webhook_host = $this->normalized_host( $webhook->url() );
```
daledupreez, line 355: Not blocking, but worth considering as a follow-up: the SSR code is only checking the host name, so it will not flag various situations that the registration logic is treating as separate, most especially sites using distinct root URL paths.

PR #4610, opened 2026-08-13
Smart Button order creation bypasses woocommerce_checkout_create_order_line_item (6773)

### LINKED ISSUE #4601: Smart Buttons should trigger standard action when adding line items to order (6773)
### Describe the Bug

The code in `WooCommerce\PayPalCommerce\Button\Helper\WooCommerceOrderCreator->configure_line_items()` creates new `WC_Order_Item_Product` instances from the incoming cart data, but bypasses any custom logic that relies on the WooCommerce Core `woocommerce_checkout_create_order_line_item` action to augment the created order item based on data that has been added to the cart.

It's worth noting that https://github.com/woocommerce/woocommerce-paypal-payments/issues/4592 is requesting a custom filter for this, but in this case there is a bigger issue in that other plugins relying on the hook above are being bypassed and resulting in incomplete details once the order has been built.

I've seen reported issues with WooCommerce Subscriptions, WooCommerce Gift Cards, WooCommerce All Products for Subscriptions, as well as [this forum thread](https://wordpress.org/support/topic/add-on-product-meta-missing-in-orders-when-using-paypal-smart-buttons/) from Advanced Product Fields (Product Addons) for WooCommerce.

### To Reproduce

1. Enable one or more plugins that hook into `woocommerce_checkout_create_order_line_item`
2. Create or modify a product for one of the related plugins
3. Ensure Smart Buttons are enabled for that product on the product page and/or cart
4. Purchase the product via the Smart Button flow
5. Confirm that all product metadata is missing from the order

### Expected Behavior

The plugin code should aim to trigger standard WooCommerce hooks to work more effectively with other WooCommerce plugins.

### Actual Behavior

The order items are created without any custom plugin metadata.

### Environment

 - WordPress Version
 - WooCommerce 10.9.4
 - PayPal Payments 4.1.1
 - Browser [e.g. Chrome, Safari] and Version
 - Any other plugins installed

### Additional Details

  ```
  ```

</details>

Dinamiko:
Fixes #4601.

Orders created through the Smart Buttons flow were built by hand in `WooCommerceOrderCreator::configure_line_items()`, which never fired WooCommerce Core's `woocommerce_checkout_create_order_line_item` hooks. As a result, third-party plugins (e.g., Subscriptions, Gift Cards, Product Add-ons...) couldn't attach their cart-item metadata, so orders came out incomplete.                                                                                                                                                                            
                                                                                                                                                                                             
This PR mirrors WC Core in configure_line_items(), build each item via the `woocommerce_checkout_create_order_line_item_object` filter, set the order on the item, and fire the `woocommerce_checkout_create_order_line_item` action before adding it. Since `CartData::items()` already carries the full original cart item array, these plugins now receive the data they need and their metadata persists.

### DIFF
--- a/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
+++ b/modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
@@ -131,15 +131,31 @@ class WooCommerceOrderCreator {
 	 * @psalm-suppress InvalidScalarArgument
 	 */
 	protected function configure_line_items( WC_Order $wc_order, CartData $cart_data, ?Payer $payer, ?Shipping $shipping ): void {
-		foreach ( $cart_data->items() as $cart_item ) {
+		foreach ( $cart_data->items() as $cart_item_key => $cart_item ) {
 			$product_id           = $cart_item['product_id'] ?? 0;
 			$variation_id         = $cart_item['variation_id'] ?? 0;
 			$quantity             = $cart_item['quantity'] ?? 0;
 			$variation_attributes = $cart_item['variation'];
 
-			$item = new WC_Order_Item_Product();
+			/**
+			 * Filters the order line item object, mirroring WooCommerce Core so plugins
+			 * can swap the item class before it is populated.
+			 *
+			 * @param WC_Order_Item_Product $item          The order line item.
+			 * @param string                $cart_item_key The cart item hash key.
+			 * @param array                 $cart_item     The cart item data.
+			 * @param WC_Order              $wc_order      The order being built.
+			 */
+			$item = apply_filters(
+				'woocommerce_checkout_create_order_line_item_object',
+				new WC_Order_Item_Product(),
+				$cart_item_key,
+				$cart_item,
+				$wc_order
+			);
 			$item->set_product_id( $product_id );
 			$item->set_quantity( $quantity );
+			$item->set_order( $wc_order );
 
 			if ( isset( $cart_item['bundled_by'] ) ) {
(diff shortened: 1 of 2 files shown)


modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
```diff
@@ -131,15 +131,31 @@ public function create_from_paypal_order( Order $order, $cart, ?array $paypal_da
...
+			$item = apply_filters(
+				'woocommerce_checkout_create_order_line_item_object',
+				new WC_Order_Item_Product(),
+				$cart_item_key,
+				$cart_item,
+				$wc_order
+			);
 			$item->set_product_id( $product_id );
 			$item->set_quantity( $quantity );
+			$item->set_order( $wc_order );
```
Narek13, line 158: Does this exist in `WC_Order_Item_Product` ?
I think Order items have `set_order_id()` only

Narek13 (commented):
Thanks. Can you please check my comment?
Maybe I am missing something but this can throw a fatal

modules/ppcp-button/src/Helper/WooCommerceOrderCreator.php
Dinamiko: @Narek13 yes, `WC_Order_Item_Product` extends `WC_Order_Item` which has [set_order](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-order-item.php#L239)

Narek13 (approved):
Thanks for working on this

PR #4614, opened 2026-08-15
Invoice prefix cannot be cleared in new settings UI plugin always falls back to wc (6557)

mmaymo:
## 🐛 What and why

A merchant who clears the Invoice Prefix field and saves still gets `WC-` prepended to every outgoing PayPal invoice ID. The empty string is persisted correctly all the way to the database, but the `api.prefix` container extension evaluated it with a truthiness check and substituted the `WC-` default, making an empty prefix **unreachable**.

This is a regression from the settings-module migration. The legacy definition used `Settings::has('prefix')`, which returns true for a stored empty string.

The naive fix — returning the value verbatim — is unsafe on its own. `api.prefix` is shared with the vault customer ID, where an empty prefix makes the plugin fabricate a bare user ID such as `5` and persist it as the merchant-scoped PayPal `customer_id`.

## ⚙️ How it works now

The `api.prefix` container key carries the merchant's configured value exactly, empty string included. `PurchaseUnitFactory::from_wc_order()` composes `$this->prefix . $order->get_order_number()`, so an empty prefix yields the bare order number.

The vault customer ID no longer reads that key. `api.repository.customer` is built from the new `api.customer-prefix` service, which resolves `api.prefix` inside its closure and substitutes `CustomerRepository::FALLBACK_PREFIX` only when the result is empty. `CustomerRepository::customer_id_for_user()` therefore keeps producing a prefixed ID for a not-yet-vaulted user.

The fallback lives in the **wiring** rather than inside `CustomerRepository`, so the constructor argument keeps an honest contract and the customer-ID prefix policy has one obvious home.

### Covered by unit tests

- ✅ With the invoice prefix stored as an empty string, `api.prefix` resolves to `''` rather than `WC-`.
- ✅ `PurchaseUnitFactory` constructed with an empty prefix produces an `invoice_id` equal to the bare order number.
- ✅ `CustomerRepository::customer_id_for_user()` yields `WC-5` for user 5 under an empty invoice prefix — never the bare `5`.
- ✅ `CustomerRepository::customer_id_for_user()` returns a stored `ppcp_customer_id` unchanged, regardless of the configured prefix.
- ✅ With a non-empty prefix configured, both the invoice ID and the fabricated customer ID use it exactly as they do today.
- ✅ A prefix containing spaces stored before validation existed survives a `SettingsModel` round trip unchanged.
- ✅ An empty prefix written to the real `woocommerce-ppcp-data-settings` option reaches `api.factory.purchase-unit` through a freshly built container and produces an unprefixed invoice ID (integration suite).

### Not covered by unit tests

- ❌ `POST /wc/v3/wc_paypal/settings` with `invoicePrefix = 'WC Store'` returns `success => false` — already covered by `SettingsRestEndpointTest.php`.
- ❌ `POST /wc/v3/wc_paypal/settings` with `invoicePrefix = ''` is accepted and persists the empty string — already covered by `SettingsRestEndpointTest.php`.
- ❌ `POST /wc/v3/wc_paypal/settings` with `invoicePrefix = 'WC-Store_1'` is accepted verbatim — already covered by `SettingsRestEndpointTest.php`.
- ❌ `InvoicePrefix.js` refuses a disallowed character on input or paste, keeps the message visible while typing, and clears it on blur — client-side behavior covered by the Jest suite in `InvoicePrefix.test.js`.
- ❌ `InvoicePrefix.js` shows no error when the field is cleared to empty — covered by `InvoicePrefix.test.js`.
- ❌ `ControlTextInput` renders without `ppcp--has-error`, `aria-invalid`, or `.ppcp-r-control-error` when no `error` prop is passed — covered by `ControlTextInput.test.js`.
- ❌ `modules/ppcp-webhooks/services.php` no longer contains the unused `$prefix` assignment — removal of an unused variable produces no observable behavior to assert.

## 🚨 What must not regress

- `PurchaseUnitFactory::__construct()` — the `string $prefix = 'WC-'` default stays. The container always passes a value; changing it only churns tests.
- `SettingsTabMigration.php:127` — must keep mapping `'prefix' => 'invoice_prefix'` verbatim. No retroactive sanitization of already-stored prefixes.
- `SettingsModel::set_invoice_prefix()` — must keep using `sanitize_text_field()` only, which preserves spaces in previously saved values.
- `ControlTextInput.js` — validation stays opt-in via the `error` / `onBlur` props; no globally applied validation.
- The five other `ControlTextInput` consumers — brand name, soft descriptor, shipping-origin postal code, and sandbox/live Client ID and Secret Key — pass neither prop and must render identically.
- `RestEndpoint::return_error()` and `data/settings/actions.js` — the silent save-failure gap (HTTP 200 with `success => false`, never inspected by the client) is out of scope.
- `wcgateway.settings.invoice-prefix` and `wcgateway.settings.invoice-prefix-random` — new-install defaults are unchanged, so new stores still receive a generated site-specific prefix.

### New container service

`api.customer-prefix` in `modules/ppcp-api-client/services.php`. Resolves `api.prefix` lazily and falls back to `CustomerRepository::FALLBACK_PREFIX`. `api.repository.customer` is its only consumer.

### DIFF
--- a/modules/ppcp-api-client/src/Repository/CustomerRepository.php
+++ b/modules/ppcp-api-client/src/Repository/CustomerRepository.php
@@ -16,6 +16,18 @@ class CustomerRepository {
 	const CLIENT_ID_MAX_LENGTH = 22;
 
 	/**
+	 * The prefix used for fabricated customer IDs when no invoice prefix is configured.
+	 *
+	 * Unlike the invoice ID, a customer ID is scoped to the PayPal merchant account
+	 * rather than the store, and it is persisted the first time a user is vaulted.
+	 * Without a prefix it would degrade to the bare WordPress user ID, so two stores
+	 * on one merchant account would fabricate the same ID for different people.
+	 *
+	 * @var string
+	 */
+	const FALLBACK_PREFIX = 'WC-';
+
+	/**
 	 * The prefix.
 	 *
 	 * @var string
--- a/modules/ppcp-api-client/services.php
+++ b/modules/ppcp-api-client/services.php
@@ -136,6 +136,13 @@ return array(
 	'api.prefix'                                     => static function (): string {
 		return 'WC-';
 	},
+	/**
+	 * The prefix for fabricated vault customer IDs.
+	 */
+	'api.customer-prefix'                            => static function ( ContainerInterface $container ): string {
+		$prefix = $container->get( 'api.prefix' );
+		return $prefix ? $prefix : CustomerRepository::FALLBACK_PREFIX;
+	},
 	'api.bearer'                                     => static function ( ContainerInterface $container ): Bearer {
 		$is_connected = $container->get( 'settings.flag.is-connected' );
 
@@ -382,7 +389,7 @@ return array(
 		return new PayeeRepository( $merchant_email, $merchant_id );
 	},
 	'api.repository.customer'                        => static function ( ContainerInterface $container ): CustomerRepository {
-		$prefix           = $container->get( 'api.prefix' );
+		$prefix = $container->get( 'api.customer-prefix' );
 		return new CustomerRepository( $prefix );
 	},
 	'api.repository.order'                           => static function ( ContainerInterface $container ): OrderRepository {
--- a/modules/ppcp-wc-gateway/extensions.php
+++ b/modules/ppcp-wc-gateway/extensions.php
@@ -49,11 +49,18 @@ return array(
 		assert( $settings_provider instanceof SettingsProvider );
 		return $settings_provider->client_secret();
 	},
+	/**
+	 * The invoice prefix exactly as the merchant configured it. An empty string is a
+	 * supported choice meaning "no prefix", so it must not be replaced by a default
+	 * here. New installations receive a generated, site-specific default instead.
+	 *
+	 * The vault customer ID does not depend on this key; it resolves
+	 * 'api.customer-prefix', which keeps a fallback of its own.
+	 */
 	'api.prefix'                     => static function ( string $previous, ContainerInterface $container ): string {
 		$settings_provider = $container->get( 'settings.settings-provider' );
 		assert( $settings_provider instanceof SettingsProvider );
-		$prefix = $settings_provider->invoice_prefix();
-		return $prefix ? $prefix : 'WC-';
+		return $settings_provider->invoice_prefix();
 	},
 	'woocommerce.logger.woocommerce' => function ( LoggerInterface $previous, ContainerInterface $container ): LoggerInterface {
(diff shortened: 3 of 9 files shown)


modules/ppcp-api-client/src/Repository/CustomerRepository.php
```diff
@@ -15,6 +15,18 @@
...
+	 * The prefix used for fabricated customer IDs when no invoice prefix is configured.
+	 *
+	 * Unlike the invoice ID, a customer ID is scoped to the PayPal merchant account
+	 * rather than the store, and it is persisted the first time a user is vaulted.
+	 * Without a prefix it would degrade to the bare WordPress user ID, so two stores
+	 * on one merchant account would fabricate the same ID for different people.
+	 *
+	 * @var string
+	 */
+	const FALLBACK_PREFIX = 'WC-';
```
stracker-phil: nit: Let's add visibility to the const

We did not do it in the past, but for new code we want to define the `public` visibility for all props, consts and methods. Existing code is often not a reliable reference 😅

stracker-phil (changes requested):
I see that we're patching "legacy spaghetti code" here and would prefer to clean up the prefix logic right away:

**Reasons**

1. services that return a string are cached by the DI layer and cannot be filtered later
2. the current service-extensions split is an artifact from the settings refactoring and not needed anymore.

**Suggested implementation**

1. Create a new class that returns the prefix and implements a `__toString()` method, same pattern as [CurrencyGetter](https://github.com/woocommerce/woocommerce-paypal-payments/blob/9f163f296ee2bd16edb05d80bfeff2c049460a74/modules/ppcp-api-client/src/Helper/CurrencyGetter.php).
2. Suggestion: Add a filter in that class so third-party code has an easy way to modify the prefix.
3. Update the DI service `api.prefix` to return the new class instance instead of a string, and drop the `api.prefix` extension in `ppcp-wc-gateway`, which the getter makes obsolete. Then fix the type hints at the 3 call sites:
    - `PurchaseUnitFactory`, `RefundProcessor` and `CustomerRepository`
4. Revert the new `api.customer-prefix` service and the `FALLBACK_PREFIX` const from this PR. An empty prefix should actually mean no prefix.


**Sample class**

_Following the `CurrencyGetter` pattern_

```php
class InvoicePrefixGetter {
      private $settings;

      public function __construct( SettingsProvider $settings ) {
              $this->settings = $settings;
      }

      public function get(): string {
              return (string) apply_filters(
                      'woocommerce_paypal_payments_invoice_prefix',
                      $this->settings->invoice_prefix()
              );
      }

      public function __toString() {
              return $this->get();
      }
}
```

PR #4615, opened 2026-08-17
Fixes for JS SDK v6 epic branch

Dinamiko:
This PR includes fixes for issues found by QA in JS SDK v6 epic branch.

1. Block cart → Pay Now: "No payment method provided."                  
  - Problem: On the block cart, approving PayPal tried to submit the checkout store from the cart page, which the Store API rejects.
  - Solution: On the block cart, create and capture the WooCommerce order server-side (like the classic cart Pay Now flow) and redirect to the order-received page, instead of triggering a checkout submit.
  2. Mini-cart Venmo missing on shop/home                 
  - Problem: v6 only loaded sitewide when the classic cart widget was active, so with a block Mini-Cart the SDK never loaded on shop/home — no buttons at all.
  - Solution: Load v6 sitewide whenever the mini-cart location is enabled (matching v5), so the SDK is present wherever the mini-cart can appear.
                                                                                                                                                                                             
  3. Classic checkout: black "Proceed to PayPal" button beside the PayPal buttons                       
  - Problem: The v5 logic that hid the native order button wasn't carried over to v6, so both showed.             
  - Solution: Hide the native #place_order button while the PayPal wallet is the selected method (kept visible for cards, which submit through it), re-applied after each checkout refresh.  
                                                                                                                                                                                             
  4. Classic checkout ACDC: cardholder-name enabled → front-end error                        
  - Problem: The v6 card SDK has no "name" field, so requesting one threw; also the injected name input reused the expiry-field class, so WooCommerce masked it as a date.
  - Solution: Stop asking the SDK for a name field — render the cardholder name as a plain input and send it to the server as the card name; removed the wrong expiry class.
                                                                                                                                                                                             
  5. Classic checkout: 3DS card payment fails
  - Problem: The card confirmation was submitted without a billing address, so 3DS/AVS verification could fail.                                                                              
  - Solution: Pass the buyer's billing address (postal code + country) into the card confirmation, as the SDK expects.                                                                       
                                                                                                                                                                                             
  6. Classic checkout: PayPal buttons show for a $0 total 
  - Problem: Button rendering never checked whether the order actually needed payment.
  - Solution: Gate the buttons on "cart needs payment", so genuinely free orders show nothing. (When a $0 order still requires a payment method — e.g. a subscription free trial — the       
  buttons remain, which is expected and matches the other gateways.)                                                                                                                         
                                                                                                                                                                                             
  7. Pay-for-order page still on v5
  - Problem: The order-pay page was excluded from v6, and its underlying flow rejected guest orders and left the black button showing.
  - Solution: Migrated the pay-for-order page to v6 (buttons + card fields), wired it to pay the existing order, and fixed the permission check so guest orders can be paid (used the correct "pay for order" capability instead of "view order"). 

  8. Block checkout ACDC: card fields missing when logos + cardholder enabled                                                                                                                
  - Problem: The unsupported name field threw during rendering and blanked the whole card-fields area.                                                               
  - Solution: Render the cardholder name as a normal text input instead of an SDK field, so the card fields render reliably.                                                                 
                                                                                                                                                                                             
  New (found during this work): block checkout doesn't show the supported-card logos
  - Problem: The "Show logos of supported cards" setting was never passed into the v6 block card method.     
  - Solution: Feed the existing card-icon list into the block card method's label so the brand logos display when the setting is on.


modules/ppcp-card-fields/src/CardFieldsModule.php
```diff
@@ -101,13 +101,15 @@ function ( $default_fields, $id ) use ( $c ) {
...
 					</p>';
 
-					// Moves new item to first position.
+					// Moves the cardholder-name field to the first position while
+					// preserving string keys (array_unshift would reindex them and
+					// break the card-number placeholder replacement below).
 					$new_field = $default_fields['card-name-field'];
 					unset( $default_fields['card-name-field'] );
-					array_unshift( $default_fields, $new_field );
+					$default_fields = array( 'card-name-field' => $new_field ) + $default_fields;
```
stracker-phil: **suggestion**: Use `array_merge()` instead of `+` / Assign the HTLM to the `$new_field` directly

**Reason**: `+` is a quite uncommon array operation, and `array_merge` would make this slightly more maintainable/clear.

Also note that we first assign a value to `$default_fields['card-name-field']` and then, 2 lines later unset() this value to re-position it in the array.

**Suggestion**
```php
if ( CreditCardGateway::ID === $id && $should_show_card_holder_name ) {
  $card_name_field = '<p class="form-row form-row-wide">...';

  // Inject the cardholder-name before all other fields.
  $default_fields = array_merge(
    array( 'card-name-field' => $card_name_field ),
    $default_fields
  );
}
```

modules/ppcp-card-fields/src/CardFieldsModule.php
```diff
@@ -166,12 +168,22 @@ function ( array $data, string $payment_method ) use ( $c ): array {
 					);
 				}
 
+				// The v6 card-fields component set is number|expiry|cvv only, so the
+				// cardholder name (when enabled) is submitted as a plain field and
+				// forwarded here as payment_source.card.name.
+				$card_name = isset( $request_data['card_name'] )
+					? substr( trim( (string) $request_data['card_name'] ), 0, 300 )
+					: '';
```
stracker-phil: **nit**: Could be shortened by using `??`

**Suggestion**
```php
// Shortest:
$card_name = substr( trim( (string) ($request_data['card_name'] ?? '' ) ), 0, 300 );

// More readable:
$card_name = (string) ($request_data['card_name'] ?? '' );
$card_name = substr( trim( $card_name ), 0, 300 );
```

modules/ppcp-sdk-v6/resources/js/blocks/V6CardFieldsComponent.js
```diff
@@ -168,6 +201,8 @@ export function V6CardFieldsComponent( {
 				context,
 				session: sessionRef.current,
 				responseTypes,
+				cardName: cardNameRef.current?.trim() || '',
+				billingAddress: billingRef.current || undefined,
```
stracker-phil: **nit**: is the `undefined` fallback intentional here?

**Reason**: In line 147 the "no address" branch uses `null`, and the condition in line 59 just checks for falsey

**Suggestion**: Use a consistent value for "no billing address", I'd lean towards `null`

modules/ppcp-sdk-v6/resources/js/blocks/V6CardFieldsComponent.js
```diff
@@ -206,16 +241,29 @@ export function V6CardFieldsComponent( {
...
+							value: cardName,
+							onChange: ( event ) =>
+								setCardName( event.target.value ),
+							placeholder: __(
+								'Cardholder name (optional)',
+								'woocommerce-paypal-payments'
+							),
+							style: { width: '100%', ...inputStyle },
+						} )
+					),
```
stracker-phil, line 267: **suggestion**: Rename this file to `.jsx` and convert this block to a React component

**Reason**: this `createElement` nesting (line 240 - 300) is not maintainable, and JSX solves this elegantly

**Sample**

```js
function NameField( { cardName, setCardName } ) {
	return (
		<div className="ppcp-sdk-v6-card-field ppcp-sdk-v6-card-field--name">
			<input
				type="text"
				className="input-text"
				value={ cardName }
				onChange={ ( event ) => setCardName( event.target.value ) }
				placeholder={ __( 'Cardholder name (optional)', 'woocommerce-paypal-payments' ) }
				style={ { width: '100%', ...inputStyle } }
			/>
		</div>
	);
}

function ExpiryCvvRow( { session, inputStyle } ) {
	return (
		<div className="ppcp-sdk-v6-card-fields__row" style={ { display: 'flex', gap: '16px' } }>
			<V6CardFieldContainer
				session={ session }
				type="expiry"
				style={ inputStyle }
				placeholder={ __( 'MM / YY', 'woocommerce-paypal-payments' ) }
				containerStyle={ { flex: 1 } }
			/>
			<V6CardFieldContainer
				session={ session }
				type="cvv"
				style={ inputStyle }
				placeholder={ __( 'CVV', 'woocommerce-paypal-payments' ) }
				containerStyle={ { flex: 1 } }
			/>
		</div>
	);
}

{ fieldsReady && (
	<Fragment>
		{ /* The v6 card-fields component set is number|expiry|cvv only, so
			the cardholder name is a plain input; its value is forwarded to
			create-order rather than confirmed through the card session. */ }
		{ hasNameField && <NameField cardName={ cardName } setCardName={ setCardName } /> }
		<V6CardFieldContainer
			session={ session }
			type="number"
			style={ inputStyle }
			placeholder={ __( 'Card number', 'woocommerce-paypal-payments' ) }
		/>
		<ExpiryCvvRow session={ session } inputStyle={ inputStyle } />
	</Fragment>
) }
```

modules/ppcp-sdk-v6/resources/js/boot.js
```diff
@@ -20,6 +20,13 @@ import { createOrder, fetchCartTotal } from './endpointsAdapter';
 import { initCardFields } from './cardFields/renderer';
 import { hasJQuery } from './utils/api';
 import { setErrorLabels } from './utils/errorHandler';
+import { setVisible } from '@ppcp-button/Helper/Hiding';
+
+// The native WC submit button, labelled "Proceed to PayPal" for the PayPal
+// gateway. It is replaced by the wallet buttons while the PayPal gateway is
+// selected, but stays as the submit for cards and every other method.
+const PLACE_ORDER_SELECTOR = '#place_order';
+const WALLET_GATEWAY_ID = 'ppcp-gateway';
```
stracker-phil: **suggestion**: Avoid the term "wallet" in this const, it's confusing

**Reason**: Internally, we refer to wallets for Apple Pay and Google Pay, not for the PPCP gateway. Introducing this term here will confuse devs in the future

rename to `const PAYPAL_GATEWAY_ID` (which the `ppcp-gateway` method actually refers to)

modules/ppcp-sdk-v6/resources/js/boot.js
```diff
@@ -198,9 +205,33 @@ import { setErrorLabels } from './utils/errorHandler';
...
+			! hasJQuery() ||
+			! [ 'checkout', 'pay-now' ].includes( config.page_context )
+		) {
+			return;
+		}
+
+		const selected = document.querySelector(
+			'input[name="payment_method"]:checked'
+		)?.value;
+		const isWallet = selected === WALLET_GATEWAY_ID;
```
stracker-phil: **suggestion**: Same as above, I would avoid using `isWallet` to identify the PayPal gateway

modules/ppcp-sdk-v6/resources/js/checkout-block.js
```diff
@@ -144,12 +144,44 @@ if ( config && config.page_context && config.continuation ) {
...
+				alignItems: 'center',
+				justifyContent: 'space-between',
+				width: '100%',
+			},
+		},
+		createElement( 'span', null, config.card_fields.title ),
+		PaymentMethodIcons &&
+			icons.length > 0 &&
+			createElement( PaymentMethodIcons, { icons, align: 'right' } )
+	);
```
stracker-phil, line 176: suggestion: Sooner or later we should rename these files to .jsx and use JSX syntax to generate HTML output instead of `create_element()`

modules/ppcp-sdk-v6/resources/js/endpointsAdapter.js
```diff
@@ -22,6 +22,34 @@ export const navigation = {
...
+/**
+ * Submits the pay-for-order form after selecting the PayPal gateway, so the
+ * approved order is captured by the PayPal gateway (not whichever method radio
+ * happens to be checked) and WC issues the order-received redirect.
+ *
+ * @throws {Error} When jQuery or the pay-order form is unavailable.
+ */
+function submitPayOrderForm() {
+	if ( typeof jQuery === 'undefined' ) {
+		throw new Error( 'Could not submit the order.' );
```
stracker-phil, line 38: **nit**: Add a reason (or an additional `console.error` call) to make the problem easier to debug, if this is ever thrown

Dinamiko:
@stracker-phil I applied the changes, all changes are style/naming only, no behavior changes:
- `CardFieldsModule.php`: `cardholder-name` field now injected via `array_merge` with a local `$card_name_field` var.
- `boot.js`: renamed `WALLET_GATEWAY_ID` → `PAYPAL_GATEWAY_ID` and `isWallet` → `isPayPalGateway`. 
- `V6CardFieldsComponent.js`: billing address now passed as null (not undefined) for the "no address" case.
- `endpointsAdapter.js`: added pconsole.errorp context to both `submitPayOrderForm` throws.

stracker-phil (approved):
LGTM 👍

PR #4619, opened 2026-08-18
reCAPTCHA rejection logging produces no logs unless plugin logging is on (5551)

Narek13:
The reCAPTCHA integration has its own "Log rejected attempts" setting, but `Recaptcha::log_rejection()` wrote through the shared `woocommerce.logger.woocommerce` service — and `ppcp-wc-gateway/extensions.php:58-61` replaces that with a `NullLogger` unless the plugin-wide Troubleshooting → Logging setting is on. Enabling only the reCAPTCHA checkbox produced no logs at all, while the rejection counter on the settings page kept climbing. That mismatch is the reported symptom.

### PR Changes

- New `fraud-protection.recaptcha.rejection-logger` service: a `WooCommerceLogger` bound to `Recaptcha::REJECTION_LOGGER_SOURCE` and independent of the plugin-wide setting (`NullLogger` when `wc_get_logger()` is unavailable).
- `Recaptcha::log_rejection()` writes through it and drops the now-redundant `'source'` context key. The other six `$this->logger` calls are plugin diagnostics and stay on the shared, gated logger.

Destination is unchanged: rejections still land in the `woocommerce-paypal-payments-recaptcha` source the settings page's "view logs" link points at, and are not duplicated into the main log when plugin logging is on.

### Steps to Test

**WooCommerce → Settings → Integration → reCAPTCHA:** Enabled ✓; **both** the v3 *and* v2 key pairs filled in (it bails unless both are present); Score Threshold `1.0` so every v3 attempt is rejected; **Log rejected attempts** ✓.

**PayPal Payments → Settings → Troubleshooting:** leave **Logging off**. The store must be fully connected — logging is force-enabled during onboarding, which masks the bug.

1. Check out with PayPal or the card gateway and submit → "CAPTCHA verification failed" notice.
2. WooCommerce → Status → Logs, source `woocommerce-paypal-payments-recaptcha`: one entry per attempt, `Rejected by v3 reCAPTCHA at … with score …`. On `dev/develop` that log source does not exist at all, while the settings-page counter still increments — that is the bug.
3. Untick **Log rejected attempts** → counter still increments, nothing logged.
4. Turn plugin-wide **Logging on** → rejection entries still go only to the reCAPTCHA source, not duplicated into `woocommerce-paypal-payments`.

A successful verification is cached for 300s and reused up to 5 times, so clear the `ppcp_recaptcha_result_*` transients if you test a passing case first. Only v3 rejections are logged.

### Changelog Entry

> Fix: the reCAPTCHA "Log rejected attempts" setting produced no logs unless the plugin-wide logging setting was enabled as well.

### DIFF
--- a/modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
+++ b/modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
@@ -39,6 +39,13 @@ class Recaptcha {
 
 	private LoggerInterface $logger;
 
+	/**
+	 * Logs the rejected attempts, independently of the plugin-wide logging setting.
+	 *
+	 * @var LoggerInterface
+	 */
+	private LoggerInterface $rejection_logger;
+
 	private PersistentCounter $rejection_counter;
 
 	private SettingsStatus $settings_status;
@@ -53,6 +60,7 @@ class Recaptcha {
 	 * @param LoggerInterface      $logger
 	 * @param PersistentCounter    $rejection_counter
 	 * @param SettingsStatus       $settings_status
+	 * @param LoggerInterface      $rejection_logger Writes to the rejection log source.
 	 */
 	public function __construct(
 		RecaptchaIntegration $integration,
@@ -61,7 +69,8 @@ class Recaptcha {
 		string $asset_version,
 		LoggerInterface $logger,
 		PersistentCounter $rejection_counter,
-		SettingsStatus $settings_status
+		SettingsStatus $settings_status,
+		LoggerInterface $rejection_logger
 	) {
 
 		$this->integration       = $integration;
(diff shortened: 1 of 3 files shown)


modules/ppcp-fraud-protection/src/Recaptcha/Recaptcha.php
```diff
@@ -39,29 +39,28 @@ class Recaptcha {
 
 	private LoggerInterface $logger;
 
+	/**
+	 * Logs the rejected attempts, independently of the plugin-wide logging setting.
+	 *
+	 * @var LoggerInterface
+	 */
+	private LoggerInterface $rejection_logger;
```
AlexP11223, line 47:
```suggestion
	/**
	 * Logs the rejected attempts, independently of the plugin-wide logging setting.
	 */
	private LoggerInterface $rejection_logger;
```

AlexP11223 (approved):

PR #4620, opened 2026-08-18
Implement Apple Pay and Google Pay in JS SDK v6 (5782)

stracker-phil:
*Changelog:* `(none - behind the disabled sdk_v6 feature flag)`

# Description

## 🎯 The goal

Add **Apple Pay and Google Pay** to the v6 SDK stack, which so far had only express buttons and card fields. Both wallets now use the v6 session API and work on product page, cart, checkout, and mini-cart.

Both wallets are described once in `walletRegistry.js` (and `WalletPlacement` on the PHP side) and driven from there, so only `applePay.js` and `googlePay.js` hold wallet-specific behavior. Placement on classic checkout is decided centrally in `gatewayPlacement.js`, because the wallet rows share "Place order" and the express wrapper.

This targets the epic branch, not `develop`. Detailed review happens before the epic merges.

## 🖼️ Visual proof

| Apple Pay | Google Pay |
|---|---|
| 2026-08-18 - Apple Pay | 2026-08-18 - Google Pay |


modules/ppcp-sdk-v6/resources/js/wallets/applePay.js
```diff
@@ -0,0 +1,293 @@
...
+		// between the sheet closing and payWithWallet's redirect.
+		spinner?.block();
+
+		try {
+			// This is what adds a viewed product to the real cart. Its total is
+			// discarded: the shopper must be charged what the sheet showed them.
+			const { purchaseUnits } = await resolveWalletTotal(
+				config,
+				context
+			);
```
Narek13, line 206: **The sheet total and the charged amount are two independent numbers, never compared.**

The comment says the resolved total is discarded so the shopper is charged what the sheet showed. It is the other way round: the sheet was opened with `sheetTotal.get()` (line 107), and what prices the order is `purchaseUnits` from this call  from `changeCart()` against the live product form on a product page, or from the WC cart when `purchaseUnits` is `[]`. Nothing reconciles them, so any drift between the cached sheet value and the cart at authorisation time is charged silently.

Google Pay avoids this structurally by resolving the total *before* `loadPaymentData`, which makes the asymmetry look accidental.

Cheapest fix that preserves the comment's promise:

```js
const { total, purchaseUnits } = await resolveWalletTotal( config, context );

if ( total && total !== sheetTotal.get() ) {
    throw new Error( 'The order total changed after the sheet opened.' );
}
```

The `throw` lands in the existing `catch`, which already reports `STATUS_FAILURE` to Apple, so the buyer is told the payment failed instead of being charged an amount they never saw.

modules/ppcp-sdk-v6/resources/js/wallets/applePay.js
```diff
@@ -0,0 +1,293 @@
...
+		// the button could never open a second sheet.
+		appleSession.oncancel = () => {
+			paying = false;
+			spinner?.unblock();
+			refreshCartUi( context );
+		};
+
+		// Presents the sheet, and only then asks for merchant validation.
+		appleSession.begin();
+	}
```
Narek13, line 155: **A throw here kills the button for the rest of the page, silently.**

`paying = true` on line 116, then `new window.ApplePaySession(...)` and `begin()` on line 145, with no `try`/`catch` anywhere in `pay()`. Both throw synchronously — consumed user gesture, malformed request, a currency or `supportedNetworks` Apple rejects. `paying` is never reset, so every later tap returns at the guard on line 103: the button does nothing, no error is shown, and only a reload recovers.

`googlePay.js` gets this right with a `finally`. Wrapping the constructor through `begin()` in a `try`/`catch` that resets `paying`, unblocks the spinner and calls `handleError` would close it. No test covers a throwing constructor.

modules/ppcp-sdk-v6/resources/js/wallets/googlePay.js
```diff
@@ -0,0 +1,167 @@
...
+		}
+		paying = true;
+
+		try {
+			// Resolved before the sheet opens: the sheet must display the same
+			// total that createOrder() then charges.
+			const { total, purchaseUnits } = await resolveWalletTotal(
+				config,
+				context
+			);
```
Narek13: **Opening and dismissing the sheet destroys the buyer's cart.**

On a product page this reaches `changeCart()`, which per its own docblock *"Empties the cart first"* — and it runs before `loadPaymentData` below. The cancel branch treats `CANCELED` as a non-event and restores nothing, nor does it call `refreshCartUi`, so the mini-cart keeps showing the old contents until the next page load.

Buyer with four items taps Google Pay on a product page to check the price, dismisses the sheet, and their cart is now that one product.

I see this is already known — `endpointsAdapter.js:122-125` has a TODO saying Google Pay should read `simulateCart()` instead. Flagging it because it is a data-loss bug the buyer cannot see or undo, so it seems worth pulling ahead of "phase 3" rather than shipping the wallet with it. Apple has the same hole on the failure path, though not on cancel.

modules/ppcp-sdk-v6/resources/js/wallets/googlePay.js
```diff
@@ -0,0 +1,167 @@
...
+
+			const paymentData = await client.loadPaymentData(
+				buildPaymentDataRequest( sessionConfig, {
+					countryCode: config.buyer_country,
+					currencyCode: config.currency,
+					total,
+				} )
+			);
```
Narek13, line 118: **No empty-total guard, unlike the Apple bridge.**

`resolveWalletTotal` returns `''` when the product branch yields no purchase units (`walletTotal.js:31`, and its test pins that), and `''` goes straight into `transactionInfo.totalPrice` with `totalPriceStatus: 'FINAL'` — a raw Google `DEVELOPER_ERROR` toast for the buyer, with the cart already emptied by then. `applePay.js:107-114` refuses in exactly this case ("Refusing beats charging a total the shopper never saw").

```suggestion
			if ( ! total ) {
				// Refusing beats charging a total the shopper never saw, as the
				// Apple bridge already does.
				throw new Error(
					'Google Pay could not determine the order total.'
				);
			}

			const paymentData = await client.loadPaymentData(
				buildPaymentDataRequest( sessionConfig, {
					countryCode: config.buyer_country,
					currencyCode: config.currency,
					total,
				} )
			);
```

modules/ppcp-sdk-v6/resources/js/wallets/googlePay.js
```diff
@@ -0,0 +1,167 @@
...
+			} );
+		} catch ( error ) {
+			// The buyer dismissing the sheet is not a failure to report.
+			if ( error?.statusCode !== 'CANCELED' ) {
+				handleError( error );
+			}
+		} finally {
+			paying = false;
+			spinner?.unblock();
+		}
```
Narek13: **The `finally` re-arms the button while the redirect is still in flight.**

It runs on the success path too, right after `payWithWallet` started a `location.assign` or submitted the checkout form. The overlay disappears, `paying` flips back to `false`, and for the second or two the navigation takes the buyer has a live Google Pay button — a second tap starts a second `createOrder`/`confirmOrder`. `applePay.js:218-219` deliberately does the opposite and keeps the spinner up.

The re-entrancy guard is defeated by its own cleanup. Resetting only on the paths that stay on the page:

```suggestion
		} catch ( error ) {
			// The buyer dismissing the sheet is not a failure to report.
			if ( error?.statusCode !== 'CANCELED' ) {
				handleError( error );
			}

			// Only on the paths that stay on the page: on success payWithWallet
			// has already started the redirect or submitted the form.
			paying = false;
			spinner?.unblock();
		}
```

Note `googlePay.test.js` pins both halves of the current behaviour ("blocks the spinner only once the sheet has closed" asserts the unblock on success; the double-click test asserts a second click is allowed after a *successful* payment), so those two tests need to move with the fix.

modules/ppcp-sdk-v6/resources/js/wallets/googlePayRequest.js
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Shipping on requires the caller to also pass an onPaymentDataChanged
+ * callback, or Google rejects the request. It does not restrict countries via
+ * shippingAddressParameters yet. Shipping off: the address comes off the
+ * payment token instead, see walletContacts.js.
+ *
+ * @param {Object}  sessionConfig                     - The v6 session config, from
+ *                                                      formatConfigForPaymentRequest().
+ * @param {Object}  transaction                       - What the buyer is about to pay.
+ * @param {string}  transaction.countryCode           - The merchant country.
```
Narek13, line 46: **This says merchant country; the caller passes the buyer's.**

`googlePay.js:114` passes `config.buyer_country`, which `SdkV6Manager.php:424-426` defines as `WC()->customer->get_billing_country()` with a base-location fallback. `applePayRequest.js:40` uses `sessionConfig.merchantCountry` for the same concept, so the two wallets disagree about what this field means.

modules/ppcp-sdk-v6/src/Helper/WalletConfig.php
```diff
@@ -0,0 +1,107 @@
...
+	 * Neither wallet can be offered for subscriptions (they have no vaulting).
+	 */
+	private function is_product_supported( string $context ): bool {
+		$contains_subscription = $this->subscription_helper->locations_with_subscription_product();
+
+		// The product page is judged by what it displays; every other page by the
+		// cart contents.
+		$location = 'product' === $context ? 'product' : 'cart';
+
+		return false === ( $contains_subscription[ $location ] ?? false );
```
Narek13: **Renewals slip through this gate, so wallets render on subscription renewals.**

`locations_with_subscription_product()` returns three keys, and `cart` is deliberately `cart_contains_subscription() && ! $cart_contains_renewal`, while `payorder` is the key that carries `$cart_contains_renewal`. Mapping every non-product context onto `cart` therefore reads `false` for a renewal: `should_render( 'pay-now' )` returns true, `wallet_script_data()` emits `styles['pay-now']`, and the button renders on the pay-for-order page of a renewal — the case this class's docblock says must be blocked, since neither wallet vaults.

v5 checks `payorder` explicitly, and `script_data()` never exposes the array, so the JS cannot compensate.

```suggestion
		// The product page is judged by what it displays.
		if ( 'product' === $context ) {
			return false === ( $contains_subscription['product'] ?? false );
		}

		// Every other page by the cart, plus payorder: it is the only key that
		// flags a renewal, which `cart` deliberately excludes.
		return false === ( $contains_subscription['cart'] ?? false )
			&& false === ( $contains_subscription['payorder'] ?? false );
```

`ApplePayConfigTest` / `GooglePayConfigTest` hard-code `'payorder' => false` and never pass `'pay-now'`, which is why this is green.

modules/ppcp-sdk-v6/src/Endpoint/SimulateCartEndpoint.php
```diff
@@ -0,0 +1,87 @@
...
+		if ( ! apply_filters( 'woocommerce_paypal_payments_simulate_cart_enabled', true ) ) {
+			wp_send_json_error(
+				array(
+					'name'    => '',
+					'message' => 'Cart simulation is disabled.',
+					'code'    => 0,
+					'details' => array(),
+				)
+			);
+		}
```
Narek13, line 67: **Missing `return`, so the kill-switch can be bypassed** and the filter needs its `@param`.

`wp_send_json_error()` ends in `wp_die()` in production, but `wp_die_ajax_handler` is a filter; where something makes it non-exiting, execution falls through to product parsing and two JSON bodies are emitted. The PR's own test documents the fall-through.

```suggestion
		/**
		 * The filter that switches cart simulation off, honoured here too so a
		 * merchant who disabled it does not get it back through this endpoint.
		 *
		 * @param bool $enabled Whether cart simulation is enabled.
		 */
		if ( ! apply_filters( 'woocommerce_paypal_payments_simulate_cart_enabled', true ) ) {
			wp_send_json_error(
				array(
					'name'    => '',
					'message' => 'Cart simulation is disabled.',
					'code'    => 0,
					'details' => array(),
				)
			);

			return;
		}
```

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -175,12 +235,104 @@ public function render_wrapper(): void {
...
+	 * buyer can pay, because eligibility is only knowable client-side: a browser
+	 * without a saved card, or without Apple Pay set up, would otherwise be
+	 * offered a row that cannot complete. Same attribute shape as v5's
+	 * GooglePayButton, which the v6 JS removes the same way its PaymentButton did.
+	 */
+	public function render_wallet_gateway_wrappers(): void {
+		foreach ( $this->wallets as $wallet ) {
+			if ( ! $this->is_wallet_gateway( $wallet ) ) {
+				continue;
+			}
```
Narek13: **The row is hidden under a weaker condition than the one that reveals it, so it can stay hidden with no button.**

`is_wallet_gateway()` checks only "classic checkout + gateway available". `revealWalletGateway()` runs on the JS render path, which additionally needs `styles['checkout']`, i.e. `should_render( 'checkout' )` — so the two differ by availability and the subscription gate. With a subscription in the cart (or via the `payorder` gap in `WalletConfig`), PHP hides the Apple Pay row, the JS never boots, and the row is gone for the rest of the page with no other way to select that gateway. v5 prints its equivalent style only behind the full eligibility gate.

```suggestion
		foreach ( $this->wallets as $wallet ) {
			// Both conditions: the row must only be hidden when the JS will also
			// reveal it, and revealWalletGateway() runs on the render path, which
			// needs should_render( 'checkout' ) as well.
			if ( ! $this->is_wallet_gateway( $wallet )
				|| ! $wallet->config->should_render( 'checkout' ) ) {
				continue;
			}
```

Narek13 (commented):
Thanks a lot @stracker-phil for working on this huge PR. I posted some doubts which I found, but could be false positives since I wasn’t able to test the behavior manually. If you think everythingis ok, I’ll be happy to approve it.
Also Nothing covers the new SdkV6Manager behaviour:
- `render_wallet_gateway_wrappers()` output
-  `wallet_script_data()` shape
- `should_load_on_current_page()` wallet branches.

modules/ppcp-sdk-v6/src/Endpoint/SimulateCartEndpoint.php
stracker-phil: Ignoring this. Though technically correct, this is a non-issue for us:

If a WooCommerce shop really implements a filter that does not terminate after `wp_die()`then this endpoint is the smallest problem they have.

I'd treat every `wp_die()` call as a contract that guarantees the request to terminate. The documented fall-through is a test-environment-issue (since the stub does not terminate the test suite)

modules/ppcp-sdk-v6/src/Helper/WalletConfig.php
stracker-phil: Looks correct, thanks for the suggestion. Updated the PR

modules/ppcp-sdk-v6/resources/js/wallets/googlePayRequest.js
stracker-phil: Thanks, this would have been a tricky bug to track!

modules/ppcp-sdk-v6/resources/js/wallets/applePay.js
stracker-phil: Looks like a valuable addition to harden the payment method. thanks for pointing it out - added to the PR

modules/ppcp-sdk-v6/resources/js/wallets/googlePay.js
stracker-phil: This looks like a theoretical problem; I could not find a branch that really provides an empty total.

Apple Pay and Google Pay are different in this respect: While GP fetches the total async/delayed, AP needs the total to be present at page load, which is why it has a different protection here. Claude agrees: "The asymmetry is intentional here"

modules/ppcp-sdk-v6/resources/js/wallets/applePay.js
stracker-phil: Looks very theoretical, and I do not think we should throw here. The totals will be identical, though the formatting or data type can be different while the amount being equal, eg "10" vs "10.00" should never throw

This asymmetry is intentional and one of the current few structural differences between Google Pay and Apple Pay.

Narek13 (approved):
Thanks @stracker-phil for providing the explainations and for implementing the changes.
LGTM

PR #4622, opened 2026-08-20
Add the BCDC card button to the SDK v6 module (6783) 

stracker-phil:
*Changelog:* (none - part of the in-progress v6 SDK migration; the parent branch carries the user-facing entry)

# Description

The black **Basic Card** button (BCDC) was the last classic-checkout payment row still on v5. This branch brings it over to the v6 SDK, where PayPal exposes it as *Guest Payments*: its own SDK component, session factory and custom elements, so it gets its own renderer rather than joining the shared express-button stack.

Supporting changes that fell out of it:

- Order approval now names the paying gateway instead of assuming `ppcp-gateway`, so a BCDC payment reaches `CardButtonGateway`.
- Recoverable declines surface as a notice through the guest session's `onWarn`, leaving the cart and the button intact.
- The card-field styles are split into box styles (applied on our own page) and text styles (handed to the SDK, inside a PayPal iframe), with a filter for merchant overrides of the latter.

## 🔍 What to review

- **BCDC row visibility** is decided by the gateway list, not by the express-button location settings, and is deliberately suppressed for subscription carts: the v6 guest component has no vaulting equivalent, so `CardButtonGateway`'s hosted-card fallback takes over there.

## 🧪 How to verify

1. With standard card payments enabled, complete a BCDC payment on classic checkout and on the pay-for-order page; confirm the order is attributed to the card-button gateway.
2. Put a subscription in the cart: no card button, "Place order" remains.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/cardButton/renderCardButton.js
+++ b/modules/ppcp-sdk-v6/resources/js/cardButton/renderCardButton.js
@@ -1,0 +1,141 @@
+/**
+ * Renders the Basic Card button (BCDC) as its own payment-method row.
+ *
+ * Kept out of components/buttonRenderer.js, which draws into the shared express
+ * wrapper: separate files make "the card button never joins the express stack"
+ * true by construction rather than by a condition someone can relax.
+ *
+ * @package
+ */
+
+import { createOrder } from '../endpointsAdapter';
+import { handleError } from '../utils/errorHandler';
+import { FundingSources } from '../utils/fundingSources';
+import { revealWalletGateway } from '../wallets/gatewayPlacement';
+
+/**
+ * The event the SDK's card button dispatches on click. Documented contract; a
+ * plain 'click' listener only works by way of the element's shadow internals.
+ */
+const CLICK_EVENT = 'bcdc-click';
+
+/**
+ * Builds the button element and its required container, detached.
+ *
+ * The button checks on connect that its parent is a
+ * <paypal-basic-card-container>, so the tree must be assembled before any of it
+ * enters the document.
+ *
+ * @param {Object}  styles         - Style config from the card_button subtree.
+ * @param {?string} [buyerCountry] - ISO 3166-1 alpha-2 buyer country.
+ * @return {{container: HTMLElement, button: HTMLElement}} The detached pair.
+ */
+function createCardButtonElements( styles, buyerCountry ) {
+	const container = document.createElement( 'paypal-basic-card-container' );
+	const button = document.createElement( 'paypal-basic-card-button' );
+
+	if ( buyerCountry ) {
+		button.buyerCountry = buyerCountry;
+	}
+
+	if ( styles?.borderRadius ) {
+		button.style.setProperty(
+			'--paypal-button-border-radius',
+			styles.borderRadius
+		);
+	}
+
+	if ( styles?.height ) {
+		button.style.height = styles.height;
+	}
+
+	// The element ships a fixed 225px width, where v5's card button spanned the
+	// checkout column. An outer declaration wins over its own :host rule.
+	if ( styles?.width ) {
+		container.style.width = styles.width;
+		button.style.width = styles.width;
+	}
+
+	container.appendChild( button );
+
+	return { container, button };
+}
+
+/**
+ * Renders the card button into its gateway row, if it belongs on this page.
+ *
+ * Idempotent, because it runs again on every checkout update: WooCommerce
+ * replaces the whole order-review DOM, which takes our container with it.
+ *
+ * @param {Object}   config         - The wc_ppcp_sdk_v6 config object.
+ * @param {Function} ensureSessions - Resolves the sessions for a context.
+ * @return {Promise<void>} Resolves once the pass is done.
+ */
+export async function initCardButton( config, ensureSessions ) {
+	if ( ! config.card_button?.enabled ) {
+		return;
+	}
+
+	const findEmptyWrapper = () => {
+		const el = document.querySelector( config.card_button.wrapper );
+		return el && el.childElementCount === 0 ? el : null;
+	};
+
+	if ( ! findEmptyWrapper() ) {
+		return;
+	}
+
+	const context = config.page_context;
+	const { map } = await ensureSessions( context );
+
+	// Asked again after the await, never only before it: the initial render and
+	// the updated_checkout that follows it overlap, and a check that straddles
+	// the await lets both through, appending the button twice.
+	const target = findEmptyWrapper();
+	if ( ! target ) {
+		return;
+	}
+
+	// No session means the buyer is not eligible for cards; the row stays hidden
+	// and "Place order" stays visible, so the checkout still offers a way to pay.
+	const session = map[ FundingSources.CARD ];
+	if ( ! session ) {
+		return;
+	}
+
+	const { container, button } = createCardButtonElements(
+		config.card_button.styles,
+		config.buyer_country
+	);
+	target.appendChild( container );
+
+	button.addEventListener( CLICK_EVENT, async () => {
+		try {
+			await session.start(
+				// Naming the target beats the SDK's "last clicked button"
+				// fallback, ambient state on a page that also carries the
+				// express buttons. 'auto' is the only mode this session takes.
+				{ presentationMode: 'auto', targetElement: button },
+				createOrder(
+					config,
+					context,
+					FundingSources.CARD,
+					undefined,
+					config.card_button.payment_method
+				)
+			);
+		} catch ( error ) {
+			handleError( error );
+		}
+	} );
+
+	// After insertion, so the row counts as rendered when the placement logic
+	// decides what to do with "Place order".
+	revealWalletGateway(
+		{
+			id: config.card_button.payment_method,
+			wrapper: config.card_button.wrapper,
+		},
+		config
+	);
+}
--- a/modules/ppcp-sdk-v6/resources/js/cardFields/cardFieldStyles.js
+++ b/modules/ppcp-sdk-v6/resources/js/cardFields/cardFieldStyles.js
@@ -113,9 +113,10 @@ export function cardFieldStyles( field ) {
  * The text-only style set for the SDK's `style.input`, which reaches inside a
  * PayPal-hosted field.
  *
- * @param {HTMLElement} field - The existing WC input being replaced.
+ * @param {HTMLElement} field       - The existing WC input being replaced.
+ * @param {Object}      [overrides] - Merchant overrides from card_fields.styles.
  * @return {Object} The style object.
  */
-export function hostedFieldTextStyles( field ) {
-	return pickComputed( field, TEXT_PROPERTIES );
+export function hostedFieldTextStyles( field, overrides ) {
+	return { ...pickComputed( field, TEXT_PROPERTIES ), ...overrides };
 }
--- a/modules/ppcp-sdk-v6/resources/js/cardFields/cardFieldStyles.js
+++ b/modules/ppcp-sdk-v6/resources/js/cardFields/cardFieldStyles.js
@@ -1,15 +1,31 @@
 /**
- * Computes the `style.input` object for the v6 card-fields component.
+ * Computes the style objects for the v6 card-fields component.
  *
- * Unlike v5's paypal.CardFields() (see CardFieldsHelper.js), the v6 SDK
- * takes camelCase JS style properties (fontSize, not font-family) and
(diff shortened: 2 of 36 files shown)


modules/ppcp-sdk-v6/resources/js/cardButton/renderCardButton.js
```diff
@@ -0,0 +1,141 @@
...
+					context,
+					FundingSources.CARD,
+					undefined,
+					config.card_button.payment_method
+				)
+			);
+		} catch ( error ) {
+			handleError( error );
+		}
+	} );
```
Narek13: **Two issues in this handler.**

**1. The click stores an unapproved order in the WC session.** The generic `createOrder()` always sends `save_order_in_session: 1` (`endpointsAdapter.js:172`), which `SdkV6Module` turns into `session_handler->replace_order()` — at click time, before card data exists and before approval. The card-fields path deliberately avoids exactly this, and says why at `endpointsAdapter.js:357`.

It matters more for BCDC than for ACDC because the documented fallback here is `CardButtonGateway::process_payment()` via the visible "Place order", and `OrderProcessor::order_is_ready_for_process()` treats a `CREATED` order as processable. Buyer opens the card sheet, closes it (`onCancel` → `refreshCartUi` is a no-op outside the `product` context, so nothing clears the session), then clicks "Place order" → capture attempted against an unapproved order. The `PayPalOrderMissingException` → hosted-card fallback never fires, because the session is not empty.

**2. No re-entrancy guard.** `googlePay.js` carries a `paying` flag whose comment reads *"Google fires onClick per tap, so a double tap opens two sheets"*. There is none here, and the `createOrder()` promise is built **eagerly** at line 119, so a double tap fires two order-creation POSTs regardless of what `session.start()` does — two PayPal orders for one cart, and the second `start()` typically rejects over a working card sheet.

The guard is self-contained:

```suggestion
	let paying = false;

	button.addEventListener( CLICK_EVENT, async () => {
		// One event per tap: without this a double tap starts two
		// createOrder() POSTs, since the promise below is built eagerly.
		if ( paying ) {
			return;
		}
		paying = true;

		try {
			await session.start(
				// Naming the target beats the SDK's "last clicked button"
				// fallback, ambient state on a page that also carries the
				// express buttons. 'auto' is the only mode this session takes.
				{ presentationMode: 'auto', targetElement: button },
				createOrder(
					config,
					context,
					FundingSources.CARD,
					undefined,
					config.card_button.payment_method
				)
			);
		} catch ( error ) {
			handleError( error );
		} finally {
			paying = false;
		}
	} );
```

Note I have mirrored `googlePay.js`'s `finally` deliberately — but that release-on-success is the same thing I flagged on #4620 (the button re-arms while the redirect is in flight). Whatever you settle on there should apply here too. A spinner would also match the sibling flows; `Spinner` is not imported in this file today.

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -399,6 +415,55 @@ public function is_card_fields_enabled( ?string $location = null ): bool {
...
+		}
+
+		// No row for subscription carts: the v6 guest component has no
+		// equivalent of the SDK URL vault param v5 uses here, so the button
+		// would take a payment that can never renew. Not a dead end — without a
+		// button "Place order" stays visible, and CardButtonGateway falls back
+		// to PayPal's hosted card checkout.
+		if ( $this->subscription_helper->cart_contains_subscription() ) {
+			return false;
+		}
```
Narek13, line 529: **The subscription suppression does not hold on pay-for-order**, which is one of the two contexts `is_card_button_enabled()` admits (`:429`).

There the cart is empty, so `cart_contains_subscription()` is always false and the row prints. WC Subscriptions deliberately does not filter gateways on that screen, and `CardButtonGateway::is_available()` only refuses free-trial carts — so nothing else stops it.

Reachable path: a subscription's initial payment fails → the customer follows the "pay now" link in the WooCommerce email → the guest button takes the payment with no vaulted token, and renewals later fail. Which is precisely what the comment above says must not happen.

The companion helper exists and is paired with `cart_contains_subscription()` elsewhere in the codebase:

```suggestion
		// No row for subscription carts: the v6 guest component has no
		// equivalent of the SDK URL vault param v5 uses here, so the button
		// would take a payment that can never renew. Not a dead end — without a
		// button "Place order" stays visible, and CardButtonGateway falls back
		// to PayPal's hosted card checkout. order_pay_contains_subscription()
		// covers pay-for-order, where the cart is empty.
		if ( $this->subscription_helper->cart_contains_subscription()
			|| $this->subscription_helper->order_pay_contains_subscription() ) {
			return false;
		}
```

No PHPUnit case covers BCDC on pay-now at all, which is what hides this — the only subscription case is a cart.

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -399,6 +415,55 @@ public function is_card_fields_enabled( ?string $location = null ): bool {
...
+		return in_array( $location, array( 'checkout', 'pay-now' ), true )
+			&& $this->card_payments_configuration->is_bcdc_enabled();
+	}
+
+	/**
+	 * Whether the BCDC button renders as its own payment-method row here.
+	 *
+	 * Asks the gateway list rather than re-deriving its policy, which already
+	 * covers the checkout button location being off, ACDC outside Mexico,
+	 * free-trial carts and zero-total carts.
```
Narek13: **Zero-total carts are not covered by the gateway list**, contrary to this docblock. `CardButtonGateway::is_available()` refuses only *free-trial* carts, via `is_free_trial_cart()`, which requires a subscription product — a 100%-off coupon on an ordinary product leaves the gateway available.

Because the BCDC hooks are registered outside the `determine_render_places()` branches, and `woocommerce_review_order_after_payment` fires even when WooCommerce skips the payment-method list, a $0 checkout gets a black card button under "Place order" for an order that needs no payment. `cart_needs_payment()` (and `$order->needs_payment()` for pay-now) would close it.

Minor, same function: the `is_block_context()` guard on line 445 is unreachable — `is_card_button_enabled()` already restricts to `checkout|pay-now`, and block contexts are `cart-block|checkout-block`. It mirrors the same redundancy in `is_wallet_gateway()`.

modules/ppcp-sdk-v6/src/Helper/CardFieldStyles.php
```diff
@@ -0,0 +1,53 @@
...
+
+class CardFieldStyles {
+
+	/**
+	 * The styles a merchant wants applied to the text inside the hosted fields.
+	 *
+	 * A card field's own box sits on the merchant's page and is stylable with
+	 * plain CSS through the .ppcp-sdk-v6-card-field class. Its text is not: that
+	 * lives in a PayPal-hosted iframe CSS cannot reach across, so the only way
+	 * in is the SDK's style.input object, which has to travel as script data.
```
Narek13: **The CSS this promises is not in the branch.** I checked at this head: `modules/ppcp-sdk-v6/resources/` contains only `js`, the module's webpack entry lists no stylesheet, and the only references to `.ppcp-sdk-v6-card-field` are the two lines in `mountField.js` that add the class plus this docblock.

So the box styling has no owner. `mountField.js` hands the SDK `hostedFieldTextStyles()` (no `background`, `border`, `borderRadius`, `boxShadow`) and applies only width/height itself, and `cardFieldStyles()` — the full set — is no longer imported by either renderer. Those four properties used to reach the field through `style.input`.

Result on a default theme: card number, expiry and CVV render borderless and transparent between still-bordered WC billing inputs and the still-bordered cardholder-name input, which does get the box set (`V6CardFieldsComponent.js:294`). Same on classic and blocks.

Either ship the stylesheet or apply the box subset to the container element the plugin owns.

modules/ppcp-sdk-v6/resources/js/cardFields/cardFieldStyles.js
```diff
@@ -46,42 +62,61 @@ const ALLOWED_PROPERTIES = [
...
+ * The text-only style set for the SDK's `style.input`, which reaches inside a
+ * PayPal-hosted field.
+ *
+ * @param {HTMLElement} field       - The existing WC input being replaced.
+ * @param {Object}      [overrides] - Merchant overrides from card_fields.styles.
+ * @return {Object} The style object.
+ */
+export function hostedFieldTextStyles( field, overrides ) {
+	return { ...pickComputed( field, TEXT_PROPERTIES ), ...overrides };
+}
```
Narek13: **Merchant overrides bypass the allow-list**, because they are spread last. A filter returning `border`, `background`, `borderRadius` or `boxShadow` goes straight into `style.input` — re-injecting into PayPal's iframe exactly the box decoration the `BOX_PROPERTIES` split just removed, and the SDK accepts those, so it paints a frame the theme never drew. `CardFieldStyles::overrides()` only validates that keys and values are strings.

```suggestion
export function hostedFieldTextStyles( field, overrides ) {
	// Overrides go through the same allow-list as the computed styles:
	// a box property here would paint a frame inside PayPal's iframe that
	// the merchant's own page never drew.
	const allowed = Object.fromEntries(
		Object.entries( overrides || {} ).filter( ( [ property ] ) =>
			TEXT_PROPERTIES.includes( property )
		)
	);

	return { ...pickComputed( field, TEXT_PROPERTIES ), ...allowed };
}
```

(Assumes `TEXT_PROPERTIES` is an array — adjust if it is a map.) Worth logging the rejected keys too, since a silently dropped override is hard for a merchant to diagnose.

modules/ppcp-sdk-v6/resources/js/utils/errorHandler.js
```diff
@@ -67,3 +67,32 @@ export function handleError( error ) {
...
+
+	const wrapper =
+		document.querySelector( '.woocommerce-notices-wrapper' ) ||
+		document.querySelector( '.woocommerce' ) ||
+		document.body;
+
+	new ErrorHandler( errorLabels.generic_error || '', wrapper ).message(
+		errorLabels.card_declined
+	);
+}
```
Narek13: **Every warning is reported as a card decline.** `warning.code` is read for nothing but the `console.warn`, and the handler then always renders `errorLabels.card_declined` — while the docblock above says it covers *"a card decline or a validation failure"*. A buyer who mistypes an expiry date is told their card was declined.

Switching on the code (with the decline label as the specific case and the generic error as the fallback) would fix it without changing the deliberate no-clear/no-refresh behaviour, which reads as the right call.

Two smaller things while you are here:

- No test covers a non-decline code, so the mislabelling is invisible to CI — and `errorHandler.test.js`'s "leaves existing notices alone" case pins the stacking, so repeated declines produce repeated identical bullets.
- The notice renders into `.woocommerce-notices-wrapper`, i.e. outside the SDK's card form. If the premise in the docblock is right and the buyer is still inside that form, the message may be behind PayPal's own surface — worth eyeballing on a real decline.

modules/ppcp-sdk-v6/src/SdkV6Module.php
```diff
@@ -284,6 +284,30 @@ static function () use ( $manager ): void {
...
+		// gateway-availability question. DisableGateways couples the two today,
+		// but relying on that coincidence would break once it stops being true.
+		/**
+		 * The action name that the PayPal buttons use for rendering on the checkout page.
+		 * Shared with the v5 SmartButton so a single override relocates both stacks.
+		 */
+		$hook = (string) apply_filters(
+			'woocommerce_paypal_payments_checkout_button_renderer_hook',
+			'woocommerce_review_order_after_payment'
+		);
```
Narek13: **These renderer-hook filters are now applied twice per request** — `woocommerce_paypal_payments_checkout_button_renderer_hook` at `:259` and again here, and `..._pay_order_renderer_hook` at `:281` and `:306`.

Any third-party callback that is not pure (a counter, logging, or one that calls `remove_filter` on first invocation) returns a different hook the second time, and the express wrapper and the BCDC row then land on different hooks — the "row on one hook, button on another" split this placement logic is trying to avoid.

Resolving each hook name once above and reusing the variable for both `add_action()` calls removes the hazard entirely.

Also: the two new `apply_filters()` docblocks here and at `:301-304` describe the hook but omit the `@param` for the filtered value, which `.claude/docs/code-quality.md` requires per argument. The new `CardFieldStyles.php` one gets this right.

Narek13 (approved):
Thanks for working on this @stracker-phil 
1 doubt related to subscriptions which desereves your attention.
Everything else looks good to me, thank you 💯

PR #4625, opened 2026-08-21
WC Subscriptions Integration (5788)

Dinamiko:
Brings the SDK v6 checkout to v5 parity for subscription vaulting and returning‑buyer saved PayPal, on classic and Blocks.                                                                 
                                                                                                                                                                                             
  - Free‑trial ($0) subscriptions: use the Vault v3 save‑without‑purchase flow instead of a $0 order (new sessions/freeTrialSave.js, wired into Blocks, classic, and pay‑for‑order); express 
    funding limited to PayPal on those carts.                                                                                                                                                
  - Regular subscriptions: PayPal/Venmo and card orders vault for renewals via the existing flow.                                                                                            
  - "Save to account": auto‑checked and disabled on subscription card carts (classic + Blocks), matching v5.                                                                                 
  - Saved‑PayPal vault component restored under v6 — classic (standalone renderer in ppcp-vault-component) and Blocks (ppcp-gateway saved‑token method reusing the existing components);     
    boot.js shows "Place order" for a selected saved token.                                                                                                                                  
                                                                                                                                                                                             
Renewals unchanged (server‑side). Native PayPal Subscriptions out of scope, will be handled on next JS SDK v6 iteration.

### DIFF
--- a/modules/ppcp-blocks/resources/js/paypal-config.js
+++ b/modules/ppcp-blocks/resources/js/paypal-config.js
@@ -273,12 +273,13 @@ export const handleApproveSubscription = async (
 			throw new Error( config.scriptData.labels.error.generic );
 		}
 
-		// A native PayPal subscription is already created and approved at this point,
-		// so there is no separate order to review or capture: complete the checkout
-		// directly instead of routing through the order-review continuation. This
-		// matches the classic checkout, whose subscription onApprove submits the
-		// order without a final-review step, and avoids leaving the buyer on the
-		// continuation page when final review is enabled.
+		// Native PayPal subscriptions always complete directly here, bypassing the
+		// order-review continuation. Unlike a one-time order, the subscription is
+		// already created and approved at this point and there is no capturable order
+		// to review, so routing through the continuation leaves the buyer unable to
+		// finish. This intentionally overrides "Require final confirmation" for native
+		// subscriptions on block checkout, matching the classic checkout (whose
+		// subscription onApprove submits the order without a final-review step).
 		setGotoContinuationOnError( true );
 		onSubmit();
 	} catch ( err ) {
--- a/modules/ppcp-blocks/resources/js/paypal-config.js
+++ b/modules/ppcp-blocks/resources/js/paypal-config.js
@@ -273,12 +273,14 @@ export const handleApproveSubscription = async (
 			throw new Error( config.scriptData.labels.error.generic );
 		}
 
-		if ( ! shouldskipFinalConfirmation() ) {
-			location.href = getCheckoutRedirectUrl();
-		} else {
-			setGotoContinuationOnError( true );
-			onSubmit();
-		}
+		// A native PayPal subscription is already created and approved at this point,
+		// so there is no separate order to review or capture: complete the checkout
+		// directly instead of routing through the order-review continuation. This
+		// matches the classic checkout, whose subscription onApprove submits the
+		// order without a final-review step, and avoids leaving the buyer on the
+		// continuation page when final review is enabled.
+		setGotoContinuationOnError( true );
+		onSubmit();
 	} catch ( err ) {
 		console.error( err );
 
--- a/modules/ppcp-button/src/Endpoint/ApproveSubscriptionEndpoint.php
+++ b/modules/ppcp-button/src/Endpoint/ApproveSubscriptionEndpoint.php
@@ -243,10 +243,15 @@ class ApproveSubscriptionEndpoint implements EndpointInterface {
 
 		// A native subscription's related order does not reliably expose a resolved
 		// payer id at approval time, so an absent id on either side is treated as
-		// "cannot compare" rather than a failure — the subscription is already bound
-		// to this session by validate_subscription() (status + custom_id ownership +
-		// plan match). Only a present, differing payer (an order approved by another
-		// account) is rejected.
+		// "cannot compare" rather than a failure. Only a present, differing payer (an
+		// order approved by another account) is rejected.
+		//
+		// Security note: this binding is best-effort. The subscription itself is
+		// session-bound by validate_subscription() (status + custom_id ownership +
+		// plan match), but the order is bound to the subscriber only when both payer
+		// ids resolve here — validate_custom_id_ownership() returns early for an order
+		// whose custom_id lacks the CustomIds::CUSTOMER_ID_PREFIX, so such an order
+		// with an unresolved payer is not independently tied to the subscriber.
 		if ( '' === $subscriber_payer_id || '' === $order_payer_id ) {
 			return;
 		}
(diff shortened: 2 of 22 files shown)


modules/ppcp-button/src/Endpoint/ApproveSubscriptionEndpoint.php
```diff
@@ -241,7 +241,17 @@ private function validate_order_belongs_to_subscriber( Order $order, \stdClass $
 		$payer               = $order->payer();
 		$order_payer_id      = $payer ? $payer->payer_id() : '';
 
-		if ( '' === $subscriber_payer_id || '' === $order_payer_id || $order_payer_id !== $subscriber_payer_id ) {
+		// A native subscription's related order does not reliably expose a resolved
+		// payer id at approval time, so an absent id on either side is treated as
+		// "cannot compare" rather than a failure — the subscription is already bound
```
Narek13: This relaxes the 4.1.0 hardening for **all merchants** — `ppcp-button` isn't behind `PCP_SDK_V6_ENABLED`, and the PR description scopes native PayPal Subscriptions out.

The "already bound by `validate_subscription()`" rationale is best-effort rather than guaranteed: `validate_custom_id_ownership()` returns early when the custom_id lacks the `pcp_customer_` prefix. So an order with no prefix *and* an unresolved payer now reaches `replace_order()` with nothing tying it to the subscriber.

I think the change itself is correct — the false positive is real, I hit this exact error on a native-subscription cart purchase this week. But could it land as its own PR/ticket with a security note, and the comment soften "is already bound" to "is bound on a best-effort basis"?

modules/ppcp-blocks/resources/js/paypal-config.js
```diff
@@ -273,12 +273,14 @@ export const handleApproveSubscription = async (
 			throw new Error( config.scriptData.labels.error.generic );
 		}
 
-		if ( ! shouldskipFinalConfirmation() ) {
-			location.href = getCheckoutRedirectUrl();
-		} else {
-			setGotoContinuationOnError( true );
-			onSubmit();
-		}
+		// A native PayPal subscription is already created and approved at this point,
```
Narek13: Same concern: `ppcp-blocks` isn't feature-flagged, so this ships to everyone. Merchants with "Require final confirmation on checkout" enabled silently lose the review step for native PayPal subscriptions on block checkout.

The classic-parity argument holds for the checkout page (`CheckoutActionHandler` does click `#place_order` directly), but this overrides an explicit merchant setting. If intended, worth its own ticket — native PayPal Subscriptions are scoped out of this PR.

modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -217,6 +238,26 @@ public function enqueue(): void {
...
+				'checkout'  => false,
+				'pay-now'   => false,
+				'mini-cart' => false,
+			);
+		}
+
+		// Activate is_cart()/is_checkout() on classic-shortcode block pages;
+		// otherwise this only happens as a side effect of constructing the
+		// (discarded) v5 SmartButton.
+		$this->context->init_context();
```
Narek13: Redundant — `SdkV6Module.php:103` already calls `init_context()` on `wp`, immediately before `register_render_hooks()` reaches this method, with nearly the same comment. The docblock at line 235 also still states the caller is responsible for it.

Suggest dropping the call and keeping the docblock contract.

modules/ppcp-sdk-v6/resources/js/sessions/freeTrialSave.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {string} [paymentMethod] - The WC gateway the token belongs to.
+ * @return {Promise<void>} Resolves once the token has been stored.
+ */
+export async function exchangeSetupToken( config, vaultSetupToken, paymentMethod ) {
+	if ( config.user?.is_logged ) {
+		const body = {
+			vault_setup_token: vaultSetupToken,
+			is_free_trial_cart: config.is_free_trial_cart ? '1' : '',
+		};
+		if ( paymentMethod ) {
```
Narek13: `payment_method` is added to the body but `CreatePaymentToken` never reads it — it detects card vs PayPal from the API result. Dead parameter, and the JSDoc implies it matters. Suggest dropping it (and the third argument) unless the endpoint is going to consume it.

modules/ppcp-sdk-v6/src/Blocks/V6PaymentMethod.php
```diff
@@ -76,13 +103,52 @@ public function get_payment_method_script_handles() {
...
+		}
+
+		// Saved-PayPal selector for returning buyers: the v5 smart-button stack
+		// that carried this in blocks is disabled under v6, so surface the same
+		// vault-component data here for checkout-block.js to register a saved-token
+		// method (mirrors the classic checkout carve-out).
+		if ( $this->vault_data && $this->vault_eligibility && ( $this->vault_eligibility )() ) {
+			$vault = $this->vault_data->add_localized_data( array() );
+			if ( isset( $vault['vault_component'] ) ) {
+				$data['vault_component']   = $vault['vault_component'];
```
Narek13: PHPCS: *Equals sign not aligned correctly; expected 1 space but found 3.* (0 errors, 1 warning on this file.)

```suggestion
				$data['vault_component'] = $vault['vault_component'];
```

modules/ppcp-vault-component/resources/js/checkout.js
```diff
@@ -0,0 +1,111 @@
...
+		updateUi();
+	}
+
+	function init() {
+		if ( typeof jQuery !== 'undefined' ) {
+			jQuery( document.body ).on( 'updated_checkout', onUpdatedCheckout );
+			jQuery( document.body ).on(
+				'updated_checkout payment_method_selected',
+				updateUi
+			);
```
Narek13: `updated_checkout` is bound twice, so `updateUi` runs twice per update (`onUpdatedCheckout` already calls it). Guarded by `isRendered()` so it's harmless, just redundant:

```suggestion
			jQuery( document.body ).on( 'updated_checkout', onUpdatedCheckout );
			jQuery( document.body ).on( 'payment_method_selected', updateUi );
```

Narek13 (commented):
Thanks for working on this @Dinamiko. I have some doubts and concerns. Please check the comments, as everything else looks good to me 💯

Dinamiko:
@Narek13 I added the following changes:
 1. `ApproveSubscriptionEndpoint`: kept the payer-validation fix (confirmed false positive); reworded the comment to "best-effort" and added a security note.                                                                                                        
  2. `paypal-config.js (Blocks)`: kept the direct-completion behavior for native PayPal subscriptions, the order-review continuation can't complete a subscription.                                                                                                               
  3. `SdkV6Manager`: removed the redundant `init_context()` call in `determine_render_places()`.                                                                                                  
  4. `freeTrialSave.js`: dropped the dead payment_method argument from `exchangeSetupToken` (+ card call sites and tests).                                                                      
  5. `checkout.js` (vault-component): removed the duplicate `updated_checkout` binding.

Narek13 (approved):
Thanks @Dinamiko. LGTM

PR #4629, opened 2026-08-24
Fix "Subscription could not be loaded" log on order screens (6706)

mmaymo:
# 🐛 What and why

Opening any WooCommerce order in wp-admin wrote a **`Subscription could not be loaded.` debug entry**, complete with a full backtrace, to the WooCommerce log.

The `add_meta_boxes` callback in `PayPalSubscriptionsModule` resolved the edited object into an order, then threw that object away and looked the subscription up again by id via `wcs_get_subscription()`. On an order screen the id is an order id, never a subscription id, so the lookup always failed. `wcs_get_subscription()` writes that debug entry itself before returning `false`.

## ⚙️ How it works now

WooCommerce Subscriptions registers the `shop_subscription` order type via `wc_register_order_type()` with `'class_name' => 'WC_Subscription'`. That means `wc_get_order()` — and the object HPOS hands the hook directly — is **already a `WC_Subscription`** on the subscription screen, under both posts and HPOS storage.

The callback now keeps that resolved object and applies a single `instanceof WC_Subscription` guard. Anything else returns early. This deletes the only code path reaching the WooCommerceLogger call inside `wcs_get_subscription()`, and removes one redundant order lookup per admin page view.

This mirrors how WooCommerce Subscriptions gates the same function in `WCS_Meta_Box_Related_Orders::output_rows()`: check the type first, and use `wcs_get_subscriptions_for_order()` on the order branch.

### Covered by unit tests

- ✅ Opening an ordinary non-subscription order writes no `Subscription could not be loaded.` entry.
- ✅ Opening a renewal order writes no `Subscription could not be loaded.` entry.
- ✅ On the subscription screen, a subscription with non-empty `ppcp_subscription` meta still gets the `ppcp_paypal_subscription` meta box, and `woocommerce-subscription-schedule` is still removed.
- ✅ On the subscription screen, a subscription with absent or empty `ppcp_subscription` meta gets no meta box, and the schedule box is left alone.
- ✅ No `ppcp_paypal_subscription` meta box is registered on any order screen, including renewal orders, under either HPOS or legacy post storage.

Each log test opens with a control assertion — it calls `wcs_get_subscription()` directly and asserts the spy caught the message — so a misconfigured logger fails loudly instead of passing vacuously.

The suite was mutation-tested with four hand-written mutants, all four killed: reverting this fix, applying the PR #4228 fallback, dropping the `ppcp_subscription` guard, and removing the `add_meta_box()` call each turn it red.

### DIFF
--- a/modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
+++ b/modules/ppcp-paypal-subscriptions/src/PayPalSubscriptionsModule.php
@@ -10,7 +10,6 @@ declare(strict_types=1);
 namespace WooCommerce\PayPalCommerce\PayPalSubscriptions;
 
 use Exception;
-use WC_Order;
 use WC_Product;
 use WC_Product_Subscription_Variation;
 use WC_Subscription;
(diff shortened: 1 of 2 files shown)


stracker-phil (approved):
LGTM 👍

PR #4630, opened 2026-08-24
fix(button): honor Pay Now for subscriptions bought from the product page

Narek13:
### Description

The approve endpoint can skip the order review and create the WC order itself, but only when the request carries `should_create_wc_order`. `CartActionHandler` sends it; `SingleProductActionHandler` never did — it also ignored the response body and hardcoded a redirect to checkout. So Pay Now had no effect for PayPal Subscriptions bought from a product page, while the cart page honored it. This is a gap in the integration, not a PayPal Subscriptions limitation.

Two parts:

- Send the flag, using the same vaulting and payment source guard the cart page applies, and follow the `order_received_url` the endpoint returns. With Pay Now off no such URL comes back and the shopper is sent to checkout exactly as before.
- Populate the cart **before** the approval rather than after it. The endpoint builds the WC order from `WC()->cart`, and `ChangeCartEndpoint::add_products()` empties the cart before adding the product, so the previous ordering would have built the order from an empty cart. This reorder is the only behavioral risk in the change and is covered by the abandon case below.

No server-side change was needed — `Context::is_checkout()` is already false for a product-page AJAX call, the same as it is for the cart page.

### Steps to Test

Setup:

1. WooCommerce Subscriptions active, and WooCommerce → Settings → Subscriptions → **Accept Manual Renewals off**.
2. PayPal Payments → Settings: **Save PayPal and Venmo off** (required to get the PayPal Subscriptions plan mode) and **Pay Now Experience on**.
3. Create a product of type **Simple subscription** with a price. In Product data → General, tick **Connect Product to PayPal Subscriptions Plan**, fill in a Plan Name, publish. Confirm the Product and Plan IDs appear afterwards.
4. Make sure the smart button is enabled on the product page.

Fix:

5. Logged in as a customer, open the product page and hard-reload. Open DevTools → Network.
6. Click the PayPal button and approve in the sandbox popup.
7. **Expected:** you land on the order received page, with no checkout stop. `ppc-change-cart` fires *before* `ppc-approve-subscription`; the approve payload contains `"should_create_wc_order": true` and the response contains `order_received_url`.
8. In WooCommerce → Orders, confirm the new order contains the subscription product at the correct total.

Regressions:

9. Turn Pay Now off, hard-reload, buy from the product page → lands on checkout in continuation mode, as before the fix.
10. Pay Now on, buy from the **cart** page → order received page, unchanged.
11. Click the button on the product page and close the PayPal popup without approving → cart unchanged, no order created.

### Changelog Entry

> Fix - Honor the Pay Now setting for PayPal Subscriptions purchased from a product page, instead of always redirecting to the checkout page for confirmation.

### DIFF
--- a/modules/ppcp-button/resources/js/modules/ActionHandler/SingleProductActionHandler.js
+++ b/modules/ppcp-button/resources/js/modules/ActionHandler/SingleProductActionHandler.js
@@ -24,45 +24,33 @@ class SingleProductActionHandler {
 					custom_id: this.config.subscription_custom_id,
 				} );
 			},
-			onApprove: ( data, actions ) => {
-				fetch( this.config.ajax.approve_subscription.endpoint, {
-					method: 'POST',
-					credentials: 'same-origin',
-					body: JSON.stringify( {
-						nonce: this.config.ajax.approve_subscription.nonce,
-						order_id: data.orderID,
-						subscription_id: data.subscriptionID,
-					} ),
-				} )
-					.then( ( res ) => {
-						return res.json();
-					} )
-					.then( () => {
-						const products = this.getSubscriptionProducts();
+			onApprove: async ( data ) => {
+				// The cart has to hold the subscription product before the
+				// approval is sent: the approve endpoint builds the WC order
+				// from WC()->cart, and the change-cart endpoint empties the
+				// cart before adding the product to it.
+				const cartResult = await this.changeCartToSubscription();
+				if ( ! cartResult.success ) {
+					console.log( cartResult );
+					throw Error( cartResult.data.message );
+				}
 
-						fetch( this.config.ajax.change_cart.endpoint, {
-							method: 'POST',
-							headers: {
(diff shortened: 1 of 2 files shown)


stracker-phil (approved):
LGTM - the new logic looks correct and solves the problem. I only have one minor and optional suggestion to better align with the `CartActionHandler` workflow:

---

`SingleProductActionHandler.js:74` .. the new helper method `approvalRedirectUrl()` always returns a redirect URL, even in a failure case. An error response has no `order_received_url`, so the customer is **sent o the checkout page, with no message**.

Difference: `CartActionHandler.js:36-39` does check for the flag and throws `data.data.message`. The PR's stated goal is parity with the cart page, so the divergence appears to be unintended.

Suggestion:

```js
approvalRedirectUrl( response ) {
    if ( ! response.success ) {
        throw Error( response.data.message );
    }
    return response.data?.order_received_url || this.config.redirect;
}
```

PR #4632, opened 2026-08-24
Orders paid via woo payments or bacs are overwritten as ppcp gateway when stale pay pal order meta exists (6698)

### LINKED ISSUE #4519: Orders paid via WooPayments or BACS are overwritten as ppcp-gateway when stale PayPal order meta exists (6698)
### Describe the Bug

WooCommerce orders completed using payment methods other than PayPal are being stored with `payment_method = ppcp-gateway`.

We observed this with two different payment methods:

1. An order paid via WooPayments.
2. An order placed using Direct Bank Transfer (BACS).

In both cases, `_ppcp_paypal_order_id` was present on the order, apparently from a previous PayPal checkout attempt. The payment method title and transaction data still indicate the payment method that was actually selected, but the internal WooCommerce payment method ID was overwritten with `ppcp-gateway`.

This appears to be possibly related to the handler introduced by PR https://github.com/woocommerce/woocommerce-paypal-payments/pull/4349:

In `WCGatewayModule::register_block_express_payment_method_handler()`, the plugin hooks into `woocommerce_before_order_object_save` and runs logic equivalent to:

```php
if ( ! $order->get_meta( PayPalGateway::ORDER_ID_META_KEY ) ) {
    return;
}

$payment_method = $order->get_payment_method();

if ( $payment_method && strpos( $payment_method, 'ppcp-' ) === 0 ) {
    return;
}

if ( $order->get_payment_method() === CreditCardGateway::ID ) {
    return;
}

$order->set_payment_method( PayPalGateway::ID );
```
My hypothesis is that as result, any order containing `_ppcp_paypal_order_id` may be changed to ppcp-gateway during a later save, even when its current payment method is woocommerce_payments or bacs.
The presence of a PayPal order ID alone does not necessarily mean that PayPal was the final payment method. A PayPal order or other stale data may have been created during an earlier or abandoned checkout attempt before the customer selected another gateway.

Affected WooPayments order:
```
payment_method:       ppcp-gateway
payment_method_title: Karte
transaction_id:       pi_[redacted]
paypal_order_id:      [present, redacted]
paypal_source:        paypal
paypal_payer_email:   [present, redacted]
```
The pi_ transaction ID is a WooPayments/Stripe PaymentIntent, while the stored gateway ID is ppcp-gateway.

Affected BACS order:
```
payment_method:       ppcp-gateway
payment_method_title: Banküberweisung
transaction_id:       [empty]
paypal_order_id:      [present, redacted]
paypal_source:        [empty]
paypal_payer_email:   [empty]
```
The title and empty transaction data are consistent with BACS, but the internal gateway ID is ppcp-gateway.

### To Reproduce
We have observed this in production.

### Expected Behavior

PayPal Payments should not overwrite the payment method when the order already has a valid non-PayPal gateway

### Actual Behavior

The plugin overwrites a non-PayPal payment method with ppcp-gateway.

### Environment
 - WordPress Version 7.0
 - WooCommerce Version 10.9.4
 - Plugin Version 4.1.1
 - Browser [e.g. Chrome, Safari] and Version -> any/not relevant
 - Any other plugins installed: WooPayments: 10.9.0

mmaymo:
## 🐛 What and why

WooCommerce orders paid with a gateway other than PayPal were being stored with `payment_method = ppcp-gateway`. Two production orders confirmed it: one paid via WooPayments (title `Karte`, transaction ID a Stripe `pi_` PaymentIntent) and one via Direct Bank Transfer (title `Banküberweisung`, no transaction ID). Both carried a **PayPal order ID left over from an earlier or abandoned checkout attempt**, and the presence of that meta alone was enough to rewrite the gateway on the next save.

The payment method title and transaction data still pointed at the real gateway, so the orders looked correct in the admin list while the internal gateway ID was wrong — breaking refunds, reporting, and any integration that keys off `payment_method`.

## ⚙️ How it works now

`register_block_express_payment_method_handler()` keeps a request-scoped `$marked_order_id`, shared by reference between its two closures. The `woocommerce_rest_checkout_process_payment_with_context` closure records the WC order ID at the moment it routes the payment to `PayPalGateway::ID`; the `woocommerce_before_order_object_save` closure then acts **only on that order, in that request**.

Everything else is unchanged: the listener still requires `PayPalGateway::ORDER_ID_META_KEY` and still leaves any deliberate `ppcp-` prefixed method — such as `CreditCardGateway::ID` — alone.

### Covered by tests

These are integration tests (`tests/integration/`), not unit tests — the listener is a WordPress hook closure and needs a real `WC_Order` save to exercise. All 6 acceptance criteria are covered; 11 tests, 13 assertions.

- ✅ A `bacs` order carrying a stale PayPal order ID keeps `bacs`.
- ✅ A `woocommerce_payments` order carrying a stale PayPal order ID keeps its gateway *and* its `pi_` transaction ID.
- ✅ An order the block express checkout routed through PayPal is still stored as `ppcp-gateway`, and the payment context itself is switched too.
- ✅ An order carrying the PayPal order ID meta that this request never routed through PayPal is untouched — whatever its gateway, including an empty one.
- ✅ An order already on a `ppcp-` prefixed method is never rewritten to `ppcp-gateway`.
- ✅ An order without the PayPal order ID meta is never modified.

### Not covered by tests

- ❌ Two different orders marked within one PHP request — `$marked_order_id` holds a single ID, so only the last would keep the guard. The block checkout produces one order per request, so the current flow cannot reach this; noted in the test file so a future checkout change does not reintroduce it silently.

### Worth knowing during review

- Mutation testing was run manually against the fix (8 mutants, all killed). 
- Pre-existing, not fixed here: the context closure reads `$payment_data['funding_source']` unconditionally, emitting an undefined-index warning when the key is absent.

### DIFF
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -947,7 +947,7 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 	 */
 	private function register_block_express_payment_method_handler( ContainerInterface $c ): void {
 		// Request-scoped guard for the save listener below: the ID of the WC order this request
-		// routed through PayPal, zero when no such switch happened.Documented in
+		// routed through PayPal, zero when no such switch happened. Documented in
 		// tests/integration/PHPUnit/WcGateway/BlockExpressPaymentMethodHandlerTest.php.
 		$marked_order_id = 0;
 
--- a/modules/ppcp-wc-gateway/src/WCGatewayModule.php
+++ b/modules/ppcp-wc-gateway/src/WCGatewayModule.php
@@ -946,9 +946,14 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 	 * Ensures PayPal handles block-checkout express payments even when another PCP gateway is sorted first in WC Settings → Payments.
 	 */
 	private function register_block_express_payment_method_handler( ContainerInterface $c ): void {
+		// Request-scoped guard for the save listener below: the ID of the WC order this request
+		// routed through PayPal, zero when no such switch happened.Documented in
+		// tests/integration/PHPUnit/WcGateway/BlockExpressPaymentMethodHandlerTest.php.
+		$marked_order_id = 0;
+
 		add_action(
 			'woocommerce_rest_checkout_process_payment_with_context',
-			function ( $context ) use ( $c ): void {
+			function ( $context ) use ( $c, &$marked_order_id ): void {
 				$payment_data = (array) ( $context->payment_data ?? array() );
 
 				if ( empty( $payment_data['paypal_order_id'] ) ) {
@@ -976,6 +981,9 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 					$funding_source = $payment_data['funding_source']
 						?: ( $session_handler->funding_source() ?: 'paypal' );
 					$order->set_payment_method_title( $funding_source_renderer->render_name( $funding_source ) );
+
+					$marked_order_id = $order->get_id();
+
 					$order->save();
 				}
 			},
@@ -985,8 +993,11 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 
 		add_action(
 			'woocommerce_before_order_object_save',
-			function ( $order ): void {
+			function ( $order ) use ( &$marked_order_id ): void {
 				if ( ! $order instanceof WC_Order ) {
+					return;
+				}
+				if ( ! $marked_order_id || $order->get_id() !== $marked_order_id ) {
 					return;
 				}
 				if ( ! $order->get_meta( PayPalGateway::ORDER_ID_META_KEY ) ) {
@@ -994,9 +1005,6 @@ class WCGatewayModule implements ServiceModule, ExtendingModule, ExecutableModul
 				}
 				$payment_method = $order->get_payment_method();
 				if ( $payment_method && strpos( $payment_method, 'ppcp-' ) === 0 ) {
-					return;
-				}
-				if ( $order->get_payment_method() === CreditCardGateway::ID ) {
 					return;
 				}
 				$order->set_payment_method( PayPalGateway::ID );
--- a/tests/integration/PHPUnit/WcGateway/BlockExpressPaymentMethodHandlerTest.php
+++ b/tests/integration/PHPUnit/WcGateway/BlockExpressPaymentMethodHandlerTest.php
@@ -1,0 +1,332 @@
+<?php
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\Tests\Integration\WcGateway;
+
+use Automattic\WooCommerce\StoreApi\Payments\PaymentContext;
+use WC_Order;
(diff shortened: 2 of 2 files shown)


modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -985,20 +993,20 @@ function ( $context ) use ( $c ): void {
 
 		add_action(
 			'woocommerce_before_order_object_save',
-			function ( $order ): void {
+			function ( $order ) use ( &$marked_order_id ): void {
 				if ( ! $order instanceof WC_Order ) {
 					return;
 				}
+				if ( ! $marked_order_id || $order->get_id() !== $marked_order_id ) {
```
Narek13: This narrows what the listener covers, and I think it drops the most common path.

The context closure returns at `if ( ( $context->payment_method ?? '' ) === PayPalGateway::ID )` **before** setting the marker. So an order WooCommerce already routed to `ppcp-gateway` — the normal case when PayPal is sorted first — is never marked, and this guard bails. Previously the re-assert applied there too, since it needed only the meta.

Deliberate? Quite possibly: the listener exists for the "another PCP gateway sorted first" case, which still marks correctly, so what is being dropped may never have been load-bearing. Worth confirming, because the failure mode is quiet — the order keeps whatever gateway reset it, meta present, nothing logged.

If it is intended, one line in the method comment saying the re-assert covers only the re-routed case would save the next reader the same question.

modules/ppcp-wc-gateway/src/WCGatewayModule.php
```diff
@@ -976,6 +981,9 @@ function ( $context ) use ( $c ): void {
 					$funding_source = $payment_data['funding_source']
```
Narek13: Minor, and you already call it out as pre-existing: no `isset` guard, so this warns when the key is absent. Fair to leave, though you are editing the line below and `?? ''` would close it here.

modules/ppcp-wc-gateway/src/WCGatewayModule.php
Narek13, line 1000: `$marked_order_id` is set only when this request actually switched the payment method to PayPal. But there is an earlier return for the case where no switch was needed:

```php
if ( ( $context->payment_method ?? '' ) === PayPalGateway::ID ) {
    return;   // already PayPal — nothing to switch, and no marker set
}
```

That is the normal case when PayPal is sorted first: WooCommerce has already picked `ppcp-gateway`, the closure has nothing to do, and it leaves before marking the order. This guard then makes the listener skip that order.

Before this PR the listener did run for it, because all it asked for was the meta.

So the re-assert now covers orders that had to be switched, but no longer covers orders that arrived as PayPal already.

That is probably fine as the listener was added for the "another PCP gateway is sorted first" case, and those orders still get marked. Worth a check though, because if it is wrong, nothing says so: the order quietly keeps whatever gateway overwrote it, the meta is still there, and nothing is logged.

If it is intended, one line in the method comment saying the re-assert applies only to re-routed orders would spare the next reader the same question.

modules/ppcp-wc-gateway/src/WCGatewayModule.php
Narek13, line 981: Minor, and you already call it out as pre-existing: no `isset` guard, so this warns when the key is absent. Fair to leave, though you are editing the line below and `?? ''` would close it here.

Narek13 (approved):
Thanks for the PR. I left 1 question, plus a minor.

PR #4641, opened 2026-08-26
enable SDK v6 feature flag for new store setups (6828)

Dinamiko:
SDK v6 was gated by a code/env flag defaulting off, so there was no way to give it to new stores while leaving existing stores (including on reconnect) on v5.                    
                                                                                                                                                                                             
This PR records a one-time `woocommerce-ppcp-sdk-v6-eligible` option at install/update, `yes` for fresh installs, `no` for updates and have `modules.php` load the v6 module when it's `yes` (env/filter overrides still apply). Since it's set once and never recomputed, disconnect/reconnect keeps existing stores on v5; the option is also cleaned up on uninstall.


Narek13 (approved):
LGTM

PR #4642, opened 2026-08-26
Register Apple Pay and Google Pay as block express methods (6788)

stracker-phil:
*Changelog:* (none - v6 SDK module, disabled by default)

# Description

## 🎯 The goal

Apple Pay and Google Pay render on classic pages and the v6 express areas, but not in the Cart and Checkout blocks. This adds them as Blocks express payment methods, driven by the same two bridges the classic pages already use rather than a second implementation.

React owns an empty container and never reconciles what the bridge puts inside it. What the block surface knows better than PHP did, such as the live shipping requirement and the address already entered, is passed down as overrides.

## 🔍 What to review

- **Gateway attribution.** The express methods register under their own gateway ids, but the order is created against `ppcp-gateway`, matching the classic express surfaces. This affects reporting and refunds, and is deliberately unchanged.
- **The subscription gate is now client-side.** Each wallet advertises its own gateway's `supports` rather than PayPal's, and Blocks re-evaluates on every cart update. An unavailable gateway falls back to `products` alone, offering the wallet on fewer carts, never more.
- **A failed render is not an incapable browser.** A bridge that throws on the way to the button is retried a few times; only the wallet answering "cannot pay this way" drops the express row.

## 🧪 How to verify

1. On Block Checkout with both wallets enabled, add a subscription product without reloading. Both buttons disappear; removing it brings them back.
2. With an address already in the checkout form, tap Apple Pay. The sheet opens on that address.
3. Open the Apple Pay sheet and cancel it. One button remains, and a second tap opens a new sheet.
4. Block the Google `pay.js` host in devtools, load Block Checkout, then unblock it. The button appears without a reload.


modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -357,63 +358,106 @@ public function render_mini_cart_wrapper(): void {
...
+
+		return $placement->is_gateway;
+	}
+
+	/**
+	 * The gateways WooCommerce offers for the current cart, keyed by id.
+	 *
+	 * @return array<string, WC_Payment_Gateway>
+	 */
+	private function available_gateways(): array {
```
Narek13, line 402: Not memoized, and it now has a second caller. `is_method_gateway()` and the new `gateway_supports()` both reach this once per placement, so with two wallets that is up to **four** full `get_available_payment_gateways()` passes per render, each looping every gateway and calling `is_available()`.

Worth fixing here specifically because #4640 (PCP-6739) just landed a fix for unnecessary gateway/product-status resolution causing repeat `merchant-integrations` calls — PPCP gateways' `is_available()` can reach `ProductStatus`. The results are cached so I don't think this adds HTTP calls, but it is needless work on a render path.

```php
private ?array $available_gateways = null;

private function available_gateways(): array {
	if ( null !== $this->available_gateways ) {
		return $this->available_gateways;
	}
	…
}
```

Same per-request memoization `is_method_gateway()` already does via `$placement->is_gateway`.

Dinamiko (approved):

Narek13 (approved):
Really nice work and a great PR. Looks good to me overall. I only have one suggestion that I think is worth your attention.

I came across it because I recently worked on #4640, and I see some potential risks here.

Narek13 (approved):

PR #4650, opened 2026-08-27
feat: send the shopper's details when creating a PayPal order from the block checkout

Narek13:
PayPal identifies a buyer partly by their email address. The block checkout sent none, so a card PayPal had never seen had nothing to be matched against, and the first attempts were declined until a fingerprint built up from those very attempts — hitting every new customer once, and nobody who had paid before.

From the merchant's PayPal contact, quoted in the ticket:

> With no card history and no email address, our risk system has nothing to identify the buyer with, so the first two attempts were declined.

The block create-order request carried neither `payer` nor `form`, so `CreateOrderEndpoint::payer()` had nothing to build from and `OrderEndpoint::create()` attached no payer.

## Changes

- **`cartPayerData()`** reads the shopper's email, name, phone and billing address from the `wc/store/cart` store, and is used by the block create-order paths on **both stacks**.
- **v5 blocks** — card fields and PayPal.
- **v6** — the payer was built only under `context === 'checkout'`, and a block checkout reports `'checkout-block'`, so that branch never ran and v6 sent nothing at all. Both create-order paths now pick their source by context: billing inputs on classic, cart store on blocks, nothing elsewhere.

**Nothing is sent without an email.** `Payer::to_array()` always emits `name`, so a payer carrying only an email would post empty strings and risk the order being rejected — worse than the missing signal. The endpoint's own `! empty( $payer->email_address() )` guard stays the floor.

A structured `payer` is used rather than the classic path's `form_encoded`: `PayerFactory::from_checkout_form()` throws on an invalid email, which would break create-order for a half-typed address, and it avoids populating `$data['form']`, which is now persisted to the session.

## Naming

There were two `payerData()` helpers — one reading classic DOM inputs, one reading the cart store — with the same file name, and v6 imported the first while the block modules used the second. The cart-store one is now `cartPayerData()` in `CartPayerData.js`, beside its DOM counterpart where both stacks can reach it.

## Steps to Test

1. Block checkout with card fields (ACDC), DevTools on the Network tab. Fill in billing including an email and submit.
2. Inspect the `ppc-create-order` payload: **expect** a `payer` with `email_address`, `name` and the address. Before this it held only `nonce`, `context`, `payment_method` and `save_payment_method`.
3. Repeat with the block PayPal button — same expectation.
4. Repeat both **with SDK v6 enabled**.
5. Leave the email empty and submit: **expect** no `payer` key at all, rather than one with empty values.
6. Classic checkout is unchanged — confirm an order still completes.

Verified by hand on v5 card fields. Steps 3–5 are covered by unit tests; **v6 has not been exercised by hand**.

The declines themselves cannot be verified from a sandbox: they depend on PayPal's risk engine, a genuinely new card and no prior fingerprint. What this demonstrably changes is that the email is now sent.

## Changelog Entry

> Fix - Send the shopper's email and billing details to PayPal when an order is created from the block checkout, so first payments with a new card are not declined for lack of a buyer signal.


modules/ppcp-blocks/resources/js/Helper/PayerData.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * The payer details the block checkout can supply when creating a PayPal order.
+ *
+ * PayPal's risk engine identifies a buyer partly by their email address. Without
+ * one, a card it has never seen has nothing to be matched against, and the first
+ * attempts on a brand-new card are declined until a fingerprint builds up. The
+ * classic checkout has always sent this along with the serialised form; the block
+ * checkout sent nothing, so the same shopper fared worse there.
```
AlexP11223: would be great to review and maybe simplify comments like this. For example, the last sentence is kind of irrelevant after the PR is merged.

modules/ppcp-blocks/resources/js/Helper/PayerData.js
```diff
@@ -0,0 +1,75 @@
...
+ * Builds the payer object for a create-order request.
+ *
+ * Returns null unless an email is known: PayPal is told nothing rather than told
+ * something empty. Payer::to_array() always emits the name, so a payer carrying
+ * only an email would send two empty strings and risk the whole order being
+ * rejected, which is worse than the missing signal this exists to provide.
+ *
+ * @return {?Object} The payer, or null when there is nothing worth sending.
+ */
+export function payerData() {
```
AlexP11223: There is already a function with this name and it is used, for example, in the V6 module. 
https://github.com/woocommerce/woocommerce-paypal-payments/blob/2bfb561a30ab22d96bce2ab9d70e4516a5481f2d/modules/ppcp-sdk-v6/resources/js/endpointsAdapter.js#L13

Not sure if it maybe makes sense to consider some kind of renaming or refactoring to avoid possible confusions.

AlexP11223 (approved):

PR #4654, opened 2026-08-28
Tailor the .claude/ tooling for a general audience (6754)

stracker-phil:
*Changelog:* (none - internal Claude tooling)

# Description

## 🎯 The goal

Make the Claude Code setup in `.claude/` generally applicable, rather than tied to one specific workflow.

The diff is confined to `.claude/`: agents and skills are renamed to describe what they do, workflow-specific and duplicated ones are dropped, `/clean-up-code` and `/md-writer` are added, frontmatter is corrected throughout, and `.claude/docs/README.md` documents the result.

## 🔍 What to review

`.claude/docs/README.md` is the fastest way in: it lists every skill, agent, hook, and rule with a short summary and a link to its source.


.claude/agents/ci.md
```diff
@@ -4,8 +4,7 @@ description: Runs the project's tests and linters (PHP unit, JS unit, integratio
 color: orange
 model: haiku
 background: true
-tools: Read, Glob, Grep, Bash(ddev npm run *)
-disallowedTools: Edit, Write, NotebookEdit, WebFetch, WebSearch, Skill, ToolSearch, EnterWorktree, ExitWorktree, Monitor, TaskStop, TodoWrite, SendMessage
+tools: Read, Glob, Grep, Bash
```
Dinamiko, line 7: Not sure about this... it previously could only run `ddev npm run ...;` but it can now execute any shell command which is a bit risky, if the goal was to permit the documented npm run ... commands, prefer a scoped Bash(npm run *) / Bash(ddev *) rather than full Bash unless strictly necessary.

.claude/skills/describe-pr/SKILL.md
```diff
@@ -1,66 +1,73 @@
...
+What problem(s) does the PR address (bug fix, feature, refactoring, ...)?
+Which files contain the core change?
+Does it touch or introduce business logic?
+Severity of the PR - eg. is it documentation or cosmetic (low), impact admin UI or front-end copy (mid), does it impact money flow (high)?
+Which is the single most important entry point that a reviewer should see?
 
 ### 3. Draft the title and description
 
-Follow [guideline.md](guideline.md) for the title rule, the mandatory changelog first line, the heading catalog, tone, writing style, and a worked example. Read it before drafting.
+Draft a crisp, and easy to understand PR description. Follow `${CLAUDE_SKILL_DIR}/guideline.md` for the title rule, the mandatory changelog first line, the heading catalog, tone, writing style, and a worked example. Read it before drafting.
```
Dinamiko, line 47: I think that variables (e.g. `${CLAUDE_SKILL_DIR}`) in a skill's body prose don't work, they're literal text, please check it to be sure if they can be used or not.

.claude/skills/md-writer/SKILL.md
```diff
@@ -0,0 +1,48 @@
+---
+name: md-writer
+description: Write Markdown documentation in alignment with project standards. Always use this skill to create or update .md files.
+paths:
+- "**/*.md"
```
Dinamiko, line 5: It seems that `paths` setting only works on rules, it's just ignored silently in skills and automatic triggering never happens, please double check this to ensure if the setting can be used or not in skills.

Dinamiko (changes requested):

.claude/agents/ci.md
stracker-phil: This was a mis-configuration in the first place.

An agent does not accept the `Bash(ddev npm run *)` syntax and the behavior is undocumented in this case; in my tests, it falls back to just allowing Bash

This is a subtle difference between skills and agents:
- Agents have `tools: Bash` - effect: the agent can use bash
- Skills have `allowed-tools: Bash(npm run *)` - effect: The "num run" commands never ask for user permission, it does NOT grant "bash" access if the agent does not have that tool already

You can test this:

In Claude code, enter "Manual" mode and use a prompt like this
```
Tell the "ci" agent verbatim: "Run the bash command 'ls' and respond with the
verbatim output you see" as I want to verify if that agent has access to "ls"
and if this triggers a prompt or not
```

.claude/skills/describe-pr/SKILL.md
stracker-phil: Here is a list with variables that can be inserted into Claude skills:

https://code.claude.com/docs/en/skills#available-string-substitutions

The CLAUDE_SKILL_DIR resolves to _The directory containing the skill’s SKILL.md file._

.claude/skills/md-writer/SKILL.md
stracker-phil: Did this not work for you in a local test?

I tested it once and it was working before creating the PR - the `paths` frontmatter is also documented in https://code.claude.com/docs/en/skills#frontmatter-reference and is described as

_Glob patterns that limit when this skill is activated. ... When set, Claude **loads the skill automatically** only when working with files matching the patterns_

Dinamiko (approved):

PR #4659, opened 2026-08-31
PayPal gateway not visible on block checkout when WC Subscription Auto renewal are enabled (6844)

Dinamiko:
On Block Checkout, the standard PayPal gateway was hidden whenever the cart contained a subscription, "Accept Manual Renewals" was enabled, and Save/Vault PayPal was disabled.

The `placeOrderEnabled` flag in `PayPalPaymentMethod::get_payment_method_data()` only allowed subscription carts when a vault token could be saved, so with vaulting off the gateway was never registered by the block.                                                                                                                                                                   
                                                                                                                                                                                             
This PR broads that condition to also allow the manual-renewal case (`accept_manual_renewals()`), reusing the existing SubscriptionHelper method. This matches `resolve_subscription_mode()`, which already treats manual renewals + vaulting-off as a plain one-time Orders API payment, so the gateway now shows and processes a standard redirect.


Narek13 (approved):
LGTM

PR #4664, opened 2026-09-01
Classic checkout - When Fastlane enabled selecting PayPal as guest button Proceed to PayPal visible (6850)

Dinamiko:
On classic checkout with Fastlane and SDK v6 enabled, a guest selecting the PayPal gateway saw the standard "Proceed to PayPal" place-order button rendered alongside the PayPal smart buttons. The v6 SDK hides that button with an inline `display:none !important`, but AXO's `AxoManager.rerender()` re-showed it via `jQuery .show()` whenever a non-AXO gateway was selected and `.show()` strips the inline style. (v5 was unaffected because it hides via a stylesheet class jQuery .show() can't override.)                                                          
                                                                                                                                                                                             
In `AxoManager.rerender()`, guarded the `defaultSubmitButton.show()` call so AXO no longer re-shows `#place_order` on v6 pages (detected via `fastlaneSdkV6Config()`), deferring ownership of that button to the v6 SDK. AXO still hides it when the Fastlane gateway itself is active, and v5 behavior is unchanged.


Narek13 (approved):
Great work! LGTM, thanks

PR #4667, opened 2026-09-01
Block checkout - When Fastlane enabled and v6 active ACDC is also present (6851)

Dinamiko:
On block checkout with Fastlane enabled and SDK v6 active, guests saw two "Debit & Credit Cards" rows, Fastlane's and a redundant standalone ACDC card. v6 replaces the legacy per-gateway block methods with a single JS entry but never carried over the v5 guard that hides the standalone ACDC card when Fastlane renders. Since enabling Fastlane requires ACDC, the v6 card flag (keyed only off `is_acdc_enabled()`) always registered the extra card row.                                                                                                      
                                                                                                                                                                                             
This PR gates the v6 `card_fields.enabled` flag on Fastlane not being active, reusing the existing `is_fastlane_enabled()` check (which already encodes the v5 conditions: guest, no subscription, supported context). Scoped to only the JS flag, so the v5 ACDC block stays suppressed, leaving exactly one card row (Fastlane). Logged-in, subscription, and order-pay cases keep the ACDC card, matching legacy behavior.

### DIFF
--- a/modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
+++ b/modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
@@ -1044,7 +1044,12 @@ class SdkV6Manager {
 			$pay_later_button['mini-cart'] = $this->is_pay_later_button_enabled( 'mini-cart' );
 		}
 
-		$card_fields_enabled = $this->is_card_fields_enabled();
+		// Fastlane replaces the standalone ACDC card row when it renders (guest,
+		// non-subscription checkout), matching the v5 advanced-card block guard.
+		// Scoped to this JS flag only: is_card_fields_enabled() still gates the v5
+		// card block's suppression, so a page keeps exactly one card option.
+		$card_fields_enabled = $this->is_card_fields_enabled()
+			&& ! $this->is_fastlane_enabled();
 
 		$messages_settings_location = $this->messages_settings_location();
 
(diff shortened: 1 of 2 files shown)


modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
```diff
@@ -1044,7 +1044,12 @@ public function script_data(): array {
 			$pay_later_button['mini-cart'] = $this->is_pay_later_button_enabled( 'mini-cart' );
 		}
 
-		$card_fields_enabled = $this->is_card_fields_enabled();
+		// Fastlane replaces the standalone ACDC card row when it renders (guest,
```
Narek13, line 1047: Two independent predicates now decide the same thing.

On classic checkout AXO already removes the ACDC gateway (`AxoModule.php:113-134`) via `AxoApplies::should_render_fastlane()`, gated on `axo.eligible` and wrapped in a public filter, `woocommerce_paypal_payments_axo_hide_credit_card_gateway`. This new flag uses a different predicate, `FastlaneConfig::should_render()`.

They agree today, but nothing enforces it, and the documented filter is the obvious way to break it: a merchant who keeps the ACDC gateway visible alongside Fastlane now gets a gateway row with no card form and no SDK component behind it.

Deriving both from one predicate - or gating this flag on the same filtered decision AXO makes - would remove the coupling.

Same shape as the guard in #4664: a v6 flag re-deriving a condition another module already owns. Might be worth one shared helper rather than solving it case by case.

Narek13 (approved):
Thanks for working on this. LGTM, just 1 doubt worth your attention.
However I am going to approve, since I am not sure about this and it is not a blocker, so please feel free to ignore

PR #4668, opened 2026-09-01
fix(order-endpoints): copy cart fees onto the order built from an approved PayPal order

Narek13:
### Description

On the express-checkout paths that build the WooCommerce order directly from the approved PayPal order - Google Pay and Apple Pay on the block checkout, and any express button on the cart or product page, whenever Pay Now is enabled - a cart fee was silently dropped from the created order. `WooCommerceOrderCreator` copied line items, shipping and coupons, but never fees.

The buyer still approved the correct, fee-inclusive amount in the wallet sheet, but the order that got built and charged was off by the fee: undercharged for a surcharge, overcharged for a negative fee. Confirmed live with a -12 fee: the buyer approved 98.00 in the Google Pay sheet and was charged 110.00.

`configure_fees()` now mirrors `WC_Checkout::create_order_fee_lines()`, adding a `WC_Order_Item_Fee` per cart fee and firing the same `woocommerce_checkout_create_order_fee_item` action third-party plugins already rely on. `CartData` carries the fees through its snapshot via an optional constructor argument, so existing callers and stored snapshots are unaffected. The snapshot falls back to the `ppcp_fees` session cache when the cart wasn't recalculated this request (the same source the PayPal purchase unit's own line items are built from), since a session-restored cart carries its fee totals but not the fees themselves.

Not tested under multisite.

### Steps to Test

Requires Pay Now enabled, since that determines whether this class or WooCommerce itself builds the order.

1. Add a cart fee via `woocommerce_cart_calculate_fees`, e.g. `$cart->add_fee( 'testFee', -12, false );`.
2. Add a physical product to the cart and go to the block checkout. Confirm the fee shows in the order summary and the total reflects it.
3. Click an express button (Google Pay or Apple Pay) and approve.
4. Check the created order: it should have a fee line item, and its total should match what was shown at checkout.
5. Regression: repeat with Pay Now disabled. The order is built by WooCommerce as before and should be unaffected.
6. Regression: repeat with a positive fee (e.g. a surcharge) to confirm the merchant is no longer undercharged either.

### Documentation

### Changelog Entry

> Fix: Cart fees are no longer dropped from the order when paying with an express button (Google Pay, Apple Pay) on the block checkout, or from the cart/product page, with Pay Now enabled.

### DIFF
--- a/modules/ppcp-button/src/Session/CartData.php
+++ b/modules/ppcp-button/src/Session/CartData.php
@@ -20,6 +20,13 @@ class CartData {
 	 */
 	protected array $coupons;
 
+	/**
+	 * The cart fees, each normalized to an array as in CartDataFactory.
+	 *
+	 * @var array<string, array<string, mixed>>
+	 */
+	protected array $fees = array();
+
 	protected bool $needs_shipping;
 
 	protected int $user_id;
@@ -36,19 +43,22 @@ class CartData {
 	 * @param bool                                $needs_shipping
 	 * @param int                                 $user_id
 	 * @param string                              $cart_hash
+	 * @param array<string, array<string, mixed>> $fees Optional, so existing callers keep working.
 	 */
 	public function __construct(
 		array $items,
 		array $coupons,
 		bool $needs_shipping,
 		int $user_id,
-		string $cart_hash
+		string $cart_hash,
+		array $fees = array()
 	) {
 		$this->items          = $items;
 		$this->coupons        = $coupons;
 		$this->needs_shipping = $needs_shipping;
 		$this->user_id        = $user_id;
 		$this->cart_hash      = $cart_hash;
+		$this->fees           = $fees;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


modules/ppcp-button/src/Session/CartData.php
```diff
@@ -36,19 +43,22 @@ class CartData {
 	 * @param bool                                $needs_shipping
 	 * @param int                                 $user_id
 	 * @param string                              $cart_hash
+	 * @param array<string, array<string, mixed>> $fees Optional, so existing callers keep working.
```
AlexP11223, line 46:
```suggestion
	 * @param array<string, array<string, mixed>> $fees Optional.
```

I sometimes encounter an issue that Claude is trying to keep some things backward-compatible assuming that there are some external users of a public class or something like that (which is not something we guarantee, and usually the class history has many recent modifications breaking such compatibility already).

In this case, it is probably fine to make it optional. But I would maybe at least remove this comment part, because the only remaining callers seem to be some tests, and this explanation probably is not something that should be kept in the code.

AlexP11223 (approved):

PR #4669, opened 2026-09-01
Fix webhook registration hitting the wrong API host right after onboarding (174)

Biont:
*Changelog:* `Fix - PayPal connection no longer gets stuck in an endless "complete registration" loop #4669`

# Description

## 🐛 The problem

Right after a merchant completes PayPal onboarding, the settings screen can get stuck showing "PayPal Payments is almost ready" and prompting to complete registration again, in a loop. The webhook registration request goes to the wrong host - the pre-connection Woo partner-proxy instead of PayPal's API - and 404s.

## 💡 Why it happens

Webhook registration fires synchronously, in the same request, right after the merchant's connection state flips to "connected." The container that produces the API host and the webhook endpoint was caching that value for the rest of the request the first time anything asked for it - which, on this exact request, usually happens *before* the connection completes. Every later call in that same request, including the webhook registration that was supposed to use the new connection, kept using the stale, pre-connection host.

## ✅ The fix

The three container entries responsible for the API host and the webhook endpoint/registrar (`api.host`, `api.endpoint.webhook`, `webhook.registrar`) now resolve fresh on every lookup instead of being cached for the request's lifetime. Every other API endpoint that depends on the host keeps the existing cached behavior, since none of them are invoked synchronously mid-connection the way webhook registration is.

## 🧪 Tests

Added `WebhookEndpointTest` (the class had no coverage before) for `create`/`list`/`delete`/`simulate`/`verify_event`, plus a wiring regression test that boots the real container and asserts the three affected services resolve fresh while sibling host-dependent endpoints stay cached - so either side of this regressing silently gets caught.

## 🔍 What to review

Only the three keys directly involved in webhook registration were changed; the ~16 other endpoints that depend on the API host intentionally still cache it, since none of them are invoked synchronously in reaction to a connection change. Manual verification of both connection flows (OAuth "Return to Store" and manual API credentials) is still outstanding - see below.

## 🧪 How to verify

1. Connect a store via the OAuth "Return to Store" flow. Confirm the settings screen does not loop back to "complete registration."
2. Connect a store using manual API credentials. Confirm the same.


Biont:
I am not 100% happy with this solution, worth flagging for reviewers.

An earlier version had `WebhookEndpoint` resolve its host at call time instead of at construction (first directly against `ConnectionState`, then behind a narrower `ApiHostResolver` wrapper). That worked, but it made `WebhookEndpoint` the one endpoint, out of about seventeen near-identical siblings, built on a different pattern, with nothing in its own code explaining why.

The `factories()` wiring keeps every endpoint class structurally identical and moves the exception to the one place it's actually legible: the container wiring. It's not fully satisfying either. The core problem is _Config/Settings changes mid-request - and object instances working off stale information_ - a problem that is not isolated to this bugfix in particular.

AlexP11223:
But it sounds like the first solution was better. Because now it is difficult to guarantee that it works and it adds complexity, you have to know that the endpoints like that has to be retrieved from the container only after the onboarding state update, so it can easily become not like that if initialized in some earlier hook etc.

I think it is better and simpler to have some kind of factory (of a factory), such as just a `callable` returning the string from the `api.host` factory, or I guess this `ApiHostResolver` wrapper that was mentioned. And then the endpoints that have a chance of ending up in the onboarding update request use this service.

Biont:
@AlexP11223 

> the first solution was better

Allow me to expand on the previous comment then: The main problem with that was that it created an exception to a **very strong** convention: Literally _all_ of our code that interacts with the PayPal API expects a `$this->host` property passed as a constructor arg. [`RequestTrait`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/b0a6037115dab44a44c97d16efc95d3c3e50170a/modules/ppcp-api-client/src/Endpoint/RequestTrait.php#L17) alone is used by **30 classes** and codifies its usage by expecting it in [RequestTrait::request_response_string](https://github.com/woocommerce/woocommerce-paypal-payments/blob/b0a6037115dab44a44c97d16efc95d3c3e50170a/modules/ppcp-api-client/src/Endpoint/RequestTrait.php#L130-L131)

Rewriting the trait to no longer demand the property is one thing. Rewriting [`WebhookEndpoint`](https://github.com/woocommerce/woocommerce-paypal-payments/blob/ddb0ae3b11e23e7408ba572f48deefccbf14b3b8/modules/ppcp-api-client/src/Endpoint/WebhookEndpoint.php#L25) to resolve the `$host` at runtime is a different thing: This class along sits among 15 sibling that are **all** written against `$this->host`. 

Whatever necessitates this change in `WebhookEndpoint` is also true for all the other classes in principle. Yes: most of them are unlikely to be caught in the middle of a config update, but we cannot know that for all eternity. So for me it was a choice between a sizeable refactor for a targeted fix - or staying true to an obvious convention.


> you have to know that the endpoints like that has to be retrieved from the container only after the onboarding state updated

You're right on this: I created a way to receive a correct instance after a config change, but it does not help consumers that already fetched an instance before the change.
I think my view got clouded when I saw what it would take to pull this off cleanly - see section above.

---

So what is the middle-ground? I understand that this PR isn't airtight yet. Remaining options are:

1. Deferring the webhook registration to a separate HTTP request (lots of new error surface, more complicated for manual credential input) 
2. Adding a resolver for `WebhookEndpoint` only - breaking the convention
3. Migrating all of `**\EndpointyThings\*` to resolve the host at runtime - installing a new convention

Gonna go with Option 2 based on your feedback.

Biont:
Pushed: went with Option 2 (`d3fff3fd4`).

`ApiHostResolver` (`modules/ppcp-api-client/src/Helper/ApiHostResolver.php`) wraps `ConnectionState` and exposes exactly one thing, `host(): string`, resolved fresh on every call instead of once at construction. `WebhookEndpoint`'s constructor now takes `ApiHostResolver $host_resolver` instead of `string $host`, and all five of its call sites resolve the host at the moment they build the request URL, not before.

To avoid cascading into the `RequestTrait` convention @Biont flagged (`$this->host` used by ~30 classes): the only place `RequestTrait::request_response_string()` touched `$this->host` was to decide whether a request body was an OAuth token exchange and should be skipped from logging. Swapped that for a `strpos($url, 'v1/oauth2/token')` check — same intent, no dependency on any endpoint having a `$host` property at all. So the other ~15 sibling endpoints, and the other ~29 `RequestTrait` consumers, are untouched. `WebhookEndpoint` is now the one exception, which is what Option 2 was.

@AlexP11223 your point about needing to know an endpoint has to be pulled from the container only after the state update still applies here in the narrow sense: `WebhookEndpoint` itself is now safe regardless of when it's constructed relative to onboarding, since it no longer freezes anything at construction time. What's still true is that nothing stops a *different* endpoint from ending up in the same synchronous connect-then-fire path in the future without anyone remembering this is the one place that needed the fix. Left as a noted, accepted risk rather than solved - matches where the discussion above landed.

Tests: new `ApiHostResolverTest` covers the four connection/environment combinations, plus a regression test that resolves the host, mutates the same `ConnectionState` instance mid-test, and asserts the second resolution picks up the change. `WebhookEndpointTest` got the equivalent regression test at the endpoint level (two `list()` calls against a resolver mock returning two different hosts). `WebhookModuleWiringTest` - the wiring-level test written for the factories() shape - was deleted, since its premise (services() vs. factories() memoization) no longer applies to anything in this PR.

---
For transparency: This comment was written by claude

Biont:
Small footnote, and a little ironic given the "very strong convention" point above: CI just caught `CartEndpoint::$host` (`modules/ppcp-wc-gateway/src/StoreApi/Endpoint/CartEndpoint.php`) as dead code. It only ever existed as a dummy property to satisfy `RequestTrait`'s implicit `$this->host` requirement - `CartEndpoint` never actually used it for a request. Now that the trait no longer touches `$this->host` at all, PHPStan flagged it as write-only and it's been removed. One less class quietly carrying the convention's cost for no real reason.

modules/ppcp-api-client/src/Helper/ApiHostResolver.php
```diff
@@ -0,0 +1,52 @@
...
+ * an API consumer's concern.
+ */
+class ApiHostResolver {
+
+	/**
+	 * The connection state.
+	 *
+	 * @var ConnectionState
+	 */
+	private $connection_state;
```
AlexP11223:
```suggestion

	private ConnectionState $connection_state;
```

AlexP11223:
PHPStan complains about an unused property, since it is no longer used by the trait.

AlexP11223 (approved):

PR #4670, opened 2026-09-01
Free trial subscription - PayPal gateway is missing on block checkout and PayPal button is missing on classic checkout (6853)

Dinamiko:
For free-trial subscriptions (initial cart total $0) with SDK v6 active, the PayPal gateway was missing on block checkout and the PayPal button was missing on classic checkout. v6's zero-total gating treated these carts as "no payment needed", even though they still require the PayPal save-without-purchase (vaulting) flow.

Solution: Carved free-trial subscription carts out of v6's zero-total gating in two places:                                                                                                
  - Classic: `SdkV6Manager::determine_render_places()` now renders the checkout button wrapper for free-trial carts (previously gated on `cart->needs_payment()`, which is false at $0).         
  - Block: the express `canMakePayment` in `checkout-block.js` now short-circuits to allow PayPal on free-trial carts, instead of running the amount-based eligibility check that rejects a $0 amount.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/test/checkout-block.test.js
+++ b/modules/ppcp-sdk-v6/resources/js/test/checkout-block.test.js
@@ -13,8 +13,14 @@ jest.mock(
 	{ virtual: true }
 );
 
-jest.mock( '../sdkLoader', () => ( { loadSdkV6: jest.fn() } ) );
-jest.mock( '../eligibility', () => ( { checkEligibility: jest.fn() } ) );
+const mockLoadSdkV6 = jest.fn();
+jest.mock( '../sdkLoader', () => ( {
+	loadSdkV6: ( ...args ) => mockLoadSdkV6( ...args ),
+} ) );
+const mockCheckEligibility = jest.fn();
+jest.mock( '../eligibility', () => ( {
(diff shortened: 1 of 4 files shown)


modules/ppcp-sdk-v6/resources/js/test/checkout-block.test.js
```diff
@@ -167,4 +173,41 @@ describe( 'checkout-block', () => {
...
+			);
+			expect( mockLoadSdkV6 ).not.toHaveBeenCalled();
+			expect( mockCheckEligibility ).not.toHaveBeenCalled();
+		} );
+
+		test.each( [
+			[ { paypal: true }, true ],
+			[ { paypal: false }, false ],
+		] )(
+			'on a non-free-trial cart, resolves to %s when eligibility reports %s',
```
Narek13: I think the `%s` placeholders are in the opposite order to the arguments, so this title prints backwards.

The rows are `[ eligibility, expected ]` and the callback destructures them in that order, so the first `%s` receives the eligibility object and the second receives the boolean, no ?

Narek13 (approved):
Thanks. LGTM. Added a small doubt in the test

PR #4674, opened 2026-09-02
Variable product smart buttons stay enabled with unselected attributes - SDK v6 (6868)

Dinamiko:
On classic variable-product pages with PayPal SDK v6, the express buttons (PayPal/Venmo/Pay Later/Google Pay) rendered fully enabled even when no variation was selected and "Add to cart" was disabled. The v6 bootstrap had no equivalent of v5's `SingleProductBootstrap` gating, and the v6 stylesheet (its own `gateway.css`, not ppcp-button's) lacked the `.ppcp-disabled` rule.                   
                                                                                                                                                                                             
This PR adds a small `productButtonGate` module that, on product pages, watches the native `.single_add_to_cart_button` state and toggles the wrapper's disabled state via the shared `ButtonDisabler`, dimming the buttons and routing a click to WooCommerce's own validation. Also added the missing `.ppcp-disabled` rule to the v6 stylesheet so the dimming and click-blocking actually apply.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/css/gateway.scss
+++ b/modules/ppcp-sdk-v6/resources/css/gateway.scss
@@ -22,3 +22,20 @@
 		width: 100%;
 	}
 }
+
+// Gating the express stack until the product form is ready to be purchased
+// (e.g. a variable product with no variation selected). The wrapper carries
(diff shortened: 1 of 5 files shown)


Narek13 (approved):

PR #4680, opened 2026-09-03
Add frontend logging and more granular log prefixes (5664)

stracker-phil:
*Changelog:* `Add - Frontend errors can now be reported to the WooCommerce log #4680`

# Description

We did not have a way to log pure front-end errors yet (like a wallet payment error).

Two main issues:
1. Anything that goes wrong in the browser reached the console at best but stays hidden for merchants/support/developers
2. Such errors were exceptionally intransparent on mobile devices - even for developers - as the browser console is difficult to access there

## ✅ What this adds

**A frontend log endpoint.** `FrontendLogEndpoint` takes a tag, an event slug, a detail string and a level, and writes one `[tag] event: detail` line through the plugin's usual logger. It is generic: any module's frontend can report through it, and the tag keeps each frontend's lines apart.

**Better log prefixes.** Log entries have a type-portion that helps us to identify the type of request; available types are

- `#1234.CLI -` request triggered via CLI
- `#1234.CRON -` scheduled background task
- `#1234.REST -` any REST endpoint
- `#1234.WC -` a `?wc-ajax=` request
- `#1234.AJAX -` a generic admin-ajax request
- `#1234.FRONT -` a new front-end log entry

## 🎯 First consumer: the wallets

In this PR only Apple Pay and Google Pay report payment issues in v5 and v6. These issues are hard to trace, as they happened in the browser without ever reaching the WP backend. This gap is now closed.

## 🔁 Refactoring opportunity

Note that `ppcp-axo` already has its own frontend log endpoint, `ppc-frontend-logger`, which does much the same job. It is left untouched here. Folding it onto the generic endpoint retires a public WC-AJAX action, which belongs in its own PR.

## 🧪 How to verify

1. Enable **Logging** in the PayPal settings, then open the classic checkout.
2. In the browser console:

```js
const ajax =
	window?.PayPalCommerceGateway?.ajax?.frontend_log ??
	window?.wc_ppcp_sdk_v6?.ajax?.frontend_log;

await fetch( ajax.endpoint, {
	method: 'POST',
	credentials: 'same-origin',
	headers: { 'Content-Type': 'application/json' },
	body: JSON.stringify( {
		nonce: ajax.nonce,
		tag: 'manual',
		event: 'test',
		message: 'HTTP 503 Service Unavailable',
	} ),
} ).then( ( response ) => response.json() );
```

3. **WooCommerce > Status > Logs**, source `woocommerce-paypal-payments`, carries `#1234.FRONT - [manual] test: HTTP 503 Service Unavailable`, with no `[New Request]` line before it.

## Screenshot

2026-09-16 - new-log-format


modules/ppcp-order-endpoints/src/Endpoint/FrontendLogEndpoint.php
```diff
@@ -0,0 +1,111 @@
...
+				// [New Request] entry would.
+				add_filter(
+					'woocommerce_paypal_payments_log_request_kind',
+					static function (): string {
+						return 'FRONT';
+					}
+				);
+				add_filter( 'woocommerce_paypal_payments_skip_new_request_log', '__return_true' );
+
+				$this->logger->log( $this->level( $data ), $this->line( $data ) );
```
Narek13: Nothing limits how often this can be called, and the nonce is in the page source, so any visitor can drive it. Each call appends a line to the merchant's log.

That is narrower than it first looks — `woocommerce.logger.woocommerce` is a `NullLogger` unless logging is enabled, so a default store writes nothing. But logging is exactly what support turns on while diagnosing a problem, which is when a store can least afford a log filling the disk. And this PR makes the endpoint the one every frontend is meant to use, with `ppcp-axo`'s to be folded in later, so whatever guard it has is the guard they all get.

There is already a tool for it in the tree: `SdkV6\Helper\RateLimiter`, transient-based with a max per window.

Worth noting `ppcp-axo`'s existing logger has no limit either, so this is not a regression — but it is the moment the surface becomes generic and documented, which seems like the moment to decide.

modules/ppcp-order-endpoints/src/Endpoint/FrontendLogEndpoint.php
```diff
@@ -0,0 +1,111 @@
...
+		$event  = $this->string_field( $data, 'event' ) ?: 'unknown';
+		$detail = $this->string_field( $data, 'message' );
+
+		$line = sprintf( '[%1$s] %2$s', $tag, $event );
+
+		if ( '' !== $detail ) {
+			$line .= ': ' . $detail;
+		}
+
+		return substr( $line, 0, self::MAX_LINE_LENGTH );
```
Narek13: `substr` counts bytes, so a line that reaches the cap mid-character writes a broken UTF-8 sequence to the log. Easy for a wallet error carrying a non-ASCII name or address to hit.

`mb_substr` would cut on character boundaries instead.

modules/ppcp-order-endpoints/src/Endpoint/FrontendLogEndpoint.php
```diff
@@ -0,0 +1,111 @@
...
+				);
+				add_filter( 'woocommerce_paypal_payments_skip_new_request_log', '__return_true' );
+
+				$this->logger->log( $this->level( $data ), $this->line( $data ) );
+			}
+
+			wp_send_json_success();
+		} catch ( NonceValidationException $error ) {
+			// Fire-and-forget endpoint, response data is never parsed, no need to indicate failures.
+			wp_send_json_success();
```
Narek13: Returning success on a nonce failure is a reasonable call for fire-and-forget — but it means the endpoint drops reports silently in one case worth thinking about: a page served from a cache with a stale nonce.

That is a plugin whose pages are often cached, and the reports being dropped are the ones you would most want when diagnosing exactly that sort of environment. Logging the rejection server-side, even at debug, would leave a trace that something tried and was turned away.

Narek13 (approved):
Thanks for working on this. Added 1 minor suggestion which you can ignore.

modules/ppcp-order-endpoints/src/Endpoint/FrontendLogEndpoint.php
stracker-phil: Right, and in this case I chose the `substr()` intentionally - the `MAX_LINE_LENGTH` is 1024, i.e., 1 kb of data max.

The only reason to trim the line is to prevent bloat in the log file, and the first 1024 bytes of a message usually give enough insight into an error to understand what happens. We're logging error messages and front-end API status details, which are usually plain English and much shorter than the max length

Narek13: Thanks for the explaination. Agreed :)

PR #4682, opened 2026-09-03
fix(sdk-v6): keep Pay Later messaging on one stack on home and shop

Narek13:
### Description

On home and shop, whether the Pay Later banner appeared came down to the **mini-cart button** — a setting with nothing to do with messaging. With the mini-cart enabled, the sitewide fallback in `SdkV6Manager::resolve_should_load()` claimed the page and the v5 stack was swapped for `DisabledSmartButton`, so the banner vanished. With it disabled, v5 stayed and drew the banner. Same site, same v6, opposite results — that inconsistency is the actual defect behind the report, rather than the banner being visible as such.

v6 has no message hook for home or shop (`messages_render_hook()` handles checkout, cart, product and pay-now only), so it cannot render one there yet.

Per the discussion on this: v6 owns messaging wherever it is active, rather than splitting it with v5 per page. These pages are now claimed whenever v5 would have had a message for them, so the banner is withheld consistently instead of depending on an unrelated toggle. **Stored configuration is untouched**, so the message returns once v6 renders these locations as text (planned for 4.1.4).

Scoped to an empty page context, so the block and page-context locations keep deciding without consulting messaging eligibility — an invariant `testShouldLoadOnCurrentPageInBlockContextsIsUnaffectedByMessagingEligibility` already pins.

**Impact.** No signature or hook changes. Behaviour change: with v6 active, a merchant who had Pay Later messaging configured for home/shop no longer sees it on those pages. That was already the case whenever the mini-cart button was enabled; this makes it consistent. Everything else is untouched — the classic and block checkout, cart, product and pay-now messaging all decide exactly as before, and the mini-cart itself is unaffected.

Not tested under multisite.

### Steps to Test

Requires the v6 SDK active and Pay Later messaging enabled for Home and Shop (on a site upgraded from v5 the settings migration carries those over; on a fresh site Home defaults to off).

1. Set the **mini-cart** button to **disabled** in the Styling tab — this is the combination that leaked the banner.
2. Visit the home page and the shop page. No Pay Later banner should appear. On `dev/release/4.1.3` the banner renders on both.
3. Enable the mini-cart button and revisit both pages. Still no banner — the point of the change is that the two settings now agree.
4. Regression: confirm Pay Later messaging still renders where v6 supports it — classic checkout, cart and product pages.
5. Regression: confirm the mini-cart button itself still renders and pays.
6. Regression: with the v6 SDK disabled, confirm the v5 banner still renders on home and shop as before.

Note the v6 client-token endpoint is rate limited — repeated page loads return 429 and the buttons stop rendering entirely, which looks like a failure but is not.

### Documentation

### Changelog Entry

> Fix: Pay Later messaging on the home and shop pages no longer appears or disappears depending on the mini-cart button setting when the v6 SDK is active.

### DIFF
--- a/modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
+++ b/modules/ppcp-sdk-v6/src/Assets/SdkV6Manager.php
@@ -573,6 +573,24 @@ class SdkV6Manager {
 			return true;
 		}
 
+		// Home and shop, where v5 places a Pay Later message and v6 has no
+		// message hook of its own yet. Whether v5 rendered there came down to the
+		// unrelated mini-cart setting: with the mini-cart on, the fallback below
+		// claimed the page and v5 went dark; with it off, v5 stayed and drew the
+		// banner. Claim the page either way, so one stack owns messaging
+		// everywhere rather than v6 and v5 splitting it per page. The message is
+		// withheld until v6 can draw it itself, at which point should_load_messages()
+		// above claims the page first and this branch stops being reached.
+		//
+		// Scoped to the empty page context so the block and page-context
+		// locations keep deciding without consulting messaging, as
+		// testShouldLoadOnCurrentPageInBlockContextsIsUnaffectedByMessagingEligibility
+		// pins.
+		$message_location = $page_location ? '' : $this->context->location();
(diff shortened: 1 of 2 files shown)


danieldudzic (approved):
That code comment in `SdkV6Manager.php` seems quite lengthy, but I guess it actually communicates useful context.

LGTM

PR #4686, opened 2026-09-04
PayPal Subscriptions + Pay Now Experience: "Cart changed after subscription approval" error (6886)

Dinamiko:
With PayPal Subscriptions + Pay Now enabled, buying a subscription from the cart button always failed with "Cart changed after subscription approval" and a Failed order. The subscription replay-guard compares the order's cart hash against the hash stored at approval, but the express order-creation helper (`WooCommerceOrderCreator::create_from_paypal_order()`) never set a cart hash so the empty hash never matched. It only affected Pay Now, since final-review checkout lets WooCommerce set the hash itself.                                       
                                                                                                                                                                                             
`create_from_paypal_order()` now sets the order's cart hash from the CartData (same value captured at approval), mirroring WooCommerce core's checkout. Express orders now pass the guard while it stays effective against genuinely changed/replayed carts.


Narek13 (approved):
Thanks. LGTM

PR #4688, opened 2026-09-04
Add a BCDC redirect gateway to Block Checkout (6882)

danieldudzic:
### Description

The BCDC button has been removed from Block Checkout (express button section) under SDK v6, leaving merchants with BCDC and no ACDC with no card option there at all. This adds the redirect gateway to replace it: a "Debit & Credit Cards" row whose "Proceed to PayPal" button sends the buyer to PayPal's hosted **guest card** page.

Also resolves reports that BCDC does not appear on block checkout, though not its account-specific 403 from find-eligible-methods.

Includes an adjacent fix in `ReturnUrlEndpoint`: it resolved the order's gateway from available gateways, but continuation mode leaves only PayPal available, so any non-PayPal gateway that sent the buyer there failed with "Payment gateway is unavailable" on return. Now falls back to registered gateways (affects AXO too), and is untested because the method is private behind a caller that ends in `exit()`.

### Screenshots

| Before - BCDC Missing | After - BCDC as a regular gateway |
| :--- | :--- |
|bcdc_missing|fixed_bcdc|

### Steps to Test

Requires the v6 SDK flag, BCDC enabled, ACDC disabled. **Use a private window** - PayPal shows the login screen to buyers it recognises from cookies, whatever the landing page says.

1. Set the PayPal landing page to **"Login page"** - the setting this gateway must override.
2. As a guest, add a product and open the **block** checkout. Express Checkout shows PayPal only, no card button.
3. Select **Debit & Credit Cards** - the order button reads **Proceed to PayPal**.
4. Click it. PayPal shows **card fields**, with the billing address pre-filled from checkout.
5. Complete payment. You land on order-received; the order is `processing` with payment method **Debit & Credit Cards** (`ppcp-card-button-gateway`), not PayPal.
6. Repeat with the **PayPal** method - it must still show the **login** screen, proving the override is scoped to cards.
7. Regression, classic checkout: the "Debit & Credit Cards" row still renders PayPal's card button inline.
8. Regression, block cart: no card option appears.

### DIFF
--- a/modules/ppcp-sdk-v6/resources/js/checkout-block.js
+++ b/modules/ppcp-sdk-v6/resources/js/checkout-block.js
@@ -212,6 +212,10 @@ if ( config && config.page_context && config.continuation ) {
 	 *                                          supports. Deliberately without
 	 *                                          a default, so a wallet cannot
 	 *                                          inherit PayPal's list.
+	 * @param {string}   [args.label]           - The buyer-facing label, for a
+	 *                                          method whose name is the
+	 *                                          merchant's to set rather than a
+	 *                                          brand.
 	 */
 	const registerExpress = ( {
 		name,
@@ -220,8 +224,9 @@ if ( config && config.page_context && config.continuation ) {
 		content,
 		isDeviceCapable,
 		features,
+		label: labelOverride,
 	} ) => {
-		const label = fundingSourceLabel( fundingSource );
+		const label = labelOverride || fundingSourceLabel( fundingSource );
 
 		registerExpressPaymentMethod( {
 			/*
@@ -242,7 +247,7 @@ if ( config && config.page_context && config.continuation ) {
 			label: createElement( 'div', null, label ),
 			ariaLabel: label,
 			content,
-			edit: createElement( V6EditorPreview, { fundingSource } ),
+			edit: createElement( V6EditorPreview, { fundingSource, label } ),
 			canMakePayment: async ( { cartTotals } = {} ) => {
 				if ( isDeviceCapable && ! isDeviceCapable() ) {
 					return false;
@@ -312,6 +317,34 @@ if ( config && config.page_context && config.continuation ) {
 			features: settings.supported_features,
 		} );
 	}
+}
+
+// BCDC. A plain redirect method here, with no SDK involvement: the v6 guest
+// session renders its card form inline, which needs an express placement that
+// WooCommerce then disables, so "Place order" hands off to PayPal instead.
+// CardButtonGateway already does that - with no approved order in the session it
+// creates one and returns PayPal's hosted checkout URL to redirect to.
+//
+// Skipped in continuation mode, like the card fields.
+if ( config?.card_button?.block_method && ! config.continuation ) {
+	const cardButtonDescription = __(
+		'Pay securely with your credit or debit card. You will be redirected to PayPal to complete your payment.',
+		'woocommerce-paypal-payments'
+	);
+
+	registerPaymentMethod( {
+		name: config.card_button.payment_method,
+		label: createElement( 'div', null, config.card_button.title ),
+		ariaLabel: config.card_button.title,
+		content: createElement( 'div', null, cardButtonDescription ),
+		edit: createElement( 'div', null, cardButtonDescription ),
+		canMakePayment: () => true,
+		supports: {
+			features: gatewayFeatures( config.card_button.supported_features ),
+			// PayPal's hosted card page cannot vault into WooCommerce.
+			showSaveOption: false,
+		},
+	} );
 }
 
 /**
--- a/modules/ppcp-sdk-v6/resources/js/checkout-block.js
+++ b/modules/ppcp-sdk-v6/resources/js/checkout-block.js
@@ -340,17 +340,14 @@ if ( config && config.page_context && config.continuation ) {
 	}
 }
 
-// BCDC. A plain redirect method here, with no SDK involvement: the v6 guest
-// session renders its card form inline, which needs an express placement that
-// WooCommerce then disables, so "Place order" hands off to PayPal instead.
-// CardButtonGateway already does that - with no approved order in the session it
-// creates one and returns PayPal's hosted checkout URL to redirect to.
-//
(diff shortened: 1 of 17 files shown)


modules/ppcp-sdk-v6/resources/js/checkout-block.js
```diff
@@ -314,6 +339,39 @@ if ( config && config.page_context && config.continuation ) {
...
+		placeOrderButtonLabel: placeOrder.text,
+		canMakePayment: () => true,
+		supports: {
+			features: gatewayFeatures( config.card_button.supported_features ),
+			// PayPal's hosted card page cannot vault into WooCommerce.
+			showSaveOption: false,
+		},
+	} );
+
+	registerPlaceOrderLabel( cardButtonId, placeOrder.text );
```
Narek13: This is the clash with #4704 (PCP-6888 + PCP-6501, also 4.1.4). Disclosure: that PR is mine.

It changes `sdk-v6.blocks.place-order-data` to return `array( 'enabled' => … )` only — `text` and `description` are gone. Whichever of us merges second, the effects here are quiet rather than loud:

- `registerPlaceOrderLabel()` returns early on `! label`, so no filter is registered and the helper becomes dead code.
- `placeOrderButtonLabel: placeOrder.text` (line 363) and `placeOrderButtonDescription: placeOrder.description` both become `undefined`.
- The button reads "Place order", so steps 3 and 5 of the test plan no longer describe what happens.
- `registers a placeOrderButtonLabel checkout filter … when place_order.text is set` keeps passing, because the fixture sets `text` explicitly. A green test over a path production can no longer produce is the part I would least like to leave behind.

Nothing breaks for a shopper. But it needs a merge-order decision rather than a code change, and the product question — PCP-6888 says never relabel, this adds a relabeller — is worth putting to Niklas before either lands.

modules/ppcp-wc-gateway/src/Processor/OrderProcessor.php
```diff
@@ -234,15 +234,25 @@ public function process_captured_and_authorized( WC_Order $wc_order, Order $orde
...
-	 * @param string   $funding_source The funding source (e.g. 'paypal', 'venmo').
+	 * @param WC_Order    $wc_order The WC order.
+	 * @param string      $funding_source The funding source (e.g. 'paypal', 'venmo').
+	 * @param string|null $landing_page An ExperienceContext::LANDING_PAGE_* value to
+	 *                                  use instead of the merchant's setting.
 	 * @return Order
 	 * @throws RuntimeException If order creation fails.
 	 */
-	public function create_order( WC_Order $wc_order, string $funding_source = 'paypal' ): Order {
+	public function create_order( WC_Order $wc_order, string $funding_source = 'paypal', ?string $landing_page = null ): Order {
```
Narek13: The new parameter is optional, so every caller is fine. Subclasses are not.

`OrderProcessor` is non-final, container-wired as `wcgateway.order-processor`, and this method is public. PHP requires an overriding child to declare at least the parent's parameters, so any subclass still carrying the two-parameter `create_order()` now **fatals on load**, not at call time:

> Declaration of X::create_order(WC_Order, string) must be compatible with OrderProcessor::create_order(WC_Order, string, ?string)

AGENTS.md treats overridable classes as contracts and asks for a backward-compatibility statement in the description for exactly this. The body has none at the moment, and it would want to cover all three surfaces this PR touches: this signature, the `ReturnUrlEndpoint` behaviour change, and `ExperienceContextBuilder::with_landing_page()` (that last one is purely additive and fine).

The refactor itself reads well — building the context once and conditionally layering the landing page is clearer than the nested version it replaces.

modules/ppcp-wc-gateway/src/Endpoint/ReturnUrlEndpoint.php
```diff
@@ -329,6 +329,11 @@ private function get_payment_gateway( string $payment_method ) {
 			return $available_gateways[ $payment_method ];
 		}
 
-		return null;
+		// Returning with an approved order puts the checkout in continuation mode,
+		// where DisableGateways offers PayPal alone - so the gateway that sent the
+		// buyer to PayPal is missing from the list above on the way back.
+		$registered_gateways = WC()->payment_gateways->payment_gateways();
```
Narek13: The fix is right for the problem, and I confirmed `WC()->payment_gateways` is already dereferenced on line 326 directly above, so this adds no new global-state assumption.

One widening worth naming in the comment: `payment_gateways()` returns *registered* gateways, which includes ones the merchant has since **disabled** — so the fallback is broader than "gateways hidden by continuation mode". For a buyer returning from PayPal that is probably what you want, since the order already exists and needs completing either way. Worth a sentence saying so, so the next reader does not have to work out whether it was deliberate.

On "untested because the method is private behind a caller that ends in `exit()`": the blocker is the `exit()`, not the visibility. The repo already leans on subclass-for-test seams in several places (`SpyablePayPalGateway` in `WcGatewayTest`, among others), so making this `protected` would make the lookup directly testable without touching the caller at all. Optional — but untested-by-construction code tends to stay that way.

modules/ppcp-sdk-v6/resources/js/checkout-block.js
Narek13, line 372: This is the clash with #4704. Disclosure: that PR is mine.

It changes `sdk-v6.blocks.place-order-data` to return `array( 'enabled' => … )` only `text` and `description` are gone. Whichever of us merges second, the effects here are quiet rather than loud:

- `registerPlaceOrderLabel()` returns early on `! label`, so no filter is registered and the helper becomes dead code.
- `placeOrderButtonLabel: placeOrder.text` (line 363) and `placeOrderButtonDescription: placeOrder.description` both become `undefined`.
- The button reads "Place order", so steps 3 and 5 of the test plan no longer describe what happens.
- `registers a placeOrderButtonLabel checkout filter … when place_order.text is set` keeps passing, because the fixture sets `text` explicitly.

Narek13 (approved):
Thanks a lot for working on this, @danieldudzic . The PR looks good to me.
My only concern is a possible merge conflict that needs your attention.

PR #4690, opened 2026-09-06
Fix negative tax_total when a discount is not reported by WooCommerce (6762)

danieldudzic:
### Description

`AmountFactory` treats `get_total()` as authoritative and pushes any gap between it and the summed breakdown into the tax line. That was sized for 1 cent inclusive-tax rounding. When a discount is not reported by the standard getters, a plugin calling `set_total()` directly or an order read before `get_total_discount()` populates, the whole discount lands there and drives tax negative. PayPal then rejects the Level 2 card data with `CANNOT_BE_NEGATIVE` on `tax_total`.

Now, when the gap is deeper than the tax, the gap is booked as a discount and the tax is kept, floored at zero. The floor matters because a negative fee can make WooCommerce report negative tax on its own, which would otherwise be passed straight through. Both values stay non-negative and the breakdown still sums to the total.

Only `from_wc_order()` is affected. Both cart methods derive the total from components, so they were never exposed.

Related: #4689 fixes the Account Funds case at the source (complementary, no shared files). 

### Steps to Test

Requires store country US, currency USD, ACDC enabled, and Level 2/3 processing on, otherwise the Level 2 data is never built.

The simplest deterministic trigger is the Account Funds scenario from #4689, since a draft-order read is not reproducible on demand:

1. With Account Funds active, give a test customer 200.00 store credit
2. Add a product costing more than that, for example 399.98
3. On the classic checkout, apply the credit so part of the total is still charged
4. Pay with Debit & Credit Cards using a sandbox card
5. Check the newest `woocommerce-paypal-payments` log under WooCommerce > Status > Logs

On `dev/develop` the request carries a negative `level_2.tax_total` and PayPal rejects it. On this branch the tax is the real amount and the gap appears as `discount`.

### DIFF
--- a/modules/ppcp-api-client/src/Factory/AmountFactory.php
+++ b/modules/ppcp-api-client/src/Factory/AmountFactory.php
@@ -216,13 +216,23 @@ class AmountFactory {
 		} else {
 			$wc_total              = (float) $order->get_total();
 			$wc_total_cents        = (int) round( $wc_total * 100 );
+			$discount_cents        = (int) round( $discount_value * 100 );
 			$component_total_cents = (int) round( $item_total_val * 100 )
 				+ (int) round( $shipping_val * 100 )
 				+ (int) round( $taxes_val * 100 )
-				- (int) round( $discount_value * 100 );
-			$taxes_cents           = (int) round( $taxes_val * 100 ) + ( $wc_total_cents - $component_total_cents );
-			$taxes                 = new Money( $taxes_cents / 100, $currency );
-			$total                 = new Money( $wc_total, $currency );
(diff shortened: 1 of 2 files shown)


stracker-phil (approved):
LGTM

---

Note and additional context:

This PR hardens the `AmountFactory` so an inconsistent order still produces a transaction PayPal accepts.

It is not doing or correcting a tax calculation. Where the incoming order is already fiscally inconsistent, the payload carries that inconsistency forward: the reported tax is WooCommerce's own figure, and the unexplained remainder is booked as a discount so every field stays non-negative and the breakdown still sums to the total.

PR #4694, opened 2026-09-08
Add snapshot mechanism to rebuild shipping line items (6763)

mmaymo:
## 🐛 What and why

When an Advanced Card Processing payment is declined by the processor, the WooCommerce order is marked `failed` but kept, and the session still points at it via `order_awaiting_payment`. On retry, WooCommerce resumes that same order: it wipes the line items and rebuilds them from the cart. The rebuild always restores the shipping *total*, but restores the shipping *line item* only when the previously chosen shipping method still resolves to a rate in the recalculated packages.

When that lookup misses, the order ends up with a non-zero shipping total and **no shipping line item at all**. The buyer is charged the full amount including shipping, and every debug log line shows the correct shipping breakdown — so nothing looks wrong — but the merchant is left with an order they cannot fulfil, invoice or report shipping against. A merchant log for order 22028 confirms both attempts patched the same `custom_id`, proving the order was resumed rather than recreated.

## ⚙️ How it works now

`ResumedOrderShippingRestorer::snapshot()` runs on `woocommerce_resume_order`, which WooCommerce fires immediately before `WC_Order::remove_order_items()` — while the shipping items still exist. It records `method_title`, `method_id`, `instance_id`, `total`, `total_tax`, `taxes`, `tax_status` and item meta as plain data, keyed by order ID in a private property that lives for the request only.

`ResumedOrderShippingRestorer::restore()` runs on `woocommerce_checkout_order_processed`, after the rebuild has been saved and before the gateway's `process_payment()`. It re-fetches the order and acts **only** when `get_shipping_total()` is non-zero and `get_shipping_methods()` is empty — the exact divergence — then re-adds fresh `WC_Order_Item_Shipping` objects and calls `save()`. No total is recalculated, so the amount `AmountFactory::from_wc_order()` sends to PayPal is untouched.

The whole body of both methods sits inside a `try`/`catch ( Throwable )`. A failure logs an error naming the order ID and the unbacked shipping total, then returns — leaving exactly today's behaviour. There is no branch that can reject, block or fail a payment.

### Covered by unit tests

All seven acceptance criteria are covered by `ResumedOrderShippingRestorerTest` (7 tests, 35 assertions, green):

- ✅ All seven shipping-item fields are captured before `remove_order_items()` deletes them, and reproduced exactly on the restored item — including taxes on a taxable rate.
- ✅ A resumed order left with an 18.45 shipping total and zero shipping methods gets exactly one persisted `WC_Order_Item_Shipping` matching the original method title, method ID, instance ID and total.
- ✅ A resumed order whose rebuild already restored the line item is left alone — same item count, same item ID, no duplicate.
- ✅ A resumed order with a 0.00 shipping total and no shipping items gets nothing added.
- ✅ A fresh checkout with no `woocommerce_resume_order` leaves the order untouched, shipping total included.
- ✅ A throw during the restore is caught and logged with the order ID, and later callbacks on `woocommerce_checkout_order_processed` still run, so `process_payment()` is reached.
- ✅ `get_shipping_total()` and `get_total()` are byte-identical strings before and after a successful restoration.

Note that three of these (already-has-items, zero-total, no-resume) also pass without the fix. They are regression guards against an over-reaching implementation, not red-by-design proofs.

### Mutation-tested

The suite was checked against **13 hand-written mutants** of `ResumedOrderShippingRestorer` — one per behaviour the criteria claim. All 13 are killed: dropping either guard, capturing an empty snapshot, skipping `save()`, losing any one restored field, narrowing the `catch`, silencing the log, dropping the order ID from the message, restoring without a snapshot, and recalculating totals.

One mutant initially survived and is worth a reviewer's attention. Inserting `calculate_totals()` before `save()` did **not** fail the byte-identical test, because recalculation happened to arrive at the same numbers — the assertion was satisfied by coincidence rather than by the invariant. That test now pins the order total to a sentinel value recalculation cannot reproduce, so any recalculation inside the restorer is detectable. The shipping total is still compared against its real rebuilt value.

### Not covered by unit tests

- ❌ End-to-end behaviour against a real processor decline (`FraudProcessorResponse` 1300/1330/1335) — the tests reproduce the resulting order state via the real checkout resume path, but do not drive an actual declined capture.
- ❌ Multisite. The restorer reads and writes only order data through `wc_get_order()` and `WC_Order::save()`, so it carries no site-scoped option or path assumptions, but it was not tested under multisite.

### Constructor signature changed

None. `ResumedOrderShippingRestorer` is new: `__construct( LoggerInterface $logger )`, wired from `woocommerce.logger.woocommerce`.

### New hook consumers

Two callbacks are attached at priority 10, to WooCommerce core hooks (`woocommerce_resume_order`, `woocommerce_checkout_order_processed`) — not to plugin hooks. No `ppcp_*` or `woocommerce_paypal_payments_*` hook is added, removed, renamed or reordered, so no third-party callback contract changes.

### Test base class signature changed

`IntegrationMockedTestCase::bootstrapModule()` gained an optional second parameter, `array $overriddenExtensions = []`. Existing callers are unaffected.

This was necessary, and is worth a reviewer's attention: **overriding a service ID in `bootstrapModule()` does not reliably replace that service**. `modules/ppcp-wc-gateway/extensions.php` extends `woocommerce.logger.woocommerce` with an extender that ignores `$previous` and rebuilds the logger from scratch, so a service-level override is silently discarded. The failure-path test needs a mock logger, so the override is now registered as an extension — the test module is added last, so its extension runs last and wins. Any other test that assumed a service override survives module extensions has the same latent problem.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Helper/ResumedOrderShippingRestorer.php
+++ b/modules/ppcp-wc-gateway/src/Helper/ResumedOrderShippingRestorer.php
@@ -1,0 +1,229 @@
+<?php
+/**
+ * Restores the shipping line items WooCommerce drops when it resumes a failed order.
+ *
+ *
+ * This helper snapshots the shipping items while they still exist and re-adds them after
+ * the rebuild, only when the divergence actually happened.
+ *
+ * @package WooCommerce\PayPalCommerce\WcGateway\Helper
+ */
+
+declare(strict_types=1);
+
+namespace WooCommerce\PayPalCommerce\WcGateway\Helper;
+
+use Psr\Log\LoggerInterface;
+use Throwable;
+use WC_Order;
+use WC_Order_Item_Shipping;
+
+/**
+ * ResumedOrderShippingRestorer class.
+ */
+class ResumedOrderShippingRestorer {
+
+	/**
+	 * The logger.
+	 *
+	 * @var LoggerInterface
+	 */
+	private $logger;
+
+	/**
+	 * The shipping line items captured before the rebuild, keyed by order ID.
+	 *
+	 * Lives for the duration of the request only: the snapshot is taken and consumed
+	 * within a single checkout submission.
+	 *
+	 * @var array<int, array<int, array<string, mixed>>>
+	 */
+	private $snapshots = array();
+
+	/**
+	 * ResumedOrderShippingRestorer constructor.
+	 *
+	 * @param LoggerInterface $logger The logger.
+	 */
+	public function __construct( LoggerInterface $logger ) {
+		$this->logger = $logger;
+	}
+
+	/**
+	 * Records the shipping line items of an order that is about to be rebuilt.
+	 *
+	 * Hooked to `woocommerce_resume_order`, which fires immediately before
+	 * WC_Order::remove_order_items() deletes them.
+	 *
+	 * @param int $order_id The ID of the order being resumed.
+	 * @return void
+	 */
+	public function snapshot( int $order_id ): void {
+		try {
+			if ( $order_id <= 0 ) {
+				return;
+			}
+
+			$wc_order = wc_get_order( $order_id );
+			if ( ! $wc_order instanceof WC_Order ) {
+				return;
+			}
+
+			$items = array();
+
+			foreach ( $wc_order->get_shipping_methods() as $item ) {
+				if ( ! $item instanceof WC_Order_Item_Shipping ) {
+					continue;
+				}
+
+				$items[] = array(
+					'method_title' => $item->get_method_title(),
+					'method_id'    => $item->get_method_id(),
+					'instance_id'  => $item->get_instance_id(),
+					'total'        => $item->get_total(),
+					'total_tax'    => $item->get_total_tax(),
+					'taxes'        => $item->get_taxes(),
+					'tax_status'   => $item->get_tax_status(),
+					'meta'         => $this->meta_of( $item ),
+				);
+			}
+
+			if ( ! $items ) {
+				return;
+			}
+
+			$this->snapshots[ $order_id ] = $items;
+		} catch ( Throwable $exception ) {
+			unset( $this->snapshots[ $order_id ] );
+
+			$this->logger->error(
+				sprintf(
+					'Could not snapshot the shipping line items of resumed order %1$d: %2$s',
+					$order_id,
+					$exception->getMessage()
+				)
+			);
+		}
+	}
+
+	/**
+	 * Re-adds the snapshotted shipping line items when the rebuild dropped them.
+	 *
+	 * Restoration is best-effort: a failure is logged and swallowed so that checkout and
+	 * payment proceed exactly as they do today.
+	 *
+	 * @param int $order_id The ID of the processed order.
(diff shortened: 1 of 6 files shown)


modules/ppcp-wc-gateway/src/Helper/ResumedOrderShippingRestorer.php
```diff
@@ -0,0 +1,229 @@
...
+				}
+
+				$items[] = array(
+					'method_title' => $item->get_method_title(),
+					'method_id'    => $item->get_method_id(),
+					'instance_id'  => $item->get_instance_id(),
+					'total'        => $item->get_total(),
+					'total_tax'    => $item->get_total_tax(),
+					'taxes'        => $item->get_taxes(),
+					'tax_status'   => $item->get_tax_status(),
```
Narek13, line 86: `tax_status` is captured here but never restored as `item_from()` doesn't set it, since `WC_Order_Item_Shipping::get_tax_status()` is hardcoded to `taxable`.

So this line stores something nothing reads. Worth dropping.

modules/ppcp-wc-gateway/src/Helper/ResumedOrderShippingRestorer.php
```diff
@@ -0,0 +1,229 @@
...
+
+			$wc_order = wc_get_order( $order_id );
+			if ( ! $wc_order instanceof WC_Order ) {
+				unset( $this->snapshots[ $order_id ] );
+
+				return;
+			}
+
+			$shipping_total = (string) $wc_order->get_shipping_total();
+			if ( (float) $shipping_total <= 0.0 ) {
```
Narek13, line 134: This early return, and the `get_shipping_methods()` one below it, leave the snapshot in `$this->snapshots`. Every other exit point clears it.

Narek13 (commented):
Thanks a lot for working on this @mmaymo.

One thing I would like your read on before this goes in.

**FYI:**
`restore()` re-adds the snapshotted item totals, but the order's shipping *total* came from the rebuild, and the two can be describing different shipping choices.

The guard checks only that the shipping total is non-zero and that there are no shipping methods. It never compares the snapshot against the total it is backing. So:

1. Card is declined, order is left `failed`, items still reflect the buyer's first shipping choice.
2. Buyer changes the shipping method and resubmits.
3. `woocommerce_resume_order` fires, the snapshot captures the **old** method and its total.
4. The rebuild sets the shipping total from the **current** cart, and misses on the rate lookup — the bug this PR fixes.
5. `restore()` re-adds the old item verbatim.

The order then carries, say, a shipping total of 12.00 with a line item reading "Flat rate 5.00". Today that order has no line item at all; afterwards it has one that looks authoritative and disagrees with the total the buyer was charged. For a change whose whole purpose is that merchants can fulfil and invoice against the shipping line, a confidently wrong line seems worse than a missing one.

Two ways out, whichever fits your intent: skip the restore when the snapshotted totals do not sum to `get_shipping_total()` and log that instead, or keep restoring the method identity but take the amount from the order rather than the snapshot. The first preserves the "only act on the exact divergence" principle the rest of the class follows.

Also please check the CI

PR #4695, opened 2026-09-08
fix(wc-gateway): ignore captures that never took money when reading the transaction id

Narek13:
### Description

`get_paypal_order_transaction_id()` returned the first capture's id without looking at its status.

A vaulted-card subscription renewal is processed during order creation, so a decline comes back as an order whose first capture is already `DECLINED`. `RenewalHandler::handle_paypal_order()` stores that id as the WooCommerce transaction id at `:491-499`, and only then handles the status at `:501` and marks the order failed.

An order carrying that meta is then treated as already paid. `OrderProcessor::verify_order_can_be_processed()` refuses any order with a transaction id, whatever its status:

```php
if ( $wc_order->get_transaction_id() ) {
    $this->logger->info( 'Order #%d already has transaction ID "%s", skipping payment processing.' );
    return false;
}
```

So Subscriptions' automatic retries and the shopper's own order-pay page are both skipped, and the failed renewal cannot be paid again without the meta being deleted by hand.

The trait now scans for the first capture that is not `DECLINED` or `FAILED`. Measured against the exact PayPal order shape a declined renewal returns:

| Captures on the order | Before | After |
|---|---|---|
| `DECLINED` | `6VV364076U2109212` — stored, order blocked | `null`, order stays payable |
| `FAILED` | id stored | `null` |
| `COMPLETED` | id stored | unchanged |
| `PENDING` | id stored | unchanged, deliberately |
| `DECLINED` then `COMPLETED` | the **declined** id | the completed id |

That last row is a second, quieter bug the same change fixes: a declined capture used to mask a successful one on the same order.

`PENDING` is deliberately still returned, since it may yet settle and its id is the reference to keep; so are the refunded states, which describe money that did move. Denied authorizations are now skipped for the same reason as declined captures — with dead captures passed over, the authorization branch became reachable where it previously was not.

Fixing the shared trait covers all ten callers, including the renewal handler, the card gateway's vaulted-card finalization, the PayPal gateway and the four order-tracking integrations.

**Why only half the reported proposal.** The report also suggested loosening `verify_order_can_be_processed()` so a stale transaction id cannot block an order that still returns `needs_payment()`. That is not included: it removes the guard against double-processing, and in a race where a capture succeeded but the order status has not caught up, the order reads as needing payment and could be charged twice. This change stops the bad data being written; that one would make a safety check tolerate bad data. Worth revisiting with a narrower condition if the team wants belt-and-braces.

**What was not executed.** The declined-renewal flow itself. Staging a real decline needs a sandbox card that declines on a vaulted merchant-initiated transaction, which is a different path from an interactive decline. The unit tests and the before/after table above pin the decision precisely, but `RenewalHandler::handle_paypal_order()` was verified by reading rather than running. The reporter observed the full flow on a live store (order `#251686`), including the log line quoted above.

No signature or hook changes. Not tested under multisite.

### Steps to Test

The behaviour can be checked directly, without staging a decline:

1. Build a PayPal `Order` whose purchase unit has `payments.captures[0]` with status `DECLINED`, and pass it to `get_paypal_order_transaction_id()`. It should return `null`; on `dev/develop` it returns the declined capture's id.
2. Repeat with `COMPLETED`, `PENDING`, `REFUNDED` and `PARTIALLY_REFUNDED` — each should still return its id.
3. Repeat with a `DECLINED` capture followed by a `COMPLETED` one — it should return the completed id.

End to end, if a card that declines on renewal is available:

4. Create a subscription paid by a vaulted card, make the renewal decline, and trigger the renewal.
5. Confirm the renewal order is `failed` and has **no** `_transaction_id`.
6. Retry through the customer's order-pay page. The attempt should proceed, rather than logging `Order #… already has transaction ID "…", skipping payment processing.`
7. Regression: a successful renewal still records its capture id as the transaction id.

### Documentation

### Changelog Entry

> Fix: A declined subscription renewal no longer records the declined transaction against the order, which previously blocked every later retry and the customer's own payment page.

### DIFF
--- a/modules/ppcp-wc-gateway/src/Processor/TransactionIdHandlingTrait.php
+++ b/modules/ppcp-wc-gateway/src/Processor/TransactionIdHandlingTrait.php
@@ -12,6 +12,9 @@ namespace WooCommerce\PayPalCommerce\WcGateway\Processor;
 use Exception;
 use Psr\Log\LoggerInterface;
 use WC_Order;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\AuthorizationStatus;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\Capture;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\CaptureStatus;
 use WooCommerce\PayPalCommerce\ApiClient\Entity\Order;
 
 /**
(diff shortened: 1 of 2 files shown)


modules/ppcp-wc-gateway/src/Processor/TransactionIdHandlingTrait.php
```diff
@@ -78,16 +81,38 @@ public function get_paypal_order_transaction_id( Order $order ): ?string {
...
+				continue;
+			}
+
 			return $capture->id();
 		}
 
-		$authorization = $payments->authorizations()[0] ?? null;
-		if ( $authorization ) {
+		foreach ( $payments->authorizations() as $authorization ) {
+			if ( $authorization->status()->is( AuthorizationStatus::DENIED ) ) {
```
stracker-phil: nit: Why is the above condition extracted to a helper method but this isn't?

We could keep the code more consistent if we apply the same pattern in both checks.

Sample:

```php
foreach ( $payments->captures() as $capture ) {
	if ( $this->is_valid_capture( $capture ) ) {
		return $capture->id();
	}
}

foreach ( $payments->authorizations() as $authorization ) {
	if ( $this->is_valid_payment( $authorization ) ) {
		return $authorization->id();
	} 
}
```

stracker-phil (approved):
LGTM - only one nit about code layout, but the PR can be merged without cleanup

PR #4696, opened 2026-09-08
Hide PayPal gateway for free trial cart when no saved PayPal account

Dinamiko:
In Blocks checkout, a free-trial subscription cart showed the non-express PayPal gateway. Selecting it and clicking "Proceed to PayPal" failed with "No saved PayPal account." because that flow needs an already-vaulted PayPal/Venmo token.                                                                                                                             
                                                                                                                                                                                             
This PR hides the non-express PayPal gateway (`placeOrderEnabled = false`) when the cart is a free-trial cart and the customer has no saved PayPal/Venmo token. The PayPal express button still shows, so the customer can vault their account and place the order.

### DIFF
--- a/modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
+++ b/modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
@@ -271,6 +271,23 @@ class WooCommercePaymentTokens {
 	}
 
 	/**
+	 * Whether the given WP user has a saved PayPal or Venmo payment token.
+	 *
+	 * @param int $user_id WP user id.
+	 * @return bool
+	 */
+	public function has_paypal_or_venmo_token( int $user_id ): bool {
+		foreach ( $this->customer_tokens( $user_id ) as $token ) {
+			$name = $token['payment_source']->name() ?? '';
+			if ( 'paypal' === $name || 'venmo' === $name ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
 	 * Creates WC payment tokens for the given WP user id using PayPal payment tokens as source.
 	 *
 	 * @param array $customer_tokens PayPal customer payment tokens.
--- a/tests/PHPUnit/WcPaymentTokens/WooCommercePaymentTokensTest.php
+++ b/tests/PHPUnit/WcPaymentTokens/WooCommercePaymentTokensTest.php
@@ -113,4 +113,58 @@ class WooCommercePaymentTokensTest extends TestCase
 		// THEN the failure is treated the same as having no tokens.
 		$this->assertSame([], $result);
 	}
+
+	/**
+	 * Builds a customer token whose payment_source reports the given name, matching the
+	 * shape returned by PaymentTokensEndpoint::payment_tokens_for_customer().
+	 *
+	 * @param string $payment_source_name Name reported by the token's payment source.
+	 */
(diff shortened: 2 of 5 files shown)


modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
```diff
@@ -270,6 +270,23 @@ public function customer_tokens( int $user_id ): array {
 		return $customer_tokens;
 	}
 
+	/**
+	 * Whether the given WP user has a saved PayPal or Venmo payment token.
+	 *
+	 * @param int $user_id WP user id.
+	 * @return bool
+	 */
+	public function has_paypal_or_venmo_token( int $user_id ): bool {
```
Narek13, line 279: This adds an **uncached remote PayPal API call to the block checkout render path**. `customer_tokens()` calls `payment_tokens_for_customer()` with no caching or per-request memoization, and `get_payment_method_data()` runs on every checkout render — so this fires again on every cart update for any logged-in customer with a free-trial cart.

It also **fails closed**: `customer_tokens()` swallows `RuntimeException` and returns `[]`, so a transient API error hides the row from a customer who *does* have a saved account. That is the safe direction and they can still use the express button, but it is worth being deliberate about it and memoizing the result per request at minimum.

modules/ppcp-wc-payment-tokens/src/WooCommercePaymentTokens.php
```diff
@@ -270,6 +270,23 @@ public function customer_tokens( int $user_id ): array {
...
+		foreach ( $this->customer_tokens( $user_id ) as $token ) {
+			$name = $token['payment_source']->name() ?? '';
```
Narek13, line 281: The `?? ''` is dead code, and it guards the wrong thing. `PaymentSource::name()` is declared `: string` (`PaymentSource.php:49`), so the coalesce can never fire. The actual hazards — an absent `payment_source` key, or a null value — both fatal *before* `??` is reached, so this reads as protection that is not there.

`create_wc_tokens()` just above does the same unguarded access, so the shape is presumably reliable. If so, drop the misleading `??`. If not, guard the object instead:

```php
foreach ( $this->customer_tokens( $user_id ) as $token ) {
    $source = $token['payment_source'] ?? null;
    if ( $source instanceof PaymentSource
        && in_array( $source->name(), array( 'paypal', 'venmo' ), true ) ) {
        return true;
    }
}
```

tests/PHPUnit/WcPaymentTokens/WooCommercePaymentTokensTest.php
```diff
@@ -113,4 +113,58 @@ public function testReturnsEmptyArrayWhenTheEndpointFails()
...
+			->shouldReceive('payment_tokens_for_customer')
+			->with('CUST-1')
+			->andReturn([$this->tokenWithPaymentSourceName($payment_source_name)]);
+
+		$result = $this->sut->has_paypal_or_venmo_token(42);
+
+		$this->assertSame($expected, $result);
+	}
+
+	public function payment_source_name_provider(): array
```
Narek13, line 149: Nit: the provider covers `paypal` / `venmo` / `card`, but not a token entry missing `payment_source` entirely which is the failure mode from the `?? ''` comment above. Worth a row once that line is settled.

Narek13 (approved):
Thanks for working on this. LGTM. I left a few comments, but I don’t think any of them are blockers, so I’m approving the PR.

PR #4374, opened 2026-05-14
[RSM-108] Add initial abilities for PayPal Payments for WooCommerce (6418)

rtio:
RSM-108

### Description

Registers WooCommerce PayPal Payments with the WordPress Abilities API so agents can interact with the extension without needing direct REST knowledge. **4 read abilities** ship in this PR; every ability is gated behind the `woocommerce_paypal_payments_abilities_enabled` filter (default `false`) so the registration scaffolding ships without committing to a final shape.

#### Read abilities (`readonly: true`, `idempotent: true`, `destructive: false`)

- `woocommerce-paypal-payments/get-connection-status` — merchant PayPal connection state (connected? sandbox? merchant ID, email, seller type). API credentials (`clientId`, `clientSecret`) are stripped from the payload before returning. **Reference ability.**
- `woocommerce-paypal-payments/get-payment-methods` — every PayPal payment gateway with enabled state, dependency edges, and warning messages.
- `woocommerce-paypal-payments/get-order-tracking` — shipment tracking entries (carrier, tracking number, status) for a WooCommerce order. Takes `wc_order_id` (required, integer ≥ 1).
- `woocommerce-paypal-payments/get-paypal-order` — full PayPal order (status, intent, purchase units, payment source, captures, authorizations) for a given PayPal order ID or WooCommerce order ID. Description carries an explicit "may return payer PII" notice so callers can apply downstream handling.

**Dropped from initial scope:** `get-settings`, `get-webhook-status` (commit 190dc4431), and `get-last-webhook-event` (commit d77482f96). The two Shape-2 backings (`SettingsRestEndpoint`, `WebhookSettingsEndpoint`) and the Shape-3 backing (`WebhookEventStorage::get_data()`) remain in place — only the Abilities API surface is dropped for now.

#### Write abilities

None — Phase 1 is read-only by design (RSM-108 constraint).

#### Architecture

- New module under `modules/ppcp-abilities/` following the plugin's Syde Modularity convention. Adds `composer.json`, `module.php`, `services.php`, `AbilitiesModule`, `Abilities_Registrar` coordinator, `Domain/AbstractPpcpAbility` base, and one `Domain/<Name>.php` per ability. Wired into `modules.php`.
[...]

PR #4698, opened 2026-09-08
Convert ppcp-abilities to Syde Modularity DI (6695)

mmaymo:
## 🐛 What and why

The `ppcp-abilities` module (merged in PR #4374) was built entirely from static classes, justified by a false premise in `AbstractPpcpAbility`: that the Abilities API "requires callable arrays, forcing static methods with no per-instance DI hook". WooCommerce only requires static `get_name()` and `get_registration_args()` on the definition class — the callbacks inside the returned args accept *any* callable, instance methods included.

Because of that premise the module bypassed **Syde Modularity** DI. It reached services through `PPCP::container()` — the only service-locator use in the entire plugin — declared `ServiceModule` against a `services.php` that returned an empty array, carried static test-reset seams, and round-tripped through `rest_do_request()` to reach two REST controllers that are already container services in this same plugin.

The practical risk was duplication and drift: identifier resolution, PII projection and envelope redaction lived in static methods that could not be injected, decorated via `extensions.php`, or reused. Three review comments raising this (3389446108, 3389479416, 3389858345) were never answered.

## ⚙️ How it works now

`AbilitiesModule::run( ContainerInterface $c )` resolves `abilities.registrar` and calls `AbilityHandlers::set()` with a map of `AbilityNames` constant → factory closure, plus `array( $registrar, 'can_manage_woocommerce' )` as the shared permission gate. Each `Domain\Get*::get_registration_args()` returns `AbilityHandlers::callback( self::get_name() )` as its `execute_callback`.

That callback is a closure with the signature `function ( $input = null )` — matching the removed static `execute( $input = null )` arity exactly. It resolves its handler **on invocation**, memoizes it for the rest of the request, and wraps resolution in `try/catch( Throwable )`.

The handlers take their backing services by constructor: `CommonRestEndpoint`, `PaymentRestEndpoint`, `OrderTrackingEndpoint` and `OrderEndpointCached`, alongside `EnvelopeParser` and a PSR-3 logger. `rest_do_request()` and the service locator are gone.

### Covered by unit tests

- ✅ `services.php` returns a non-empty array registering a registrar plus one handler per ability, each receiving its backing endpoint and a `LoggerInterface`.
- ✅ No file under `modules/ppcp-abilities/` references `PPCP::container()`, and `resolve_service()` no longer exists.
- ✅ `run()` resolves the registrar from the container and invokes it on the instance — a strict container double serving *only* `abilities.registrar` proves nothing else is resolved.
- ✅ `AbilitiesRegistrar` is instantiable, its init guard is instance state, and `reset_initialized_for_testing()` is gone.
- ✅ `AbstractPpcpAbility` holds no static logger; every runtime log call goes through a constructor-injected logger.
- ✅ Each `Domain\Get*` still implements `AbilityDefinition` with static `get_name()`/`get_registration_args()`, bound to a DI handler.
- ✅ The Shape-2 handlers call `CommonRestEndpoint::get_merchant_details()` and `PaymentRestEndpoint::get_details()` on injected instances; no file calls `rest_do_request()`.
- ✅ The loader gate uses `class_exists( AbilitiesLoader::class )`, with no quoted FQCN left.
- ✅ With `woocommerce_paypal_payments_abilities_enabled` at its `false` default, nothing registers and the loader filter callback is never added.
- ✅ With the flag on, all four abilities register under unchanged slugs with the `manage_woocommerce` gate; `clientId`/`clientSecret` stay stripped, and `payer`/`payment_source`/purchase-unit shipping stay stripped unless `include_payer_pii` is true.
- ✅ A `success=false` envelope yields a `WP_Error` carrying no upstream PayPal text, with the raw message and details written to the PSR-3 logger.
- ✅ `AbilitiesModule::run()` contains no reference to any class under `Abilities\Domain\`; the map is keyed by `AbilityNames` constants.
- ✅ Booting where `AbilityDefinition` is undefined completes without a fatal and leaves no `Domain\Get*` class in `get_declared_classes()` — driven as a real subprocess, since the PHPUnit bootstrap always defines the interface.
- ✅ Each `Domain\Get*::get_name()` returns the matching `AbilityNames` constant, so slugs cannot drift from map keys.
- ✅ Building the registration args for all four abilities resolves no backing endpoint service.
- ✅ A container that throws yields `WP_Error 'woocommerce_paypal_payments_service_unavailable'` with the `Throwable` logged via `error_log()`, and no exception escapes into the registration action.
- ✅ The execute callback forwards its input to the handler and memoizes the resolved handler for the request.
- ✅ `ModuleStructureTest.php` is tracked in git and asserts `AbilitiesModule.php` carries no `Domain\` code reference.

### Constructor signature changed

- `AbilitiesRegistrar::__construct( ?callable $loader_gate = null )` — was a purely static class. The optional gate is a test seam; the service is wired as `abilities.registrar`.
- New handler services, all wired in `services.php`: `GetConnectionStatusHandler( CommonRestEndpoint, EnvelopeParser, LoggerInterface )` from `settings.rest.common`; `GetPaymentMethodsHandler( PaymentRestEndpoint, EnvelopeParser, LoggerInterface )` from `settings.rest.payment`; `GetOrderTrackingHandler( OrderTrackingEndpoint, LoggerInterface )` from `order-tracking.endpoint.controller`; `GetPaypalOrderHandler( OrderEndpointCached, LoggerInterface )` from `api.endpoint.order.cached`.
- `EnvelopeParser::__construct( LoggerInterface $logger )` — new collaborator, wired as `abilities.envelope-parser`.

### Deprecated, not removed

`AbstractPpcpAbility::CATEGORY_SLUG` and `AbilitiesRegistrar::CATEGORY_SLUG` are kept as `@deprecated 4.1.0` aliases of `AbilityNames::CATEGORY_SLUG`, so external code reading either constant keeps working.

### Removed from `@internal` classes

`AbstractPpcpAbility::set_logger()`, `logger()`, `reset_logger_for_testing()`, `resolve_service()` and `delegate_to_rest_controller()` are gone, as is `AbilitiesRegistrar::reset_initialized_for_testing()`. All were `@internal` static members of a module marked `@internal` and gated behind a feature flag that defaults to `false`, so no ability has ever registered on a stock install.


modules/ppcp-abilities/src/Handler/GetConnectionStatusHandler.php
```diff
@@ -0,0 +1,132 @@
...
+class GetConnectionStatusHandler {
+
+	/**
+	 * Fields dropped before returning to the agent. clientId/clientSecret are
+	 * the OAuth API credentials (admin-only); an agent could log them verbatim.
+	 * The merchant `id`/email stay — agents need them to reason about the account.
+	 *
+	 * @var array<int, string>
+	 */
+	private const REDACTED_FIELDS = array( 'clientId', 'clientSecret' );
```
Narek13: This is a denylist, while the top-level projection right below it is an allowlist (`merchant` + `features` only) — so the two halves fail in opposite directions. Against today's `merchant_info_map` the redaction is complete; I checked all 8 fields. But add a ninth that happens to be sensitive and it ships to the agent by default.

AGENTS.md names this exact drift: *"audit the ability's registration for required updates — annotations, input/output schema, description, and the redaction list on `GetConnectionStatus`. Drift between the ability and its backing is the failure mode the feature flag exists to contain."*

An allowlist of the six intended fields would make that drift fail closed instead of open. Cheap to do here, and it removes the need for anyone to remember the rule.

modules/ppcp-abilities/src/Handler/GetPaypalOrderHandler.php
```diff
@@ -0,0 +1,175 @@
...
+	 * Project the PayPal Order payload to the agent shape, stripping the payer
+	 * block, payment_source, and per-purchase-unit shipping unless
+	 * $include_payer_pii.
+	 *
+	 * @param array $payload           Decoded PayPal Order payload.
+	 * @param bool  $include_payer_pii Pass payer + shipping through when true.
+	 * @return array
+	 */
+	private function project_order( array $payload, bool $include_payer_pii ): array {
+		if ( $include_payer_pii ) {
```
Narek13, line 116: Worth considering a server-side gate as the outer bound, e.g.

```php
$allowed = (bool) apply_filters( 'woocommerce_paypal_payments_abilities_allow_payer_pii', false );
if ( $include_payer_pii && $allowed ) {
    return $payload;
}
```

so the input flag can only narrow within what the site permits. Not blocking, but it is the one place where the surface hands out PII on request.

modules/ppcp-abilities/src/AbilityHandlers.php
```diff
@@ -0,0 +1,143 @@
...
+		};
+	}
+
+	/**
+	 * The shared permission gate. Fails closed when nothing was bound: an
+	 * unwired abilities surface must never be readable.
+	 *
+	 * @return callable
+	 */
+	public static function permission_callback(): callable {
```
Narek13: Asymmetry worth a look: `callback()` returns a closure that resolves on invocation, but this resolves eagerly and bakes the result into the registration args. Ordering makes that safe today — `run()` always calls `set()` before `init()`, and Woo's loader runs later still — so this is latent rather than broken.

The consequence is that a later `set()` would update execution but not permissions, which is a surprising split if anyone ever rebinds. Returning a deferring closure here too would make the pair consistent and cost nothing.

modules/ppcp-abilities/src/Helper/EnvelopeParser.php
```diff
@@ -0,0 +1,105 @@
...
+	 * extra top-level keys (e.g. CommonRestEndpoint's merchant/features) can
+	 * reuse the redaction without having those keys discarded by `data`
+	 * extraction.
+	 *
+	 * @param array $payload        Decoded REST envelope.
+	 * @param bool  $redact_message See unwrap().
+	 * @return \WP_Error|null WP_Error on success=false; null otherwise.
+	 */
+	public function error_or_null( array $payload, bool $redact_message = true ): ?\WP_Error {
+		if ( ! array_key_exists( 'success', $payload ) || false !== $payload['success'] ) {
```
Narek13, line 74: Strict comparison means `0`, `'false'` and `null` all count as success. The plugin's own envelope always sets a real boolean, so this is fine in practice.

`return_success()` output is filterable, and any callback in the chain can return the wrong type. A falsy check would be safer here, at the cost of treating an absent value as failure. Your call which way you would rather be wrong.

modules/ppcp-abilities/src/Handler/GetConnectionStatusHandler.php
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * @var EnvelopeParser
+	 */
+	private $envelope;
+
+	/**
+	 * @var LoggerInterface
+	 */
+	private $logger;
```
Narek13, line 47:
```suggestion
	private LoggerInterface $logger;
```

I sugeest doing this for all properties

modules/ppcp-abilities/src/Handler/GetConnectionStatusHandler.php
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * @param CommonRestEndpoint $endpoint The backing merchant-details endpoint.
+	 * @param EnvelopeParser     $envelope The shared envelope parser.
+	 * @param LoggerInterface    $logger   The plugin's PSR-3 logger.
+	 */
```
Narek13, line 53: We decided to not add these constructor comments as they do not convey any meaningful info

modules/ppcp-abilities/src/Handler/GetOrderTrackingHandler.php
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * @var OrderTrackingEndpoint
+	 */
+	private $endpoint;
+
+	/**
+	 * @var LoggerInterface
+	 */
+	private $logger;
```
Narek13, line 35: Same here
```suggestion
	private OrderTrackingEndpoint $endpoint;

	private LoggerInterface $logger;
```

modules/ppcp-abilities/src/Handler/GetOrderTrackingHandler.php
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * @param OrderTrackingEndpoint $endpoint The backing tracking endpoint.
+	 * @param LoggerInterface       $logger   The plugin's PSR-3 logger.
+	 */
```
Narek13, line 40: Same here. These are redundant

modules/ppcp-abilities/src/Handler/GetPaymentMethodsHandler.php
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * @var LoggerInterface
+	 */
+	private $logger;
+
+	/**
+	 * @param PaymentRestEndpoint $endpoint The backing payment-details endpoint.
+	 * @param EnvelopeParser      $envelope The shared envelope parser.
+	 * @param LoggerInterface     $logger   The plugin's PSR-3 logger.
+	 */
```
Narek13, line 44: Same here :)

Narek13 (approved):
Thanks a lot for the great PR, @mmaymo! The code looks good to me. I’ve just added a couple of suggestions, mostly around adding types for the class properties and removing redundant comments. I’m going to approve it since none of them are blockers.
I would suggest to check all the new classes and the types of properties.

PR #4699, opened 2026-09-08
Keep the express shipping methods when PayPal has no usable address yet (6898)

stracker-phil:
*Changelog:* `Fix - Express PayPal buttons now complete the order instead of redirecting to the checkout page #4699`

# Description

## 🐛 The problem

With the Pay Now experience on SDK v6, the PayPal button on product and cart pages sent the buyer to the WooCommerce checkout page instead of completing the order right after approval.

## 💡 Why it happens

Before approval, PayPal gives us no usable buyer address. The plugin took that as "no shipping details at all" and discarded the **shipping methods** along with the address. An order without shipping methods cannot be completed in the popup, so PayPal falls back to the redirect.

## ✅ The fix

An unusable address no longer costs the shipping methods. They are sent on their own, and the places that decide the shipping preference and assemble the request treat "methods, but no address yet" as a normal state.

## 🔍 What to review

Shipping details without an address are new to this plugin. Everything downstream reads the address back from PayPal's response, so the only path worth a second look is the one that patches an existing PayPal order.

No signature, hook or script handle changed.

## 🧪 How to verify

(testable in incognito window to start with an uninitialized session)

1. Enable the Pay Now experience with SDK v6, then add a physical product to the cart as a guest.
2. Click the PayPal button on the product page and on the cart page and approve in the popup: the order finishes without a stop at the checkout page.
3. On classic checkout with a complete address in the form, place an order through the PayPal button: shipping stays the address that was entered.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -32,6 +32,7 @@ patchwork.json
 README.md
 AGENTS.md
 CLAUDE.md
+.claude/
 E2E-TESTS.md
 wordpress_org_assets*
 .DS_Store
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Changelog ***
 
+= 4.1.3 - XXXX-XX-XX =
+* TBD
+
 = 4.1.2 - 2026-08-04 =
 * Enhancement - Advanced onboarding options for onboarding wizard #4496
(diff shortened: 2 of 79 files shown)


Narek13 (commented):
LGTM. Thanks

Narek13 (approved):

PR #4704, opened 2026-09-10
Keep WooCommerce's place order button label

Narek13:
The plugin renamed WooCommerce's "Place order" button to "Proceed to PayPal" and added a matching description below it. The label is wrong for every funding source that does not return to PayPal, and the button belongs to WooCommerce, so both are gone.

Applies to Classic and Block checkout, and to both SDK v5 and v6.

## Changes

- The `wcgateway.place-order-button-text` and `...-description` services are **removed**, together with the two filters they wrapped. They existed only to supply strings the plugin no longer sets.
- Classic: `PayPalGateway` and `CardButtonGateway` no longer take or store a label. WooCommerce names its own button.
- Blocks: `placeOrderButtonLabel` and the Checkout Actions filter are registered only when an override is present. The filter matters because it is what the block actually reads.
- v6's continuation branch set the label explicitly to defend against v5's filter reaching it. Nothing renames the button now, so that workaround is gone.

## Replacing the removed extension points

`get_payment_method_data()` on **both** the v5 and v6 payment methods now returns through a new filter:

```php
add_filter(
	'woocommerce_paypal_payments_blocks_payment_method_data',
	function ( array $data ): array {
		$data['placeOrderButtonLabel'] = 'Pay now';
		return $data;
	}
);
```

One hook over the whole payload replaces two that each carried a single string, and it reaches every key rather than those two. Classic checkout has no equivalent because it does not go through `get_payment_method_data()`; renaming the button there is `woocommerce_order_button_text`'s job, which is the correct hook for it.

## Backward compatibility

This is a deliberate break, not a deprecation.

- **Constructor signatures changed on three container-reachable classes**: `PayPalGateway`, `CardButtonGateway` and `PayPalPaymentMethod` each lose a parameter. This is the sharpest edge here, since a subclass calling `parent::__construct()` fatals on load. Anything extending or directly instantiating them needs updating.
- `woocommerce_paypal_payments_place_order_button_text` and `woocommerce_paypal_payments_place_order_button_description` are **removed**. Callbacks stop applying silently. Nothing in the payment flow depends on them: in the worst case a custom label reverts to "Place order".
- `placeOrderButtonText` and `placeOrderButtonDescription` no longer appear in the block payment method data, which is visible to third-party JS that reads them.
- New public hook `woocommerce_paypal_payments_blocks_payment_method_data`, fired by both payment methods.

## Test steps

With "Pay now" disabled so the non-express PayPal row appears:

1. Classic checkout, select PayPal: the button reads "Place order", not "Proceed to PayPal", and no description appears beneath it.
2. Block checkout, same check, under both SDK v5 and v6.
3. Start an express payment (PayPal, Venmo, Pay Later or Google Pay), approve, and land back on block checkout in order review: the button reads "Place order".
4. Add the filter snippet above and confirm the label applies on block checkout under both SDKs.

## Verification

`npm run test:unit-js` 1065/1065 pass. PHPUnit sits at the pre-existing baseline of 9 errors, unchanged from a clean checkout of the base commit and unrelated to this work. PHPCS and PHPStan clean on every changed file. Each commit was checked out and tested individually, so the suite is green at every step rather than only at the tip.


modules/ppcp-sdk-v6/services.php
```diff
@@ -280,9 +280,7 @@
 			$usable_for_cart = ! $subscription_helper->cart_contains_subscription() || $can_vault;
 
 			return array(
-				'enabled'     => $offer_method && $usable_for_cart,
-				'text'        => (string) $container->get( 'wcgateway.place-order-button-text' ),
-				'description' => (string) $container->get( 'wcgateway.place-order-button-description' ),
+				'enabled' => $offer_method && $usable_for_cart,
 			);
```
stracker-phil: Suggestion: Rename this service and return a boolean instead of an array that only contains a single boolean flag

stracker-phil (approved):
Nice work :) LGTM

PR #4720, opened 2026-09-16
feat: load the SDK v6 frontend by default

Narek13:
SDK v6 shipped to fresh installs only: every store that already existed had `'no'` written against `woocommerce-ppcp-sdk-v6-eligible`, and the flag read as opt-in. This turns it around — v6 loads unless something said no.

## Changes

- **The flag reads as default-on.** `'yes' === get_option(…)` becomes `'no' !== get_option(…)`. The `woocommerce.feature-flags…sdk_v6_enabled` filter and `PCP_SDK_V6_ENABLED` are untouched, so v5 stays reachable.
- **A one-shot migration hands existing stores over**, behind its own marker.
- **The install-time write is gone.** With an absent option meaning "on", writing `'yes'` on a fresh install said the same thing twice. The option is now purely the opt-out record.
- **No country is withheld any more.** `MerchantCountrySupport`'s default list goes from `array( 'MX' )` to `array()` — Mexico was the only entry, held back against an issue since fixed, and it ships with the rollout.

## On the country list

The class and its `woocommerce_paypal_payments_sdk_v6_unsupported_countries` filter are kept rather than deleted, so a country can still be withdrawn without a release while the rollout is in flight. Expected to go in the first 2027 release, once the rollout is globally complete.

The migration runs that check itself rather than leaving it to `SdkV6Module`, which only evaluates it on `woocommerce_paypal_payments_authenticated_merchant` — a store connected months ago would never re-evaluate it. With an empty list the check passes for everyone; it matters only if a country is filtered back in mid-rollout.

## Backward compatibility

No signature, hook or handle changes.

- **Every existing store moves to the v6 frontend** on update — the whole merchant base, not just new installs.
- **Mexican merchants move too**, having previously been held on v5.
- **A downgrade to 4.1.x does not return a store to v5.** The option is left at `'yes'`, which that version reads as v6-on. Deliberate: deleting it instead would let the older install path rewrite `'no'`, and with the marker set the migration could never put it right.
- **`woocommerce-ppcp-sdk-v6-eligible` changes meaning**, from "was eligible at install" to "opted out".

Not tested under multisite.

## Test steps

1. **Existing store** — set `woocommerce-ppcp-version` older and the eligible option to `no`, load any page: it becomes `yes`, the marker is set, v6 loads.
2. **One-shot** — force it back to `no` with the marker in place: it stays `no`.
3. **Fresh install** — no option at all: v6 loads.
4. **Mexican merchant** — merchant country `MX` in `woocommerce-ppcp-data-common`: handed over like any other.
5. **Escape hatches** — `add_filter( 'woocommerce.feature-flags.woocommerce_paypal_payments.sdk_v6_enabled', '__return_false' )` falls back to v5; filtering a country into `woocommerce_paypal_payments_sdk_v6_unsupported_countries` still withholds it.

Note: a mu-plugin forcing the feature flag with `__return_true` masks the whole change. Check for one before concluding the gate works.


stracker-phil (approved):
LGTM

stracker-phil (approved):

PR #4725, opened 2026-09-18
Free trial subscription - PayPal gateway is missing on block checkout and PayPal button is missing on classic checkout (6853 v3)

Dinamiko:
On the Blocks checkout customers could only use the express button, not the place-order flow., the PayPal gateway row was hidden for free-trial ($0) subscriptions (PCP-6900). A first-time buyer had no saved PayPal account, and the native "Place order" button couldn't vault one, a $0 order can't be created at PayPal, and vaulting only happened through the express button , so showing the row produced a "No saved PayPal account." failure. 

This PR mades the native "Place order" button vault a first-time free-trial buyer via a server-side redirect:                                                                                                                                                                                        
  - Gateway (`PayPalGateway::process_payment`): when a free-trial order has no saved token, it now applies a filter to get a PayPal vault-approval URL and redirects there instead of failing. 
  - New vault-redirect service + return endpoint (`ppcp-save-payment-methods`): creates a Vault v3 setup token (return URL carries the order id + a one-time nonce), redirects to PayPal to approve; on return, exchanges the setup token for a stored payment token and completes the $0 order through the existing free-trial branch.                                              
  - Un-hid the row for free-trial carts in both Blocks stacks: v5 (PayPalPaymentMethod, reverting PCP-6900's gate) and v6 (`checkout-block.js`, `regularRowAllowedForCart`).


modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+			return;
+		}
+
+		$setup_token_id = (string) $wc_order->get_meta( self::SETUP_TOKEN_META );
+		if ( ! $setup_token_id ) {
+			$this->fail( __( 'Payment information is missing. Please try placing your order again.', 'woocommerce-paypal-payments' ) );
+			return;
+		}
+
+		$customer_id = is_user_logged_in()
```
Narek13: This reads the customer from whoever is browsing, but the gateway you hand off to keys on the order:

```php
// here
get_user_meta( get_current_user_id(), '_ppcp_target_customer_id', true )

// PayPalGateway, a few lines into the same flow
WC_Payment_Tokens::get_customer_tokens( $wc_order->get_customer_id(), self::ID );
```

On the happy path they are the same person, so this works. They come apart in two ways:

**The session is lost over the PayPal round trip.** `is_user_logged_in()` is false on return, `$customer_id` is `''`, and the account is vaulted with no PayPal customer attached so the next vault for that buyer starts a second PayPal customer rather than adding to the first.

**Someone else opens the return URL while logged in.** The nonce is a bearer token in a query string: it reaches PayPal, comes back through the browser, and lands in history. Whoever holds it gets the token vaulted against *their* `_ppcp_target_customer_id`, on an order that is not theirs.

Reading it off the order instead closes both, and matches what the gateway already does:

```php
$order_customer = $wc_order->get_customer_id();
$customer_id    = $order_customer
    ? (string) get_user_meta( $order_customer, '_ppcp_target_customer_id', true )
    : '';
```

Worth considering a `$wc_order->get_customer_id() === get_current_user_id()` check too when the order has an owner, so a leaked nonce cannot complete a logged-in buyer's order at all.

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+
+		// One-time use: drop the stored token/nonce so the return URL cannot be replayed.
+		$wc_order->delete_meta_data( self::SETUP_TOKEN_META );
+		$wc_order->delete_meta_data( self::RETURN_NONCE_META );
+		$wc_order->save();
+
+		// Hand the exchanged token to the gateway's existing free-trial branch, which
+		// stores it against the customer and completes the $0 order. Works for both
+		// logged-in and guest buyers (the branch keys off the order's customer id).
+		WC()->session->set( 'ppcp_guest_payment_for_free_trial', $result );
```
Narek13, line 132: `WC()->session` is dereferenced without a guard. It should be there on a `wc-ajax` request, but this endpoint is reached by a browser redirect from PayPal rather than by your own JS, so it is worth not assuming — the codebase guards it elsewhere, e.g. `CardFieldsModule`:

```php
function_exists( 'WC' ) && WC()->session instanceof \WC_Session
```

A null here is a fatal on the buyer's return, after the token has already been exchanged.

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+
+			$result = $this->payment_method_tokens_endpoint->create_payment_token( $payment_source, $customer_id );
+		} catch ( Exception $exception ) {
+			$this->logger->error( "Free-trial vault return: could not exchange setup token for WC order {$wc_order_id}: " . $exception->getMessage() );
+			$this->fail( __( 'Could not save the PayPal account. Please try again.', 'woocommerce-paypal-payments' ) );
+			return;
+		}
+
+		// One-time use: drop the stored token/nonce so the return URL cannot be replayed.
+		$wc_order->delete_meta_data( self::SETUP_TOKEN_META );
```
Narek13, line 115: These run before `process_payment()`, so a transient failure there leaves the buyer with a dead return URL: the retry lands on "Payment session expired" and they have to place the order again, even though the account was vaulted successfully.

Clearing before the exchange is the safe order against replay, so this may be the trade you want. Just flagging that the failure path costs the buyer the vaulting they just approved.

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+use Exception;
+use Psr\Log\LoggerInterface;
+use WC_Order;
+use WooCommerce\PayPalCommerce\ApiClient\Endpoint\PaymentMethodTokensEndpoint;
+use WooCommerce\PayPalCommerce\ApiClient\Entity\PaymentSource;
+use WooCommerce\PayPalCommerce\WcGateway\Gateway\PayPalGateway;
+
+/**
+ * Class FreeTrialVaultReturnEndpoint
+ */
```
Narek13: Redundant
```suggestion
```

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+
+	/**
+	 * The order meta key holding the approved setup token id.
+	 */
+	const SETUP_TOKEN_META = '_ppcp_free_trial_setup_token';
+
+	/**
+	 * The order meta key holding the one-time return nonce.
+	 */
+	const RETURN_NONCE_META = '_ppcp_free_trial_vault_nonce';
```
Narek13:
```suggestion
	/**
	 * The order meta key holding the approved setup token id.
	 */
	public const SETUP_TOKEN_META = '_ppcp_free_trial_setup_token';

	/**
	 * The order meta key holding the one-time return nonce.
	 */
	public const RETURN_NONCE_META = '_ppcp_free_trial_vault_nonce';
```

I suggest to always add visibilities

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+	 * @var PaymentMethodTokensEndpoint
+	 */
+	private $payment_method_tokens_endpoint;
+
+	/**
+	 * The logger.
+	 *
+	 * @var LoggerInterface
+	 */
+	private $logger;
```
Narek13:
```suggestion
	private PaymentMethodTokensEndpoint $payment_method_tokens_endpoint;

	private LoggerInterface $logger;
```

modules/ppcp-save-payment-methods/src/Endpoint/FreeTrialVaultReturnEndpoint.php
```diff
@@ -0,0 +1,183 @@
...
+
+	/**
+	 * FreeTrialVaultReturnEndpoint constructor.
+	 *
+	 * @param PaymentMethodTokensEndpoint $payment_method_tokens_endpoint The payment method tokens endpoint.
+	 * @param LoggerInterface             $logger                         The logger.
+	 */
```
Narek13:
```suggestion
```

Redundant

modules/ppcp-save-payment-methods/src/FreeTrialVaultRedirect.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @var LoggerInterface
+	 */
+	private $logger;
+
+	/**
+	 * FreeTrialVaultRedirect constructor.
+	 *
+	 * @param PaymentMethodTokensEndpoint $payment_method_tokens_endpoint The payment method tokens endpoint.
+	 * @param LoggerInterface             $logger                         The logger.
+	 */
```
Narek13: Same here. Please see My suggestions for `FreeTrialVaultReturnEndpoint` class

Narek13 (changes requested):
Thanks @Dinamiko for working on this. Added some suggestions and doubts.
Please also check the CI and merge conflicts

Dinamiko:
@Narek13 thanks for the feedback, I've just applied them

Narek13 (approved):
Thanks, LGTM

